PR #228, opened 2019-12-04
Added Header map to Rest request

royjit:

### DIFF
--- a/Amplify/Categories/API/Request/RESTOperationRequest.swift
+++ b/Amplify/Categories/API/Request/RESTOperationRequest.swift
@@ -19,7 +19,7 @@ public struct RESTOperationRequest: AmplifyOperationRequest {
     public let path: String?
 
     /// Request headers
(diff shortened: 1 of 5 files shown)


Amplify/Categories/API/Request/RESTOperationRequest.swift
```diff
@@ -30,12 +33,14 @@ public struct RESTOperationRequest: AmplifyOperationRequest {
     public init(apiName: String?,
                 operationType: RESTOperationType,
                 path: String? = nil,
+                headers: [String: String] = [:],
```
lawmicha: any reason to do this over `headers: [String: String]? = nil`? it would be more consistent with queryParameters

royjit: Just to avoid another check for optional. 
I guess, it is better to make it same as query, will update my PR.

PR #229, opened 2019-12-04
[Draft] [Not working] Add subspec to AmplifyPlugins to pick up AWSPluginsCore automatically

lawmicha:
### Add the "AWSPluginsCore" subspec to AmplifyPlugins podspec.
this puts the dependency (AWSPluginsCore) under the AmplifyPlugins pod so that they do not need to have a dependency on a pod that is external. As can be seen in the change, from
```
ss.dependency 'AWSPluginsCore', AMPLIFY_VERSION // external
to
ss.dependency 'AmplifyPlugins/AWSPluginsCore', AMPLIFY_VERSION // same pod
```
This is the developer facing pod and technically we could name this whatever we want like `AmplifyPluginsCore`. The developer could also install this pod (pod "AmplifyPlugins/AWSPluginsCore") if they really wanted to, perhaps for GraphQL document helpers of some sort, and it is not a plugin.

### AWSPluginsCore.podspec, AmplifyTestCommon.podspec
We still have the AWSPluginsCore.podspec and AmplifyTestCommon.podspec. These do not need to be published and are used internally by the plugin's via Podfile to maintain separate workspaces per Plugin. We need these local pods to match the targets, ie. AWSPluginsCore's source files matches the AWSPluginsCore target. Same for AmplifyTestCommon and AmplifyTestCommon/AWSPluginsTestCommon. 

## Minor changes
- Moved CoreMLPredictionsPlugin under AmplifyPlugins. This means docs and samples will change to  `pod "AmplifyPlugins/CoreMLPredictionsPlugin"` 

### Testing done

```
pod "Amplify", :path => "~/aws-amplify/amplify-ios"
pod "AmplifyPlugins/AWSAPIPlugin", :path => "~/aws-amplify/amplify-ios"
```
Not working due to missing `import AWSPluginsCore` within a file in AWSAPIPlugin..


lawmicha:
need to investigate
```
/Users/distiller/amplify-ios/AmplifyPlugins/Core/AWSPluginsTestCommon/MockAWSAuthService.swift:10:8: error: no such module 'AWSPluginsCore'
```

Amplify workspace contains the AmplifyPluginsTestCommon target which is importing AWSPluginsCore from local (not pod)

lawmicha:
closing this for now, and keeping the branch

PR #232, opened 2019-12-04
feat(ops): Pipeline improvements, deploy step

iartemiev:
* Running all unit test jobs in parallel
* Create test output xml for better test reporting in circleci
* Added deploy step to push podspecs to trunk using fastlane actions (enabled for release branch)

Will be adding conventional commits support next to automatically bump semver 


Gemfile.lock
```diff
@@ -0,0 +1,218 @@
+GEM
```
royjit: Should we commit this file?

fastlane/Fastfile
```diff
@@ -0,0 +1,43 @@
...
+  "AmplifyTestCommon.podspec",
+  "AWSPluginsCore.podspec",
+  "CoreMLPredictionsPlugin.podspec"
+]
+
+platform :ios do
+  desc "Bump pod versions"
+  lane :bump_podspecs_patch do
+
+    pods.each { |pod| version_bump_podspec(path: pod, bump_type: "patch") }
```
royjit: Does this bump the version for pod? Pod version and our github code base version should match

iartemiev: Yes but that action is not actually utilized at the moment. I just wrote it with the intention of using it down the road. 

The plan is to add Git tagging and automatic semver versioning with conventional commits to the pipeline.

fastlane/report.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuites>
+  <testsuite name="fastlane.lanes">
+    
+
```
royjit: nit: Remove extra lines

Gemfile.lock
iartemiev: I committed it mainly to enable caching gems in the pipeline. I checksum Gemfile.lock for the cache key. Could probably go off just the Gemfile though, especially since I'm only using those gems in the pipeline itself

fastlane/report.xml
iartemiev: Actually, this file shouldn't have been committed at all, that was just from my running the job locally on my machine. I'll delete it

PR #236, opened 2019-12-04
Moved out predictions podspec

royjit:
*Issue #, if available:*

*Description of changes:*


lawmicha:
samples and docs should need a follow up to replace `import AmplifyPlugins/AWSPredictionsPlugin` to just `import AWSPredictionsPlugin" i think

royjit:
> samples and docs should need a follow up to replace `import AmplifyPlugins/AWSPredictionsPlugin` to just `import "AWSPredictionsPlugin"` i think

Yes we need to update

Issue #239, opened 2019-12-06
Model based APIs for API Category check against existence of DataStore


lawmicha:
since removing the syncable flag in the models (they weren't being model generated at the moment anyways). and the change to datastore to always treat non-system models as syncable, then API category's Model based APIs are broken. API needs a way to check if DataStore exists, and if it does not, treat the models non-syncable. If datastore exists, developer should use datastore

lawmicha:
https://github.com/aws-amplify/amplify-ios/pull/238/files#r355501833

palpatim:
Related to #177

jamesonwilliams:
> API needs a way to check if DataStore exists

Why does API need to know of the existence of DataStore? Shouldn't DataStore depend on API, but API know nothing of DataStore?

lawmicha:
> > API needs a way to check if DataStore exists
> 
> Why does API need to know of the existence of DataStore? Shouldn't DataStore depend on API, but API know nothing of DataStore?

Right, originally the Model Based API in API plugin had a dependency on ModelRegistry, which then makes it look like API depends on if sync enabled is true which is DataStore's use case. this code has been refactored and no longer checks for sync-enabled. consumers now have to explicitly construct requests which have sync fields or not.

lawmicha:
i believe this can be closed- Using API's GraphQLRequest builders with APIPlugin directly should be done on an API that does not have conflict resolution enabled. The original issue was opened with the idea that the builders should fail to construct the request if the API does have conflict resolution enabled, but this would have just moved the failure earlier. Instead, we can keep this logic out of the builders and leave it to the request to fail. As such, developers will rarely get into this situation since Amplify CLI will only create the conflict resolution enabled API by asking if the API should be enabled for DataStore.

Issue #177, opened 2019-11-27
Use existence of API config to trigger sync start


palpatim:
Currently we're using the existence of syncable models to start syncing; instead use the Android method of detecting whether an API is configured

drochetti:
This was fixed at some point before GA. Currently DataStore starts syncing based on the presence of a configured `Amplify.API`.

PR #249, opened 2019-12-10
Convert Amplify.API to protocol

palpatim:
**Note: based on branch `palpatim/format-and-lint`. Will rebase against master when that branch is merged in.

- Separated Amplify.API into APICategory protocol and AmplifyAPICategory
  implementation; fixed API client behavior conformances that used to be
  implemented only on the API category class but not declared in the Client
  Behavior protocol
- Refactored "Resettable" behaviors for all categories

*Issue #, if available:*

Lays groundwork for fixing #177, #239

*Description of changes:*


AmplifyTests/CategoryTests/API/APICategoryConfigurationTests.swift
```diff
@@ -250,13 +256,64 @@ class APICategoryConfigurationTests: XCTestCase {
...
+
+        let amplifyConfig = AmplifyConfiguration(api: categoryConfig)
+        try Amplify.configure(amplifyConfig)
+
+        guard let category = Amplify.API as? AmplifyAPICategory else {
+            XCTFail("Could not cast Amplify.API as AmplifyAPICategory")
+            return
+        }
+
+        XCTAssertTrue(category.isConfigured)
```
lawmicha: LGTM, on a related note; I'll do some refactoring with how dataStore actually syncs to the cloud via API plugin to remove the dependency on checking Model.isSyncable when adding selection set. there should be enough information to construct the request with the sync-able fields. This is needed for API plugin's Model based APIs because they are not sync-able by default. Then `category.isConfigured` check is only used for datastore to check if it should sync to the cloud. Datastore constructing the graphQL request should have enough information to add sync-able fields and shouldn't need to know whether or not the category is configured.

1. Developer using API by itself should never provision a sync-able API
2. Developer using DataStore should only provision a sync-able API

PR #252, opened 2019-12-11
Remove dependency on ModelRegistry for syncability of models

lawmicha:
### Major changes
This introduces a flag for whether a GraphQLDocument helper method should contain the sync-able fields. Methods like GraphQLSyncQuery by default will be true, and methods like GraphQLGetQuery can be used with syncEnabled parameter.

### Minor changes
- Moved MinimalGraphQLDeleteMutation over to PluginsCore as GraphQLDeleteSyncMutation
- Updated GraphQLListQuery to include limit and nextToken, in anticipation for ModelBased List query to allow limit/nextToken.

### Testing done
- Ensured the Sync provisioned backend in API integration tests is working as expected, and so went to datastore plugin and updated the calls where necessary, like
```
let document = GraphQLSubscription(of: modelType,
                                   type: subscriptionType,
                                   syncEnabled: true)
// adding syncEnabled true parameter
```
- Ensure Model based APIs are working. by default the models are not syncable until specified to be syncable. since Model based API's should never be used with sync and developer should use DataStore if using sync provisioned backend

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
@@ -48,12 +48,17 @@ class GraphQLSyncBasedTests: XCTestCase {
         let uuid = UUID().uuidString
         let testMethodName = String("\(#function)".dropLast(2))
         let title = testMethodName + "Title"
-        guard let post = createPost(id: uuid, title: title) else {
+        guard let createdPost = createPost(id: uuid, title: title) else {
             XCTFail("Failed to create post with version 1")
             return
         }
+
         let updatedTitle = title + "Updated"
-        let modifiedPost = Post(id: post.id, title: updatedTitle, content: post.content, createdAt: Date())
+
+        let modifiedPost = Post(id: createdPost.model.id,
+                                title: updatedTitle,
+                                content: createdPost.model.content,
+                                createdAt: Date())
 
         let completeInvoked = expectation(description: "request completed")
         var responseFromOperation: GraphQLResponse<MutationSync<AnyModel>>?
@@ -102,7 +107,7 @@ class GraphQLSyncBasedTests: XCTestCase {
         }
 
         XCTAssertEqual(mutationSync.model["title"] as? String, updatedTitle)
-        XCTAssertEqual(mutationSync.model["content"] as? String, post.content)
+        XCTAssertEqual(mutationSync.model["content"] as? String, createdPost.model.content)
         XCTAssertEqual(mutationSync.syncMetadata.version, 2)
     }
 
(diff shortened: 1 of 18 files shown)


AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
```diff
@@ -48,12 +48,17 @@ class GraphQLSyncBasedTests: XCTestCase {
         let uuid = UUID().uuidString
         let testMethodName = String("\(#function)".dropLast(2))
         let title = testMethodName + "Title"
-        guard let post = createPost(id: uuid, title: title) else {
+        guard let createMutation = createPost(id: uuid, title: title) else {
```
palpatim: Is `createPost` creating a Post model or a CreatePost mutation? Please align the names appropriately

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/GraphQLDocument+Subscription.swift
```diff
@@ -16,11 +16,14 @@ public struct GraphQLSubscription: GraphQLDocument {
     public let documentType = GraphQLDocumentType.subscription
     public let modelType: Model.Type
     public let subscriptionType: GraphQLSubscriptionType
+    public let syncEnabled: Bool
 
     public init(of modelType: Model.Type,
-                type subscriptionType: GraphQLSubscriptionType) {
+                type subscriptionType: GraphQLSubscriptionType,
+                syncEnabled: Bool = false) {
```
wooj2: I'm not sure if it makes sense to expose a syncEnabled flag for this class, right?  Seems like if we have a subscription with a model, syncEnabled has to be true, right?

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
lawmicha: updating to `createdPost`

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/GraphQLDocument+Subscription.swift
lawmicha: a backend can be provisioned with/without sync so the graphql document can be constructed with sync related fields or without by using this `syncEnabled` parameter

lawmicha:
> My read on this is that syncability is an internal concept, which seems appropriate. The way we're representing it seems odd to me--why do some types have separate "Sync" flavors, while others are extended by virtue of an incoming boolean parameter? I'd like us to think a little more about this to see if this is the right way to model it. But as an incremental change to enable the functionality for now, LGTM.

Yes I agree, it is odd that there are specific types to build in the sync fields and then there are types that just take in the syncEnabled parameter. I can see these types getting another overhaul, especially with the @auth directive.

PR #256, opened 2019-12-12
Reprovision Storage Integration test backend

lawmicha:
### Major changes
- reprovisioned storage resource in test account
- Adding readme for integration test set up

### Minor changes
- Moved AuthHelper to AmplifyTestCommon for API and Storage tests


AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/BlogPostCommentGraphQLWithAPIKey/BlogPostCommentGraphQLWithAPIKeyTests.swift
```diff
@@ -248,8 +249,8 @@ class BlogPostCommentGraphQLWithAPIKeyTests: XCTestCase {
         return blog
     }
 
-    func createPost(postBlogId: String, title: String) -> Post? {
-        var post: Post?
+    func createPost(postBlogId: String, title: String) -> AWSAPICategoryPluginTestCommon.Post? {
+        var post: AWSAPICategoryPluginTestCommon.Post?
```
palpatim: What is the difference between the two `Post` flavors in AmplifyTestCommon and AWSAPICategoryPluginTestCommon? Any opportunity to collapse those to reduce confusion? If not, consider renaming one of the models altogether so we're not surprised by unexpected divergence in behavior.

AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/AWSS3StoragePluginAccessLevelTests.swift
```diff
@@ -19,8 +20,8 @@ class AWSS3StoragePluginAccessLevelTests: AWSS3StoragePluginTestBase {
 
     // This is a run once function to set up users then use console to verify and run rest of these tests.
     func testSetUpOnce() {
-        signUpUser(username: user1)
-        signUpUser(username: user2)
+        AuthHelper.signUpUser(username: user1, password: password)
```
palpatim: We need to move the password out of the source file and into a setup run from the CLI. As it stands, if an attacker obtains our user pool ID (which remember is not considered sensitive), they could use the test storage bucket to store and distribute files.

AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/AWSS3StoragePluginTestBase.swift
```diff
@@ -11,143 +11,33 @@ import AWSMobileClient
...
-                 }
-             }
-         }
-     }
-     */
-    let bucket: JSONValue = "xxxxxxx-devo"
-    let region: JSONValue = "us-east-1"
-    let networkTimeout = TimeInterval(180) // 180 seconds to wait before network timeouts
-    static let largeDataObject = Data(repeating: 0xff, count: 1_024 * 1_024 * 6) // 6MB
+    static let largeDataObject = Data(repeating: 0xff, count: 1_024 * 1_024 * 6) // 6MBb
```
palpatim: What's an "MBb"? Did you mean to specify Mebibyte "MiB"?

AmplifyTestCommon/Helpers/AuthHelper.swift
```diff
@@ -13,8 +13,8 @@ class AuthHelper {
         let callbackInvoked = DispatchSemaphore(value: 0)
 
         AWSMobileClient.default().initialize { userState, error in
-            if let error = error {
-                fatalError("Error initializing AWSMobileClient. Error: \(error.localizedDescription)")
+            if let error = error as? AWSMobileClientError {
```
palpatim: Why only fatal on AWSMobileClientErrors?

AmplifyTestCommon/Helpers/AuthHelper.swift
```diff
@@ -69,4 +71,17 @@ class AuthHelper {
...
+        AWSMobileClient.default().signOut()
+    }
+
+    static func getIdentityId() -> String {
+        let task = AWSMobileClient.default().getIdentityId()
+        task.waitUntilFinished()
+        if let error = task.error {
+            fatalError("Could not get identityId, with error \(error)")
+        }
+        return task.result! as String
```
wooj2: generally we comment when we use a `!`

AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/README.md
```diff
@@ -1,16 +1,42 @@
...
-8. Copy the newly generated awsconfiguration.json file to AmplifyTestApp
-9. Replace the bucket and region variables under AWSPlugins/AWSS3StoragePluginIntegrationTests/AWSS3StoragePluginTestBase.swift with the values in your awsconfiguration.json file
-10. You can now run most of the integration tests
-11. To run the AWSS3StoragePluginAccessLevelTests
-    1. Go to the signUpUser function at the bottom of the file and add the following parameter to the signUp function call: userAttributes: ["email":"sample@sample.com"]
-    2. Run testSetUpOnce()
-    3. Go to the AWS console for the associated account -> Cognito -> Manage user pools -> Select the account name you created during amplify add storage step -> Click on Users and Groups on the left -> Click each of the two test usernames and confirm them
+## Storage Integration Tests
+
+The following steps show how to set up Storage with unauthenticated and authenticated access. The auth configured will be Cognito UserPool. This set up is used to run the tests in AWSS3StoragePluginFunctionalTests
```
wooj2: s/show/demonstrate

change:
The auth configured will be Cognito UserPool
to
In the case of authenticated access, we will be using Cognito UserPools.  Both unauthenticated and authenticated configurations are used to execute the AWSS3StoragePluginFunctionalTests.

AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/AWSS3StoragePluginTestBase.swift
```diff
@@ -175,4 +65,22 @@ class AWSS3StoragePluginTestBase: XCTestCase {
...
+    static func getBucketFromConfig(forResource: String) throws -> String {
+        guard let path = Bundle.main.path(forResource: forResource, ofType: "json") else {
+            throw "Could not get \(forResource).json"
+        }
+        let url = URL(fileURLWithPath: path)
+        let data = try Data(contentsOf: url)
+        let json = try JSONDecoder().decode(JSONValue.self, from: data)
+        guard let bucket = json["storage"]?["plugins"]?["awsS3StoragePlugin"]?["bucket"] else {
+            throw "Could not retrieve bucket from config"
+
```
wooj2: remove extra space

AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/AWSS3StoragePluginAccessLevelTests.swift
lawmicha: depending on what the attacker obtains, couldn't they also just create their own user as well? I thought that if `awsconfiguration.json` and `amplifyconfiguration.json` is never exposed then we are good, and the username and password for this user is specific for the user pool and wouldn't provide any value to the attacker.


if we need to move this out, i can follow a similar pattern and have another file like `credentials.json` that gets pulled down like https://github.com/aws-amplify/aws-sdk-ios/blob/f3967e2e2a648e8a114ef5dd50bef5e075668693/AWSAuthSDK/AWSAuthSDKTestAppUITests/AWSDropInUIUserPoolTests.swift#L24

AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/AWSS3StoragePluginTestBase.swift
lawmicha: haha that was `6MB`, and when i press CMD+B but my CMD button failed me, reverting change

AmplifyTestCommon/Helpers/AuthHelper.swift
lawmicha: i didn't think AWSMobileClient returns anything other than AWSMobileClientErrors, so i was casting to it and extracting the `message` out for a friendly local debugging message

AmplifyTestCommon/Helpers/AuthHelper.swift
lawmicha: updating code to remove force cast

AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/README.md
lawmicha: thanks, updating

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/BlogPostCommentGraphQLWithAPIKey/BlogPostCommentGraphQLWithAPIKeyTests.swift
lawmicha: One was hand crafted without using model gen to make it work with the document based GraphQL requests when building the API plugin and testing the default Blog/Post/Comment schema provided by amplify CLI, and the other is the model generated Post/Comment from the DataStore post/comment schema that was hand crafted.

i think the work i want to capture is to 
1. use the default blog/post/comment schema provided by amplify-cli for model gen (update schema in AmplifyTestCommon)
2. model gen and place the model files in AmplifyTestCommon
3. Remove the hand crafted types in AWSAPICategoryPluginTestCommon.
4. migrate the test cases in `BlogPostCommentGraphQLWithAPIKeyTests` over to GraphQLModelBasedTests. The tests are using the document based APIs, so we should take the models, and the GraphQLDocument helper classes to construct the GraphQLRequest for the document based APIs.
5. `GraphQLModelBasedTests` can probably renamed to something more specific like `BlogPostCommentAPITests`

this would show that using the amplify-cli schema works with model gen, how to call API, using the model based APIs and the document based APIs.

lawmicha: since this PR is mostly to get storage up and running, including some of the refactoring of the common test helper classes over to core, will capture any API work in separate issue: https://github.com/aws-amplify/amplify-ios/issues/263

AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/AWSS3StoragePluginAccessLevelTests.swift
palpatim: We treat things like pool IDs with different sensitivity than credentials like password. E.g., if we log a user pool ID, the config file doesn't need to be exposed. If this class is setting up tests that allow guest access, we need to protect the credentials. A `credentials.json` pattern or similar would be one way to do that, but let's talk about it IRL to make sure we don't end up with a catch-all for "everything that doesn't fit in a config file." :)

AmplifyTestCommon/Helpers/AuthHelper.swift
palpatim: I follow your logic, but it relies a lot of knowledge of the code paths that lead up to this, which in AWSMobileClient is a little fraught right now. Consider a pattern that logs the `message` if it's an AWSMobielClientError and the `localizedDescription` otherwise.

AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/AWSS3StoragePluginAccessLevelTests.swift
lawmicha: discussed with tim and in the current state of things, we'll do a couple of things to make things more secure and less prone to abuse:
1. credentials.json to contain user/pass. this can be stored in the same pattern as the configuration files, ie. `AWSS3StorageIntegrationTests-credentials.json`. 
2. Not as urgent for API plugin but will do the same
3. add `*credentials.json` to .gitignore 
4. bucket policy updated to expire asap, looks like 1 day is the shortest possible.
5. The tests need to be prone to parallel execution. S3 keys defined in the tests should contain some unique identifier. there are many cases where the tests may be run in parallel. Multiple circleCI runs and developer using test account locally (though ideally developer should create their own resources using the readme instructions).

lawmicha: ExpireObjectsRule

AmplifyTestCommon/Helpers/AuthHelper.swift
lawmicha: adding check for AWSMobileClientError, else fatalError with localizedDescription

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/GraphQLWithUserPoolIntegrationTests.swift
```diff
@@ -17,9 +17,23 @@ class GraphQLWithUserPoolIntegrationTests: XCTestCase {
...
+    static var user1: String!
+    static var password: String!
 
     override func setUp() {
         do {
+
+            let credentials = try TestConfigHelper.retrieveCredentials(
+                forResource: GraphQLWithUserPoolIntegrationTests.credentials)
+
+            if credentials["user1"] == nil || credentials["password"] == nil {
```
palpatim: nit: prefer `guard let` when you're trying to get actual values:

```swift
guard let userName = credentials["user1"], let password = credentials["password"] else {...}
```

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/GraphQLWithUserPoolIntegrationTests.swift
```diff
@@ -42,18 +56,19 @@ class GraphQLWithUserPoolIntegrationTests: XCTestCase {
         Amplify.reset()
     }
 
-    let user1 = "user1@GraphQLWithUserPoolIntegrationTests.com"
-    let password = "Abc123@@!!"
+
     // This is a run once function to set up users then use console to verify and run rest of these tests.
     func testSetUpOnce() {
```
palpatim: Since running this by itself isn't sufficient to actually provision the test user, let's remove it altogether in favor of including CLI commands or console instructions in the test setup README.

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/GraphQLWithUserPoolIntegrationTests.swift
```diff
@@ -17,9 +17,23 @@ class GraphQLWithUserPoolIntegrationTests: XCTestCase {
...
     static let amplifyConfiguration = "GraphQLWithUserPoolIntegrationTests-amplifyconfiguration"
     static let awsconfiguration = "GraphQLWithUserPoolIntegrationTests-awsconfiguration"
+    static let credentials = "GraphQLWithUserPoolIntegrationTests-credentials"
+    static var user1: String!
+    static var password: String!
 
     override func setUp() {
         do {
+
+            let credentials = try TestConfigHelper.retrieveCredentials(
```
palpatim: Put this into the static `setUp` since you only have to read the credentials file once

AmplifyPlugins/Storage/AWSS3StoragePluginFunctionalTests/AWSS3StoragePluginTestBase.swift
```diff
@@ -17,11 +17,27 @@ class AWSS3StoragePluginTestBase: XCTestCase {
...
     static let largeDataObject = Data(repeating: 0xff, count: 1_024 * 1_024 * 6) // 6MB
 
+    static var user1: String!
+    static var user2: String!
+    static var password: String!
+
     override func setUp() {
         do {
+
+            let credentials = try TestConfigHelper.retrieveCredentials(forResource: AWSS3StoragePluginTestBase.credentials)
```
palpatim: make this a static set up method?

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/GraphQLWithUserPoolIntegrationTests.swift
lawmicha: updating readme with instructions on setting up a user and removing disabled method

PR #266, opened 2019-12-17
Integrate retryability for outgoing mutation queue

wooj2:
Work-in-progress.  need to find a way to plumb through the hostname from the client config

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
@@ -1,0 +1,77 @@
+//
+// Copyright 2018-2019 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+import Foundation
+import AWSPluginsCore
+import Combine
+
+@available(iOS 13.0, *)
+final class MutationRetryNotifier {
+    private var nextSyncTimer: DispatchSourceTimer?
+    private var handlerQueue = DispatchQueue.global(qos: .default)
+    var retryMutationCallback: () -> Void
+    private var reachabilitySubscription: Subscription?
+
+    init(advice: RequestRetryAdvice,
+         networkReachabilityPublisher: AnyPublisher<ReachabilityUpdate, DataStoreError>?,
+         retryMutationCallback: @escaping BasicClosure) {
+        self.retryMutationCallback = retryMutationCallback
+
+        networkReachabilityPublisher?.subscribe(self)
+
+        let deadline = DispatchTime.now() + advice.retryInterval
+        scheduleTimer(at: deadline)
+    }
+
+    deinit {
+        cancel()
+    }
+
+    private func scheduleTimer(at deadline: DispatchTime) {
+        nextSyncTimer = DispatchSource.makeOneOffDispatchSourceTimer(deadline: deadline, queue: handlerQueue) {
+            self.notifyCallback()
+        }
+        nextSyncTimer?.resume()
+    }
+
+    func cancel() {
+        reachabilitySubscription?.cancel()
+        nextSyncTimer?.cancel()
+    }
+
+    func notifyCallback() {
+        // Call the cancel routine as the purpose of retry is fulfilled
+        cancel()
+        retryMutationCallback()
+    }
+}
+
+@available(iOS 13.0, *)
+extension MutationRetryNotifier: Subscriber {
+    func receive(subscription: Subscription) {
+        log.verbose(#function)
+        reachabilitySubscription = subscription
+        subscription.request(.unlimited)
+    }
+
+    func receive(_ reachabilityUpdate: ReachabilityUpdate) -> Subscribers.Demand {
+        if reachabilityUpdate.isOnline {
+            notifyCallback()
+            return .none
+        }
+        return .unlimited
+    }
+
+    func receive(completion: Subscribers.Completion<DataStoreError>) {
+        log.verbose(#function)
+        reachabilitySubscription?.cancel()
+    }
+}
+
+@available(iOS 13.0, *)
+extension MutationRetryNotifier: DefaultLogger { }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
@@ -13,19 +13,30 @@ import AWSPluginsCore
 /// Publishes a mutation event to the specified Cloud API. Upon receipt of the API response, validates to ensure it is
 /// not a retriable error. If it is, attempts a retry until either success or terminal failure. Upon success or
 /// terminal failure, publishes the event response to the appropriate ModelReconciliationQueue subject.
+@available(iOS 13.0, *)
 class SyncMutationToCloudOperation: Operation {
 
     private weak var api: APICategoryGraphQLBehavior?
     private let mutationEvent: MutationEvent
     private var mutationOperation: GraphQLOperation<MutationSync<AnyModel>>?
+    private let networkReachabilityPublisher: AnyPublisher<ReachabilityUpdate, DataStoreError>?
     private let completion: GraphQLOperation<MutationSync<AnyModel>>.EventListener
+    private var mutationRetryNotifier: MutationRetryNotifier?
+    private var requestRetryablePolicy: RequestRetryablePolicy
+    private var currentAttemptNumber: Int
 
-    init(mutationEvent: MutationEvent, api: APICategoryGraphQLBehavior,
+    init(mutationEvent: MutationEvent,
+         api: APICategoryGraphQLBehavior,
+         networkReachabilityPublisher: AnyPublisher<ReachabilityUpdate, DataStoreError>?,
+         currentAttemptNumber: Int = 1,
+         requestRetryablePolicy: RequestRetryablePolicy? = RequestRetryablePolicy(),
          completion: @escaping GraphQLOperation<MutationSync<AnyModel>>.EventListener) {
         self.mutationEvent = mutationEvent
         self.api = api
+        self.networkReachabilityPublisher = networkReachabilityPublisher
         self.completion = completion
-
+        self.currentAttemptNumber = currentAttemptNumber
+        self.requestRetryablePolicy = requestRetryablePolicy ?? RequestRetryablePolicy()
         super.init()
     }
 
@@ -42,6 +53,13 @@ class SyncMutationToCloudOperation: Operation {
         sendMutationToCloud()
     }
 
+    override func cancel() {
+        mutationOperation?.cancel()
+        mutationRetryNotifier?.cancel()
+        let apiError = APIError.unknown("Operation cancelled", "")
+        finish(result: .failed(apiError))
+    }
+
     private func sendMutationToCloud() {
         guard !isCancelled else {
             mutationOperation?.cancel()
@@ -51,14 +69,6 @@ class SyncMutationToCloudOperation: Operation {
         }
 
         log.debug(#function)
-        guard let api = api else {
-            // TODO: This should be part of our error handling routines
-            log.error("\(#function): API unexpectedly nil")
-            let apiError = APIError.unknown("API unexpectedly nil", "")
-            finish(result: .failed(apiError))
-            return
-        }
-
         guard let mutationType = GraphQLMutationType(rawValue: mutationEvent.mutationType) else {
             let dataStoreError = DataStoreError.decodingError(
                 "Invalid mutation type",
@@ -67,11 +77,19 @@ class SyncMutationToCloudOperation: Operation {
                 match any known GraphQL mutation type. Ensure you only send valid mutation types:
                 \(GraphQLMutationType.allCases)
                 """
-                )
+            )
             log.error(error: dataStoreError)
+            let apiError = APIError.unknown("Invalid mutation type", "", dataStoreError)
+            finish(result: .failed(apiError))
             return
         }
 
+        if let apiRequest = createAPIRequest(mutationType: mutationType) {
+            makeAPIRequest(apiRequest)
+        }
+    }
+
+    func createAPIRequest(mutationType: GraphQLMutationType) -> GraphQLRequest<MutationSync<AnyModel>>? {
         let request: GraphQLRequest<MutationSync<AnyModel>>
         do {
             if mutationType == .delete {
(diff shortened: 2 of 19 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    init(retryAttemptNumber: Int,
+         advice: RequestRetryAdvice,
+         retryMutationCallback: @escaping () -> Void) {
+        self.retryAttemptNumber = retryAttemptNumber
+        self.retryMutationCallback = retryMutationCallback
+
+        NetworkReachabilityNotifier.shared?.publisher().subscribe(self)
+
+        let deadline = DispatchTime.now() + advice.retryInterval!
```
palpatim: Let's set up a sane default rather than force-unwrapping

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
```diff
@@ -0,0 +1,117 @@
...
+    private var retryAttemptNumber: Int
+
+    private var handlerQueue: DispatchQueue = DispatchQueue.global(qos: .utility)
+    var retryMutationCallback: () -> Void
+    private var reachabilitySubscription: Subscription?
+
+    init(retryAttemptNumber: Int,
+         advice: RequestRetryAdvice,
+         retryMutationCallback: @escaping () -> Void) {
+        self.retryAttemptNumber = retryAttemptNumber
```
palpatim: We only use `retryAttemptNumber` for logging, and the function in which it's logged is immediately invoked from the initializer, so no need to maintain this as a property. In fact, I'd argue that there's no reason to have it at all in this class.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
```diff
@@ -51,14 +66,12 @@ class SyncMutationToCloudOperation: Operation {
...
         log.debug(#function)
-        guard let api = api else {
-            // TODO: This should be part of our error handling routines
-            log.error("\(#function): API unexpectedly nil")
-            let apiError = APIError.unknown("API unexpectedly nil", "")
-            finish(result: .failed(apiError))
-            return
+        if let apiRequest = createAPIRequest() {
+            makeAPIRequest(apiRequest)
         }
```
palpatim: If the request is `nil` do we need to recover or propagate an error?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
```diff
@@ -82,11 +95,21 @@ class SyncMutationToCloudOperation: Operation {
...
+        return request
+    }
 
-        log.verbose("\(#function) sending mutation with sync data: \(request)")
-        mutationOperation = api.mutate(request: request) { asyncEvent in
+    func makeAPIRequest(_ apiRequest: GraphQLRequest<MutationSync<AnyModel>>) {
+        guard let api = api else {
+            // TODO: This should be part of our error handling routines
+            log.error("\(#function): API unexpectedly nil")
+            let apiError = APIError.unknown("API unexpectedly nil", "")
```
palpatim: Consider the "nil API" extension of DataStoreError (the actual symbol escapes me for the moment)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
```diff
@@ -139,7 +172,36 @@ class SyncMutationToCloudOperation: Operation {
...
+    }
+
+    private func scheduleRetry(advice: RequestRetryAdvice) {
+        log.verbose("\(#function) scheduling retry for mutation")
+        mutationRetryNotifier = MutationRetryNotifier(retryAttemptNumber: currentAttemptNumber,
+                                                      advice: advice) {
+                                                        self.sendMutationToCloud()
+                                                        self.mutationRetryNotifier = nil
+        }
+        currentAttemptNumber += 1
```
palpatim: I know these mutation operations are enqueued on a serial Operations queue; is any of the internal work being done in a way that makes this not-thread-safe? My quick scan makes me think everything is pretty well serialized, but wanted you to confirm since you're closer to this code.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/NetworkReachabilityNotifier.swift
```diff
@@ -16,14 +16,41 @@ struct ReachabilityUpdate {
 
 @available(iOS 13.0, *)
 class NetworkReachabilityNotifier {
+    private(set) static var shared: NetworkReachabilityNotifier?
```
palpatim: I'm just now realizing that this is singleton pattern isn't going to work if we want reachability notifiers for different APIs (e.g., one each for a GraphQL and a REST endpoint). I think we're going to need these to be straight up instances. We could consider a Factory that maintains cached reachability notifiers per host and creates or returns those instances at need.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
```diff
@@ -0,0 +1,129 @@
...
+import AWSPluginsCore
+import Combine
+
+@available(iOS 13.0, *)
+final class MutationRetryNotifier {
+    private var nextSyncTimer: DispatchSourceTimer?
+    private var retryAttemptNumber: Int
+    private var retryOperationQueue: OperationQueue
+
+    private var handlerQueue: DispatchQueue = DispatchQueue.global(qos: .utility)
```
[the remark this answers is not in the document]
wooj2: Per discussion, removing DispatchQueue all together.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
wooj2: Thanks! I was using it at first to control time-outs, but eventually opted for mocking out the entire retry policy class.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/NetworkReachabilityNotifier.swift
wooj2: Thanks for pointing this out.. I will revert this.

Hmm, i think we still want the API Category to own the NetworkReachability stuff since it's closest to knowing what the endpoint is... but this means we'll either need to do something like expose a Combine publisher to the Datastore category so that our MutationRetryNotifier can subscribe (or something to that effect)

I'll dig into this.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
wooj2: Just to double check, your question revolves around executing multiple SyncMutationToCloudOperations running at the same time such that we could run into a situation where our code is not thread safe?  (not that we'd ever want to do this, because we want to drain the mutation queue one at a time in case any of the data changes)

But... I looked at the code and thought about implications of instance variables if we were to create multiple operations and run them at the same time, and didn't see a situation where we would be overwriting data.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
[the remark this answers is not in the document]
wooj2: copy and pasted .utility-- but i agree, no reason to make this .utility unless it is needed.  Updating to .default and removing DispatchQueue type declaration.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
```diff
@@ -0,0 +1,129 @@
...
+         advice: RequestRetryAdvice,
+         retrySignalCallback: @escaping () -> Void) {
+        self.retryAttemptNumber = retryAttemptNumber
+        self.retrySignalCallback = retrySignalCallback
+
+        self.retryOperationQueue = OperationQueue()
+        retryOperationQueue.maxConcurrentOperationCount = 1
+        retryOperationQueue.name = "com.amazonaws.OutgoingMutationOperationQueue.retryOperationQueue"
+
+        NetworkReachabilityNotifier.shared?.publisher().subscribe(self)
```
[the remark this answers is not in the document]
wooj2: Removing shared entirely from this class.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/NetworkReachabilityNotifier.swift
wooj2: In the short term, I'm going to make the NetworkReachabilityNotifier as an optional -- until we figure out out how to plumb this between API and Datastore.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
wooj2: Tried using this:
```
            let nilAPIError = DataStoreError.nilAPIHandle()
```
but using `.failed` is expecting an APIError, not a Datastore error... leaving this for now.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
wooj2: oup, good call out, for now, lets fail, and if we need to build out a recovery mechanism, we can at a later time.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
wooj2: Sane defaults sounds good.  Our max retry duration is about 5 minutes.. somehow 2.5 minutes seems too long, how about 60 seconds?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/NetworkReachabilityNotifier.swift
```diff
@@ -43,10 +70,10 @@ class NetworkReachabilityNotifier {
     deinit {
         reachability?.stopNotifier()
         NotificationCenter.default.removeObserver(self)
-        reachabilityPublisher.send(completion: Subscribers.Completion<Never>.finished)
+        reachabilityPublisher.send(completion: Subscribers.Completion<DataStoreError>.finished)
     }
 
-    func publisher() -> AnyPublisher<ReachabilityUpdate, Never> {
+    func publisher() -> AnyPublisher<ReachabilityUpdate, DataStoreError> {
```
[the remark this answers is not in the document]
wooj2: done.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
```diff
@@ -0,0 +1,129 @@
...
+    func receive(completion: Subscribers.Completion<DataStoreError>) {
+        log.verbose(#function)
+        reachabilitySubscription?.cancel()
+    }
+}
+
+@available(iOS 13.0, *)
+extension MutationRetryNotifier: DefaultLogger { }
+
+extension DispatchSource {
```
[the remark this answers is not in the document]
wooj2: added to amplify core.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/MutationRetryNotifier.swift
```diff
@@ -0,0 +1,129 @@
...
+        retryOperationQueue.addOperation {
+            // Use of MutationRetryNotifier assumes that retriable has been set to true
+            // and a retryInterval has already been calculated
+            let deadline = DispatchTime.now() + advice.retryInterval!
+            self.scheduleTimer(at: deadline)
+        }
+    }
+
+    deinit {
+        cancel()
```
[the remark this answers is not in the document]
wooj2: will do.

wooj2:
Assuming circle ci tests pass, we should be able to merge these changes into master.  With the following outstanding items to handle:
* Update model to not wake up on 429 errors
* Add tests around deinit for mutation retry notifier
* Dig into errors for reachability
* Migrate reachability into API category and figure out a way to expose the publisher to datastore.

wooj2:
Squashed commits.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
palpatim: No, I'm actually concerned about multiple operations happening on a single operation--e.g., a network callback being invoked at the same time as we're trying to make the outgoing call b/c of a retry.

wooj2: I see... hmmm.. I suppose it is possible that our timer could wake up at the same exact time that we have a receive() through combine (e.g. a network event).

In this case, I think a pragmatic solution could be to introduce something (atomic write to a variable?) in our `notifyCallback` function which allows the "first event" to win.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
```diff
@@ -67,11 +77,19 @@ class SyncMutationToCloudOperation: Operation {
...
+            )
             log.error(error: dataStoreError)
+            let apiError = APIError.unknown("Invalid mutation type", "", dataStoreError)
+            finish(result: .failed(apiError))
             return
         }
 
+        if let apiRequest = createAPIRequest(mutationType: mutationType) {
+            makeAPIRequest(apiRequest)
+        }
```
wooj2: This is where the bug got introduced. Based on the comment of "bubbling up errors" i had introduced (unwittingly not noticing that we were already bubbling up the error):
```
  if let apiRequest = createAPIRequest(mutationType: mutationType) {
            makeAPIRequest(apiRequest)
+  } else {
+        finish(result: .failed(...)
+ }
```
This basically was causing us to call finish twice, which caused the unit tests to crash(because expectation's `fulfill()` calls were being called  more than expected.

So, to remedy this, i made createAPI request simpler by moving out the checking of the graphQLMutation type,  so we can easily see that the function is indeed handling the bubbling up of those errors.

PR #267, opened 2019-12-17
CI pipeline improvements

iartemiev:
This PR adds a fully functional unstable release pipeline. 
Stable still needs some minor tweaks that will be added in a later PR.

I have tested this extensively using a private clone of amplify-ios and a private spec repo. Happy to grant anyone access to these if they want to try it out before merging.

I will create a Quip doc detailing the process. 

Here's an overview:

* When a PR is merged into master the pipeline will release it to CocoaPods trunk under an `unstable` prerelease tag that conforms to https://semver.org/#spec-item-9

* Once deployed, customers will be able to install the latest unstable version by specifying e.g. `pod 'Amplify', '~> 0.9.1-unstable'` in their Podfile

* Versioning example: if the latest "stable" tag is `0.9.0`, the next prerelease will be `0.9.1-unstable.0`, followed by `0.9.1-unstable.1`. The following "stable" version will be `0.9.1` or `0.10.0` or `1.0.0` depending on the analysis of commit messages using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) spec.

* If Conventional Commits are not followed, version is bumped to patch by default

* The pipeline uses [fastlane](https://fastlane.tools/) to automate versioning and releasing pods to CocoaPods Trunk. Utilizing a combination of built-in fastlane actions and custom functionality.

* Each version creates a corresponding git tag. This is necessary for CocoaPods to be able to reference the source code. 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -150,6 +150,25 @@ jobs:
       - store_test_results:
           path: build/reports
   
+  deploy_unstable:
+    <<: *defaults
+    description: deploy pods to trunk
+    steps: 
+      - *restore_repo
+      - restore_gems
+      - check_bundle
+      - run:
+          # without this step, the pipeline hangs during `pod repo update` after pushing (see Fastfile -> :release_pods)
+          # TODO: look into caching this 
+          name: Clone CocoaPods Repo
+          command: |
+            curl https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash -s cf
+      - run:
+          name: Release pods with unstable tag
+          command: bundle exec fastlane unstable
+      - store_artifacts:
+          path: fastlane/report.xml
+
   deploy:
     <<: *defaults
     description: deploy pods to trunk
@@ -157,13 +176,16 @@ jobs:
       - *restore_repo
       - restore_gems
       - check_bundle
-      # - run:
-      #     name: Bump pod versions
-      #     command: bundle exec fastlane bump_podspecs_patch
+      - run:
+          # without this step, the pipeline hangs during `pod repo update` after pushing (see Fastfile -> :release_pods)
+          name: Clone CocoaPods Repo
+          command: |
+            curl https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash -s cf
       - run:
           name: Release pods
-          command: bundle exec fastlane release_pods
-
+          command: bundle exec fastlane release
+      - store_artifacts:
+          path: fastlane/report.xml
 
 workflows:
   build_test_deploy:
@@ -217,11 +239,11 @@ workflows:
           scheme: AWSDataStoreCategoryPlugin
           requires:
             - install_gems
-      - deploy:
+      - deploy_unstable:
           filters:
             branches:
               only:
-                - release
+                - master
           requires:
             - build_test_amplify
             - unit_test_api
--- a/AWSPluginsCore.podspec
+++ b/AWSPluginsCore.podspec
@@ -21,11 +21,13 @@ Pod::Spec.new do |s|
   s.source       = { :git => 'https://github.com/aws-amplify/amplify-ios.git', :tag => s.version}
   
   s.requires_arc = true
+  s.swift_versions = '5.1'
   
+  AMPLIFY_VERSION = '0.9.0'
   AWS_SDK_VERSION = "~> 2.12.2"
 
   s.source_files = 'AmplifyPlugins/Core/AWSPluginsCore/**/*.swift'
-  s.dependency 'Amplify', '0.9.0'
+  s.dependency 'Amplify', AMPLIFY_VERSION
   s.dependency 'AWSMobileClient', AWS_SDK_VERSION
 
 end
--- a/AmplifyTestCommon.podspec
+++ b/AmplifyTestCommon.podspec
@@ -6,26 +6,29 @@
 #  To see working Podspecs in the CocoaPods repo see https://github.com/CocoaPods/Specs/
 #
 
-Pod::Spec.new do |spec|
+Pod::Spec.new do |s|
+  s.name         = "AmplifyTestCommon"
+  s.version      = "0.9.0"
+  s.summary      = "Test resources used by different targets"
+  s.description  = "Provides different test resources and mock methods"
 
-  spec.name         = "AmplifyTestCommon"
-  spec.version      = "0.9.0"
-  spec.summary      = "Test resources used by different targets"
-  spec.description  = "Provides different test resources and mock methods"
-
-  spec.homepage     = "https://aws.amazon.com/amplify/"
-  spec.license      = 'Apache License, Version 2.0'
-  spec.author       = { 'Amazon Web Services' => 'amazonwebservices' }
-  spec.platform     = :ios, '11.0'
-  spec.source       = { :git => 'https://github.com/aws-amplify/amplify-ios.git', :tag => spec.version}
+  s.homepage     = "https://aws.amazon.com/amplify/"
+  s.license      = 'Apache License, Version 2.0'
+  s.author       = { 'Amazon Web Services' => 'amazonwebservices' }
+  s.platform     = :ios, '11.0'
+  s.source       = { :git => 'https://github.com/aws-amplify/amplify-ios.git', :tag => s.version}
   
-  spec.requires_arc = true
+  s.requires_arc   = true
+  s.swift_versions = '5.1'
 
-  spec.source_files = 'AmplifyTestCommon/**/*.swift'
-  spec.dependency 'Amplify', '0.9.0'
+  AMPLIFY_VERSION = '0.9.0'
+  AMPLIFY_PLUGINS_CORE_VERSION = '0.9.0'
 
-  spec.subspec 'AWSPluginsTestCommon' do |subspec|
-    subspec.source_files = 'AmplifyPlugins/Core/AWSPluginsTestCommon/**/*.swift'
-    spec.dependency 'AWSPluginsCore', '0.9.0'
+  s.source_files = 'AmplifyTestCommon/**/*.swift'
+  s.dependency 'Amplify', AMPLIFY_VERSION
+
+  s.subspec 'AWSPluginsTestCommon' do |ss|
+    ss.source_files = 'AmplifyPlugins/Core/AWSPluginsTestCommon/**/*.swift'
+    s.dependency 'AWSPluginsCore', AMPLIFY_PLUGINS_CORE_VERSION
   end
 end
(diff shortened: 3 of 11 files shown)


iartemiev:
For anyone reviewing, the meat of the changes are in the fastlane directory and fastlane/Fastfile is a good place to look for a high-level overview of the process

fastlane/actions/analyze_commits_canary.rb
```diff
@@ -0,0 +1,399 @@
...
+        Actions.lane_context[SharedValues::RELEASE_NEXT_VERSION] = next_version
+        Actions.lane_context[SharedValues::RELEASE_NEXT_PRERELEASE_VERSION] = "#{next_preid}.#{next_pre_ver}"
+
+        success_message = "Next version (#{next_version}) is higher than last version (#{version}). This version should be released."
+        UI.success(success_message) if is_next_version_releasable
+
+        is_next_version_releasable
+      end
+
+      def self.is_codepush_friendly(params)
```
palpatim: As above, this reads like a simple evaluation, but you're also updating `lane_context`. Consider breaking the side effects out

fastlane/actions/analyze_commits_canary.rb
```diff
@@ -0,0 +1,399 @@
...
+        is_codepush_friendly(params)
+        is_next_version_releasable
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Finds a tag of last release and determinates version of next release"
```
palpatim: s/determinates/determines/

fastlane/actions/analyze_commits_canary.rb
```diff
@@ -0,0 +1,399 @@
...
+      end
+
+      def self.details
+        "This action will find a last release tag and analyze all commits since the tag. It uses conventional commits. Every time when commit is marked as fix or feat it will increase patch or minor number (you can setup this default behaviour). After all it will suggest if the version should be released or not."
```
palpatim: nit: consider one of the multiline string syntaxes (e.g., `%q` or heredoc) to make this easier to read without lots of horizontal scrolling

fastlane/actions/analyze_commits_canary.rb
```diff
@@ -0,0 +1,399 @@
...
+      end
+
+      def self.return_value
+        # If your method provides a return value, you can describe here what it does
+        "Returns true if the next version is higher then the last version"
+      end
+
+      def self.authors
+        # So no one will ever forget your contribution to fastlane :) You are awesome btw!
+        ["xotahal"]
```
palpatim: Is this accurate?

fastlane/actions/podspec_constant_versioning.rb
```diff
@@ -0,0 +1,78 @@
...
+    class PodspecConstantVersioningAction < Action
+      def self.run(params)
+        podspec = params[:podspec]
+        constant = params[:constant]
+        value = params[:value]
+        
+        podspec_contents = File.read(podspec)
+
+        # TODO: improve RegEx
+        if !podspec_contents.match(/(?<=#{constant})(\s*=\s*["'].*)/)
```
palpatim: 1. Consider making the regex a separate variable
1. Regexes can get real hairy, real quick. Consider making it a commented regex with the `/x` flag to aid future devs
1. If you adopt a convention of declaring all versions to be updated in a Podspec preamble, ignore this note. Otherwise: this action appears to be limited to updating versions like `AMPLIFY_VERSION = "1.2.3"`. Are you also going to invoke this for updating the podspec's base `s.version` declaration? If so, will `constant` need to be metacharacter escaped since it includes a dot?

fastlane/actions/podspec_constant_versioning.rb
```diff
@@ -0,0 +1,78 @@
...
+          UI.error("#{constant} not present or doesn't have an explicit value in #{podspec}")
+          return
+        end
+
+        new_value = " = '#{value}'"
+        
+        podspec_new_contents = podspec_contents.gsub(/(?<=#{constant})(\s*=\s*["'].*)/, new_value)
```
palpatim: I think this is the same regex as used in the match? If so, definitely want this in a separate variable

fastlane/actions/podspec_constant_versioning.rb
```diff
@@ -0,0 +1,78 @@
...
+        UI.success("successfully modified #{constant} to value #{value} in #{podspec}")
+            
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "This action will modify the value of a constant in your podspec."
```
palpatim: This is more general: it looks to me like it will change the value of any `key=value` pattern from any file?

fastlane/amplify_pods.rb
```diff
@@ -0,0 +1,63 @@
+module AmplifyPods
```
palpatim: How are we keeping these in sync with newly-added targets?

fastlane/helper/semantic_release_helper.rb
```diff
@@ -0,0 +1,97 @@
...
+        command = "git log --pretty='#{pretty}' --reverse #{start}..HEAD"
+        Actions.sh(command, log: false).chomp
+      end
+
+      def self.parse_commit(params)
+        commit_subject = params[:commit_subject].strip
+        commit_body = params[:commit_body]
+        releases = params[:releases]
+        codepush_friendly = params[:codepush_friendly]
+        pattern = /^(docs|fix|feat|chore|style|refactor|perf|test)(\((.*)\))?(!?)\: (.*)/
```
palpatim: 1. As above, consider multiline commented regexes.
1. Why does the colon need to be escaped here?
1. The capture of the scope probably wants to be non-greedy, or else it will fail on commits like `feat(lang): add Chinese (simplified) language`
1. Probably want a `\s*` rather than a hardcoded single space between the colon and the message capture

fastlane/helper/semantic_release_helper.rb
```diff
@@ -0,0 +1,97 @@
...
+        breaking_change_pattern = /BREAKING CHANGES?: (.*)/
+        codepush_pattern = /codepush?: (.*)/
```
palpatim: As above, probably want a `\s*` rather than a hardcoded single space between colon and capture

fastlane/helper/semantic_release_helper.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        matched = commit_subject.match(pattern)
```
palpatim: consider destructuring the match into named variables to make it easier to understand. Alternately, construct your regex using named captures

fastlane/actions/analyze_commits_canary.rb
iartemiev: I forked this guy's plugin to use as a base and then added the canary/prerelease logic to it. I might separate it out into another plugin and just open source it myself.

fastlane/amplify_pods.rb
iartemiev: At the moment, this list would need to get updated any time a target is added

AmplifyTestCommon.podspec
```diff
@@ -6,26 +6,29 @@
...
-  spec.dependency 'Amplify', '0.9.0'
+  s.source_files = 'AmplifyTestCommon/**/*.swift'
+  s.dependency 'Amplify', AMPLIFY_VERSION
 
-  spec.subspec 'AWSPluginsTestCommon' do |subspec|
-    subspec.source_files = 'AmplifyPlugins/Core/AWSPluginsTestCommon/**/*.swift'
-    spec.dependency 'AWSPluginsCore', '0.9.0'
+  s.subspec 'AWSPluginsTestCommon' do |ss|
+    ss.source_files = 'AmplifyPlugins/Core/AWSPluginsTestCommon/**/*.swift'
+    s.dependency 'AWSPluginsCore', AMPLIFY_PLUGINS_CORE_VERSION
```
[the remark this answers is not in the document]
iartemiev: In this case I just changed the variable from `spec` to `s`. So the file previously had it listed as a sepc dep inside the subspec block. I will figure out where it actually belongs

fastlane/Fastfile
```diff
@@ -13,31 +13,108 @@
...
-opt_out_usage
+require_relative './amplify_pods.rb'
 
+opt_out_usage
 default_platform(:ios)
 
-pods = [
-  # "Amplify.podspec",
-  # "AWSPluginsCore.podspec",
-  # "CoreMLPredictionsPlugin.podspec",
```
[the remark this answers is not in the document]
iartemiev: We are. The full list of specs is in fastlane/amplify_pods.rb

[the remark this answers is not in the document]
iartemiev: The reason those were commented out was: during the 0.9.0 deployment, those pods got deployed successfully but we were having issues with Predicitons and Plugins, so we commented these out temporarily, so as not to re-deploy them while we were troubleshooting deployment issues for the remaining pods

.circleci/config.yml
```diff
@@ -150,20 +150,42 @@ jobs:
...
+  deploy_unstable:
+    <<: *defaults
+    description: deploy pods to trunk
+    steps: 
+      - *restore_repo
+      - restore_gems
+      - check_bundle
+      - run:
+          # without this step, the pipeline hangs during `pod repo update` after pushing (see Fastfile -> :release_pods)
+          # TODO: look into caching this
```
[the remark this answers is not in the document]
iartemiev: fastlane essentially just wraps CocoaPods CLI commands, so I don't think it would be a compatibility issue. 

When I was testing this in my private repo/CircleCI account, ci would get stuck typically when trying to push the second spec, `AWSPluginsCore.podspec`. One of the first steps of that would be "Cloning spec repo" and it would hang on that. After adding the step for cloning it in advance, it stopped hanging and would push all specs successfully. 

Maybe this is somehow related to my also using a private spec repo and manually specifying sources... I'm definitely happy to try without the Clone step after we get this merged in.

AWSPluginsCore.podspec
```diff
@@ -21,11 +21,13 @@ Pod::Spec.new do |s|
   s.source       = { :git => 'https://github.com/aws-amplify/amplify-ios.git', :tag => s.version}
   
   s.requires_arc = true
+  s.swift_versions = '5.1'
   
+  AMPLIFY_VERSION = '0.9.0'
```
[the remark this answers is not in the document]
palpatim: Created https://github.com/aws-amplify/amplify-ios/issues/276 to track centralizing version handling. The results of that issue will also be consumed by followups to @wooj2's work on https://github.com/aws-amplify/amplify-ios/pull/273

iartemiev:
Closing this in favor of https://github.com/aws-amplify/amplify-ios/pull/588

PR #268, opened 2019-12-18
Moving reachability over from datastore to api category

wooj2:
Moved reachability out of datastore & updated unit tests.
Need to figure out how to expose a publisher to datastore at this point (or.. pass in the subscriber to the api category)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/MutationQueue/SyncMutationToCloudOperationTests.swift
```diff
@@ -19,12 +19,14 @@ class SyncMutationToCloudOperationTests: XCTestCase {
     let defaultAsyncWaitTimeout = 2.0
     let secondsInADay = 60 * 60 * 24
     var mockAPIPlugin: MockAPICategoryPlugin!
-    var networkReachability: NetworkReachabilityNotifier!
+
+    var reachabilityPublisher: PassthroughSubject<ReachabilityUpdate, DataStoreError>!
```
wooj2: these need to be changed over to APIErrors

wooj2: Looked into reachability errors -- the errors seem to occur during instantiation and attempting to start the reachability subsystem so... In the following Pr:
https://github.com/aws-amplify/amplify-ios/pull/269

I made two changes:
1.  I have coupled the instantiation & starting of notifier into a single success (or failure) criteria -- if they both succeed, we return an instance -- if one of them fails, then we return nil.
2.  Change the PassthroughSubject from emitting errors to Never.

AmplifyPlugins/API/Podfile
```diff
@@ -10,13 +10,15 @@ target 'AWSAPICategoryPlugin' do
 	pod 'Amplify', :path => '../../'
   pod 'AWSPluginsCore', :path => '../../'
   pod "Starscream", "~> 3.0.2"
+  pod "ReachabilitySwift", "~> 5.0.0"
```
lawmicha: also need to update AmplifyPlugins.podspec for consumers if u haven't already
```
s.subspec 'AWSDataStorePlugin' do |ss|
    ss.source_files = 'AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/**/*.swift'
    ss.dependency 'AWSPluginsCore', AMPLIFY_VERSION
    ss.dependency 'SQLite.swift', '~> 0.12.0'
    ss.dependency 'ReachabilitySwift', '~> 5.0.0' // moved to APIPlugin
  end
```

lawmicha: if you have a sample app, would be good to consume the pod again locally and see if it works

Amplify/Categories/API/Reachability/ReachabilityUpdate.swift
```diff
@@ -0,0 +1,15 @@
+//
+// Copyright 2018-2019 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+public struct ReachabilityUpdate {
```
palpatim: Does this gain us anything beyond a simple Bool?

wooj2: For now, you are right, it's just a bool, but since it is crossing the boundary between two categories, I believe that introducing a type that we own/control allows us to make it easier to make changes to the interface.  In this case, we may want to reveal additional information like the connection type (wifi, cell, microwave.. hah, etc.)

AmplifyPlugins/API/Podfile
wooj2: Thanks!  Updated the pod spec, and compiled w/ a sample app to make sure i was able to consume the pod.

Amplify/Categories/API/Reachability/ReachabilityUpdate.swift
palpatim: I'm not convinced; in order to make this an additive change, we'd have to retain the `isOnline` boolean property and add new specifiers for connection type, which would then be redundant with `isOnline`, and we don't yet have a "different connection type" story. 

If this ends up being a public API, I'll push back harder on it during the API review, but for now we can leave it as-is for an internal type just to prove functionality.

PR #269, opened 2019-12-18
Exposed reachability in API category and using for OutgoingMutationQueue retries

wooj2:

### DIFF
--- a/Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+ReachabilityBehavior.swift
+++ b/Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+ReachabilityBehavior.swift
@@ -1,0 +1,20 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Combine
+extension AmplifyAPICategory: APICategoryReachabilityBehavior {
+    @available(iOS 13.0, *)
+    public func reachabilityPublisher(for apiName: String?) throws -> AnyPublisher<ReachabilityUpdate, Never>? {
+        return try plugin.reachabilityPublisher(for: apiName)
+    }
+
+    @available(iOS 13.0, *)
+    public func reachabilityPublisher() throws -> AnyPublisher<ReachabilityUpdate, Never>? {
+        return try plugin.reachabilityPublisher(for: nil)
+    }
+}
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
@@ -1,0 +1,83 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+import Foundation
+import Combine
+import Reachability
+
+extension AWSAPIPlugin {
+    @available(iOS 13.0, *)
+    public func reachabilityPublisher() throws -> AnyPublisher<ReachabilityUpdate, Never>? {
+        return try reachabilityPublisher(for: nil)
+    }
+
+    @available(iOS 13.0, *)
+    public func reachabilityPublisher(for apiName: String?) throws -> AnyPublisher<ReachabilityUpdate, Never>? {
+        let hostName = try determineHostName(apiName: apiName)
+        if let networkReachability = reachabilityMap[hostName] {
+            return networkReachability.publisher
+        }
+        do {
+            let networkReachability = try NetworkReachabilityNotifier(host: hostName,
+                                                                      allowsCellularAccess: true,
+                                                                      reachabilityFactory: Reachability.self)
+            reachabilityMap[hostName] = networkReachability
+            return networkReachability.publisher
+        } catch {
+            Amplify.API.log.error("Unable to initialize NetworkReachabilityNotifier: \(error)")
+            return nil
+        }
+    }
+
+    private func determineHostName(apiName: String?) throws -> String {
+        if let apiName = apiName {
+            guard let baseUrl = self.pluginConfig.endpoints[apiName]?.baseURL,
+                let host = baseUrl.host else {
+                    let error = APIError.invalidConfiguration("Invalid endpoint configuration for \(apiName)",
+                                                              """
+                                                              baseURL does not contain a valid hostname
+                                                              for apiName: \(apiName)
+                                                              """
+                    )
+                    throw error
+            }
+            return host
+        }
+
+        if pluginConfig.endpoints.count > 1 {
+            let error = APIError.invalidConfiguration("Unable to determine which endpoint configuration",
+                                                      """
+                                                      Pass in the apiName to disambiguate between which endpoint
+                                                      you are requesting reachability for
+                                                      """
+            )
+            throw error
+        }
+
+        guard let configEntry = self.pluginConfig.endpoints.first else {
+            let error = APIError.invalidConfiguration("No API configurations found",
+                                                      """
+                                                      Review how the API category is being instantiated and
+                                                      configured.
+                                                      """
+            )
+            throw error
+        }
+
+        guard let host = configEntry.value.baseURL.host else {
+            let error = APIError.invalidConfiguration("Invalid endpoint configuration",
+                """
+                baseURL does not contain a valid hostname
+                """
+            )
+            throw error
+        }
+        return host
+    }
+
+}
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Resettable.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Resettable.swift
@@ -25,6 +25,10 @@ extension AWSAPIPlugin: Resettable {
 
         authService = nil
 
+        if #available(iOS 13.0, *) {
+            reachabilityMap.removeAll()
+        }
+
         subscriptionConnectionFactory = nil
 
         onComplete()
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
@@ -1,5 +1,5 @@
 //
-// Copyright 2018-2019 Amazon.com,
+// Copyright 2018-2020 Amazon.com,
 // Inc. or its affiliates. All Rights Reserved.
 //
 // SPDX-License-Identifier: Apache-2.0
@@ -9,7 +9,6 @@ import Amplify
 import AWSPluginsCore
 
 final public class AWSAPIPlugin: NSObject, APICategoryPlugin {
-
     /// The unique key of the plugin within the API category.
     public var key: PluginKey {
         return "awsAPIPlugin"
@@ -37,6 +36,23 @@ final public class AWSAPIPlugin: NSObject, APICategoryPlugin {
     /// Creating and retrieving connections for subscriptions. This will be instantiated during the configuration phase,
     /// and is clearable by `reset()`. This is implicitly unwrapped to be destroyed when resetting.
     var subscriptionConnectionFactory: SubscriptionConnectionFactory!
+
+    /// The underlying map that retains instances of NetworkReachabilityNotifier.  Using a computed property
+    /// to work around @available for use on stored properties
+    var iReachabilityMap: [String: Any]?
+
+    @available(iOS 13.0, *)
+    var reachabilityMap: [String: NetworkReachabilityNotifier] {
+        get {
+            if iReachabilityMap == nil {
+                iReachabilityMap = [String: NetworkReachabilityNotifier]()
+            }
+            return iReachabilityMap as! [String: NetworkReachabilityNotifier] // swiftlint:disable:this force_cast
+        }
+        set {
+            iReachabilityMap = newValue
+        }
+    }
 
     public init(modelRegistration: AmplifyModelRegistration? = nil,
                 sessionFactory: URLSessionBehaviorFactory? = nil) {
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
@@ -15,14 +15,14 @@ class NetworkReachabilityNotifier {
     private var reachability: NetworkReachabilityProviding?
     private var allowsCellularAccess = true
 
-    let reachabilityPublisher = PassthroughSubject<ReachabilityUpdate, DataStoreError>()
-    var publisher: AnyPublisher<ReachabilityUpdate, DataStoreError> {
+    let reachabilityPublisher = PassthroughSubject<ReachabilityUpdate, Never>()
+    var publisher: AnyPublisher<ReachabilityUpdate, Never> {
         return reachabilityPublisher.eraseToAnyPublisher()
     }
 
     public init(host: String,
                 allowsCellularAccess: Bool,
-                reachabilityFactory: NetworkReachabilityProvidingFactory.Type) {
+                reachabilityFactory: NetworkReachabilityProvidingFactory.Type) throws {
         self.reachability = reachabilityFactory.make(for: host)
         self.allowsCellularAccess = allowsCellularAccess
 
@@ -34,15 +34,14 @@ class NetworkReachabilityNotifier {
         do {
             try reachability?.startNotifier()
         } catch {
-            //TODO: Test effects of inability to start ReachabilitySwift
-            Amplify.DataStore.log.error("Unable to start notifier from ReachabilitySwift")
+            throw error
         }
     }
 
     deinit {
         reachability?.stopNotifier()
         NotificationCenter.default.removeObserver(self)
-        reachabilityPublisher.send(completion: Subscribers.Completion<DataStoreError>.finished)
+        reachabilityPublisher.send(completion: Subscribers.Completion<Never>.finished)
     }
 
     // MARK: - Notifications
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
@@ -1,5 +1,5 @@
 //
-// Copyright 2018-2019 Amazon.com,
+// Copyright 2018-2020 Amazon.com,
 // Inc. or its affiliates. All Rights Reserved.
 //
 // SPDX-License-Identifier: Apache-2.0
@@ -19,7 +19,7 @@ class SyncMutationToCloudOperation: Operation {
     private weak var api: APICategoryGraphQLBehavior?
     private let mutationEvent: MutationEvent
     private var mutationOperation: GraphQLOperation<MutationSync<AnyModel>>?
-    private let networkReachabilityPublisher: AnyPublisher<ReachabilityUpdate, DataStoreError>?
+    private var networkReachabilityPublisher: AnyPublisher<ReachabilityUpdate, Never>?
     private let completion: GraphQLOperation<MutationSync<AnyModel>>.EventListener
     private var mutationRetryNotifier: MutationRetryNotifier?
     private var requestRetryablePolicy: RequestRetryablePolicy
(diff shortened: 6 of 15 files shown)


Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+GraphQLBehavior.swift
```diff
@@ -67,4 +68,8 @@ extension AmplifyAPICategory: APICategoryGraphQLBehavior {
             plugin.subscribe(toAnyModelType: modelType, subscriptionType: subscriptionType, listener: listener)
     }
 
+    @available(iOS 13.0, *)
+    public func reachability(apiName: String, allowsCellularAccess: Bool) -> AnyPublisher<ReachabilityUpdate, Never>? {
```
lawmicha: should this be here in Model based GraphQL operations or something else like APICategoryReachabilityBehavior?

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        do {
+            let networkReachability = try NetworkReachabilityNotifier(host: host,
+                                                                      allowsCellularAccess: allowsCellularAccess,
+                                                                      reachabilityFactory: Reachability.self)
+            reachabilityMap[apiName] = networkReachability
+            return networkReachability.publisher
+        } catch {
+            return nil
```
lawmicha: can we log something here with the local `error` in the catch block? `Amplify.API.log.warn("Failed to create network reachability notifier with \(error))`

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
```diff
@@ -0,0 +1,37 @@
...
+import Reachability
+
+extension AWSAPIPlugin {
+
+    @available(iOS 13.0, *)
+    public func reachability(apiName: String, allowsCellularAccess: Bool) -> AnyPublisher<ReachabilityUpdate, Never>? {
+        guard let baseUrl = self.pluginConfig.endpoints[apiName]?.baseURL,
+            let host = baseUrl.host,
+            var reachabilityMap = self.reachabilityMap else {
+                return nil
```
lawmicha: some questions and suggestion for debugging:
- apiName: where does it come from? what if apiName cannot be found? how does the caller (assuming datastore) know the apiName? suggestion: separate the logic from retrieving the endpoint by apiName out from this line to consolidate logging for missing endpoint due to bad apiname

- guard var reachabilityMap = self.reachabilityMap else log("missing reachability map"), why is reachabilityMap optional?

AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
```diff
@@ -36,13 +36,14 @@ class NetworkReachabilityNotifier {
         } catch {
             //TODO: Test effects of inability to start ReachabilitySwift
             Amplify.DataStore.log.error("Unable to start notifier from ReachabilitySwift")
```
lawmicha: `Amplify.API.log`

AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
```diff
@@ -36,13 +36,14 @@ class NetworkReachabilityNotifier {
...
+            throw error
         }
     }
 
     deinit {
         reachability?.stopNotifier()
         NotificationCenter.default.removeObserver(self)
-        reachabilityPublisher.send(completion: Subscribers.Completion<DataStoreError>.finished)
+        reachabilityPublisher.send(completion: Subscribers.Completion<Never>.finished)
```
lawmicha: DataStoreError to APIError? or is Never correct?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
```diff
@@ -179,6 +181,17 @@ class SyncMutationToCloudOperation: Operation {
         }
     }
 
+    private func resolveReachabilityPublisher(request: GraphQLRequest<MutationSync<AnyModel>>) {
+        if networkReachabilityPublisher == nil {
+            guard let apiName = request.apiName else {
+                return
```
lawmicha: from what i remember, GraphQLRequest can be created without `apiName` and so depending on the context in which API Plugin is called (either on the GraphQL APIS or the Rest APIs), it will traverse through the endpointConfig at runtime to try and infer the single endpoint that the caller is trying to call, so to provide an experience where caller doesn't need to specify the apiName. 

My thought for this guard statement and then returning when it cannot find the apiName: we know datastore calls api plugin currently without apiName and happy path is that there's only one endpoint configured. is there a missing piece here for getting the reachability for apiName = nil?

Amplify/Categories/API/ClientBehavior/APICategoryGraphQLBehavior.swift
```diff
@@ -117,4 +119,12 @@ public protocol APICategoryGraphQLBehavior: class {
                    listener: GraphQLSubscriptionOperation<AnyModel>.EventListener?)
     -> GraphQLSubscriptionOperation<AnyModel>
 
+    /// Attempts to create and start a reachability client for a host that corresponds to the apiName, and then
+    /// returns the associated Publisher which vends ReachabiltyUpdates
+    /// - Parameters:
+    ///   - apiName: The corresponding apiName that maps to the plugin configuration
```
lawmicha: should apiName be optional? all other APIs operate on apiName being optional and then retrieving the single endpoint per endpointType (GraphQL|REST)

palpatim: What about something like `reachabilityPublisher(for apiName: String?)` which would default to "I will derive the API name if you have a single API configured".

We don't want to yet expose `allowsCellularAccess` until we get customer feedback on it. At that time we will need to plumb that decision/configuration through to the configuration of the NSURLSession for REST APIs and GraphQL queries/mutations, and to the WSS layer for GraphQL subscriptions, as well as reachability determinations. (Unless I'm misremembering and we already support this flag in the API category configuration)

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
```diff
@@ -38,6 +37,23 @@ final public class AWSAPIPlugin: NSObject, APICategoryPlugin {
     /// and is clearable by `reset()`. This is implicitly unwrapped to be destroyed when resetting.
     var subscriptionConnectionFactory: SubscriptionConnectionFactory!
 
+    /// The underlying map that retains instances of NetworkReachabilityNotifier.  Using a computed property
+    /// to work around @available for use on stored properties
+    var iReachabilityMap: [String: Any]?
```
palpatim: 1. Prefer `private var _<nameOfVar>` to indicate a shadow var backing a property.
1. But I'm not sure you need it. If you're just defining a map of `[String: SomeProtocol]`, you should be fine, as long as you define `SomeProtocol` to have no availability restrictions at the top level, and only restrict the method:

```swift
protocol NetworkReachabilityNotifier {
    @available(iOS 13.0, *)
    func whatever()
}
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
```diff
@@ -15,14 +15,14 @@ class NetworkReachabilityNotifier {
     private var reachability: NetworkReachabilityProviding?
```
palpatim: Would it add too much complexity to define the notifier behavior as a protocol without availability restrictions, and then expose a `publisher` property `@available(iOS 13.0, *)`? That way it is easy to add some other notification API without availability restrictions when the time comes to support < 13.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
```diff
@@ -179,6 +181,17 @@ class SyncMutationToCloudOperation: Operation {
...
+            }
+            //TODO: Determine where allowsCellularAccess is controlled
```
palpatim: As above, we don't expose the `allowsCellularAccess` flag yet.

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Resettable.swift
```diff
@@ -25,6 +25,10 @@ extension AWSAPIPlugin: Resettable {
 
         authService = nil
 
+        if #available(iOS 13.0, *) {
```
palpatim: See below, I think we can simplify this and do away with the availability check

AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
wooj2: I believe Never is correct.  My thought was that if we can instantiate and start the reachability subsystem, then we do not expect to be sending any errors to the subscribers.

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
wooj2: There's already one in NetworkReachabilityNotifier, but seems to log the error in this class instead... marking as resolved.

Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+GraphQLBehavior.swift
wooj2: Updated.

AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
wooj2: Thanks, updated

AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
wooj2: I agree that ideally the reachability interface wouldn't require iOS 13.0.  I think it might be a bit difficult to get in right away due to the us returning an AnyPublisher.  How about I look into this afterwards?  It doesn't seem like a high priority issue considering that this is currently only needed for the sync engine which also depends on iOS 13.

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
wooj2: reachabilityMap is a var ( and not let) because the compiler tells me that computed properties are not not allowed to be 'let' declarations, and it's a computed property because the compiler doesn't let me use @available on a stored property.

wooj2: if you are referring to the local variable reachabilityMap, then, this variable is a var because we are updating a key/value pair down below.

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
wooj2: 1.  I also prefer _<nameOfVar> , but when I tried to create _<nameOfVar> it gives me an error of:
```
Identifier Name Violation: Variable names should only contain alphanumeric characters: '_ReachabilityMap' (identifier_name)
```
2.  I am unable to follow you with regards to your suggestion with the NetworkReachabilityNotifier.  Lets sync up when we are both around.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
wooj2: done.

AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
wooj2: I tried doing this, but wasn't able to figure out how to work around all the compiler complaints :(. punting.

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Resettable.swift
wooj2: Sorry, I was unable to figure out what you were suggesting.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
```diff
@@ -179,6 +181,17 @@ class SyncMutationToCloudOperation: Operation {
...
     }
 
+    private func resolveReachabilityPublisher(request: GraphQLRequest<MutationSync<AnyModel>>) {
+        if networkReachabilityPublisher == nil {
+            guard let apiName = request.apiName else {
+                return
+            }
+            if let reachability = api as? APICategoryReachabilityBehavior {
+                networkReachabilityPublisher = reachability.reachability(apiName: apiName)
+            }
```
wooj2: Tim mentioned that we could hide behind an interface, it looking through the code, it became unclear on how to accomplish this.  Any suggestions on how to expose this functionality to other categories would be appreciated.

lawmicha: i'm not entirely sure, is it possible to move `APICategoryReachabilityBehavior` over to APIPlugin? APIPlugin can still implement the behavior but developer will not be able to call it from Amplify.API.<method>. they would also have to check that `api as? APIReachabilityBehavior` to get the methods

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
```diff
@@ -38,6 +37,23 @@ final public class AWSAPIPlugin: NSObject, APICategoryPlugin {
...
+
+    @available(iOS 13.0, *)
+    var reachabilityMap: [String: NetworkReachabilityNotifier] {
+        get {
+            if iReachabilityMap == nil {
+                iReachabilityMap = [String: NetworkReachabilityNotifier]()
+            }
+            return iReachabilityMap as! [String: NetworkReachabilityNotifier]
```
wooj2: Not super happy about this, but attempting to work around the whole iOS 13 `@available` compatibility

Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+ReachabilityBehavior.swift
```diff
@@ -0,0 +1,15 @@
...
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Combine
+extension AmplifyAPICategory: APICategoryReachabilityBehavior {
+    @available(iOS 13.0, *)
+    public func reachability(apiName: String?) throws -> AnyPublisher<ReachabilityUpdate, Never>? {
```
palpatim: As noted elsewhere, default this to `nil` so callers can just invoke `Amplify.API.reachability()`

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
palpatim: 1. D'oh! Forgot about SwiftLint not liking underscore-prefixed symbols.

Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+ReachabilityBehavior.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
 import Combine
 extension AmplifyAPICategory: APICategoryReachabilityBehavior {
     @available(iOS 13.0, *)
-    public func reachability(apiName: String?) throws -> AnyPublisher<ReachabilityUpdate, Never>? {
+    public func reachability(apiName: String? = nil) throws -> AnyPublisher<ReachabilityUpdate, Never>? {
```
lawmicha: "Defaulting paramter, but doesn't seem to work" as in, even though you this here, `Amplify.API.reachability()` complains about missing argument?

Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+ReachabilityBehavior.swift
wooj2: sync'ed offline. updated & resolving.

Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+ReachabilityBehavior.swift
```diff
@@ -0,0 +1,20 @@
...
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Combine
+extension AmplifyAPICategory: APICategoryReachabilityBehavior {
+    @available(iOS 13.0, *)
+    public func reachabilityPublisher(for apiName: String? = nil) throws -> AnyPublisher<ReachabilityUpdate, Never>? {
```
palpatim: Now that we have a no-arg version of this method, we shouldn't assign a default value, or else it's hard to reason about which code path gets executed when invoking `reachabilityPublisher()`.

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
```diff
@@ -0,0 +1,83 @@
...
+extension AWSAPIPlugin {
+    @available(iOS 13.0, *)
+    public func reachabilityPublisher() throws -> AnyPublisher<ReachabilityUpdate, Never>? {
+        return try reachabilityPublisher(for: nil)
+    }
+
+    @available(iOS 13.0, *)
+    public func reachabilityPublisher(for apiName: String?) throws -> AnyPublisher<ReachabilityUpdate, Never>? {
+        let hostName = try determineHostName(apiName: apiName)
+        if let networkReachability = self.reachabilityMap[hostName] {
```
palpatim: nit: remove unnecessary `self`

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
```diff
@@ -0,0 +1,83 @@
...
+            return networkReachability.publisher
+        }
+        do {
+            let networkReachability = try NetworkReachabilityNotifier(host: hostName,
+                                                                      allowsCellularAccess: true,
+                                                                      reachabilityFactory: Reachability.self)
+            self.reachabilityMap[hostName] = networkReachability
```
palpatim: nit: remove unnecessary `self`

AmplifyTestCommon/Mocks/MockAPICategoryPlugin.swift
```diff
@@ -178,6 +178,16 @@ class MockAPICategoryPlugin: MessageReporter, APICategoryPlugin {
             return operation
     }
 
+    @available(iOS 13.0, *)
+    public func reachabilityPublisher(for apiName: String? = nil) -> AnyPublisher<ReachabilityUpdate, Never>? {
```
palpatim: As above, remove default arg now that we have a no-arg method

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
wooj2: yea, you're right, i was being paranoid here since i got bit by `if let reachabilityMap = self.reachabilityMap` not actually storing an object in the instance map..  but yeah.. will update.

AmplifyTestCommon/Mocks/MockAPICategoryPlugin.swift
wooj2: thanks.

PR #272, opened 2019-12-19
Adding predicate to mutate for GraphQL condition input

lawmicha:
### Details
- This exposes the ability to add a condition input for callers using Model based GraphQL APIs from APIPlugin. (Model based GraphQL APIs still need to be reviewed, this change is mostly POC/expose the capability/unblock existing preview features)
- The work on the Model-to-GraphQLDocument translators  is pre-req for DataStore Delete API with predicate, DataStore Save API with predicate.

### Major changes
- GraphQLDocument helpers have now been refactored to a new class called SyncEnabledGraphQLDocument. It is a decorator class which holds a GraphQLDocument.
- Add condition input to all mutations, exposed to the client as a predicate on the mutation. 

### Minor changes
- Remove AnyModel based GraphQL APIs as datastore is now constructing the requests for Document based GraphQL APIs with AnyModel as the responseType. 

- Updated decoding logic to send error with [GraphQLError] instead of transformation error with raw response when errors can be decoded but data cannot be.

- Removed decodePath from GrpahQLDocument helpers and rely on caller to decide what the decode path is since they are also passing in responseType. Updated most callers to use `document.name`


AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/GraphQLResponseDecoder.swift
```diff
@@ -53,12 +53,17 @@ struct GraphQLResponseDecoder {
...
+
+                do {
+                    let jsonValue = JSONValue.object(data)
+                    let responseData = try decode(graphQLData: jsonValue,
+                                                  into: responseType,
+                                                  at: decodePath)
+                    return GraphQLResponse<R>.failure(.partial(responseData, responseErrors))
+                } catch let decodingError as DecodingError {
+                    return GraphQLResponse<R>.failure(.error(responseErrors))
+                }
```
wooj2: I think it make the code more readable if we avoided nested do/catch statements... what do you think?

Amplify/Categories/API/ClientBehavior/APICategoryGraphQLBehavior.swift
```diff
@@ -39,10 +39,12 @@ public protocol APICategoryGraphQLBehavior: class {
     ///
     /// - Parameters:
     ///   - model: The instance of the `Model`.
+    ///   - predicate: the condition that specifies whether the request should succeed or not
     ///   - type: The type of mutation to apply on the instance of `Model`.
     ///   - listener: The event listener for the operation
     /// - Returns: The AmplifyOperation being enqueued.
     func mutate<M: Model>(of model: M,
+                          where predicate: QueryPredicate?,
```
wooj2: In some of the unit tests, we are passing in nil --maybe it would be easier for customers to default this parameter to nil?

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+GraphQLModelBehavior.swift
```diff
@@ -12,7 +12,7 @@ public extension AWSAPIPlugin {
     func query<M: Model>(from modelType: M.Type,
                          byId id: String,
                          listener: GraphQLOperation<M?>.EventListener?) -> GraphQLOperation<M?> {
-        let request = GraphQLRequest<M>.query(from: modelType, byId: id)
+        let request = GraphQLRequest<M?>.query(from: modelType, byId: id)
```
wooj2: Sorry, need to ask.. what does adding a `?` to the type here do, and why are we adding it?

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
```diff
@@ -294,6 +294,63 @@ class GraphQLModelBasedTests: XCTestCase {
...
+                               content: createdPost.content,
+                               createdAt: createdPost.createdAt)
+        let post = Post.keys
+        var predicate = post.title == "incorrectPreviousTitle"
+        let completedWithErrorInvoked = expectation(description: "request completed with error")
+        _ = Amplify.API.mutate(of: updatedPost, where: predicate, type: .update, listener: { event in
+            switch event {
+            case .completed(let data):
+                switch data {
+                case .success(let post):
```
wooj2: Looks like we have one test which tests the .failure case, and the .success case.. i think it would be better to separate this to two separate test cases.

Amplify/Categories/API/ClientBehavior/APICategoryGraphQLBehavior.swift
palpatim: > In some of the unit tests, we are passing in nil --maybe it would be easier for customers to default this parameter to nil?

We'd have to declare defaults in the implementation; protocol declarations can't have default values assigned in method signatures

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+GraphQLModelBehavior.swift
palpatim: > Sorry, need to ask.. what does adding a `?` to the type here do, and why are we adding it?

1. "What": GraphQLRequest is a generic type whose type arg specifies the decoded response type. For model-based GraphQL APIs, we can do things like `GraphQLRequest<Post>` to specify that the request's response should be decoded into a single `Post` object, or `GraphQLRequest<[Post]>` to specify an array of `Post` objects, etc. In this case, `GraphQLRequest<M?>` says the response type is an Optional `M`, meaning the query result could be either an `M` or nil. 
1. "Why": A query isn't guaranteed to have a result. For "byId" queries, we know that a valid result will only ever have zero or one object, so we can specify the response decodes to an `Optional<M>` to accurately reflect that expectation. (In contrast, we can't predict how many results a predicate query might have, so we specify that result as an array. We *could* return `nil` for a no-result case, but instead we specify a return type of `[M]` and return an empty array, making it convenient to do things like `for foo in result { }` without having to guard for nil results.

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/GraphQLResponseDecoder.swift
palpatim: +1. I'd also favor breaking out the handling into a separate method after it grows beyond about 3 lines.

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/Mutation/GraphQLDocument+CreateMutation.swift
```diff
@@ -8,48 +8,52 @@
 import Amplify
 import Foundation
 
-public extension GraphQLMutation {
-    convenience init(of anyModel: AnyModel, type mutationType: GraphQLMutationType) {
-        self.init(of: anyModel.instance, type: mutationType)
+public extension GraphQLCreateMutation {
+    convenience init(of anyModel: AnyModel,
```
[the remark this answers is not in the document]
lawmicha: good question, i just tried to do this on the AppSync console:
```
mutation createPost($input:CreatePostInput!, $condition: ModelPostConditionInput) {
  createPost(input: $input, condition: $condition){
    content
    id
    title
  }
}

{
  "input": {
    "title": "title",
    "content": "content",
    "createdAt": "2011-11-11T11:11:11.111Z"
  },
  "condition": {
    "title": {
      "eq": "title"
    }
  }
}
```
 I can't get pass the 400 ConditionalCheckFailedException. What I think we should do here is:
- This GraphQLCreateMutation should not take in a predicate to avoid confusion on what the use case is ("conditionally update on an update/delete mutation").
- Model based GraphQL API's mutate takes in the predicate and mutation type. We can fail fast by doing client side validation.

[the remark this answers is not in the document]
palpatim: We need to go a bit deeper--why is the transformer generating create mutations with conditionals? Either it's a bug in the transformer, or there's a valid use case we're not catching. If the former, we need to get the CLI team to address. If the latter, we'll need to understand and support the use case.

Based on the [DDB docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html), and that fact that we're already explicitly setting a `attribute_not_exists` condition in the transformer-generated `onCreate*` resolvers, my belief is that conditional creates are not supported, and the transformer is incorrectly generating a mutation that allows a conditional input.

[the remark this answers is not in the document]
lawmicha: thanks for looking into the DDB docs, will confirm that this is a bug with AppSync service team. If it is, then we should for sure do client validation or do not expose a way to do a create mutation with a condition (depending on how the APIs are structured). For example, mutation type can be an Enum with create, update(condition), delete(condition). or another way is as i was mentioning earlier, to do fail on the client with validation against a call to the Model-based API mutation with create mutation type and passing in the condition.

lawmicha:
> I'm less happy with the `syncEnabled` and `syncEnabledVersion` arguments every time I see them. Sync and conflict resolution should be:
> 
> * Separate concerns from the models themselves, as they are metadata about the operation and model, not part of the model
> * Generic, such that some other provider could adopt a completely different strategy having nothing to do with versions lastChanged dates
> 
> I think we need to revisit this design.

thanks for the suggestions, i'm working on outlining a design of these classes and will review it with you guys later this week.

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/GraphQLResponseDecoder.swift
lawmicha: thanks, will do another pass on refactoring this

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
lawmicha: updating

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/Mutation/GraphQLDocument+CreateMutation.swift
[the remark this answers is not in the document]
lawmicha: re-read this comment and realized, this is amplify-cli transformation logic, which isn't AppSync service team's issue. Opening issue over in amplify-CLI repo https://github.com/aws-amplify/amplify-cli/issues/3146

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/GraphQLResponseDecoder.swift
lawmicha: updated code, pulled out the nested try-catch into sequential logic so no longer a nested try-catch block. Couldn't think of a quick way it out into separate methods and name them accurately and will leave that for another time (still have TODO in that file to do separate methods)

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+GraphQLModelBehavior.swift
wooj2: +100, thanks tim.

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/Mutation/GraphQLDocument+CreateMutation.swift
```diff
@@ -0,0 +1,57 @@
...
+public class GraphQLCreateMutation: GraphQLDocument {
+
+    public let documentType = GraphQLDocumentType.mutation
+    public let mutationType = GraphQLMutationType.create
+    public let model: Model
+    public let modelType: Model.Type
+    public let predicate: QueryPredicate?
+
+    public init(of model: Model,
+                where predicate: QueryPredicate? = nil) {
```
palpatim: I think we've agreed that there is no valid use case for predicates in Create mutations, let's remove it here

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/Query/GraphQLDocument+ListQuery.swift
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2018-2019 Amazon.com,
+// Copyright 2018-2020 Amazon.com,
```
palpatim: No good way to comment on L14: Should this class actually be public? Wouldn't the customer be doing a request based or mode-based query rather than creating documents? Or is it public so the DataStore category can create documents directly?

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/Query/GraphQLDocument+SyncQuery.swift
```diff
@@ -0,0 +1,82 @@
...
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+import Foundation
+
+public class GraphQLSyncQuery: GraphQLDocument {
```
palpatim: As above: Should this class be public?

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/SyncEnabledGraphQLDocument.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public var inputParameters: String? {
+        graphQLDocument.inputParameters
+    }
+
+    public var variables: [String: Any] {
+        var variables = graphQLDocument.variables
+
+        if let version = version, var input = variables["input"] as? [String: Any] {
```
palpatim: What, if any, guarantees do we provide around consistency of the schema demanding a `_version` and us supplying it? Is it possible for the Input document generated by the CLI to want a _version but the variable to not supply it? What happens in that case, does the service generate an error?

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/SyncEnabledGraphQLDocument.swift
```diff
@@ -0,0 +1,82 @@
...
+            input.updateValue(version, forKey: "_version")
+            variables["input"] = input
+        }
+
+        return variables
+    }
+
+    public var selectionSetFields: [SelectionSetField] {
+        if graphQLDocument.name.contains("list") {
```
palpatim: Relying on string comparison in document names is a bug waiting to happen. We need a more robust and explicit way of determining that a selection set field returns a `List`, and specifically one of the AWS-defined `List`s that have an `items` and `nextToken` field for pagination.

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
```diff
@@ -24,12 +24,28 @@ extension GraphQLRequest {
...
-                                 variables: document.variables,
-                                 responseType: M.self,
-                                 decodePath: document.decodePath)
+        switch type {
+        case .create:
+            let document = GraphQLCreateMutation(of: model, where: predicate)
+            return GraphQLRequest<M>(document: document.stringValue,
+                                     variables: document.variables,
+                                     responseType: M.self,
+                                     decodePath: document.name)
```
palpatim: If all you do is vary the `document` type in the `switch` cases, move the duplicated `return GraphQLRequest` statements outside of the switch

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
```diff
@@ -64,10 +80,11 @@ extension GraphQLRequest {
     public static func query<M: Model>(from modelType: M.Type,
                                        where predicate: QueryPredicate? = nil) -> GraphQLRequest<[M]> {
         let document = GraphQLListQuery(from: modelType, predicate: predicate)
+        // TODO: decodePath should just be document.name and return `GraphQLRequest<GraphQLListResponse<M>>`
         return GraphQLRequest<[M]>(document: document.stringValue,
                                    variables: document.variables,
-                                   responseType: [M].self,
-                                   decodePath: document.decodePath)
+                                   responseType: [M].self, // TODO: should be `GraphQLListResponse<M>`
```
palpatim: If we don't already have one let's get an issue filed for this so we can refer to it in the release notes/commit logs when we release this incremental update

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelField+GraphQL.swift
```diff
@@ -0,0 +1,46 @@
...
+import Amplify
+import Foundation
+
+/// Extension that adds GraphQL specific utilities to `ModelField`.
+extension ModelField {
+
+    /// The GraphQL name of the field.
+    var graphQLName: String {
+        if isAssociationOwner, case let .belongsTo(_, targetName) = association {
+            return targetName ?? name.pascalCased() + "Id"
```
palpatim: Enhancement for later: We should see if we can centralize name generation of these various documents, fields, input types, etc. There's a lot of string manipulation scattered throughout the graphql support code that is hard to remember and refer to later.

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSetField.swift
```diff
@@ -0,0 +1,34 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+public class SelectionSetField {
```
palpatim: Does this need class semantics (reference, inheritance)

Amplify/Categories/API/ClientBehavior/APICategoryGraphQLBehavior.swift
```diff
@@ -39,10 +39,12 @@ public protocol APICategoryGraphQLBehavior: class {
     ///
     /// - Parameters:
     ///   - model: The instance of the `Model`.
+    ///   - predicate: the condition that specifies whether the request should succeed or not
     ///   - type: The type of mutation to apply on the instance of `Model`.
     ///   - listener: The event listener for the operation
     /// - Returns: The AmplifyOperation being enqueued.
     func mutate<M: Model>(of model: M,
```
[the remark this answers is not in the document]
lawmicha: thanks, updating!

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/SyncEnabledGraphQLDocument.swift
lawmicha: the cases which the developer needs to be aware of is:
1. Create Mutation, Queries, and Subscribe do not need version. Create mutation does allow a version but will fail with internal error response from the service. There shouldn't be a use case where create mutation needs a version, but i think it might just be how the transformer code was written that all mutations including Create result in a PutItem call to DynamoDB. (similar to why a ConditionInput exists on a create mutation)
2. Update Mutation and Delete Mutation contains optional version in the variables. if not passed in, will bump existing version. If passed in, I'm assuming will do conflict resolution on service side

i tried representing this in client side logic but then decided to create a more flexible decorator which simply takes in an optional version, thus leaving it up to the caller to know what they are doing.

- currently the caller needs to know what they are doing to supply or not supply a version on the underlying graphQLDocument instance they are decorating.
- yes, it is possible . currently CLI transforms the schema and wants a _version for create/update/delete mutation. create should never be created with a version. update/delete is optional. 
- only in the create mutation scenario, have i seen an error generated by the service

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/SyncEnabledGraphQLDocument.swift
lawmicha: ah yes, the problem here is if name contains "list" where it was part of the user generated model's name then this all breaks. taking a look

lawmicha: hmm so i basically moved the responsibility of knowing whether it is a paginated selection set to the `[SelectionSet]` extension like `[SelectionSet].paginate()` will return append "items" and "nextToken" and `[SelectionSet].isPaginated() -> Bool` checks if first and last items in the array. may have to chat with you about where you see the custom `List` used

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
lawmicha: oh i see why i did this, i was getting 
`Protocol 'GraphQLDocument' can only be used as a generic constraint because it has Self or associated type requirements`, not sure how to declare the 'document' type

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
lawmicha: added https://github.com/aws-amplify/amplify-ios/issues/291

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelField+GraphQL.swift
lawmicha: Added over in "API Plugin Improvement" Issue to track this https://github.com/aws-amplify/amplify-ios/issues/292

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSetField.swift
lawmicha: no, updating :D

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/SyncEnabledGraphQLDocument.swift
palpatim: Sounds like we still have some sharp edges on this. If not already in our backlog, please add an issue to resolve this with CLI team/transformer, and let's do what we can to reduce cases where a customer might be put into a place where they have to understand the inner workings of our conflict resolution strategies.

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Resettable.swift
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2018-2019 Amazon.com,
+// Copyright 2018-2020 Amazon.com,
```
jamesonwilliams: I would do this in a separate PR, to reduce the size of this API PR, and keep the discussion focused to API changes.

lawmicha:
closing in favour of https://github.com/aws-amplify/amplify-ios/pull/309

PR #273, opened 2019-12-21
Adding amplify specific user agent, revival of PR#166 (271)

wooj2:
This pr is a continuation of the royjit@'s work here:
https://github.com/aws-amplify/amplify-ios/pull/166

* Created (mostly copy and pasted) ServiceConfiguration -> AmplifyAWSServiceConfiguration
* Updated (mostly copy and pasted) Predictions category to use this new class (which should be used when using aws-sdk-ios)
* Updated Amplify API to leverage AmplifyAWSServiceConfiguration
* URLRequestContants.swift -> URLRequestConstants
 * removed placeholder useragent
* Update for Analytics to use AmplifyAWSServiceConfiguration
* Update for Storage to use AmplifyAWSServiceConfiguration

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/RequestInterceptor/APIKeyURLRequestInterceptor.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/RequestInterceptor/APIKeyURLRequestInterceptor.swift
@@ -21,7 +21,10 @@ struct APIKeyURLRequestInterceptor: URLRequestInterceptor {
         var modifiedRequest = request
         let apiKey = apiKeyProvider.getAPIKey()
         modifiedRequest.addValue(apiKey,
-                                 forHTTPHeaderField: URLRequestContants.Header.xApiKey)
+                                 forHTTPHeaderField: URLRequestConstants.Header.xApiKey)
+        modifiedRequest.setValue(AmplifyAWSServiceConfiguration.baseUserAgent(),
+                                 forHTTPHeaderField: URLRequestConstants.Header.userAgent)
+
         return modifiedRequest
     }
 
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/RequestInterceptor/UserPoolRequestInterceptor.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/RequestInterceptor/UserPoolRequestInterceptor.swift
@@ -24,13 +24,12 @@ struct UserPoolURLRequestInterceptor: URLRequestInterceptor {
         guard let mutableRequest = (request as NSURLRequest).mutableCopy() as? NSMutableURLRequest else {
             throw APIError.unknown("Could not get mutable request", "")
         }
-
         mutableRequest.setValue(NSDate().aws_stringValue(AWSDateISO8601DateFormat2),
-                                forHTTPHeaderField: URLRequestContants.Header.xAmzDate)
-        mutableRequest.setValue(URLRequestContants.ContentType.applicationJson,
-                                forHTTPHeaderField: URLRequestContants.Header.contentType)
-        mutableRequest.setValue(URLRequestContants.UserAgent.amplify,
-                                forHTTPHeaderField: URLRequestContants.Header.userAgent)
+                                forHTTPHeaderField: URLRequestConstants.Header.xAmzDate)
+        mutableRequest.setValue(URLRequestConstants.ContentType.applicationJson,
+                                forHTTPHeaderField: URLRequestConstants.Header.contentType)
+        mutableRequest.setValue(AmplifyAWSServiceConfiguration.baseUserAgent(),
+                                forHTTPHeaderField: URLRequestConstants.Header.userAgent)
 
         let tokenResult = userPoolTokenProvider.getToken()
         guard case let .success(token) = tokenResult else {
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Constants/URLRequestConstants.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Constants/URLRequestConstants.swift
@@ -8,7 +8,7 @@
 import Foundation
 
 // TODO: remove this https://github.com/aws-amplify/amplify-ios/issues/75
-struct URLRequestContants {
+struct URLRequestConstants {
     static let appSyncServiceName = "appsync"
 
     struct Header {
@@ -20,9 +20,5 @@ struct URLRequestContants {
 
     struct ContentType {
         static let applicationJson = "application/json"
-    }
-
-    struct UserAgent {
-        static let amplify = "amplify-ios/0.9.0 Amplify"
     }
 }
--- a/AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
@@ -1,0 +1,39 @@
+//
+// Copyright 2018-2019 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import AWSCore
+
+public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
+    override public class func baseUserAgent() -> String! {
+        //TODO: Retrieve this version from a centralized location:
+        //https://github.com/aws-amplify/amplify-ios/issues/276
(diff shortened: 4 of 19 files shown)


AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/RequestInterceptor/UserPoolRequestInterceptor.swift
```diff
@@ -24,13 +24,13 @@ struct UserPoolURLRequestInterceptor: URLRequestInterceptor {
...
-        mutableRequest.setValue(URLRequestContants.ContentType.applicationJson,
-                                forHTTPHeaderField: URLRequestContants.Header.contentType)
-        mutableRequest.setValue(URLRequestContants.UserAgent.amplify,
-                                forHTTPHeaderField: URLRequestContants.Header.userAgent)
+                                forHTTPHeaderField: URLRequestConstants.Header.xAmzDate)
+        mutableRequest.setValue(URLRequestConstants.ContentType.applicationJson,
+                                forHTTPHeaderField: URLRequestConstants.Header.contentType)
+        let serviceConfiguration = AmplifyAWSServiceConfiguration(region: .Unknown, credentialsProvider: nil)
+        mutableRequest.setValue(serviceConfiguration.userAgent,
+                                forHTTPHeaderField: URLRequestConstants.Header.userAgent)
```
lawmicha: This seems a bit odd to me that we have to instantiate something to get a static value, can we just do 
`AmplifyAWSServiceConfiguration.baseUserAgent`?

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Constants/URLRequestConstants.swift
```diff
@@ -8,7 +8,7 @@
 import Foundation
 
 // TODO: remove this https://github.com/aws-amplify/amplify-ios/issues/75
-struct URLRequestContants {
+struct URLRequestConstants {
```
lawmicha: thanks :D

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
```diff
@@ -0,0 +1,37 @@
...
+    override public func copy(with zone: NSZone? = nil) -> Any {
+        return super.copy(with: zone)
+    }
+
+    override init() {
+        super.init(region: .Unknown, credentialsProvider: nil)
+    }
+
+    override public init(region regionType: AWSRegionType,
+                         credentialsProvider: AWSCredentialsProvider!) {
```
lawmicha: why with a `!` here on `AWSCredentialsProvider`?

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
```diff
@@ -0,0 +1,37 @@
...
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import AWSCore
+
+public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
+    static let baseUserAgent: String = {
```
lawmicha: [nit] either in test comment or here, i think it would be nice to have an example of what this actually generates, ie.
```
/// Returns a string representing the user agent like `Amplify-iOS/1.0.0 OS X/13.0 en-US`
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/RequestInterceptor/UserPoolRequestInterceptor.swift
wooj2: Hmm that makes way more sense... thanks updating.

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
wooj2: I noticed that you commented on this before, and I thought long and hard about putting a comment here.  I decided that putting a comment here was more work than it's worth because the code is obvious and (like all comments) require maintenance.  That being said, if you feel strongly about this, i can add a comment.

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
wooj2: to be honest, this was copy and pasted it in from Jithin.. but looking into it, it looks like it's calling super.init() which is calling:
```
- (instancetype)initWithRegion:(AWSRegionType)regionType
           credentialsProvider:(id<AWSCredentialsProvider>)credentialsProvider;
```

That being said, i think you're right in that we can get away with removing it.

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
```diff
@@ -0,0 +1,37 @@
...
+        let version = "0.9.0"
```
palpatim: Let's add a TODO to pick up the version from a central location. @iartemiev is exploring version management as part of their work into Amplify CI/CD.

https://github.com/aws-amplify/amplify-ios/issues/276

AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/RequestInterceptor/APIKeyURLRequestInterceptor.swift
```diff
@@ -21,7 +21,11 @@ struct APIKeyURLRequestInterceptor: URLRequestInterceptor {
         var modifiedRequest = request
         let apiKey = apiKeyProvider.getAPIKey()
         modifiedRequest.addValue(apiKey,
-                                 forHTTPHeaderField: URLRequestContants.Header.xApiKey)
+                                 forHTTPHeaderField: URLRequestConstants.Header.xApiKey)
+        let serviceConfiguration = AmplifyAWSServiceConfiguration(region: .Unknown, credentialsProvider: nil)
```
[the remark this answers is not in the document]
wooj2: thanks for catching this... it has already been updated when i updated per michael's comments.

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
wooj2: sounds good, i'll put a comment that references 276

AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/RequestInterceptor/APIKeyURLRequestInterceptor.swift
```diff
@@ -21,7 +21,10 @@ struct APIKeyURLRequestInterceptor: URLRequestInterceptor {
         var modifiedRequest = request
         let apiKey = apiKeyProvider.getAPIKey()
         modifiedRequest.addValue(apiKey,
-                                 forHTTPHeaderField: URLRequestContants.Header.xApiKey)
+                                 forHTTPHeaderField: URLRequestConstants.Header.xApiKey)
+        modifiedRequest.setValue(AmplifyAWSServiceConfiguration.baseUserAgent(),
```
palpatim: Nit: for methods that take no value, do not throw, and return a value, prefer a computed property

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
palpatim: As noted above, since you've made this a static property, your call sites should just use `AmplifyAWSServiceConfiguration.baseUserAgent` not `AmplifyAWSServiceConfiguration.baseUserAgent()`, unless I'm missing something.

AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/RequestInterceptor/APIKeyURLRequestInterceptor.swift
wooj2: Per offline discussion, baseUserAgent static property was overriding the static function defined in the base class, but this was too confusing, so we are changing it a function w/ the override keyword.

PR #294, opened 2020-01-17
updated readme file for cocoapods

kneekey23:
*Description of changes:*
Updated readme with cocoapod instructions


README.md
```diff
@@ -36,11 +36,43 @@ Amplify requires Xcode 11 or higher to build.
...
+        $ gem install cocoapods
+        $ pod setup
+
+    Depending on your system settings, you may have to use `sudo` for installing `cocoapods` as follows:
+
+        $ sudo gem install cocoapods
+        $ pod setup
+2. In your project directory (the directory where your `*.xcodeproj` file is), type `pod init` and open the Podfile that was created. Add the `Amplify` pod and any plugins you would like to use. Below is an example of what a podfile might look like if you were going to use the Predictions plugin.
+
+        source 'https://github.com/CocoaPods/Specs.git'
```
palpatim: We should be using the CDN rather than the github spec.

README.md
```diff
@@ -36,11 +36,43 @@ Amplify requires Xcode 11 or higher to build.
...
+        
+        platform :ios, '13.0'
```
palpatim: This platform restriction is only necessary for Predictions, or DataStore+Sync. Other customers can use 11.0

kneekey23: yes we should add another example not dependent on 13, will do in follow up PR

README.md
kneekey23: will do in follow up PR

PR #299, opened 2020-01-23
add support for DateTime and Time scalars fixes #278

### LINKED ISSUE #278: AWSDate cannot be used.
The following model fails when an instance is created.

```graphql
type Sample @model {
  id: ID!
  whatever: AWSDate!
}
```

The generated code creates:

```swift
public struct Sample: Model {
  public let id: String
  public var whatever: Date
  
  public init(id: String = UUID().uuidString,
      whatever: Date]) {
      self.id = id
      self.whatever = whatever
  }
}
```
The definition is fine, the serialisation is wrong as pushing this mutation up returns:
```
GraphQLResponseError<Sample>: GraphQL service returned a successful response containing errors: [Amplify.GraphQLError(message: "Variable \'dateOfBirth\' has an invalid value. Unable to parse `2019-12-24T04:15:57.798Z` as a valid date.", locations: Optional([Amplify.GraphQLError.Location(line: 1, column: 26)]), path: nil, extensions: nil)]
```

I think it is either not using an encoder for AWSDate or it is not using the right encoder.

The [documentation](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html) indicates the format should be:

> `1970-01-01Z`, `1970-01-01-07:00` and `1970-01-01+05:30` as valid dates. The time zone offset must either be Z (representing the UTC time zone) or be in the format ±hh:mm:ss. The seconds field in the timezone offset will be considered valid even though it is not part of the ISO 8601 standard.

In any case, checking the `AWSPluginCore` I can see we are using the `iso 8601` string encoding for date and date time ([here](https://github.com/aws-amplify/amplify-ios/blob/master/AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift#L28) but I am not quite sure what changes should I make here. I tried providing a date format similar to `1970-01-01Z` but it only crashes somewhere else with a `APIError: valueNotFound`

Changing the schema to the one below seems to fix the serialisation problem... but this needs a better fix.
 
```graphql
type Sample @model {
  id: ID!
  whatever: AWSDateTime! # changed from AWSDate to AWSDateTime
}
```

drochetti:
**Notes:**

AWS supports more granular definition for Date types while the Swift
language represents all date values with the built-in `Date` struct.

This change adds `DateTime` and `Time` for an improved and more
permissive (de)serialization support.

**References:**

- https://github.com/aws-amplify/amplify-ios/issues/278


Amplify/Categories/DataStore/Model/DataType/DataStoreError+DateScalar.swift
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+extension DataStoreError {
```
raphkim: isn't invalid date technically API error since it applies to AppSync format?

Amplify/Categories/DataStore/Model/DataType/DataStoreError+DateScalar.swift
```diff
@@ -0,0 +1,23 @@
...
+
+    public static func invalidDateFormat(_ value: String) -> DataStoreError {
+        return DataStoreError.decodingError(
+            """
+            Could not parse \(value) as a Date using the ISO8601 format.
+            """,
+            """
+            Check if the format used to parse the date is the correct one. Check
+            `DateScalarFormar` for all the options.
```
raphkim: DateScalarFormat :)

drochetti:
I rebased (instead of merge to avoid the unnecessary merge commit) and I don't understand why GitHub is adding an extra commit that should be in master to my diff. Any advice?

palpatim:
> I rebased (instead of merge to avoid the unnecessary merge commit) and I don't understand why GitHub is adding an extra commit that should be in master to my diff. Any advice?

I don't confirm that it was rebased--at least not completely; when I check out your branch, It looks like it went against release rather than master? Anyway, I checked out your feature branch and did a `git rebase master`, and it worked as expected, with only 2 commits on top of master: https://github.com/aws-amplify/amplify-ios/compare/palpatim/date-scalar-rebase?expand=1

Amplify/Categories/DataStore/Model/DataType/DateTime.swift
```diff
@@ -0,0 +1,66 @@
...
+
+/// `DateTime` is an immutable `DateScalar` object that represents a date with a time,
+/// often viewed as `yyyy-MM-dd'T'HH:mm:ss.SSSSZZZZZ`.
+///
+/// `DateTime` can be represented to nanosecond precision and it also holds a reference
+/// to a TimeZone. As all Date scalars, `DateTime` relies on the ISO8601 calendar.
+public struct DateTime: DateScalar, Comparable {
+
+    public static var iso8601DateComponents: Set<Calendar.Component> {
+        [.year, .month, .day, .hour, .minute, .second, .nanosecond, .timeZone]
```
raphkim: shouldn't this be milliseconds?

raphkim:
We should discuss where DateTime scalars should belong. Personally I think it should be in either API category or core (or wherever models are defined). It is true that DataStore category also uses these models, but the parsing error would result only if a user assigned ISO8601-incompatible type time on AppSync. This error should propagate from API to datastore rather than the other way around.

drochetti:
Closing this pull request since GH messed up the diff after a rebase. I'v learned that GH PRs don't play well with rebases... here's a SO discussion about the subject: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

I'll recreate the PR and go back to merge commits =/

PR #308, opened 2020-02-05
Add mutation events to publisher() interfaces in sync engine components to bubble up status and mutation events

wooj2:
* Compiled hello world app against this code to see mutation events being printed out w/:
```
        noteSubscription = Amplify.DataStore.publisher(for: Note.self)
            .sink(receiveCompletion: {completion in
                if case .failure(let err) =  completion { 
                    print("issue with subscription: \(err)")
                }
            }, receiveValue: { value in
                print("Subscription got this value: \(value)")
        })
```
* Ran unit tests

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -28,8 +28,8 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
     @available(iOS 13.0, *)
     var cancellable: AnyCancellable? {
         get {
-            if let iCancellable = iCancellable {
-                return (iCancellable as! AnyCancellable) //swiftlint:disable:this force_cast
+            if let iCancellable = iCancellable as? AnyCancellable {
+                return iCancellable
             }
             return nil
         }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -23,18 +23,17 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
     /// The local storage provider. Resolved during configuration phase
     var storageEngine: StorageEngineBehavior!
 
-    var iCancellable: Any?
-
+    var iStorageEngineSink: Any?
     @available(iOS 13.0, *)
-    var cancellable: AnyCancellable? {
+    var storageEngineSink: AnyCancellable? {
         get {
-            if let iCancellable = iCancellable as? AnyCancellable {
-                return iCancellable
+            if let iStorageEngineSink = iStorageEngineSink as? AnyCancellable {
+                return iStorageEngineSink
             }
             return nil
         }
         set {
-            iCancellable = newValue
+            iStorageEngineSink = newValue
         }
     }
 
@@ -94,27 +93,39 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
 
         storageEngine = try StorageEngine(isSyncEnabled: isSyncEnabled)
         if #available(iOS 13.0, *) {
-            cancellable = storageEngine.publisher.sink(receiveCompletion: { completion in
-                guard let dataStorePublisher = self.dataStorePublisher as? DataStorePublisher else {
-                    self.log.error("Data store publisher not initalized")
-                    return
-                }
-                switch completion {
-                case .failure(let dataStoreError):
-                    dataStorePublisher.send(dataStoreError: dataStoreError)
-                case .finished:
-                    dataStorePublisher.sendFinished()
-                }
+            setupStorageSink()
+        }
+    }
 
-            }, receiveValue: { event in
-                guard let dataStorePublisher = self.dataStorePublisher as? DataStorePublisher else {
-                    self.log.error("Data store publisher not initalized")
-                    return
-                }
-                if case .mutationEvent(let mutationEvent) = event {
-                    dataStorePublisher.send(input: mutationEvent)
-                }
-            })
+    @available(iOS 13.0, *)
+    private func setupStorageSink() {
+        storageEngineSink = storageEngine.publisher.sink(receiveCompletion: onReceiveCompletion(completed:),
+                                                         receiveValue: onRecieveValue(receiveValue:))
+    }
+
+    @available(iOS 13.0, *)
+    private func onReceiveCompletion(completed: Subscribers.Completion<DataStoreError>) {
+        guard let dataStorePublisher = self.dataStorePublisher as? DataStorePublisher else {
+            log.error("Data store publisher not initalized")
+            return
+        }
+        switch completed {
+        case .failure(let dataStoreError):
+            dataStorePublisher.send(dataStoreError: dataStoreError)
+        case .finished:
+            dataStorePublisher.sendFinished()
+        }
+    }
+
+    @available(iOS 13.0, *)
+    private func onRecieveValue(receiveValue: StorageEngineEvent) {
+        guard let dataStorePublisher = self.dataStorePublisher as? DataStorePublisher else {
+            log.error("Data store publisher not initalized")
+            return
+        }
+
+        if case .mutationEvent(let mutationEvent) = receiveValue {
+            dataStorePublisher.send(input: mutationEvent)
         }
     }
 
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -6,6 +6,7 @@
 //
 
 import Amplify
+import Combine
 
 final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
 
@@ -21,6 +22,21 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
 
     /// The local storage provider. Resolved during configuration phase
     var storageEngine: StorageEngineBehavior!
+
+    var iCancellable: Any?
+
+    @available(iOS 13.0, *)
+    var cancellable: AnyCancellable? {
+        get {
+            if let iCancellable = iCancellable {
+                return (iCancellable as! AnyCancellable) //swiftlint:disable:this force_cast
+            }
+            return nil
+        }
+        set {
+            iCancellable = newValue
+        }
+    }
 
     /// No-argument init that uses defaults for all providers
     public init(modelRegistration: AmplifyModelRegistration) {
@@ -77,6 +93,29 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         }
 
         storageEngine = try StorageEngine(isSyncEnabled: isSyncEnabled)
+        if #available(iOS 13.0, *) {
+            cancellable = storageEngine.publisher.sink(receiveCompletion: { completion in
+                guard let dataStorePublisher = self.dataStorePublisher as? DataStorePublisher else {
+                    self.log.error("Data store publisher not initalized")
+                    return
+                }
+                switch completion {
+                case .failure(let dataStoreError):
+                    dataStorePublisher.send(dataStoreError: dataStoreError)
+                case .finished:
+                    dataStorePublisher.sendFinished()
+                }
+
+            }, receiveValue: { event in
+                guard let dataStorePublisher = self.dataStorePublisher as? DataStorePublisher else {
+                    self.log.error("Data store publisher not initalized")
+                    return
+                }
+                if case .mutationEvent(let mutationEvent) = event {
+                    dataStorePublisher.send(input: mutationEvent)
+                }
+            })
+        }
     }
 
     public func reset(onComplete: @escaping (() -> Void)) {
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
@@ -11,13 +11,46 @@ import Foundation
 import AWSPluginsCore
 
 final class StorageEngine: StorageEngineBehavior {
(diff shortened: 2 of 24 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -22,6 +23,21 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
...
+    var iCancellable: Any?
+
+    @available(iOS 13.0, *)
+    var cancellable: AnyCancellable? {
+        get {
+            if let iCancellable = iCancellable {
+                return (iCancellable as! AnyCancellable) //swiftlint:disable:this force_cast
+            }
+            return nil
+        }
```
royjit: Avoid the force cast?
```
if let iCancellable = iCancellable as? AnyCancellable {
                return iCancellable
 }
return nil
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -11,14 +11,47 @@ import Foundation
...
     private weak var api: APICategoryGraphQLBehavior?
 
+    var iStorageEnginePublisher: Any?
+
+    var iCancellable: Any?
+    @available(iOS 13.0, *)
+    var cancellable: AnyCancellable? {
+        get {
+            if let iCancellable = iCancellable {
+                return (iCancellable as! AnyCancellable) //swiftlint:disable:this force_cast
```
royjit: Same here? Or make iCancellable to `AnyCancellable`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -24,37 +25,56 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
+    }
+
     /// Synchronizes startup operations
     let syncQueue: OperationQueue
 
     // Assigned at `setUpCloudSubscriptions`
     var reconciliationQueue: IncomingEventReconciliationQueue?
+    var reconciliationQueueFactory: IncomingEventReconciliationQueueFactory
 
     /// Initializes the CloudSyncEngine with the specified storageAdapter as the provider for persistence of
```
royjit: Update the comment.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -131,15 +153,36 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
-                                                                  api: api,
-                                                                  storageAdapter: storageAdapter)
+        reconciliationQueue = reconciliationQueueFactory(syncableModelTypes, api, storageAdapter)
+        reconciliationQueueCancellable = reconciliationQueue?.publisher.sink(receiveCompletion: { completionMode in
+            if case .failure(let error) = completionMode {
+                self.remoteSyncTopicPublisher.send(completion: .failure(error))
+            }
+            if case .finished = completionMode {
+                self.log.error("reconciliationQueue was finished.  This is unexpected")
+                let unexpectedFinish = DataStoreError.unknown("ReconcilationQueue finished unexpectedly",
```
royjit: Can we keep these error messages in a constants string file?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -131,15 +153,36 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
-        reconciliationQueue = AWSIncomingEventReconciliationQueue(modelTypes: syncableModelTypes,
-                                                                  api: api,
-                                                                  storageAdapter: storageAdapter)
+        reconciliationQueue = reconciliationQueueFactory(syncableModelTypes, api, storageAdapter)
+        reconciliationQueueCancellable = reconciliationQueue?.publisher.sink(receiveCompletion: { completionMode in
+            if case .failure(let error) = completionMode {
+                self.remoteSyncTopicPublisher.send(completion: .failure(error))
+            }
+            if case .finished = completionMode {
+                self.log.error("reconciliationQueue was finished.  This is unexpected")
```
royjit: Can we add some data/info into the error log that can help in debugging further?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
wooj2: I thought i tried all combinations of this to make this work without a force cast -- but i guess not, thanks!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
wooj2: Updated, thanks!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
wooj2: Hmm..  `reconiliationQueue` is still assigned in setupCloudSubscriptions.  Hmm not exactly sure which comment you are referring to?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
wooj2: The only reason we would hit this case would be if the reconciliationQueue sends a .finished message to the subject. E.g. `modelReconciliationQueueSubject.send(completion: .finished)` -- so really this case is here for completeness for Combine -- we aren't really expecting to hit this code.  I am also hesitant to move these strings to a constant file unless we are using this message in more than 2 places.  In this case, it is a very specific case -- that is- the reconciliationQueue has finished, which is being (only) being managed by the RemoteSyncEngine.  That being said, relooking at the message, i think it makes sense to update message to be clearer for developers.  Thanks for the input.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
royjit: Nvm, I saw the new parameters added. Thought we should mention that in the comments.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
```diff
@@ -10,14 +10,32 @@ import AWSPluginsCore
...
+
 @available(iOS 13.0, *)
 final class AWSIncomingEventReconciliationQueue: IncomingEventReconciliationQueue {
 
+    static let factory: IncomingEventReconciliationQueueFactory = { modelTypes, api, storageAdapter in
+        AWSIncomingEventReconciliationQueue(modelTypes: modelTypes, api: api, storageAdapter: storageAdapter)
+    }
+
+    private let reconciliationQueueTopic: PassthroughSubject<IncomingEventReconciliationQueueEvent, DataStoreError>
+    private var reconciliationQueueTopicCancellable = [String: AnyCancellable]()
```
palpatim: nit: I dislike mixing "declare and init" vs. "only declare" properties. Can we move the init of `reconciliationQueueTopicCancellable` into `init()`? Luckily, since you will declare the type here, the init should be as simple as `self. reconciliationQueueTopicCancellable = [:]`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
```diff
@@ -29,15 +47,31 @@ final class AWSIncomingEventReconciliationQueue: IncomingEventReconciliationQueu
...
+
+            let cancellable = queue.publisher.sink(receiveCompletion: { completed in
+                switch completed {
+                case .failure(let error):
+                    self.reconciliationQueueTopic.send(completion: .failure(error))
+                case .finished:
+                    self.reconciliationQueueTopic.send(completion: .finished)
+                }
+            }, receiveValue: { mutationEvent in
+                if case .mutationEvent(let event) = mutationEvent {
```
palpatim: I think the closure variable is a reconciliationQueueEvent or something, right? It's a bit odd to have to pattern match `.mutationEvent` on a `mutationEvent`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
```diff
@@ -29,15 +47,31 @@ final class AWSIncomingEventReconciliationQueue: IncomingEventReconciliationQueu
                 continue
             }
             reconciliationQueues[modelName] = queue
+
+            let cancellable = queue.publisher.sink(receiveCompletion: { completed in
```
palpatim: nit: I generally dislike (although I'm certainly guilty of it) inlining closure code in the init method. Can we add handlers for the sink completion & value?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -11,14 +11,47 @@ import Foundation
...
 final class StorageEngine: StorageEngineBehavior {
-
     // TODO: Make this private once we get a mutation flow that passes the type of mutation as needed
     let storageAdapter: StorageEngineAdapter
 
     private var syncEngine: RemoteSyncEngineBehavior?
 
     private weak var api: APICategoryGraphQLBehavior?
 
+    var iStorageEnginePublisher: Any?
```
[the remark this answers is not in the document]
wooj2: Agree. done, thanks

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -22,6 +23,21 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
     /// The local storage provider. Resolved during configuration phase
     var storageEngine: StorageEngineBehavior!
 
+    var iCancellable: Any?
+
+    @available(iOS 13.0, *)
+    var cancellable: AnyCancellable? {
```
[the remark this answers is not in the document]
wooj2: Sounds good. I will scrub through my changes and update accordingly.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
wooj2: Agree!  Came up with this.  I think it's pretty decent:
```
let modelReconciliationQueueSink = queue.publisher.sink(receiveCompletion: onReceiveCompletion(completed:),
                                                        receiveValue: onRecieveValue(receiveValue:))
```

then:
```
private func onReceiveCompletion(completed: Subscribers.Completion<DataStoreError>) {
...
}
```
```
private func onRecieveValue(receiveValue: ModelReconciliationQueueEvent) {
...
}
```

way cleaner! thanks!

I will also apply this to other places as well.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -77,6 +93,29 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         }
 
         storageEngine = try StorageEngine(isSyncEnabled: isSyncEnabled)
+        if #available(iOS 13.0, *) {
```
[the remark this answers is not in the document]
wooj2: Abstracted it into a function.  I'm not sure what you mean by invoke it from init(), but assuming this is functionality we may choose to build out later.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
wooj2: done! thanks!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
wooj2: Thanks for pointing this out.. an indeed, it isn't always a mutationEvent!  Perhaps it was cruft left over as i was figuring out how to model this stuff with combine.

During the refactor, i've updated this to:
```
    private func onRecieveValue(receiveValue: ModelReconciliationQueueEvent) {
        if case .mutationEvent(let event) = receiveValue {
            self.eventReconciliationQueueTopic.send(.mutationEvent(event))
        }
    }
```
I believe this should alleviate your concern.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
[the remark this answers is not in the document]
palpatim: "invoke it from init()" is my misremembering that we resolve these members during the configuration phase, not initialization phase. In any case, thanks for reading my mind :)

PR #309, opened 2020-02-05
GraphQLDocument Builder

lawmicha:
This adds a GraphQLDocument Builder class which holds a list of decorators for intercepting the document to decorate it with graphql related components using Model related data

list of decorators
- ConflictResolutionDecorator
- DirectiveDecorator
- ModelDecorator
- ModelIdDecorator
- PredicateDecorator

Usage
```
var documentBuilder = SingleDirectiveGraphQLDocumentBuilder(modelType: Post.self, operationType: .mutation)
documentBuilder.add(decorator: DirectiveDecorator(type: .update))
documentBuilder.add(decorator: ModelDecorator(model: post))
let document = documentBuilder.build()
```

The currently split between APIPlugin's use of GraphQLRequest and DataStore's are implemented in the GraphQLRequest extensions, I tried not to make many changes here other than consolidating the methods places which these methods live


AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/DirectiveDecorator.swift
```diff
@@ -0,0 +1,42 @@
...
+enum GraphQLOperationSubType {
+    case query(GraphQLQueryType)
+    case mutation(GraphQLMutationType)
+    case subscription(GraphQLSubscriptionType)
+}
+
+/// Replaces the directive name of the GraphQL document based on Amplify GraphQL operation types such as "get", "list",
+/// "sync", "create", "update", "delete", "onCreate", "onUpdate", and "onDelete". The GraphQL name is constructed based
+/// on the data from the Model schema and the operation type.
+public struct DirectiveDecorator: SingleDirectiveGraphQLDocumentDecorator {
```
palpatim: If all it's doing is replacing the name, consider renaming this to `DirectiveNameDecorator`

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelDecorator.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+/// Decorate the GraphQL document with the data from an instance of the model. This is added as a single parameter
+/// called "input" that can be referenced by other decorators to append additional document inputs. This decorator
+/// has a dependency on the `DirectiveDecorator` as it constructs the input's type using the document name.
```
palpatim: Where is the dependency on `DirectiveDecorator` used?

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/GraphQLQuery.swift
```diff
@@ -0,0 +1,35 @@
...
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+/// A concrete implementation of `SingleDirectiveGraphQLDocument` that represents a mutation operation.
```
palpatim: s/mutation/query/

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/GraphQLSubscription.swift
```diff
@@ -0,0 +1,35 @@
...
+import Foundation
+
+/// A concrete implementation of `SingleDirectiveGraphQLDocument` that represents a subscription operation.
+public struct GraphQLSubscription: SingleDirectiveGraphQLDocument {
+
+    public init(operationType: GraphQLOperationType,
+                name: String,
+                inputs: [GraphQLParameterName: GraphQLDocumentInput],
+                selectionSet: SelectionSet?) {
+        self.operationType = operationType
```
palpatim: here and above: do we need to init with an OperationType if we're creating a GraphQLSubscription ? Wouldn't the OperationType be inferrable in this init?

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/SingleDirectiveGraphQLDocument.swift
```diff
@@ -0,0 +1,111 @@
...
+
+public typealias GraphQLParameterName = String
+
+/// Represents a single directive GraphQL document. Concrete types that conform to this protocol must
+/// define a valid GraphQL operation document.
+///
+/// This type aims to provide a representation of a simple GraphQL document with its components that can be decorated
+/// and later serialized into query document and variables for a request to a GraphQL service.
+/// Therefore, documents represented by concrete implementations provide a single GraphQL
+/// operation based on a defined `Model.Type`.
```
palpatim: I'm missing how this protocol relies on Model. It seems generic enough to represent any single directive document, even if such is based on something other than a model. However, it might be worth a bit more explanation on the use cases we're trying to solve:

* Single directives rather than multiple directives
* Easily derivable and decoratable inputs and selection sets with standardized output forms, including standardized names
* ...

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/SingleDirectiveGraphQLDocument.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    /// Provides default construction of the graphQL document based on the components of the document.
+    public var stringValue: String {
+
+        var selectionSetString = ""
+        if let selectionSet = selectionSet {
+            selectionSetString = selectionSet.fields.map { $0.toString() }.joined(separator: "\n    ")
+        }
+
+        if !inputs.isEmpty {
```
palpatim: nit: change this to a `guard !inputs.isEmpty else return { your empty-case value below }`

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/SingleDirectiveGraphQLDocumentBuilder.swift
```diff
@@ -0,0 +1,49 @@
...
+            self.document = GraphQLSubscription(modelType: modelType)
+        }
+    }
+
+    public mutating func add(decorator: SingleDirectiveGraphQLDocumentDecorator) {
+        decorators.append(decorator)
+    }
+
+    public mutating func build() -> SingleDirectiveGraphQLDocument {
+        decorators.forEach { decorator in
```
palpatim: 1. Don't mutate `self.document`, or else you'll multi-decorate if someone inadvertently calls `build` multiple times
2. Consider using `reduce` for cases where you're combining elements into a single output:

    ```swift
    let decoratedDocument = decorators.reduce(document) { doc, decorator in
       decorator.decorate(doc, modelType: model.type)
    }
    return decoratedDocument
    ```

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+AnyModelWithSync.swift
```diff
@@ -0,0 +1,98 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+import Foundation
+
+public typealias SyncQueryResult = PaginatedList<AnyModel>
+public typealias MutationSyncResult = MutationSync<AnyModel>
+
+/// Extension methods that are useful for `DataStore`. The methods consist of conflict resolution related fields such
```
palpatim: > Extension methods that are useful for `DataStore`

Wouldn't these belong in DataStore category then, or do you use them in API category as well?

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/SingleDirectiveGraphQLDocumentBuilder.swift
```diff
@@ -0,0 +1,49 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+/// Helps construct a `SingleDirectiveGraphQLDocument`. Collects instances of the decorators and applies the changes
+/// on the document.
+public struct SingleDirectiveGraphQLDocumentBuilder {
```
palpatim: This is the first time we really start relying on a model. It's probably worth calling it out in the name of the type: `ModelBasedGraphQLlDocumentBuilder` or similar. Is `SingleDirective` super important to understanding the point of this type and how it would be used?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/GraphQLDocumentnputValue.swift
```diff
@@ -0,0 +1,28 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// A container to hold either an object or a value, useful for storing document inputs and allowing manipulation at
+/// the first level of the object
+public enum GraphQLDocumentInputValue {
+    case value(Any)
+    case object([String: Any?])
```
palpatim: Is it meaningful to distinguish between `value` and `Any` when `Any` could itself be a `[String: Any?]`? Seems like we should specify we're expecting a scalar value, or a String.

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelField+GraphQL.swift
```diff
@@ -0,0 +1,47 @@
...
+    var graphQLName: String {
+        if isAssociationOwner, case let .belongsTo(_, targetName) = association {
+            return targetName ?? name.pascalCased() + "Id"
+        }
+        return name
+    }
+}
+
+/// Extension that translate an array of model fields into `SelectionSet` structure.
+extension Array where Element == ModelField {
```
palpatim: We may be using this pattern elsewhere, but I'm having second thoughts about it. It feels odd to have something so local applied as an extension to a standard type like Array, even with the `ModelField` condition. Consider rewriting these as helper methods on `ModelField`, like

```swift
extension ModelField {
    static func selectionSet(for fields: [ModelField]) -> SelectionSet { ... }
    static func selectionSetFields(for fields: [ModelField]) -> [SelectionSetField] { ... }
}
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelSchema+GraphQL.swift
```diff
@@ -11,9 +11,26 @@ import Foundation
 /// Extension that adds GraphQL specific utilities to `ModelSchema`.
 extension ModelSchema {
 
-    /// The GraphQL name of the schema.
-    var graphQLName: String {
-        name
+    /// The GraphQL directive name translated from the GraphQL operation and model schema data
+    func graphQLName(operationSubType: GraphQLOperationSubType) -> String {
```
palpatim: This feels like an appropriate use of an extension: taking a custom type and extending it with more custom behavior that doesn't necessarily represent the core functionality of the type.

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelSchema+GraphQL.swift
```diff
@@ -11,9 +11,26 @@ import Foundation
...
+        let graphQLName: String
+        switch operationSubType {
+        case .mutation(let mutationType):
+            graphQLName = mutationType.rawValue + name
+        case .query(let queryType):
+            switch queryType {
+            case .list:
+                graphQLName = queryType.rawValue + name + "s"
```
palpatim: make a String helper method called `pluralize` (I actually thought we already had one somewhere), to centralize and abstract how we pluralize strings

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSet.swift
```diff
@@ -0,0 +1,111 @@
...
+//
+
+import Foundation
+import Amplify
+
+/// Contains detail about the current selection set. `pagination` indicates that it contains paginated fields.
+/// `modelField` indicates that the values represent the property names from a `ModelSchema`.
+public enum SelectionSetType {
+    case pagination
+    case modelField
```
palpatim: I don't understand why we need these if we've got decorators that manipulate the selection set

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSet.swift
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// A nested structure to contain a list of `SelectionSetField`, each of which may optionally contain another
+/// `SelectionSet`
+public struct SelectionSet {
```
palpatim: I've got some concerns about this structure; I'll ping you offline about it.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
```diff
@@ -86,14 +86,9 @@ final class InitialSyncOperation: AsynchronousOperation {
...
-                                        nextToken: nextToken,
-                                        lastSync: lastSyncTime)
-
-        let request = GraphQLRequest(document: document.stringValue,
-                                     variables: document.variables,
-                                     responseType: SyncQueryResult.self,
-                                     decodePath: document.decodePath)
+        let request = GraphQLRequest<SyncQueryResult>.syncQuery(modelType: modelType,
+                                                                nextToken: nextToken,
+                                                                lastSync: lastSyncTime)
```
palpatim: Much nicer!

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+AnyModelWithSync.swift
lawmicha: currently, API Category has a set of integration tests which uses theses against a conflict resolution enabled backend and so they will also need to move over to DataStore. I think they can go either way (thoughts around decorators being public facing or only extensions should be public), let me know what you think

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelSchema+GraphQL.swift
lawmicha: ended up removing `GraphQLOperationSubType` and creating three extensions with parameters taking in `GraphQLQueryType' `GraphQLMutationType, etc

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelField+GraphQL.swift
lawmicha: I ended up rewriting this as an extension on SelectionSet
```
convenience init(fields: [ModelField])
```
it seems to be a good spot since SelectionSets are always first created with model fields

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/GraphQLDocumentnputValue.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 /// A container to hold either an object or a value, useful for storing document inputs and allowing manipulation at
 /// the first level of the object
 public enum GraphQLDocumentInputValue {
-    case value(Any)
+    case scalarOrString(Any)
```
palpatim: As above: `scalar` should be sufficient--all non-optional types can be safely interpolated via `String(describing:)`. However, what you really want is for the incoming value to be safely representable as a primitive type (as opposed to a nested JSON-like object), rather than relying on an internal `String(describing:)` implementation that may change without warning (hello, NSData!) so you could define some protocol, and some convenience extensions to make it trivial to pass primitives into the enum value:

```swift
public enum GraphQLDocumentInputValue {
    case value(GraphQLDocumentValueRepresentable)
    case object([String: GraphQLDocumentValueRepresentable])
}

public protocol GraphQLDocumentValueRepresentable {
    var graphQLDocumentValue: String { get }
}

extension Bool: GraphQLDocumentValueRepresentable {
    public var graphQLDocumentValue: String {
        return "\(self)"
    }
}

extension Float: GraphQLDocumentValueRepresentable {
    public var graphQLDocumentValue: String {
        return "\(self)"
    }
}

extension Int: GraphQLDocumentValueRepresentable {
    public var graphQLDocumentValue: String {
        return "\(self)"
    }
}

extension String: GraphQLDocumentValueRepresentable {
    public var graphQLDocumentValue: String {
        return self
    }
}

...etc
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSet.swift
```diff
@@ -8,104 +8,102 @@
...
-    case modelField
-}
+public typealias SelectionSet = TreeNode<SelectionSetField>
 
-/// A nested structure to contain a list of `SelectionSetField`, each of which may optionally contain another
-/// `SelectionSet`
-public struct SelectionSet {
-    var fields: [SelectionSetField]
-    var type: SelectionSetType
+public class TreeNode<E> {
```
palpatim: 1. Put `TreeNode` in its own file
1. `TreeNode` is a public class, but its methods and properties are internal. Intentional?
1.  Since we're naming a type, we should probably call it [`Tree` instead of `TreeNode`](https://en.wikipedia.org/wiki/Tree_(data_structure)#Data_type_versus_data_structure)

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSet.swift
```diff
@@ -8,104 +8,102 @@
...
-    public init(fields: [SelectionSetField], type: SelectionSetType) {
-        self.fields = fields
-        self.type = type
+    init(value: E) {
+        self.value = value
     }
 
-    mutating func append(_ selectionSetField: SelectionSetField) {
-        fields.append(selectionSetField)
+    func add(child: TreeNode) {
```
palpatim: This has the non-obvious side-effect of mutating the child. Suggest you rename the method to make it more obvious to `addChild(settingParentOf child:)` or similar, and add a doc comment with a note. It's the right behavior--it's would be cumbersome to require call sites to do this explicitly--but we just want to make sure that it's obvious what's happening when we call the method.

Amplify/Core/Support/String+Casing.swift
```diff
@@ -30,4 +30,9 @@ extension String {
     public func camelCased() -> String {
         return prefix(1).lowercased() + dropFirst()
     }
+
```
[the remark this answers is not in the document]
lawmicha: updating to `String+Extensions` (hmm, couldn't think of something better lol)

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/AnyModelIntegrationTests.swift
```diff
@@ -50,7 +50,7 @@ class AnyModelIntegrationTests: XCTestCase {
 
         let callbackInvoked = expectation(description: "Callback invoked")
         var responseFromOperation: GraphQLResponse<AnyModel>?
-        _ = Amplify.API.mutate(ofAnyModel: anyPost, type: .create) { response in
+        _ = Amplify.API.mutate(of: anyPost, type: .create) { response in
```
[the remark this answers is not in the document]
lawmicha: ah yes, i still have those captured in the previous PR: https://github.com/aws-amplify/amplify-ios/blob/7db239ddb6e303fb9dad876033286f0c0ca8979a/Amplify/Categories/API/ClientBehavior/AmplifyAPICategory%2BGraphQLBehavior.swift

but would opt to do any of these in another PR. The change you see here is due to removing the mutate(ofAnyModel) API from Amplify.API. this set of test is disabled at the moment so i'm not sure if that's even the right call for the test. datastore uses the GraphQLRequest extensions so `AnyModelIntegrationTests` should actualy be removed. we test the sync/anymodel APIs in `GraphQLSyncBasedIntegrationTests`

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
```diff
@@ -98,7 +98,7 @@ extension GraphQLRequest {
         return GraphQLRequest<[M]>(document: document.stringValue,
                                    variables: document.variables,
                                    responseType: [M].self,
-                                   decodePath: document.name)
+                                   decodePath: document.name + ".items")
```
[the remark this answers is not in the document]
lawmicha: we have this bug here for fixing the model based list query API https://github.com/aws-amplify/amplify-ios/issues/291

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSet.swift
lawmicha: updating 1 and 3, for 2, `Tree` is pubic for the `typealias SelectionSet` as SelectionSet is used as a property on the `SingleDirectiveGraphQLDocument` public protocol. Everything else on the Tree is used internally.

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSet.swift
lawmicha: thanks, updating!

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/GraphQLDocumentnputValue.swift
lawmicha: thanks for providing the example! i added this in with a few modifications:

- Added `extension Decimal: GraphQLDocumentValueRepresentable` as we represent Double's as Decimal. from what i recall it something to do with rounding. current we don't have any scalar cases other than int and string, but left this in for now.
- `    case object([String: GraphQLDocumentValueRepresentable])` doesn't quite work since if you take `QueryPredicate` to GraphQL input for example, the value can be infinitely nested. so we're storing strings of strings of values in `[String: Any]`

lawmicha: so digging into this deeper, having `.scalar(GraphQLDocumentValueRepresentable)` seems to work well in this class. Adding it to the `.object[String:GraphQLDocumentValueRepresentable]` case doesn't work as it may contain arrays as well. then it will evolve to something similar to `JSONValue`.

Amplify/Core/Support/Tree.swift
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// A Tree data type with a `value` of some type `E` and `children` subtrees.
+public class Tree<E> {
+    public var value: E
+    public var children: [Tree<E>] = []
+    public weak var parent: Tree<E>?
```
palpatim: Thanks for remembering to make this `weak`

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/GraphQLDocumentnputValue.swift
```diff
@@ -10,19 +10,34 @@ import Foundation
...
+extension Bool: GraphQLDocumentValueRepresentable {
+    public var graphQLDocumentValue: String {
+        return "\(self)"
+    }
+}
 
-    public init(type: String, value: GraphQLDocumentInputValue) {
-        self.type = type
-        self.value = value
+extension Decimal: GraphQLDocumentValueRepresentable {
```
palpatim: I'm surprised we're using Decimal--I thought we had moved over to Float for our codegenned models?

lawmicha: I believe we landed on Double because it can hold up to 64 bit. then i encountered some serialization issues, a Double with value 12.3 is serialized as "12.300000000001", which is why i started using Decimal (https://github.com/aws-amplify/amplify-ios/issues/159). so still not 100% confident in which is the right type to use as looking up more details on Decimal isn't giving me many results to resolve this. Maybe there to use Double and override the serialization logic on Doubles. I think the best test is to write an integration test which has a single model with all of the various types that we support.

I believe Float in AppSync is currently returning .double as the ModelFieldType, which then gets translated to a Decimal and serialized. https://github.com/aws-amplify/amplify-ios/blob/461c3cfd8923fbbb6eb10b60495b0f4594879cf4/Amplify/Categories/DataStore/Model/ModelSchema%2BDefinition.swift#L56

lawmicha: opening an issue track adding test with all types that are supported, may be completed as part of @drochetti 's work or follow up task https://github.com/aws-amplify/amplify-ios/issues/314

PR #312, opened 2020-02-10
Add Rest API with UserPool Integration tests

lawmicha:
This adds integration tests around using APIGateway with Cognito User Pool token authentication. Updated some error handling logic to accurately bubble up the error for "User is not authenticated and tries to make a call against the API", so is reflected in the test.

Currently Amplify-CLI experience does not support creating REST API with cognito use pool authorizier, experience and feature request documented here: https://github.com/aws-amplify/amplify-cli/issues/3390

### DIFF
--- a/AmplifyPlugins/API/Podfile
+++ b/AmplifyPlugins/API/Podfile
@@ -1,7 +1,7 @@
 platform :ios, '11.0'
 use_frameworks!
 
-AWS_SDK_VERSION = "2.12.2"
+AWS_SDK_VERSION = "2.12.6"
 
 pod 'SwiftFormat/CLI'
 pod 'SwiftLint'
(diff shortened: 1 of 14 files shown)


AmplifyPlugins/API/Podfile
```diff
@@ -1,7 +1,7 @@
 platform :ios, '11.0'
 use_frameworks!
 
-AWS_SDK_VERSION = "2.12.2"
+AWS_SDK_VERSION = "2.12.6"
```
palpatim: We're on .7 now

lawmicha: will chat with @iartemiev offline about this, most places are 2.12.6, and a few like analytics is 2.12.2, so possibly there is a script that needs to be updated to catch all podfiles.

PR #313, opened 2020-02-11
Restart remote sync engine on errors emitted from API subscription

wooj2:

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
@@ -32,6 +32,11 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
         return remoteSyncTopicPublisher.eraseToAnyPublisher()
     }
 
+    private var networkReachabilityPublisher: AnyPublisher<ReachabilityUpdate, Never>?
+    private var mutationRetryNotifier: MutationRetryNotifier?
+    private let requestRetryablePolicy: RequestRetryablePolicy
+    private var currentAttemptNumber: Int
+
     /// Synchronizes startup operations
     let syncQueue: OperationQueue
 
@@ -45,7 +50,9 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
     convenience init(storageAdapter: StorageEngineAdapter,
                      outgoingMutationQueue: OutgoingMutationQueueBehavior? = nil,
                      initialSyncOrchestratorFactory: InitialSyncOrchestratorFactory? = nil,
-                     reconciliationQueueFactory: IncomingEventReconciliationQueueFactory? = nil) throws {
+                     reconciliationQueueFactory: IncomingEventReconciliationQueueFactory? = nil,
+                     networkReachabilityPublisher: AnyPublisher<ReachabilityUpdate, Never>? = nil,
+                     requestRetryablePolicy: RequestRetryablePolicy? = nil) throws {
         let mutationDatabaseAdapter = try AWSMutationDatabaseAdapter(storageAdapter: storageAdapter)
         let awsMutationEventPublisher = AWSMutationEventPublisher(eventSource: mutationDatabaseAdapter)
         let outgoingMutationQueue = outgoingMutationQueue ?? OutgoingMutationQueue()
@@ -53,13 +60,16 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
             AWSIncomingEventReconciliationQueue.init(modelTypes:api:storageAdapter:)
         let initialSyncOrchestratorFactory = initialSyncOrchestratorFactory ??
             AWSInitialSyncOrchestrator.init(api:reconciliationQueue:storageAdapter:)
+        let requestRetryablePolicy = requestRetryablePolicy ?? RequestRetryablePolicy()
 
         self.init(storageAdapter: storageAdapter,
                   outgoingMutationQueue: outgoingMutationQueue,
                   mutationEventIngester: mutationDatabaseAdapter,
                   mutationEventPublisher: awsMutationEventPublisher,
                   initialSyncOrchestratorFactory: initialSyncOrchestratorFactory,
-                  reconciliationQueueFactory: reconciliationQueueFactory)
+                  reconciliationQueueFactory: reconciliationQueueFactory,
+                  networkReachabilityPublisher: networkReachabilityPublisher,
+                  requestRetryablePolicy: requestRetryablePolicy)
     }
 
     init(storageAdapter: StorageEngineAdapter,
@@ -67,7 +77,9 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
          mutationEventIngester: MutationEventIngester,
          mutationEventPublisher: MutationEventPublisher,
          initialSyncOrchestratorFactory: @escaping InitialSyncOrchestratorFactory,
-         reconciliationQueueFactory: @escaping IncomingEventReconciliationQueueFactory) {
+         reconciliationQueueFactory: @escaping IncomingEventReconciliationQueueFactory,
+         networkReachabilityPublisher: AnyPublisher<ReachabilityUpdate, Never>?,
+         requestRetryablePolicy: RequestRetryablePolicy) {
         self.storageAdapter = storageAdapter
         self.mutationEventIngester = mutationEventIngester
         self.mutationEventPublisher = mutationEventPublisher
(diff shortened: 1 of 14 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
```diff
@@ -87,6 +91,30 @@ final class AWSIncomingEventReconciliationQueue: IncomingEventReconciliationQueu
...
+
+        if shouldIssueKill {
+            modelReconciliationQueueSinks.values.forEach { $0.cancel() }
+            reconciliationQueues.values.forEach { $0.kill() }
+            reconciliationQueues = [:]
+            modelReconciliationQueueSinks = [:]
+            return true
+        } else {
+            return false
+        }
```
wooj2: Not super proud of this.  I would have liked to receive one and only one message from any downstream recieveCompletion() on any single model reconciliation queue.  Couldn't figure out a simpler way to do this, so just used a simple semaphore to do this only once.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -32,6 +32,11 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
         return remoteSyncTopicPublisher.eraseToAnyPublisher()
     }
 
+    private var networkReachabilityPublisher: AnyPublisher<ReachabilityUpdate, Never>?
+    private var mutationRetryNotifier: MutationRetryNotifier?
```
wooj2: Will rename MutationRetryNotifier to something else in a subsequent PR

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -258,6 +284,51 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
         group.wait()
         onComplete()
     }
+
```
wooj2: Considering putting all of the retryability stuff into a category RemoteSyncEngine+Retryability.swift -- thoughts?

palpatim: I like the idea, especially since it starts to enforce the idea of Retriability as a strategy we can apply to a number of uses, not just RemoteSyncEngine. Maybe in a future PR, so as not to bloat this one?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
palpatim: I may be misunderstanding the problem, but I think what you want is to [`merge()`](https://developer.apple.com/documentation/combine/publisher/3204724-merge) the downstream publishers, and wait for completions in this class' subscriber. The `kill()` method should simply iterate over each queue as you're doing and kill it, but don't cancel the sink. When each reconciliationQueue emits a completion, the merged sink will receive its own completion, which you can then use to trigger the remaining cleanup, including notifying upstreams by emitting your own completion. Sample playground code: https://gist.github.com/palpatim/057856de01a2c0202ef6b7b4c02bff8d

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingSubscriptionEventPublisher.swift
```diff
@@ -15,22 +15,41 @@ import Combine
...
+    private func onReceiveCompletion(receiveCompletion: Subscribers.Completion<DataStoreError>) {
+        subscriptionEventSubject.send(completion: receiveCompletion)
     }
 
+    private func onReceive(receiveValue: MutationSync<AnyModel>) {
+        subscriptionEventSubject.send(.mutationEvent(receiveValue))
+    }
+
+    func kill() {
+        mapperSink?.cancel()
```
palpatim: As above, I think our `kill` contract needs to be asynchronous, and emit a completion on its publisher when it is complete.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingSubscriptionEventPublisher.swift
```diff
@@ -9,7 +9,12 @@ import Amplify
...
 
+enum IncomingSubscriptionEventPublisherEvent {
+    case mutationEvent(MutationSync<AnyModel>)
+}
+
 @available(iOS 13.0, *)
 protocol IncomingSubscriptionEventPublisher {
-    var publisher: AnyPublisher<MutationSync<AnyModel>, DataStoreError> { get }
+    var publisher: AnyPublisher<IncomingSubscriptionEventPublisherEvent, DataStoreError> { get }
+    func kill()
```
palpatim: Let's move `kill()` into a `Killable` interface; or consider just renaming `kill()` to `Cancel` and have these publishers implement `Cancellable`. The contract would be that an invocation of `kill/cancel` is asynchronous, and would cause the receiver to emit a completion event on its publisher when finished disposing of its resources.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
```diff
@@ -92,10 +92,12 @@ final class AWSModelReconciliationQueue: ModelReconciliationQueue {
             .publisher
             .sink(receiveCompletion: { [weak self] completion in
                 self?.receiveCompletion(completion)
-                }, receiveValue: { [weak self] remoteModel in
-                    self?.incomingSubscriptionEventQueue.addOperation(CancelAwareBlockOperation {
-                        self?.enqueue(remoteModel)
-                    })
+                }, receiveValue: { [weak self] recieveValue in
```
palpatim: nit: s/recieveValue/receiveValue/

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
```diff
@@ -141,6 +143,22 @@ final class AWSModelReconciliationQueue: ModelReconciliationQueue {
...
         }
     }
+
+    func kill() {
+        incomingEventsSink?.cancel()
+        incomingEventsSink = nil
+
+        incomingSubscriptionEvents.kill()
+
+        //Enqueue the rest of the items that our subscription downloaded
```
palpatim: I'm not sure this is the right behavior. The data contained in these subscriptions are going to be re-populated on the next startup by the delta/base query, and it's probably more important to exit quickly than wait for incoming mutations to apply.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
wooj2: So the problem this addresses is that one or more reconciliation queues have the ability to send a completion on their publisher, which is received by a subscriber -- in this case, this class: `AWSIncomingEventReconciliationQueue`.  For example, we have 27 persistent websockets (because we have 9 syncable models), and 5 of those websockets fail -- this would mean `AWSIncomingEventReconciliationQueue:: onReceiveCompletion()` would get called 5 times -- meaning `AWSIncomingEventReconciliationQueue:: kill()->Bool` would get called 5 times, which means we'd attempt to kill reconciliation queues that potentially are either already killed, or in the process of being killed.  This has lead to crashes, and (somehow?) memory leaks.  Therefore, I've created this Semaphore which locks out access to other executing threads.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
wooj2: Ah, good point, our initial sync operation should handle that -- I'll change these to cancel all operations.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
wooj2: thanks, updated. resolving.

wooj2:
Closing this PR..Code was migrated into 4 separate PRS:

https://github.com/aws-amplify/amplify-ios/pull/321 - migrating to a state machine
https://github.com/aws-amplify/amplify-ios/pull/322 -added retry logic to sync engine, without canceling downstream async operations
https://github.com/aws-amplify/amplify-ios/pull/323 - added in the cancelling interface
https://github.com/aws-amplify/amplify-ios/pull/324 - cancelling the downstream operations & wait for subscription to init.

PR #315, opened 2020-02-12
Updating SDK versions to 2.12.7

lawmicha:
This updates the AWS_SDK_VERSION to 2.12.7 across all Podfiles. Currently the process is manual and in the future, heard that @kneekey23 may be picking up something down the road to automatically do this. 

### DIFF
--- a/AWSPluginsCore.podspec
+++ b/AWSPluginsCore.podspec
@@ -22,7 +22,7 @@ Pod::Spec.new do |s|
   
   s.requires_arc = true
   
-  AWS_SDK_VERSION = "~> 2.12.6"
+  AWS_SDK_VERSION = "~> 2.12.7"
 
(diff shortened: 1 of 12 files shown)


lawmicha:
> I see we're using "compatible with" (`~> 2.12.7`) in some cases and "exact match" in others. We should be consistent with our usage to avoid unexpected surprises down the road. In particular, the "compatible with" flag means we don't have to manually update the version on patch version upgrades to the SDK, as long as we do a `pod update` and commit the resulting Pods & Podfile.lock

it looks like they are using "~>" in all cases, for example when we define the constant `AWS_SDK_VERSION`, it is being used as "~> AWS_SDK_VERSION"

lawmicha:
this PR is outdated now and we can update to take `~> 2.13.0` to get all patch versions of 2.13. a minor version will last a few months

Issue #246, opened 2019-12-10
Codegen failure: Enum types generates an invalid type definition as enums do not implement Codable


rodrigoelp:
Let's take the following example:

```graphql
enum Gender {
  male,
  female,
  none,
  nonBinary,
  unspecified
}

type Person @model 
  (mutations: { create: "newPerson", update: "updatePerson", delete: null } ) {

  id: ID!
  name: String!
  gender: Gender!
}
```

I would expect this generation to be pretty straight forward, but building this code will generate an error indicating `Person` is not decodable. Checking the generated definition I can see a `Person.swift`, `Person+Extensions.swift` and a `Gender.swift` file without any extension.

Fixing this issue is simple, basically just add the extension file manually and type in:
`extension Gender: Codable { }`... But I would imagine this is the responsibility of the code generator to put in place.

drochetti:
Hi @rodrigoelp, support for codable enums just landed in `master`. We will release Amplify soon along with a new version of the Amplify CLI with the codegen updated. I'll keep this issue open until the release is out.

AnruStander:
Has this been fixed yet? I have the same issue in my project at the moment.

yosuke1985:
Still same issue..
@drochetti When it will be released?

drochetti:
Hey @rodrigoelp @yosuke1985 @AnruStander we just released a stable [v1.0.0](https://github.com/aws-amplify/amplify-ios/releases/tag/v1.0.0).

Enums should work correctly in `1.0.0`. Can you try and let me know if that fixes the issue for you?

---

I appreciate your patience and the fact you tried our unstable preview and provided valuable feedback and insights with your issue report. I hope the stable release fixes things for you.

We also have a brand new Discord server for Amplify related topics. Feel free to hit me up there to talk about your use cases and challenges with the new APIs and docs. https://discord.gg/Px5YVfQ

drochetti:
Closing this issue. Feel free to open a new one if you have problems with the stable version of Amplify.

PR #317, opened 2020-02-13
add enum support to DataStore models fixes #111 and #246

### LINKED ISSUE #111: Support enums in code generated models
Given the following type:

```swift
public struct MutationEvent: Model {
    public enum MutationType: String, Codable {
        case create
        case update
        case delete
    }

    public let id: Identifier
    public let modelName: String
    public let json: String
    public let mutationType: MutationType
    public let createdAt: Date

    public init(id: Identifier = UUID().uuidString,
                modelName: String,
                data: String,
                mutationType: MutationType,
                createdAt: Date = Date()) {
        self.id = id
        self.modelName = modelName
        self.json = data
        self.mutationType = mutationType
        self.createdAt = createdAt
    }
}

extension MutationEvent {
    // MARK: - CodingKeys

    public enum CodingKeys: String, ModelKey {
        case id
        case modelName
        case json
        case mutationType
        case createdAt
    }

    public static let keys = CodingKeys.self

    // MARK: - ModelSchema

    public static let schema = defineSchema { definition in
        let mutation = MutationEvent.keys

        definition.attributes(.isSyncable, .isSystem)

        definition.fields(
            .id(),
            .field(mutation.modelName, is: .required, ofType: .string),
            .field(mutation.json, is: .required, ofType: .string),
            .field(mutation.mutationType,
                   is: .required,
                   ofType: .enum(type: MutationType.self)),
            .field(mutation.createdAt, is: .required, ofType: .dateTime)
        )
    }
}
```

Registering the type as a model fails:

```
2019-11-18 16:19:40.219199-0800 AmplifyTestApp[21061:3952734] Fatal error: Model with name MutationType could not be found.: file /Users/schmelte/src/github-repos/amplify-ios/Amplify/Categories/DataStore/Model/ModelSchema+Definition.swift, line 70
```

It looks like it's not going to be as simple as just detecting the `type` in `ModelField.typeDefinition`, since the incoming `type` on the field is `MutationType`, and I don't see any path to find that it's actually an `enum`.

drochetti:
**Notes:**

- introduced a `PersistentEnum` protocol to indicate enums that can be persisted
  - they are `Codable`
  - must provide a `rawValue` String
- changed `ModelRegistry` to accept enum registration
- updated the DataStore plugin to handle `PersistentEnum`

### DIFF
--- a/Amplify/Categories/DataStore/Model/Model+Enum.swift
+++ b/Amplify/Categories/DataStore/Model/Model+Enum.swift
@@ -1,0 +1,32 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// Protocol that represents a `Codable` Enum that can be persisted and easily
+/// integrate with remote APIs since it must have a raw `String` value.
+///
+/// That means only simple enums (i.e. the ones that don't have arguments) can be used
+/// as model properties.
+///
+/// - Example:
+///
+/// ```swift
+/// public enum PostStatus: String, PersistentEnum {
+///     case draft
+///     case published
+/// }
+/// ```
+public protocol PersistentEnum: Codable {
+
+    /// The `String` representation. Make your enum conform to `String`
+    /// to get an automatic `rawValue`.
+    /// See the [Enumeration Language Guide](https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html#ID149)
+    /// for details on enum with raw values.
+    var rawValue: String { get }
+
+}
--- a/Amplify/Categories/DataStore/Model/Model+Enum.swift
+++ b/Amplify/Categories/DataStore/Model/Model+Enum.swift
@@ -23,8 +23,7 @@ import Foundation
 /// ```
 public protocol PersistentEnum: Codable {
 
-    /// The `String` representation. Make your enum conform to `String`
-    /// to get an automatic `rawValue`.
+    /// The `String` representation. Make your enum conform to `String` to get an automatic `rawValue`.
     /// See the [Enumeration Language Guide](https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html#ID149)
     /// for details on enum with raw values.
     var rawValue: String { get }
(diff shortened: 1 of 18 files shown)


AmplifyTestCommon/Models/schema.graphql
```diff
@@ -15,3 +16,9 @@ type Comment @model {
     createdAt: AWSDateTime!
     post: Post @connection(name: "PostComment")
 }
+
+enum PostStatus {
+    deleted
```
lawmicha: Since this is updated, will have to update the provisioned backend for the API test as well

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/ModelSchema+SQLite.swift
```diff
@@ -116,6 +116,8 @@ extension ModelField: SQLColumn {
                 // values we accept, but always output the same format.
                 return value.iso8601Date?.iso8601String
             }
+        case .enum:
+            return binding as? String
```
lawmicha: Synced with Daniel on this, opening issue to refactor some of this conversion logic into some converter class to be used to take model field types and values into SQLite and GraphQL components

Amplify/Categories/DataStore/Model/Model+Enum.swift
```diff
@@ -0,0 +1,32 @@
...
+///
+/// - Example:
+///
+/// ```swift
+/// public enum PostStatus: String, PersistentEnum {
+///     case draft
+///     case published
+/// }
+/// ```
+public protocol PersistentEnum: Codable {
```
drochetti: naming: should this be called `PersistableEnum` to follow a more "Swifty" naming pattern?

AmplifyTestCommon/Models/schema.graphql
lawmicha: i plan on pulling this PR locally and updating the API integration tests which will use the new Post model with an updated provisioned backend, so we can push the changes in one go without having to follow up on another PR. i think this will be a good test to ensure that this case functions in with a real backend

drochetti:
Please, look at this PR instead: https://github.com/aws-amplify/amplify-ios/pull/334

PR #319, opened 2020-02-16
[WIP] Remote sync revamp

wooj2:
Don't review this.  This is actually four separate commits I will be splitting up in the very near future.

Uploading this to a branch in the event my computer gets hit by a bus.

Notable updates:
- remote sync engine was moved to a serial statemachine
- handle a sendCompletion (failure) in order to kick the state machine to an errored/clean up restart 
- Added cancel interface to cancel event reconciliation queue, model reconciliation queue, listeners, mappers, etc.
- Added sending connection information from API in order to decide when subscribeInitialization succeeds (e.g., if we can't make a subscription connection, then we shouldn't be doing an initialSync)

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
@@ -1,0 +1,41 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+import Combine
+import Foundation
+
+@available(iOS 13.0, *)
+extension RemoteSyncEngine {
+    @available(iOS 13.0, *)
+    func onReceiveCompletion(receiveCompletion: Subscribers.Completion<DataStoreError>) {
+        if case .failure(let error) = receiveCompletion {
+            remoteSyncTopicPublisher.send(completion: .failure(error))
+        }
+        if case .finished = receiveCompletion {
+            let unexpectedFinishError = DataStoreError.unknown("ReconcilationQueue sent .finished message",
+                                                               AmplifyErrorMessages.shouldNotHappenReportBugToAWS(),
+                                                               nil)
+            remoteSyncTopicPublisher.send(completion: .failure(unexpectedFinishError))
+        }
+    }
+
+    @available(iOS 13.0, *)
+    func onReceive(receiveValue: IncomingEventReconciliationQueueEvent) {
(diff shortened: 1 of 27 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
```diff
@@ -0,0 +1,56 @@
...
+import Amplify
+import Combine
+import Foundation
+
+@available(iOS 13.0, *)
+extension RemoteSyncEngine {
+    @available(iOS 13.0, *)
+    func onReceiveCompletion(receiveCompletion: Subscribers.Completion<DataStoreError>) {
+        let semaphore = DispatchSemaphore(value: 1)
+        semaphore.wait()
```
kneekey23: guessing this gets called multiple times and you only want it called once? can you removeDuplicates() on the sink that calls this

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngineBehavior.swift
```diff
@@ -7,7 +7,7 @@
 
 import Amplify
 import Combine
-
+//How much of this we will use vs trash, i have no idea
```
kneekey23: lol

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
wooj2: Great callout -- and actually this may not be needed at all

What I was actually trying to protect against is not multiple calls on a single subscription - but I think there was a bug in my clean up logic, where i was only pausing the queue (rather than destroying it), which lead to multiple onReceiveCompletions.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngineBehavior.swift
wooj2: yeah, whooops that was removed . in a subsequent pr :P

wooj2:
Closing this pr.. as i have split this PR into 4 distinct PRs:
https://github.com/aws-amplify/amplify-ios/pull/321 - migrating to a state machine
https://github.com/aws-amplify/amplify-ios/pull/322 -added retry logic to sync engine, without canceling downstream async operations
https://github.com/aws-amplify/amplify-ios/pull/323 - added in the cancelling interface
https://github.com/aws-amplify/amplify-ios/pull/324 - cancelling the downstream operations & wait for subscription to init.

PR #321, opened 2020-02-16
Update RemoteSyncEngine to adopt a StateMachine to allow for easier control of start/stop/retry

wooj2:

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
@@ -1,0 +1,41 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+import Combine
+import Foundation
+
+@available(iOS 13.0, *)
+extension RemoteSyncEngine {
+    @available(iOS 13.0, *)
+    func onReceiveCompletion(receiveCompletion: Subscribers.Completion<DataStoreError>) {
+        if case .failure(let error) = receiveCompletion {
+            remoteSyncTopicPublisher.send(completion: .failure(error))
+        }
+        if case .finished = receiveCompletion {
+            let unexpectedFinishError = DataStoreError.unknown("ReconcilationQueue sent .finished message",
+                                                               AmplifyErrorMessages.shouldNotHappenReportBugToAWS(),
+                                                               nil)
+            remoteSyncTopicPublisher.send(completion: .failure(unexpectedFinishError))
+        }
+    }
+
+    @available(iOS 13.0, *)
+    func onReceive(receiveValue: IncomingEventReconciliationQueueEvent) {
+        switch receiveValue {
+        case .started:
+            remoteSyncTopicPublisher.send(.subscriptionsActivated)
+            if let api = self.api {
+                stateMachine.notify(action: .activatedCloudSubscriptions(api, mutationEventPublisher))
+            }
+        case .paused:
+            remoteSyncTopicPublisher.send(.subscriptionsPaused)
+        case .mutationEvent(let mutationEvent):
+            remoteSyncTopicPublisher.send(.mutationEvent(mutationEvent))
+        }
+    }
+}
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+State.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+State.swift
@@ -1,0 +1,53 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+import Combine
+@available(iOS 13.0, *)
+extension RemoteSyncEngine {
+
+    /// States are descriptive, they say what is happening in the system right now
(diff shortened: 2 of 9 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
```diff
@@ -0,0 +1,41 @@
...
+
+import Amplify
+import Combine
+import Foundation
+
+@available(iOS 13.0, *)
+extension RemoteSyncEngine {
+    @available(iOS 13.0, *)
+    func onReceiveCompletion(receiveCompletion: Subscribers.Completion<DataStoreError>) {
+        if case .failure(let error) = receiveCompletion {
```
palpatim: prefer using `switch` rather than serial `if` to model mutually exclusive choices like picking between enum cases.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -79,120 +89,101 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
         return mutationEventIngester.submit(mutationEvent: mutationEvent)
     }
 
     // MARK: - Startup sequence
-
     private func pauseSubscriptions() {
         log.debug(#function)
         reconciliationQueue?.pause()
+
+        remoteSyncTopicPublisher.send(.subscriptionsPaused)
```
palpatim: Is `reconciliationQueue.pause()` synchronous, or do we need to pass in a completion handler to `pause()`, or is that something done on a subsequent PR?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -79,120 +89,101 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
+        stateMachine.notify(action: .pausedSubscriptions)
     }
 
     private func pauseMutations() {
         log.debug(#function)
         outgoingMutationQueue.pauseSyncingToCloud()
+
         remoteSyncTopicPublisher.send(.mutationsPaused)
+        if let api = self.api, let storageAdapter = self.storageAdapter {
```
palpatim: Is there an error condition that pertains to not having an API or storageAdapter?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+State.swift
```diff
@@ -0,0 +1,53 @@
...
+    enum State {
+        case notStarted
+
+        case pauseSubscriptions
+        case pauseMutationQueue
+        case initializeSubscriptions(APICategoryGraphQLBehavior, StorageEngineAdapter)
+        case performInitialSync
+        case activateCloudSubscriptions
+        case activateMutationQueue(APICategoryGraphQLBehavior, MutationEventPublisher)
+        case notifySyncStarted
```
palpatim: nit: states should describe what is happening. Consider changing to gerund phrases to indicate ongoing activity (e.g., `pausingSubscriptions`) or declarative phrases to indicate something more long-lived (like you already have with, `syncEngineActive`). Disclaimer: I am not a grammarist, so I may be using the wrong names here. :)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
wooj2: pause is currently a synchronous operation.  If pause becomes an area where we can't reliably stop all downstream components, then this may need to become async, but for now, I believe keeping this as a synchronous operation which halts further progression of remote sync engine seems to currently work.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
wooj2: There is a guard/else for storage adapter.  Inside of the start function, if the storageAdapter is available, we send:
```
        remoteSyncTopicPublisher.send(.storageAdapterAvailable)
```
In a subsequent PR I can change this so that we can check for both storage adapter and api and send status up that both of these are available (or perhaps that we are ready to start the sync engine)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
wooj2: This was changed over to a switch statement:
https://github.com/aws-amplify/amplify-ios/pull/324/files

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
wooj2: Updated the code to set self.api = api after checking for a nil storage adapter.  api in the context that i'm referring to is guaranteed to be not nil.  See PR: https://github.com/aws-amplify/amplify-ios/pull/329

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+State.swift
wooj2: Updated: https://github.com/aws-amplify/amplify-ios/pull/329

PR #322, opened 2020-02-16
Added retry/restart logic to RemoteSyncEngine without cancel downstream components

wooj2:
This handles logic to restart the remote sync engine, without actually canceling downstream components.

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Action.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Action.swift
@@ -23,10 +23,12 @@ extension RemoteSyncEngine {
         case activatedCloudSubscriptions(APICategoryGraphQLBehavior, MutationEventPublisher)
         case activatedMutationQueue
         case notifiedSyncStarted
+        case cleanedUp(AmplifyError?)
+        case scheduleRestart(AmplifyError?)
 
         // Terminal actions
         case receivedCancel
-        case errored(AmplifyError)
+        case errored(AmplifyError?)
 
         var displayName: String {
             switch self {
@@ -46,6 +48,10 @@ extension RemoteSyncEngine {
                 return "activatedMutationQueue"
             case .notifiedSyncStarted:
                 return "notifiedSyncStarted"
+            case .cleanedUp:
+                return "cleanedUp"
+            case .scheduleRestart:
+                return "scheduleRestart"
             case .receivedCancel:
                 return "receivedCancel"
             case .errored:
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
@@ -13,14 +13,11 @@ import Foundation
 extension RemoteSyncEngine {
     @available(iOS 13.0, *)
     func onReceiveCompletion(receiveCompletion: Subscribers.Completion<DataStoreError>) {
-        if case .failure(let error) = receiveCompletion {
-            remoteSyncTopicPublisher.send(completion: .failure(error))
-        }
-        if case .finished = receiveCompletion {
-            let unexpectedFinishError = DataStoreError.unknown("ReconcilationQueue sent .finished message",
-                                                               AmplifyErrorMessages.shouldNotHappenReportBugToAWS(),
-                                                               nil)
-            remoteSyncTopicPublisher.send(completion: .failure(unexpectedFinishError))
+        switch receiveCompletion {
+        case .failure(let error):
+            stateMachine.notify(action: .errored(error))
+        case .finished:
+            stateMachine.notify(action: .errored(nil))
         }
     }
 
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Retryable.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Retryable.swift
@@ -1,0 +1,64 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+import Foundation
+
+@available(iOS 13.0, *)
+extension RemoteSyncEngine {
+
+    func resetCurrentAttemptNumber() {
+        currentAttemptNumber = 1
+    }
+
+    func scheduleRestart(error: AmplifyError?) {
+        let advice = getRetryAdviceIfRetryable(error: error)
+        if advice.shouldRetry {
+            scheduleRestart(advice: advice)
+        } else {
+            if let error = error {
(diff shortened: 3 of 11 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Retryable.swift
```diff
@@ -0,0 +1,64 @@
...
+            if let error = error {
+                remoteSyncTopicPublisher.send(completion: .failure(DataStoreError.api(error)))
+            } else {
+                remoteSyncTopicPublisher.send(completion: .finished)
+            }
+        }
+
+    }
+
+    private func getRetryAdviceIfRetryable(error: Error?) -> RequestRetryAdvice {
```
palpatim: nit: "ifRetryable" is redundant, since the advice could be "don't retry".

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngineBehavior.swift
```diff
@@ -17,6 +17,7 @@ enum RemoteSyncEngineEvent {
     case subscriptionsActivated
     case mutationQueueStarted
     case syncStarted
+    case cleanedUp
```
palpatim: BTW, while I'm thinking about it, I like the idea you're promoting (intentionally or not) of normalizing our naming: `State` can represent a point-in-time view of a system, and could be thought of as something managed internally. `Event` can represent an external report of a state transition, even if it doesn't literally represent a change from one `State` to another.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
```diff
@@ -13,14 +13,11 @@ import Foundation
...
-        if case .finished = receiveCompletion {
-            let unexpectedFinishError = DataStoreError.unknown("ReconcilationQueue sent .finished message",
-                                                               AmplifyErrorMessages.shouldNotHappenReportBugToAWS(),
-                                                               nil)
-            remoteSyncTopicPublisher.send(completion: .failure(unexpectedFinishError))
+        switch receiveCompletion {
+        case .failure(let error):
+            stateMachine.notify(action: .errored(error))
+        case .finished:
+            stateMachine.notify(action: .errored(nil))
```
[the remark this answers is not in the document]
wooj2: API category gives an .finished completion when we disconnect.  Seems like we can introduce a new state called "finish" which automatically attempts to restart (rather than looking at the error that comes back to make some smart decisions on when/if to restart)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Retryable.swift
wooj2: Updated, thanks!

wooj2: addressed: https://github.com/aws-amplify/amplify-ios/pull/329

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Action.swift
```diff
@@ -23,10 +23,12 @@ extension RemoteSyncEngine {
         case activatedCloudSubscriptions(APICategoryGraphQLBehavior, MutationEventPublisher)
         case activatedMutationQueue
         case notifiedSyncStarted
+        case cleanedUp(AmplifyError?)
+        case scheduleRestart(AmplifyError?)
 
         // Terminal actions
         case receivedCancel
-        case errored(AmplifyError)
+        case errored(AmplifyError?)
```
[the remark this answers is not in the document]
wooj2: addressed: https://github.com/aws-amplify/amplify-ios/pull/329

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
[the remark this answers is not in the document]
wooj2: addressed: https://github.com/aws-amplify/amplify-ios/pull/329

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -231,10 +249,20 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
         stateMachine.notify(action: .activatedMutationQueue)
     }
 
+    private func cleanup(error: AmplifyError?) {
+        // TODO:
+        // handle clean up here
```
drochetti: nit: put the message on the same line of the TODO so IDE displays the warning accordingly

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Retryable.swift
```diff
@@ -0,0 +1,64 @@
...
+        return advice
+    }
+
+    private func scheduleRestart(advice: RequestRetryAdvice) {
+        log.verbose("\(#function) scheduling retry for restarting remote sync engine")
+        resolveReachabilityPublisher()
+        mutationRetryNotifier = MutationRetryNotifier(advice: advice,
+                                                      networkReachabilityPublisher: networkReachabilityPublisher) {
+                                                        self.mutationRetryNotifier = nil
+                                                        self.stateMachine.notify(action: .receivedStart)
```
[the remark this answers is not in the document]
wooj2: Hmm indentation seems fine to me? i'm selecting everything and hitting "Ctrl + i", and there doesn't seem to be any updates to the formatting of the code...

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Retryable.swift
```diff
@@ -0,0 +1,64 @@
...
+        currentAttemptNumber += 1
+    }
+
+    private func resolveReachabilityPublisher() {
+        if networkReachabilityPublisher == nil {
+            if let reachability = api as? APICategoryReachabilityBehavior {
+                do {
+                    networkReachabilityPublisher = try reachability.reachabilityPublisher()
+                } catch {
+                    log.error("\(#function): Unable to listen on reachability: \(error)")
```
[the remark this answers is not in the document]
wooj2: I believe this code is correct.  If the networkReachabilityPublisher does not initialized due to some exception being thrown, then our retry logic will be based purely on the exponential backoff (versus exponential backoff w/ using the status of reachability).

I looked into the reachability code, and it seems that it is possible for reachability to fail during initialization, so, we kept the interface to return an optional instance.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
wooj2: Ah! I didn't know that.  I will do that in the future, but this TODO has already been implemented in a subsequent PR.

PR #324, opened 2020-02-16
Propagate connection status up, wait on subscription init, fix unit tests

wooj2:
In this case we are waiting for the subscriptions to properly initialize the subscriptions, which means, we have the need to make sure that the sockets are connected, prior to doing our initial sync.

I'm not particularly proud how we are handling connected sockets & propagating that status up -- I will be refactoring this shortly -- but the good news is that if i explicitly kill sockets in API, this code seems to allow us to bring and remote sync engine up and down! :)

Known issues
. API is not passing back errors (working with lawmicha@ on this.)
. API is storing connections and not killing them (have some code which cleans the connections on a cancel() of an Amplify API operation) (working with lawmicha@ on this)
. OutgoingMutationQueue occasionally receives and invalid state transition -- Haven't dug into it yet, but I will.

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
@@ -13,22 +13,32 @@ import Foundation
 extension RemoteSyncEngine {
     @available(iOS 13.0, *)
     func onReceiveCompletion(receiveCompletion: Subscribers.Completion<DataStoreError>) {
-        let semaphore = DispatchSemaphore(value: 1)
-        semaphore.wait()
-        if case .syncEngineActive = stateMachine.state {
-            switch receiveCompletion {
-            case .failure(let error):
-                stateMachine.notify(action: .errored(error))
-            case .finished:
-                stateMachine.notify(action: .errored(nil))
-            }
+        switch stateMachine.state {
+        case .initializeSubscriptions:
+            notifyError(receiveCompletion: receiveCompletion)
+        case .syncEngineActive:
+            notifyError(receiveCompletion: receiveCompletion)
+        default:
+            break
         }
-        semaphore.signal()
+    }
+
+    @available(iOS 13.0, *)
+    func notifyError(receiveCompletion: Subscribers.Completion<DataStoreError>) {
+        switch receiveCompletion {
+        case .failure(let error):
+            stateMachine.notify(action: .errored(error))
+        case .finished:
+            stateMachine.notify(action: .errored(nil))
+        }
     }
 
     @available(iOS 13.0, *)
     func onReceive(receiveValue: IncomingEventReconciliationQueueEvent) {
         switch receiveValue {
+        case .initialized:
+            remoteSyncTopicPublisher.send(.subscriptionsInitialized)
+            stateMachine.notify(action: .initializedSubscriptions)
         case .started:
             remoteSyncTopicPublisher.send(.subscriptionsActivated)
             if let api = self.api {
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
@@ -24,6 +24,8 @@ extension RemoteSyncEngine {
 
             case (.initializeSubscriptions, .initializedSubscriptions):
                 return .performInitialSync
+            case (.initializeSubscriptions, .errored(let error)):
+                return .cleanup(error)
 
             case (.performInitialSync, .performedInitialSync):
                 return .activateCloudSubscriptions
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
@@ -198,8 +198,8 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
             receiveCompletion: onReceiveCompletion(receiveCompletion:),
             receiveValue: onReceive(receiveValue:))
 
-        remoteSyncTopicPublisher.send(.subscriptionsInitialized)
-        stateMachine.notify(action: .initializedSubscriptions)
+        //Notifying the publisher & state machine are handled in:
+        // RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
     }
 
     private func performInitialSync() {
@@ -253,8 +253,10 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
     }
 
     private func cleanup(error: AmplifyError?) {
-        // TODO:
-        // handle clean up here
+        reconciliationQueue?.cancel()
+        reconciliationQueue = nil
+        outgoingMutationQueue.pauseSyncingToCloud()
+
         remoteSyncTopicPublisher.send(.cleanedUp)
         stateMachine.notify(action: .cleanedUp(error))
     }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
@@ -29,6 +29,7 @@ final class AWSIncomingEventReconciliationQueue: IncomingEventReconciliationQueu
     }
 
     private var reconciliationQueues: [String: ModelReconciliationQueue]
+    private var reconciliationQueueConnectionStatus: [String: Bool]
 
     init(modelTypes: [Model.Type],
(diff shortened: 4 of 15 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
```diff
@@ -23,47 +23,51 @@ final class IncomingAsyncSubscriptionEventPublisher: Cancellable {
     typealias Event = AsyncEvent<SubscriptionEvent<GraphQLResponse<Payload>>, Void, APIError>
 
     private var onCreateOperation: GraphQLSubscriptionOperation<Payload>?
-    private let onCreateListener: GraphQLSubscriptionOperation<Payload>.EventListener
+    private var onCreateListener: GraphQLSubscriptionOperation<Payload>.EventListener?
```
palpatim: Here and with other listeners: I'm not sure why this needs to be an optional var? I don't see that you unassign it anywhere?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
```diff
@@ -93,11 +93,7 @@ final class AWSModelReconciliationQueue: ModelReconciliationQueue {
             .sink(receiveCompletion: { [weak self] completion in
                 self?.receiveCompletion(completion)
                 }, receiveValue: { [weak self] receiveValue in
```
palpatim: 1. nit: arg name is probably better as `value` 
1. For future: I poked at this a bit today, and I don't think we need to `weak self` inside the closure--as long as we're properly disposing of the cancellable, and as long as our owner is properly disposing of us, we don't get a retain cycle from this closure. When I did this originally I didn't investigate it first, and it looks like I was overabundantly cautious (or as we call it, "paranoid")

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ModelReconciliationQueue.swift
```diff
@@ -12,6 +12,7 @@ import Combine
 enum ModelReconciliationQueueEvent {
     case started
     case paused
```
palpatim: Why don't we need the model name on these events, if we need it below?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
```diff
@@ -24,6 +24,8 @@ extension RemoteSyncEngine {
 
             case (.initializeSubscriptions, .initializedSubscriptions):
                 return .performInitialSync
+            case (.initializeSubscriptions, .errored(let error)):
+                return .cleanup(error)
```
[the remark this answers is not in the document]
wooj2: I needed to find a way to pass the error from the completion to the retry block (since it goes from completion -> .errored/finished -> scheduleRestart) so passing error through these enum values was one way to do this.  If there's a better way to do this, i'm all ears.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -198,8 +198,8 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
             receiveCompletion: onReceiveCompletion(receiveCompletion:),
             receiveValue: onReceive(receiveValue:))
 
-        remoteSyncTopicPublisher.send(.subscriptionsInitialized)
-        stateMachine.notify(action: .initializedSubscriptions)
+        //Notifying the publisher & state machine are handled in:
+        // RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
```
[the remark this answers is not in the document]
wooj2: Done.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -253,8 +253,10 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
     }
 
     private func cleanup(error: AmplifyError?) {
-        // TODO:
-        // handle clean up here
+        reconciliationQueue?.cancel()
+        reconciliationQueue = nil
+        outgoingMutationQueue.pauseSyncingToCloud()
```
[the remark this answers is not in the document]
wooj2: Absolutely, this is handled here:
https://github.com/aws-amplify/amplify-ios/pull/326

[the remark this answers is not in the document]
wooj2: In this case, I dont think we should cancel the upstream subscription (being the StorageEngine).  I feel like the only reason why we should cancel upstream is if there is a catastrophic error that happens -- e.g. the sync engine is going to stop retrying, API is not valid, or perhaps, the storage engine is nil.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
```diff
@@ -83,8 +85,16 @@ final class AWSIncomingEventReconciliationQueue: IncomingEventReconciliationQueu
...
 
     private func onRecieveValue(receiveValue: ModelReconciliationQueueEvent) {
-        if case .mutationEvent(let event) = receiveValue {
-            self.eventReconciliationQueueTopic.send(.mutationEvent(event))
+        switch receiveValue {
+        case .mutationEvent(let event):
+            eventReconciliationQueueTopic.send(.mutationEvent(event))
+        case .connected(let modelName):
+            reconciliationQueueConnectionStatus[modelName] = true
+            if reconciliationQueueConnectionStatus.count == reconciliationQueues.count {
```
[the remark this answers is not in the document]
wooj2: Regarding the refactor:
I'm not sure if i understand your point?  The event Reconciliation queue manages N queues, and we need to make sure that each one of the queues is fully connected prior to sending status to the RemoteSyncEngine that we are fully connected.

Regarding the potential bug:
Ah, yes, set is exactly what we need here.. updating.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
wooj2: It is because we are using `self.` inside the closure.  Setting to var allows us use `self` prior to all the `let`'s being assigned.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
wooj2: thanks for the clarification, i'll put this on the todo list.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ModelReconciliationQueue.swift
wooj2: Valid point - they could be added, and probably will be at some point in time, but the only consumer of this class is currently `IncomingEventReconciliationQueueFactory`, and we aren't using the `started` and `paused` events.  These were initially added because I was planning to do unit testing around these class (which I have not gotten around to doing :))

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+IncomingEventReconciliationQueueEvent.swift
```diff
@@ -13,22 +13,32 @@ import Foundation
...
+        case .initializeSubscriptions:
+            notifyError(receiveCompletion: receiveCompletion)
+        case .syncEngineActive:
+            notifyError(receiveCompletion: receiveCompletion)
+        default:
+            break
+        }
+    }
+
+    @available(iOS 13.0, *)
```
drochetti: question: if the extensions is marked with `@available(iOS 13.0, *)` does each method also need it?

wooj2: Hmm good call out, looks like it isn't needed!

wooj2: https://github.com/aws-amplify/amplify-ios/pull/329/files

PR #325, opened 2020-02-16
[DO NOT COMMIT] Hacks used to test remote sync engine restarting

wooj2:
This is probably not useful, as you could probably imagine the code i used to hack API to work on the restarting of the remote sync engine, but i figure i give this to you just in case you were curious how I hacked up API.

I'm hoping we can come up with something similar so that
1.  Errors are passed back to the data store category
2. Sockets are destroyed upon no other operations are using that particular socket

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
@@ -46,6 +46,7 @@ final public class AWSGraphQLSubscriptionOperation<R: Decodable>: GraphQLSubscri
         } else {
             super.cancel()
         }
+        subscriptionConnectionFactory.cancel()
     }
 
     override public func main() {
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Subscription/SubscriptionConnection/AppSyncSubscriptionConnection+EventHandler.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Subscription/SubscriptionConnection/AppSyncSubscriptionConnection+EventHandler.swift
@@ -65,6 +65,9 @@ extension AppSyncSubscriptionConnection {
             serialSubscriptionQueue.async {[weak self] in
                 // Move all subscriptionItems to disconnected but keep them in memory.
                 self?.subscriptionItems.forEach { _, subscriptionItem in
+                    if let error = error {
+                       subscriptionItem.dispatch(error: .networkError("", nil, error))
+                    }
                     switch subscriptionItem.subscriptionConnectionState {
                     case .connecting, .connected:
                         subscriptionItem.setState(.disconnected)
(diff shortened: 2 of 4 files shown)


AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
```diff
@@ -46,6 +46,7 @@ final public class AWSGraphQLSubscriptionOperation<R: Decodable>: GraphQLSubscri
         } else {
             super.cancel()
         }
+        subscriptionConnectionFactory.cancel()
```
lawmicha: some thoughts around this

SubscriptionFactory is specific implementation of using the RealTimeSubscriptionConnection code. for example, AppSyncClient has their own and Amplify API has their own, with methods that retrieve connections based on Amplify specific classes like `EndpointConfig`. 

So the package that is shared between Amplify API Plugin and AppSyncClient is without the factory class, each consumer of the package will build the `subscriptionConnectionFactory` class however they like. in our case, we manage apiName to connectionProviders (which represent a single websocket connection that can be subscribed on).

lawmicha: looking at this change, we're trying to solve the problem with a socket still open even after all subscriptions have unsubscribed, right?

- SubscriptionFactory holds the single socket, and vend connections to the caller (SubscriptionOperation). the SubscriptionOperation holds onto the single subscription on that socket and also has a reference to the connectionProvider via that subscription connection.

if we call cancel on an operation.cancel() call, and remove the mapping, that means we the subscription factory loses reference to all of the connection providers.

will have to think about this further to solve the +1 subscription connection, and then -1 subscription connection scenario, check that there are no remaining subscription connections on the connection provider and do a websocket disconnect call

wooj2: Reference counting on the subscription connection seems like a reasonable approach (along with taking into account the state of the socket).  Just curious, is it common for a subscription to end without a socket to be disconnected (e.g. i'm trying to understand just how important it is to implement reference counting, or perhaps adopt a more aggressive approach where -- if one subscription has an issue, we kill the entire socket along with all other subscriptions on that socket)?

The use case I was basing the datastore restart code was around turning off wifi, in which case, the socket would get closed, and all subscriptions would be invalidated at that point.

lawmicha: i think the common use case we need to make sure that when there are two subscriptions on a socket, and one is being cancelled or the view is deinitialized or something then manages to disconnect the second subscription, the first subscription is still connected and socket still needs to be connected.

lawmicha: thanks for bringing this to my attention, and i'll look into this a bit more once i get the consolidation work completed (https://github.com/aws-amplify/amplify-ios/pull/327)

lawmicha: added PR that builds on top of the one above https://github.com/aws-amplify/amplify-ios/pull/333 that should address the issue

PR #326, opened 2020-02-17
Fix two memory leaks when restarting remote sync engine

wooj2:
Found two memory leaks using the memory graph debugger/visualizer.

Repro:
setup a data store app, go into airplane mode to force subscription connections to fail, then go back online to restart subscription connections

Observed behavior:
Instances of AWSIncomingSubscriptionEventPublisher were referencing itself
IncomingAsyncSubscriptionEventToAnyModelMapper references dangling due to subscriptions and.


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingSubscriptionEventPublisher.swift
```diff
@@ -52,6 +52,7 @@ final class AWSIncomingSubscriptionEventPublisher: IncomingSubscriptionEventPubl
         mapperSink = nil
 
         asyncEvents.cancel()
+        mapper?.cancel()
```
drochetti: ok, so this answers the question I had in the previous PR. Nice!

PR #332, opened 2020-02-20
Switch mutex to serial work queue & remove unused serial queue

wooj2:
- Minor refactor to avoid using mutex/DispatchSemaphore
- Removed serial operation queue that was no longer being used in RemoteSyncEngine

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
@@ -34,8 +34,8 @@ final class IncomingAsyncSubscriptionEventPublisher: Cancellable {
     private var onDeleteListener: GraphQLSubscriptionOperation<Payload>.EventListener?
     private var onDeleteConnected: Bool
 
-    private let connectionStatusMutex: DispatchSemaphore
-    private var combinedConnectionStatus: Bool {
+    private let connectionStatusQueue: OperationQueue
+    private var combinedConnectionStatusIsConnected: Bool {
         return onCreateConnected && onUpdateConnected && onDeleteConnected
     }
 
@@ -46,7 +46,10 @@ final class IncomingAsyncSubscriptionEventPublisher: Cancellable {
         self.onCreateConnected = false
         self.onUpdateConnected = false
         self.onDeleteConnected = false
-        self.connectionStatusMutex = DispatchSemaphore(value: 1)
+        self.connectionStatusQueue = OperationQueue()
+        connectionStatusQueue.name = "com.amazonaws.Amplify.RemoteSyncEngine.\(modelType.modelName).IncomingAsyncSubscriptionEventPublisher"
+        connectionStatusQueue.maxConcurrentOperationCount = 1
+        connectionStatusQueue.isSuspended = false
 
         let incomingSubscriptionEvents = PassthroughSubject<Event, DataStoreError>()
         self.incomingSubscriptionEvents = incomingSubscriptionEvents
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
```diff
@@ -155,11 +161,11 @@ final class IncomingAsyncSubscriptionEventPublisher: Cancellable {
         onUpdateOperation = nil
         onUpdateListener = nil
 
-
         onDeleteOperation?.cancel()
         onDeleteOperation = nil
         onDeleteListener = nil
 
+        self.connectionStatusQueue.cancelAllOperations()
```
palpatim: nit: remove extraneous `self`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
```diff
@@ -79,55 +82,58 @@ final class IncomingAsyncSubscriptionEventPublisher: Cancellable {
     func onCreateListenerHandler(event: Event) {
         log.verbose("onCreateListener: \(event)")
```
palpatim: It seems like we're repeating the guts of this method in `onCreateListenerHandler`, `onUpdateListenerHandler`, and `onDeleteListenerHandler`?

wooj2: Thanks for calling this out.  I struggled at first to figure out how to do this better -- ended up with an update.  Feel free to suggest something better.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
wooj2: done.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
```diff
@@ -81,56 +81,47 @@ final class IncomingAsyncSubscriptionEventPublisher: Cancellable {
...
     func onDeleteListenerHandler(event: Event) {
         log.verbose("onDeleteListener: \(event)")
+        let onDeleteConnectionOp = CancelAwareBlockOperation {
+            self.onDeleteConnected = self.determineConnectionStatusOn(event: event)
+            self.sendConnectionEventIfConnected(event: event)
+        }
+        genericListenerHandler(event: event, cancelAwareBlock: onDeleteConnectionOp)
+    }
+
+    func determineConnectionStatusOn(event: Event) -> Bool {
```
palpatim: For Boolean return types, prefer `isXXX(for event: Event)` or similar naming that [asserts something about the receiver](https://swift.org/documentation/api-design-guidelines/#boolean-assertions) and reads like an English question. For this, maybe something like `isConnected(for event: Event)` or `isConnectionValid(for event: Event)`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
```diff
@@ -81,56 +81,47 @@ final class IncomingAsyncSubscriptionEventPublisher: Cancellable {
...
+        if case .inProcess(.connection(.connected)) = event {
+            return true
+        }
+        return false
+    }
+
+    func sendConnectionEventIfConnected(event: Event) {
+        if combinedConnectionStatus {
```
palpatim: I don't immediately see where this is named, but again, Booleans should be named with an `isXXX` pattern.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
wooj2: thanks for calling this out, changed to:
`isConnectionStatusConnected(for event: Event)`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
wooj2: This is a computed property located on line 38, and in this case, i don't agree with following the isXXX pattern.  That being said, I think the name could be better, so I changed this to:
````
combinedConnectionStatusIsConnected
```

wooj2: To be clear, i think the `isXXX` pattern works well for singular cases (and thanks for calling it out above), but in this case, we are referring to a case which is computed from multiple components, and the wording of:
```
if isCombinedConnectionStatus {
}
```
seems less clear than
```
if combinedConnectionStatusIsConnected
```

At this point, I'm going to bias for action and push in since I have approval, with the understanding that I'm open to debate on changing the name at any given time.

PR #333, opened 2020-02-20
Add websocket disconnect when all subscriptions are closed

lawmicha:
*Issue #, if available:*
An underlying websocket remains open even after all subscriptions are closed. 

*Description of changes:*
Forcefully close the websocket connection when there are no more subscriptions on the socket, by 
1. setting internal state of the connection provider to disconnected
2. call socket.disconnect (Starscream method).

We had decided on this approach of cleaning up the socket so that new subscriptions will trigger fresh instantiations of the socket, and thus connection handshake process. That way, the internal state of the connection provider is more often than not synchronzied to the actual state of the socket connection.

Take internet connectivity for an example. If the socket gets into a bad state while the connection provider does not receive a disconnect event, then our internal state maintains a connected state. We then rely on the keep alive to check the staleness of the connection which will take 5 minutes before the internal state is deemed not connected.

Now let's assume a new subscription is attempted on the existing broken socket which has an internal state of connected, we simply write a message on the socket and react to the callbacks such as `didReceiveMessage`. It's not clear exactly what the behavior is for Starscream socket implementation and if we never get a message back, the subscription remains in a connecting state and never transitions to connected.

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests+Subscriptions.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests+Subscriptions.swift
@@ -1,0 +1,149 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+import AWSMobileClient
+@testable import AWSAPICategoryPlugin
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSAPICategoryPluginTestCommon
+@testable import AppSyncSubscriptionClient
+
+/// Subscription related stress/load tests
+extension GraphQLModelBasedTests {
+
+    /// Ensure that the underlying websocket socket is disconnected when the number of subscriptions on the socket
+    /// reaches zero. This test first creates multiple subscriptions, does a mutation to make sure data is flowing,
+    /// cancel the operations, and makes sure the underlying socket is disconnected. Given this new state of the system
+    /// where there is an existing connection provider, run a similar test by create a new subscription, which in turn
+    /// uses the same connection provider instance but will create a new websocket connection.
+    ///
+    /// - Given: Connected subscriptions
+    /// - When:
+    ///    - All subscription operations are cancelled
+    /// - Then:
+    ///    - Underlying websocket is disconnected
+    func testAllSubscriptionsCancelledShouldDisconnectTheWebsocket() {
+        let connectedInvoked = expectation(description: "Connection established")
+        connectedInvoked.expectedFulfillmentCount = 3
+        let disconnectedInvoked = expectation(description: "Connection disconnected")
+        disconnectedInvoked.expectedFulfillmentCount = 3
+        let completedInvoked = expectation(description: "Completed invoked")
+        completedInvoked.expectedFulfillmentCount = 3
+        let progressInvoked = expectation(description: "progress invoked")
+        progressInvoked.expectedFulfillmentCount = 3
+
+        let operation1 = Amplify.API.subscribe(from: Post.self, type: .onCreate) { event in
+            if case let .inProcess(response) = event, case let .connection(state) = response {
+                if case .connected = state {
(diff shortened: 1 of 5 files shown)


AppSyncSubscriptionClient/AppSyncSubscriptionClient/ConnectionProvider/AppsyncRealtimeConnection/RealtimeConnectionProvider+Websocket.swift
```diff
@@ -10,11 +10,11 @@ import Foundation
 extension RealtimeConnectionProvider: AppSyncWebsocketDelegate {
 
     public func websocketDidConnect(provider: AppSyncWebsocketProvider) {
+        lastKeepAliveTime = DispatchTime.now()
```
wooj2: I think that keep-alive messages arrive on `WebsocketDidReceiveMessage`, so I'm a bit surprised to see this in the websocketDidConnect method.

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests+Subscriptions.swift
```diff
@@ -0,0 +1,149 @@
...
+@testable import AWSAPICategoryPluginTestCommon
+@testable import AppSyncSubscriptionClient
+
+/// Subscription related stress/load tests
+extension GraphQLModelBasedTests {
+
+    /// Ensure that the underlying websocket socket is disconnected when the number of subscriptions on the socket
+    /// reaches zero. This test first creates multiple subscriptions, does a mutation to make sure data is flowing,
+    /// cancel the operations, and makes sure the underlying socket is disconnected. Given this new state of the system
+    /// where there is an existing connection provider, run a similar test by create a new subscription, which in turn
```
wooj2: You might want to rephrase this paragraph (feel free to reject/accept, and/or update where is necessary)

The purpose of this test is to ensure that all websockets can be successfully created, exercised and terminated while keeping a single connection provider in memory.  Specifically, the following test exercises the following:
1.  Create a new connection provider
2.  Create multiple subscriptions
3.  Make a mutation and validate that it succeeds
4.  Cancel the operations
5.  Ensure the socket is disconnected
6.  Ensure connection provider is still in memory
7.  Repeat Steps 2-6 with the existing connection provider.

lawmicha: thanks for the improved wording! will take this with next iteration

AppSyncSubscriptionClient/AppSyncSubscriptionClient/ConnectionProvider/AppsyncRealtimeConnection/RealtimeConnectionProvider+Websocket.swift
lawmicha: `lastKeepAliveTime` is technically "last time we got any data back" which includes graphql responses, connection messages, errors, etc. Once the keepAlive process checks that current time is greater than ("lastKeepAliveTime" + 5 minutes) then it is determined to be stale and sets the internal connection provider state to .notConnected and best effort call to `websocket.disconnect()`. Since introducing this change to check count of subscriptions on the socket, we hold onto the connection provider instance while keeping the websocket disconnected. 

Looking at this further, originally I added this so to account for the case where a websocket connection is established again at a later time. but in fact we don't need this since the when we process the response from the websocket service, we will also update it then. so it's sort of redundant to update it twice, once at connect and once at websocketDidReceivedData. I will remove this check here

lawmicha:
since we have decided on separate package, this PR no longer accurate. However, now I am faced with making a change over in https://github.com/aws-amplify/aws-appsync-realtime-client-ios and then having integration tests over in Amplify repo. Next steps will probably be
1. Make this change over in the appsync-realtime-client repo. there are no integration tests there so perhaps it will be useful to add more unit test or integration tests there.
2. I have a meeting with ivan to set up circleCI + fastlane for automated test runs + pod release
3. Publish "1.0.1" of AppSyncRealTimeClient and consume it in amplify, and add this integration test back at the same time

lawmicha:
PR over in shared package https://github.com/aws-amplify/aws-appsync-realtime-client-ios/pull/8

Issue #318, opened 2020-02-14
Add converter class for model into SQLite and GraphQL components


lawmicha:
Synced with Daniel on this, opening issue to refactor some of this conversion logic into some converter class to be used to take model field types and values into SQLite and GraphQL components

_Originally posted by @lawmicha in https://github.com/aws-amplify/amplify-ios/pull/317_

drochetti:
This was mostly done by the PR #334. I'll close this one for now and we can track future improvements of the serialization code in a new ticket.

Issue #314, opened 2020-02-12
Add DataStore/API integration test for a model with all supported types


lawmicha:
This would be a follow up task after @drochetti's work Date/Enum/Primitive support, We should add an integration test which contains a model with all of the supported types

lawmicha:
we should also make sure mutating and querying for AppSync Floats with values that are at the upper limit is working or if there are some limitations, document what is the maximum value possible

lawmicha:
we could take https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyTestCommon/Models/QPredGen.swift and use the schema
```
/*
 Generated from:
 type QPredGen @model {
   id: ID!
   name: String!
   myBool: Boolean
   myDouble: Float
   myInt: Int
   myString: String
   myDate: AWSDate
   myDateTime: AWSDateTime
   myTime: AWSTime
 }
 */
```
in both of the integration tests in DataStore and API

kjones:
Other possible scalars to test: https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html
Are all these supported?

Other scenarios could include all combinations of nullable fields. 
```graphql
type TestModel @model {
  id: ID!
  testInt: Int!
  nullableInt: Int
  intList: [Int!]!
  intNullableList: [Int!]
  nullableIntList: [Int]!
  nullableIntNullableList: [Int]
}
```
Enums and nested types have also been problematic in the past so having tests for them might be useful.
```graphql
type EnumTestModel @model {
  enumVal: TestEnum!
  nullableEnumVal: TestEnum
  enumList: [TestEnum!]!
  enumNullableList: [TestEnum!]
  nullableEnumList: [TestEnum]!
  nullableEnumNullableList: [TestEnum]
}

enum TestEnum {
  VALUE_ONE
  VALUE_TWO
}

type NestedTypeTestModel @model {
  nestedVal: Nested!
  nullableNestedVal: Nested
  nestedList: [Nested!]!
  nestedNullableList: [Nested!]
  nullableNestedList: [Nested]!
  nullableNestedNullableList: [Nested]
}

type Nested {
  valueOne: Int
  valueTwo: String
}
```

lawmicha:
Thanks @kjones, those are great examples. I've used them almost exactly in this PR: https://github.com/aws-amplify/amplify-ios/pull/1135

Issue #240, opened 2019-12-09
codegen fails to generate code for non @model dependencies 


sebsto:
When using a model as :

```graphql
type Coordinate
{
    longitude: Float
    latitude: Float
}


type Landmark
    @model
    @auth(rules: [ {allow: private, provider: userPools, operations: [ read ] } ])
{
    id: ID!
    name: String!
    category: String
    city: String
    state: String
    isFeatured: Boolean
    isFavorite: Boolean
    park: String
    coordinates: Coordinate 
    imageName: String
}
```

Codegen fails with 

```error
Unknown type Coordinate for field coordinates. Did you forget to add the @model directive
Error: Unknown type Coordinate for field coordinates. Did you forget to add the @model directive
    at AppSyncSwiftVisitor.getNativeType (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-codegen-appsync-model-plugin/src/visitors/appsync-visitor.ts:298:13)
    at /usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-codegen-appsync-model-plugin/src/visitors/appsync-swift-visitor.ts:33:32
    at Array.forEach (<anonymous>)
    at /usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-codegen-appsync-model-plugin/src/visitors/appsync-swift-visitor.ts:32:34
    at Array.forEach (<anonymous>)
    at AppSyncSwiftVisitor.generateStruct (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-codegen-appsync-model-plugin/src/visitors/appsync-swift-visitor.ts:27:46)
    at AppSyncSwiftVisitor.generate (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-codegen-appsync-model-plugin/src/visitors/appsync-swift-visitor.ts:21:22)
```

As a workaround, I added @model on Coordinate and a @connection annotation on Landmark.  This is NOT want I want to model.  I do not want Coordinate to be stored in a DynamoDB table, this is just a type definition.  In previous version Coordinate was a JSON object stored as a property into the Landmark table.  I want to keep this behaviour.

Soek22:
Hello @sebsto ,

1. This has been marked as `BUG` regarding `API` here: https://github.com/aws-amplify/amplify-cli/issues/3026 .

[...]

PR #334, opened 2020-02-21
add support for Enum and non-model types

drochetti:
**Issues:** #111 #240 #246 #318 #314 (Partially, API is still missing)

**Notes:**

- added support for SQLite to handle Enums
- added support for SQLite to handle non-model types:
  - custom Codable structs
  - codable arrays
- added tests to make sure the right values are represented as SQLite bindings

**Tests:**

- added a model that has all types of fields to make it easier to
visualize how data is represented on SQLite

**Pending:**

- More documentation
- GraphQL cleanup on API category
- More tests
- Merge with the latest Sync Engine changes


AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Models/SQLModelValueConverterTests.swift
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    func testModelWithEveryTypeConversionToBindings() {
+
+        // non-model type
+        let nonModelExample = ExampleNonModelType(someString: "string", someEnum: .foo)
+        let nonModelJSON = "{\"someString\":\"string\",\"someEnum\":\"foo\"}"
+
+        // example model
+        let example = ExampleWithEveryType(id: "df5dd4a4-34f8-4974-8a37-2617cf8dafe1",
```
lawmicha: I went and provisioned the backend with NonModel type, it does not create any mutations or queries, and just defines the type. Seems like the usecase is for attaching it to an existing type that contains the model directive. for example:
```
type Todo @model {
  id: ID!
  name: String!
  description: String
  innerTodo: JustTodo
}

type JustTodo {
  name: String!
  description: Int
}
```
```
mutation createTodo($input: CreateTodoInput!) {
  createTodo(input: $input) {
    description
  	innerTodo {
      name
      description
    }
  }
}
```

lawmicha:
Non-blocking comment, but just want to share my learnings for the backend provisioning.  When i provisioned the backend with NonModel type in the schema, it does not create any mutations or queries, and just defines the type. Seems like the usecase is for attaching it to an existing type that contains the model directive. for example:
```
type Todo @model {
  id: ID!
  name: String!
  description: String
  innerTodo: JustTodo
}

type JustTodo {
  name: String!
  description: Int
}
```
```
mutation createTodo($input: CreateTodoInput!) {
  createTodo(input: $input) {
    description
  	innerTodo {
      name
      description
    }
  }
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Models/SQLModelValueConverterTests.swift
lawmicha: nice, this is a Model instance containing a `nonModelField` which is what i expected to be the use case. are we still supporting DataStore.save on NonModelTypes or it is always on a Model which may or may not contain non model fields?

drochetti: You can only save a `Model`, the non-model types can be used as values of a `Model`. That is also true on the AppSync level.

Amplify/Categories/DataStore/Model/Model+Enum.swift
```diff
@@ -0,0 +1,30 @@
...
+///
+/// - Example:
+///
+/// ```swift
+/// public enum PostStatus: String, ModelEnum {
+///     case draft
+///     case published
+/// }
+/// ```
+public protocol ModelEnum: Codable {
```
drochetti: To @palpatim (and anyone else interested).

We talked about relying solely on the `RawRepresentable` for this. However, I could not make it work properly with our current StorageEngine implementation. `RawRepresentable` has an `associatedtype` (named `RawValue`) which blocks me from doing `RawRepresentable.Type` for `ModelFieldDefinition` in the schema (`.enum(MyEnum.Type)`).

I'm open to suggestions!

Amplify/Categories/DataStore/Model/Schema/ModelValueConverter.swift
```diff
@@ -0,0 +1,59 @@
...
+    ///   - fieldType: the type of the `Model` field
+    /// - Returns: the converted value to the expected `ModelFieldType`
+    func convertToSource(from target: TargetType, fieldType: ModelFieldType) throws -> SourceType
+
+}
+
+/// Extension with reusable JSON encoding/decoding utilities.
+extension ModelValueConverter {
+
+    static var decoder: JSONDecoder {
```
palpatim: nit: name property `jsonDecoder` since some day we may have a `cborDecoder` or `fooDecoder`...

Amplify/Categories/DataStore/Model/Schema/ModelValueConverter.swift
```diff
@@ -0,0 +1,59 @@
...
+        JSONDecoder(dateDecodingStrategy: ModelDateFormatting.decodingStrategy)
+    }
+
+    static var encoder: JSONEncoder {
```
palpatim: as above, name `jsonEncoder`

Amplify/Core/Support/Encodable+AnyEncodable.swift
```diff
@@ -0,0 +1,29 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+public struct AnyEncodable: Encodable {
```
palpatim: 1. If we're exposing this as a public utility, let's add some unit tests to make sure this works as contracted, and some docs defining its use cases.
1. We don't need a corresponding `AnyDecodable`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Model+SQLite.swift
```diff
@@ -23,24 +23,25 @@ extension Persistable {
...
+        let valueType = type(of: value)
+        let valueConverter = SQLiteModelValueConverter()
+        do {
+            let binding = try valueConverter.convertToTarget(from: value,
+                                                             fieldType: .from(type: valueType))
+            guard let validBinding = binding else {
+                preconditionFailure("""
+                Converting \(String(describing: value)) of type \(String(describing: valueType))
+                to a SQLite Binding returned a nil value. This is likely a bug in the
+                SQLiteModelValueConverter logic.
```
palpatim: I think we have a canned "search for issues or file one on our GitHub page" string that would be good to append to this

Amplify/Categories/DataStore/Model/Schema/ModelSchema+Definition.swift
```diff
@@ -7,36 +7,22 @@
...
     case int
     case double
     case date
     case dateTime
     case time
+    case timestamp
     case bool
-    case `enum`(Any.Type)
+    case `enum`(type: ModelEnum.Type)
+    case type(_ type: Codable.Type)
```
palpatim: Maybe `customType` or something a bit more descriptive? Do we need to differentiate that this is a `nonModel` (or something that actually reads better in English)?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/ModelValueConverter+SQLite.swift
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+}
+
+public struct SQLiteModelValueConverter: ModelValueConverter {
+
+    public typealias SourceType = Any?
+    public typealias TargetType = Binding?
+
+    public func convertToTarget(from source: Any?, fieldType: ModelFieldType) throws -> Binding? {
```
palpatim: Since these don't rely on any internal state, consider making them static methods, to avoid an unnecessary instantiation at the call site

Amplify/Categories/DataStore/Model/Schema/ModelSchema+Definition.swift
drochetti: I like `customType`. The other options I discussed with the JS team were:
- `embeddedType`
- `nestedType`

"non-model" is how we've been referring to it internally due to the lack os agreement on the naming :D

Thoughts on `.customType` vs `.embeddedType` vs `.nestedType`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/ModelValueConverter+SQLite.swift
drochetti: good call. I initially had an internal state but then refactored it to get rid of the state and didn't change it to static. Will do.

Amplify/Categories/DataStore/Model/Model+Enum.swift
drochetti: @palpatim as we talked offline, I tried making enums just conform to `Codable` and not rely on any specific protocol. I faced two issues:

1. Given the `Codable` freedom, users could have their enums being represented by Int instead of String, which then breaks our logic to save it to a `TEXT` column in SQLite. It would be possible but our code that handle `Any` type would become more complex.
2. Related to the previous point, there's one place in our code (which I intend to simplify, but it will require work) that does a `Any.Type` -> `ModelFieldType` conversion, then I had no way to tell `.type` (or `.customType`) from `.enum` unless I drop `.enum` entirely.

Those two would be solved by an `isEnum(Any.Type)` or `is ValueEnumerable.Type` (like a [proposal on Swift Evolution](https://github.com/apple/swift-evolution/pull/114#issuecomment-336590026)). But for now I believe we will need a custom protocol in order to keep our serialization code sane and move forward.

So my suggestion is we go with a type like the `ModelEnum` I defined in the first version of this PR, that defines a more strict contract for enums (i.e. it has to be represented by a `String`) and find a better name for it like you suggested. Options that I discussed with you and some other members of the team:

1. RawStringRepresentable
2. StringRepresentable
3. PersistableEnum

Thoughts?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Model+SQLite.swift
lawmicha: we have this one https://github.com/aws-amplify/amplify-ios/blob/461c3cfd8923fbbb6eb10b60495b0f4594879cf4/Amplify/Core/Support/AmplifyErrorMessages.swift

Amplify/Core/Support/Encodable+AnyEncodable.swift
drochetti: 1. Done
2. `AnyDecodable` was not needed so far, so I decided to only introduce it when/if needed

Amplify/Categories/DataStore/Model/Model+Enum.swift
palpatim: Thanks for the thoughtful due diligence on this. After discussing and my own belated realization that this is primarily a codegen/model-authoring structure, I'd favor a name like `ModelValueStringRepresentable` or something that makes it (painfully) obvious that we really only want to use this in a narrow set of circumstances. :)

Amplify/Categories/DataStore/Model/Schema/ModelSchema+Definition.swift
palpatim: "embedded" & "nested" have implications that their definitions are somehow self-contained within the parent type. I'd favor "customType", which would stand in opposition to "modelType".

Amplify/Categories/DataStore/Model/Model+Enum.swift
wooj2: Hmmm imho `ModelEnum`... doesn't sound that bad, considering we have models that conform to `Model`, so I'm not quite sure why this was veto'ed.

But if we are looking at this holistically, and considering it's a protocol (and protocols should be `-able`), and riffing off of one of drochetti's suggestions, how about `EnumPersistable`?

drochetti: Thanks for reviewing and the feedback John. Let's do this:
- I really believe (and hope) that Swift will add some type of marker protocol do all enums in the future given the community engagement in such feature (`ValueIterable` is the current proposal as you can see here: https://github.com/apple/swift-evolution/pull/114#issuecomment-336590026). So this is really something that I hope we can remove in the future
- The protocol we're introducing can (should) only be used in enums, so I like the idea of naming it in a way we communicate that constraint, so `EnumPersistable` sounds good to me
- Our goal is to remove this protocol in the future in favor of either: `Codable` or `ValueIterable`, so I think we all agree on a future refactor of this
- Last but not least, this is controlled by the codegen provided by the CLI since handwritten models is not something we support yet (it's possible but there's no documentation around it), so if we decide to rename it in the future the impact is not too big. We can also deprecate it so customer don't break, etc

PR #338, opened 2020-02-25
Fix crash when retry handler attemptNumber is >= 57

wooj2:
Issue was reported here for app sync:
https://github.com/awslabs/aws-mobile-appsync-sdk-ios/issues/339

Since we are using very similar logic, this needs to get updated for datastore as well.

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/RequestRetryablePolicyTests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/RequestRetryablePolicyTests.swift
@@ -188,6 +188,39 @@ class RequestRetryablePolicyTests: XCTestCase {
         XCTAssertEqual(retryAdvice.retryInterval, defaultTimeout)
     }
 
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/RequestRetryablePolicyTests.swift
```diff
@@ -188,6 +188,39 @@ class RequestRetryablePolicyTests: XCTestCase {
...
+
+        let retryAdvice = retryPolicy.retryRequestAdvice(urlError: retryableErrorCode,
+                                                         httpURLResponse: nil,
+                                                         attemptNumber: 31)
+
+        XCTAssert(retryAdvice.shouldRetry)
+        assertMilliseconds(retryAdvice.retryInterval, greaterThan: 214_748_364_800, lessThan: 214_748_364_900)
+    }
+
+    func testBeyonMaxValueRetryDelay() {
```
wooj2: ugh, typo. needs to be "Beyond"

PR #353, opened 2020-03-11
Datastore clear() sync implementation

wooj2:
. clear() is an async interface
. After calling clear() on DataStore, the datastore is in an idle state -- that is, the remote sync engine has stopped, all subscriptions have been severed and any publishers the customer may have been listening were called with receiveCompletion() (meaning, they will need to re-subscribe)
. Any subsequent call to query, save, delete, publisher() will restart the sync engine and publishers, mirroring JS's behavior

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
@@ -13,7 +13,7 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
     public func save<M: Model>(_ model: M,
                                completion: @escaping DataStoreCallback<M>) {
         log.verbose("Saving: \(model)")
-
+        reinitStorageEngineIfNeeded()
         // TODO: Refactor this into a proper request/result where the result includes metadata like the derived
         // mutation type
         let modelExists: Bool
@@ -55,6 +55,7 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
     public func query<M: Model>(_ modelType: M.Type,
                                 byId id: String,
                                 completion: DataStoreCallback<M?>) {
+        reinitStorageEngineIfNeeded()
         let predicate: QueryPredicateFactory = { field("id") == id }
         query(modelType, where: predicate) {
             switch $0 {
@@ -74,6 +75,7 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
     public func query<M: Model>(_ modelType: M.Type,
                                 where predicateFactory: QueryPredicateFactory? = nil,
                                 completion: DataStoreCallback<[M]>) {
+        reinitStorageEngineIfNeeded()
         storageEngine.query(modelType,
                             predicate: predicateFactory?(),
                             completion: completion)
@@ -82,6 +84,7 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
     public func delete<M: Model>(_ modelType: M.Type,
                                  withId id: String,
                                  completion: @escaping DataStoreCallback<Void>) {
+        reinitStorageEngineIfNeeded()
         storageEngine.delete(modelType,
                              withId: id,
                              completion: completion)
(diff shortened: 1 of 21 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -82,13 +84,15 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
     public func delete<M: Model>(_ modelType: M.Type,
                                  withId id: String,
                                  completion: @escaping DataStoreCallback<Void>) {
+        reinitStorageEngineIfNeeded()
```
jamesonwilliams: Instead of doing it lazy-init-style here, can you do it eagerly, right before your `completion(.successfulVoid)` calls inside of `clear(...)`?

wooj2: Hmm! I would have liked to do what you suggested.. hmmm but... 

Based on the implementation of reinitStorageEngineIfNeeded(), I don't think this is what we want, because then we would be calling `storage.startSync()` which would attempt to restart the entire SyncEngine (so... then we'd potentially rehydrate the data base we just deleted)

If reinitStorageEngine were to be changed to only resolveTheStorageEngine (and not start the sync engine), we'd still have to call to startSyncEngine on any query/delete/subscription.  At this point, i'm not seeing a huge benefit one way or the other.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -124,6 +129,41 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
...
+            // continue to attempt to kill the storage engine & publishers
+            self.storageEngine = nil
+            if #available(iOS 13.0, *) {
+                if let publisher = self.dataStorePublisher as? DataStorePublisher {
+                    publisher.sendFinished()
+                }
+            }
+            self.dataStorePublisher = nil
+
+            if let error = sError {
```
palpatim: Why not just invoke `completion(result)`? You should then be able to get rid of L141-147 as well

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -223,6 +226,23 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
...
         }
     }
+
+    func clear(completion: @escaping DataStoreCallback<Void>) {
+        guard let dbFilePath = dbFilePath else {
+            log.error("Attempt to clear DB, but file path was empty")
+            completion(.failure(causedBy: DataStoreError.invalidDatabase(path: "Database path not set", nil)))
+            return
+        }
+        connection = nil
```
palpatim: I don't remember the SQLite API, but is there a cancel or teardown that we should invoke before releasing our connection?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
wooj2: WHOA!  How did i not see that?!  Thank you!

updated to:
```
            self.storageEngine = nil
            if #available(iOS 13.0, *) {
                if let publisher = self.dataStorePublisher as? DataStorePublisher {
                    publisher.sendFinished()
                }
            }
            self.dataStorePublisher = nil
            completion(result)
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
wooj2: I looked in the following link, but didn't see anything:
https://github.com/stephencelis/SQLite.swift/blob/master/Documentation/Index.md

Hmm but, the way I arrived at this approach (`connection = nil`) is that i noticed that we needed to a call to `sqlite3_close()` -- which seems to be in the `deinit` of the `Connection` class.  So I figured this was how to 'teardown' the connection.

I'm going to leave this as-is for now.

PR #359, opened 2020-03-16
Set excludeFromBackup flag to avoid backing up sqlite3 db

wooj2:
royjit@ - Thank you so much for calling this out!!

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
@@ -20,12 +20,16 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
         guard let documentsPath = getDocumentPath() else {
             preconditionFailure("Could not create the database. The `.documentDirectory` is invalid")
         }
-        let dbFilePath = documentsPath.appendingPathComponent("\(databaseName).db")
+        var dbFilePath = documentsPath.appendingPathComponent("\(databaseName).db")
         let path = dbFilePath.absoluteString
 
         let connection: Connection
         do {
             connection = try Connection(path)
+
+            var urlResourceValues = URLResourceValues()
+            urlResourceValues.isExcludedFromBackup = true
+            try dbFilePath.setResourceValues(urlResourceValues)
         } catch {
             throw DataStoreError.invalidDatabase(path: path, error)
         }


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -20,12 +20,16 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
         guard let documentsPath = getDocumentPath() else {
             preconditionFailure("Could not create the database. The `.documentDirectory` is invalid")
```
palpatim: Out of curiosity, why do we preconditionFailure here instead of throwing? (I know this isn't touched in this PR, just wondering while I'm reviewing)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -20,12 +20,16 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
...
         }
-        let dbFilePath = documentsPath.appendingPathComponent("\(databaseName).db")
+        var dbFilePath = documentsPath.appendingPathComponent("\(databaseName).db")
         let path = dbFilePath.absoluteString
 
         let connection: Connection
         do {
             connection = try Connection(path)
+
+            var setValues = URLResourceValues()
```
palpatim: Kind of a weird name--it sounds like it wants to be a setter method.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
wooj2: Seems like the author wanted to completely halt the execution of the program since this seems like an unrecoverable error that we would have no control over?

If our interface changed to allow users to specify what path to save the db in, then maybe throwing would be more appropriate?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
wooj2: Hmm. thanks, changed to:
`urlResourceValues`

PR #361, opened 2020-03-17
Fix DataStore E2E test - testCreateMutateDelete

lawmicha:
*Description of changes:*
`DataStoreEndToEndTests.testCreateMutateDelete` was broken for two reasons:
- SelectionSet string generation logic was incorrect, fixed with added unit test
- MutationEvent does not contain model for delete mutationTypes, and the id of the model should be retrieved from mutationEvent.modelId

This also adds a README for setting up the DataStore tests and made sure the tests are passing.

### DIFF
--- a/AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLDocument/GraphQLSyncQueryTests.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLDocument/GraphQLSyncQueryTests.swift
@@ -32,7 +32,7 @@ class GraphQLSyncQueryTests: XCTestCase {
     ///     - - it contains an `filter` argument of type `ModelPostFilterInput`
     ///     - it is named `syncPosts`
     ///     - it has a list of fields with no nested models
-    func testSyncGraphQLQueryFromSimpleModel() {
+    func testSyncGraphQLQueryForPost() {
         let post = Post.keys
         let predicate = post.id.eq("id") && (post.title.beginsWith("Title") || post.content.contains("content"))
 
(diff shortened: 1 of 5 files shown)


AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLDocument/GraphQLSyncQueryTests.swift
```diff
@@ -75,4 +75,43 @@ class GraphQLSyncQueryTests: XCTestCase {
         XCTAssertEqual(document.variables["lastSync"] as? Int, 123)
     }
 
+    func testSyncGraphQLQueryForComment() {
+        var documentBuilder = ModelBasedGraphQLDocumentBuilder(modelType: Comment.self, operationType: .query)
+        documentBuilder.add(decorator: DirectiveNameDecorator(type: .sync))
+        documentBuilder.add(decorator: PaginationDecorator(limit: 100, nextToken: "token"))
+        documentBuilder.add(decorator: ConflictResolutionDecorator(lastSync: 123))
```
wooj2: One thing I noticed is odd is that if i re-arrange the decorators, the unit tests break. e.g.
```
        documentBuilder.add(decorator: ConflictResolutionDecorator(lastSync: 123))
        documentBuilder.add(decorator: DirectiveNameDecorator(type: .sync))
        documentBuilder.add(decorator: PaginationDecorator(limit: 100, nextToken: "token"))
```
will not generate the `startedat` field.  I suspect that the change you made didn't impact this though.

wooj2: What's even more odd to me is that if i don't include any of the decorators, the queryDocument looks wildly different.  I'm not super familiar with this portion of datastore -- is this expected?

lawmicha: yes the decorators are not that smart per-say, you have to arrange them in the correct order to achieve what you want. there's inheritly a dependency chain that's not enforced to allow us to iterate on this faster as we're still figuring out what we need to support. For the 80% use case, there are extensions to create `GraphQLRequest` here https://github.com/aws-amplify/amplify-ios/tree/master/AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest so in most cases, the developer does not need to arrange the decorators in the right way.

lawmicha: DataStore will use these https://github.com/aws-amplify/amplify-ios/blob/master/AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest%2BAnyModelWithSync.swift#L61

PR #362, opened 2020-03-17
DataStore.save() with condition for remote storage

lawmicha:
**Description of changes**
- Add Codable conformance to QueryPredicate to allow encodinig/decoding to JSON, stored in MutationEvent. This is accomplished with `AnyQueryPredicate` for encoding array of predicates, and custom Codable for `QueryOperator`.
- Storing `queryPredicateJson` in MutationEvent to allow decoding to `QueryPredicate` for 1. sending API requests with condition input on create/update mutations, and 2. handle conditional mutations in the `AWSMutationDatabaseAdapter: MutationEventIngester`.
-  `AWSMutationDatabaseAdapter` resolves conflicts for incoming candidate mutation events with conditions by saving the candidate. (does not drop/replace local with candidate, etc)

**DRAFT**
- This is written on top of existing PR for DataStore.save() with conditiion for local store https://github.com/aws-amplify/amplify-ios/pull/355 . we will merge this into 355 as single PR.

- Need to check for successful response from API, and that the response contains graphQL Errors like "Conditional save failed", then retrieve from remote, and update local storage

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
@@ -82,4 +82,77 @@ class DataStoreEndToEndTests: SyncEngineIntegrationTestBase {
 
         wait(for: [deleteReceived], timeout: networkTimeout)
     }
+
+    /// - Given: A post that has been saved
+    /// - When:
+    ///    - attempt to update the existing post with a condition does not match existing data
+    ///    - attempt to update the existing post with a condition that matches existing data
+    /// - Then:
+    ///    - the update with condition that does not match will return existing post data, not updated.
+    ///    - the update with condition that matches existing data will be applied and returned.
+    func testCreateThenMutateWithCondition() throws {
+        try startAmplifyAndWaitForSync()
+
+        let post = Post.keys
+        let date = Date()
+        let title = "This is a new post I created"
+        let newPost = Post(
+            title: title,
+            content: "Original content from DataStoreEndToEndTests at \(date)",
+            createdAt: date)
+
+        var updatedPost = newPost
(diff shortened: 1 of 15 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
```diff
@@ -82,4 +82,77 @@ class DataStoreEndToEndTests: SyncEngineIntegrationTestBase {
...
+
+                if mutationEvent.mutationType == GraphQLMutationType.create.rawValue {
+                    XCTAssertEqual(post.content, post.content)
+                    XCTAssertEqual(mutationEvent.version, 1)
+                    createReceived.fulfill()
+                    return
+                }
+
+                if mutationEvent.mutationType == GraphQLMutationType.update.rawValue && mutationEvent.version == 2 {
+                    XCTAssertEqual(post.content, newPost.content)
```
lawmicha: in JS, when there is an existing model, but condition fails, then "Conditional Update failed", we do not return success with the model unmodified.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
```diff
@@ -82,4 +82,77 @@ class DataStoreEndToEndTests: SyncEngineIntegrationTestBase {
...
+            return
+        }
+
+        Amplify.DataStore.save(newPost) { _ in }
+
+        wait(for: [createReceived], timeout: networkTimeout)
+
+        Amplify.DataStore.save(updatedPost, where: post.title == "some random title") { _ in }
+
+        wait(for: [updateWithNoChangeReceived], timeout: networkTimeout)
```
lawmicha: reoworking this test. current behavior from local storage work is that it will fail

PR #365, opened 2020-03-24
add pagination support to DataStore

drochetti:
**Notes:**

Introduce pagination support to the `DataStore.query` API.

- `QueryPaginationInput` holds the pagination values
- the `DataStore` plugin translates that struct to SQLite limit/offset
- add tests to make sure the new API behaves correctly under different scenarios

### DIFF
--- a/Amplify/Categories/DataStore/Query/QueryPaginationInput.swift
+++ b/Amplify/Categories/DataStore/Query/QueryPaginationInput.swift
@@ -10,16 +10,26 @@ import Foundation
 /// A simple struct that holds pagination information that can be applied queries.
 public struct QueryPaginationInput {
 
-    /// The default page size
-    public static var defaultLimit: UInt = 100
+    /// The default page size.
+    public static let defaultLimit: UInt = 100
 
+    /// The page number. It starts at 0.
     public let page: UInt
+    
+    /// The number of results per page.
     public let limit: UInt
 
 }
 
 extension QueryPaginationInput {
 
+    /// Creates a `QueryPaginationInput` in an expressive way, enabling a short
+    /// and developer friendly access to an instance of `QueryPaginationInput`.
+    ///
+    /// - Parameters:
+    ///   - page: the page number (starting at 0)
+    ///   - limit: the page size (defaults to `QueryPaginationInput.defaultLimit`)
+    /// - Returns: a new instance of `QueryPaginationInput`
     public static func page(_ page: UInt,
                             limit: UInt = QueryPaginationInput.defaultLimit) -> QueryPaginationInput {
         return QueryPaginationInput(page: page, limit: limit)
--- a/Amplify/Categories/DataStore/Query/QueryPaginationInput.swift
+++ b/Amplify/Categories/DataStore/Query/QueryPaginationInput.swift
@@ -1,0 +1,38 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// A simple struct that holds pagination information that can be applied queries.
+public struct QueryPaginationInput {
+
+    /// The default page size
+    public static var defaultLimit: UInt = 100
+
+    public let page: UInt
+    public let limit: UInt
+
+}
+
+extension QueryPaginationInput {
+
+    public static func page(_ page: UInt,
+                            limit: UInt = QueryPaginationInput.defaultLimit) -> QueryPaginationInput {
+        return QueryPaginationInput(page: page, limit: limit)
+    }
+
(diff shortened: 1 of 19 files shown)


Amplify/Categories/DataStore/Query/QueryPaginationInput.swift
```diff
@@ -0,0 +1,38 @@
...
+
+import Foundation
+
+/// A simple struct that holds pagination information that can be applied queries.
+public struct QueryPaginationInput {
+
+    /// The default page size
+    public static var defaultLimit: UInt = 100
+
+    public let page: UInt
```
palpatim: Add doc comment, especially to note that this is 0 based.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/AWSMutationDatabaseAdapter/AWSMutationDatabaseAdapter+MutationEventSource.swift
```diff
@@ -21,13 +21,15 @@ extension AWSMutationDatabaseAdapter: MutationEventSource {
         let fields = MutationEvent.keys
         let predicate = fields.inProcess == false || fields.inProcess == nil
 
+        // TODO remove this in favor of proper sorting/orderBy API
```
palpatim: Might be worthwhile for us to start linking backlog items into our TODOs, so we don't lose track of them

drochetti: Good call and I actually thought about it. We do have an item for this one, but it's a Pivotal story, which is not public. Do you think we should make these GitHub issues?

Amplify/Categories/DataStore/Query/QueryPaginationInput.swift
```diff
@@ -0,0 +1,38 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// A simple struct that holds pagination information that can be applied queries.
+public struct QueryPaginationInput {
+
+    /// The default page size
+    public static var defaultLimit: UInt = 100
```
wooj2: Seems like this should be a `let` not a `var` -- if customers want to change the limit, seems like we should force them through calling the `page` function, rather than allow them to update a `defaultLimit` (which can be changed throughout the life of an app)

Amplify/Categories/DataStore/Query/QueryPaginationInput.swift
```diff
@@ -0,0 +1,38 @@
...
+    public let page: UInt
+    public let limit: UInt
+
+}
+
+extension QueryPaginationInput {
+
+    public static func page(_ page: UInt,
+                            limit: UInt = QueryPaginationInput.defaultLimit) -> QueryPaginationInput {
+        return QueryPaginationInput(page: page, limit: limit)
```
wooj2: Hmm.. in the case where we have `page(0, limit: 0)` -- we could throw an exception.. or log some sort of error.. or.. just do nothing.
Seems like nothing could be okay -- what do you think?

Also, do we care to support an 'unlimited' parameter? (e.g. they want to specify a page, but specify an unlimited parameter)

Amplify/Categories/DataStore/Query/QueryPaginationInput.swift
drochetti: Good call!

Amplify/Categories/DataStore/Query/QueryPaginationInput.swift
drochetti: > Hmm.. in the case where we have page(0, limit: 0) -- we could throw an exception.. or log some sort of error.. or.. just do nothing.
Seems like nothing could be okay -- what do you think?

I'm willing to allow customers to do that if they want, like databases usually allow.

> Also, do we care to support an 'unlimited' parameter? (e.g. they want to specify a page, but specify an unlimited parameter)

How does a page is defined without knowing its size? That could work in case of an `offset` is used instead of a `page`, but since we  decided to calculate the offset based on the `page` and `limit`, both are needed (i.e. what range page 2 belongs to if we don't know the page size?).

wooj2: regarding page(0, limit:0) -- i was worried about unintentionally sending those values in.. but.. probably something we don't need to spend time debating.. i think it's a case where customers just need to be careful about what they're doing. :).

Regarding .unlimited:
Oops brain fart: I was thinking they might call page(0, limit:100), and then decide that based on those first 100 records, to then call page(1, limit: unlimited) -- but that doesn't make sense :)

PR #369, opened 2020-04-01
add DataStore configuration API

drochetti:
## DataStore Configuration

DataStore requires some configuration. This API is heavily based and aligned with the current JavaScript implementation: https://aws-amplify.github.io/docs/js/datastore#config.

#### Default:

```swift
// default configuration can be omitted
Amplify.add(plugin: AWSDataStorePlugin(schema: schema, configuration: .default))
```

#### Custom - only required option:

```swift
// custom configuration (only conflictHandler is required)
Amplify.add(plugin: AWSDataStorePlugin(schema: schema, configuration: .custom(
    conflictHandler: { (data, resolve) in
        let (local, remote) = data
        // implement conflict resolution logic
        // then call the resolve callback with the result
        resolve(.new(mergedModel))
    }
)))
```

#### Custom - all options (and their respective defaults):

```swift
// custom configuration with all options
Amplify.add(plugin: AWSDataStorePlugin(schema: schema, configuration: .custom(
    conflictHandler: { (data, resolve) in
        .resolve(.discard) // same as .resolve(.new(data.remote))
    },
    errorHandler: { error in
        Amplify.Logging.error(error: error)
    },
    syncInterval: 1_440,
    syncMaxRecords: 10_000,
    syncPageSize: 1_000
)))
```

#### Configuration options

The JavaScript have the same options, but the sync-related configuration have different names. We decided to slightly change the property names so they have a common prefix: `sync*`. The equivalent options would be:

| JavaScript         | iOS              |
|--------------------|------------------|
| `maxRecordsToSync` | `syncMaxRecords` |
| `fullSyncInterval` | `syncInterval`   |
| `syncPageSize`     | `syncPageSize`   |

#### Conflict Handler

The `conflictHandler` in JavaScript relies on Promises (`async/await`) to allow asynchronous code to be executed inside the callback. The solution in Swift was to pass a `resolver` function to the callback so once the resolution code is done, the `resolver` is invoked. The approach suits both synchronous and asynchronous needs within the conflict resolution logic.


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
```diff
@@ -0,0 +1,103 @@
...
+        self.syncInterval = syncInterval
+        self.syncMaxRecords = syncMaxRecords
+        self.syncPageSize = syncPageSize
+    }
+
+}
+
+extension DataStoreConfiguration {
+
+    public static let defaultSyncInterval: UInt = 1_440
```
wooj2, line 67: 1440 = 24 * 60 = every 24 minutes?

drochetti: good catch, I think the configuration demands more documentation. That interval is in minutes. In JS that's not very clear either, unless you look at the source code. Good opportunity to improve it - should we adopt a `TimeInterval` approach, with `.minutes(60)`, `.hours(24)`, etc?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/APICategoryDependencyTests.swift
```diff
@@ -72,7 +72,7 @@ extension APICategoryDependencyTests {
                                           syncEngine: syncEngine)
 
         let dataStorePublisher = DataStorePublisher()
-        let dataStorePlugin = AWSDataStorePlugin(modelRegistration: TestModelRegistration(),
+        let dataStorePlugin = AWSDataStorePlugin(schema: TestDataStoreSchema(),
                                                          storageEngine: storageEngine,
                                                          dataStorePublisher: dataStorePublisher)
```
wooj2, line 77: spacing.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/LocalSubscriptionTests.swift
```diff
@@ -53,7 +53,7 @@ class LocalSubscriptionTests: XCTestCase {
         }
 
         let dataStorePublisher = DataStorePublisher()
-        let dataStorePlugin = AWSDataStorePlugin(modelRegistration: TestModelRegistration(),
+        let dataStorePlugin = AWSDataStorePlugin(schema: TestDataStoreSchema(),
```
wooj2, line 56: spacing.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/MutationQueue/AWSMutationEventIngesterTests.swift
```diff
@@ -43,7 +43,7 @@ class AWSMutationEventIngesterTests: XCTestCase {
                                               syncEngine: syncEngine)
 
             let publisher = DataStorePublisher()
-            let dataStorePlugin = AWSDataStorePlugin(modelRegistration: TestModelRegistration(),
+            let dataStorePlugin = AWSDataStorePlugin(schema: TestDataStoreSchema(),
```
wooj2, line 46: spacing.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/RemoteSync/RemoteSyncAPIInvocationTests.swift
```diff
@@ -49,7 +49,7 @@ class RemoteSyncAPIInvocationTests: XCTestCase {
         }
 
         let dataStorePublisher = DataStorePublisher()
-        let dataStorePlugin = AWSDataStorePlugin(modelRegistration: TestModelRegistration(),
+        let dataStorePlugin = AWSDataStorePlugin(schema: TestDataStoreSchema(),
```
wooj2, line 52: spacing.

Amplify/Categories/DataStore/Model/DataStoreSchema.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
 
 /// Protocol that defines a contract between the consumer and the DataStore plugin.
 /// All models have to be registered and have an associated `version`.
-public protocol AmplifyModelRegistration {
+public protocol DataStoreSchema {
```
lawmicha, line 12: How about `SchemaRegistry` or `ModelSchema`? Having "DataStore" in APIPlugin seems odd

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
```diff
@@ -0,0 +1,112 @@
...
+    ///   - conflictHandler: a callback called when a conflict could not be resolved by the service
+    ///   - syncInterval: how often the sync engine will run (in seconds)
+    ///   - syncMaxRecords: the number of records to sync per execution
+    ///   - syncPageSize: the page size of each sync execution
+    /// - Returns: an instance of `DataStoreConfiguration` with the passed parameters.
+    public static func custom(
+        errorHandler: @escaping DataStoreErrorHandler = { error in
+            Amplify.Logging.error(error: error)
+        },
+        conflictHandler: @escaping DataStoreConflictHandler,
```
wooj2, line 93: Seems like this line should be:
```
        conflictHandler: @escaping DataStoreConflictHandler = { _, resolve in
            resolve(.discard)
        },
```

In the event that a customer wants to take the default for both the errorHandler and conflictHandler, but wants other custom fields?
```
        let configuration = DataStoreConfiguration.custom(
           syncInterval: 1.0,
           syncMaxRecords: 5,
           syncPageSize: 10)
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
```diff
@@ -0,0 +1,112 @@
...
+import Amplify
+import Foundation
+
+/// Error Handler function typealias
+public typealias DataStoreErrorHandler = (AmplifyError) -> Void
+
+/// Holds a reference to both the local `Model` and the remote one during a conflict
+/// resolution. Implementations of the `DataStoreConflictHandler` use this to decide
+/// what the outcome of a conflict should be.
+public struct DataStoreConclictData {
```
lawmicha, line 17: typo here; `DataStoreConflictData`

lawmicha: I'm discovering there is a difference here from JS. API returns conflictUnhandled with error.data as the MutationSync<AnyModel> object. This means it gets deserialized into MutationSync containing the model and syncMetadata fields like `version` and `deleted`. In JS, this is returned as error.data with the `deleted` field on the model itself and passed to the confliict handler.

I'm guessing the developer can then decide what to do with the two models based on this information. The `DataStoreConflictData` doesn't pass the "is this model deleted?" information back.


Now that brings back to the question, what should DataStore do based on the result of the ConflictHandler?

If developer decides to `discard` the local changes, and the remote model is deleted, we should probably delete it from local store. On a side note, iif the remote model is not deleted, then we update local store with the remote model and update syncMetadata item in the local DB with the remote.version

If developer decides to `keep` the local changes, and the remote model is deleted, then we should call API.update(localModel, remote.version). the model is still marked as deleted after we update it. This could possibly be a bug, how do we possibly un-delete a model?

If the developer decides to give us a `new` model, then we should call API.update(newModel, remote.version), this is the same scenario as keeping the local changes. DataStore will call API and rely on the sync path to reconcile local changes.

if the conflict occurs at the service with the remote model deleted, should we still allow the client to resolve the conflict? If they decide to keep the local model, how do we "undo" a deleted remote model?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/// The `DataStore` plugin configuration object.
+public struct DataStoreConfiguration {
+
+    /// A callback function called on unhandled errors
+    public let errorHandler: DataStoreErrorHandler
+    
+    /// A callback called when a conflict could not be resolved by the service
+    public let conflictHandler: DataStoreConflictHandler
```
jamesonwilliams, line 49: This PR doesn't "wire" these into the code, right? We're just defining the attributes and usage of the `DataStoreConfiguration`, in this PR?

I think it looks good, if so.

We should layer these two handlers in some heft docs. One particular thing that might be worth noting the Apple doc is:
1. The conflict handler runs synchronously, and will block (what actions) until complete?

I like the error handler. I like plain callbacks in general. But, why are we using an error handler callback, instead of broadcasting errors to Hub? Or, will we do both? Does the error handler work synchronously, and block other operations, until complete? If so, which ones? Can we document all of this?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
jamesonwilliams: In Java, we have this awesome [`TimeUnit`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/TimeUnit.html) conversion utility, backed into the core library since Java 7. Is there not a Swift equivalent?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -18,7 +18,11 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
     /// The Publisher that sends mutation events to subscribers
     var dataStorePublisher: DataStoreSubscribeBehavior?
 
-    let modelRegistration: AmplifyModelRegistration
+    /// The schema that holds a reference to its version and all registered models
+    let schema: DataStoreSchema
+
+    /// The DataStore configuration
+    let configuration: DataStoreConfiguration
```
wooj2, line 25: On line72, in the function `configure(using configuration: Any)`, we pass in the amplify configuration (I think), so we might want to rename this to something like dataStoreConfiguration

Amplify/Categories/DataStore/Model/DataStoreSchema.swift
wooj2: You guys probably already realized this, but if we make this change, we'll need to update the codegen tools so that it generates:
```
final public class AmplifyModels: DataStoreSchema {
```
instead of what it currently geneates:
```
final public class AmplifyModels: AmplifyModelRegistration {
```

drochetti: > How about `SchemaRegistry` or `ModelSchema`? Having "DataStore" in APIPlugin seems odd

I understand the concern, but we need to keep in mind that the model-based API integration should be revisited. Right now API imports classes and protocols from the DataStore category, so the integration is there already. We need to think of ways of making that more isolated.

That said, `ModelSchema` cannot be used because it's already used by `Model`, as the schema of each `Model`. That said, I'll think of a better name for it, `SchemaRegistry` could be one.

drochetti: > Comments about aligning naming --
> 
> For your consideration, on Android, we have two similar components which are named.
> 
> 1. The [`ModelProvider`](https://github.com/aws-amplify/amplify-android/blob/master/core/src/main/java/com/amplifyframework/core/model/ModelProvider.java#L20-L24), and its code-generated `AmplifyModelProvider` ([an example of one](https://github.com/aws-amplify/amplify-android/blob/master/testmodels/src/main/java/com/amplifyframework/testmodels/commentsblog/AmplifyModelProvider.java#L10-L15)). This component is read-only and declares what the different types of models are, and a version for their contract, as a whole.
> 2. The [`ModelSchemaRegistry`](https://github.com/aws-amplify/amplify-android/blob/master/core/src/main/java/com/amplifyframework/core/model/ModelSchemaRegistry.java#L27-L30), which accepts a collection of model classes as inputs, and then constructs and maintains [`ModelSchema`](https://github.com/aws-amplify/amplify-android/blob/master/core/src/main/java/com/amplifyframework/core/model/ModelSchema.java#L44-L48) for each, by parsing Java annotations and using Java reflection, on the `@Model`-annotated classes.
> 
> To recall, as above, Android generates the schema at runtime, not compile-time. So, there are some architectural differences. But we can probably keep the names a little tighter between the platforms even so, if we try.

thanks @jamesonwilliams for the call-out. JavaScript simply calls it `schema` (using `initSchema(schema: Schema)`. Before iOS was using `ModelRegistration` which is odd, so this is an attempt to align iOS more with JS. I'll revisit the Android and JS implementations to see if we can agree on something. I'll get in touch with you directly before I submit an update to this PR.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
drochetti: thanks for pointing that out @lawmicha 
I'll merge this PR with that limitation, but feel free to update the `conflictHandler` signature to match your needs if you still need the extra data. After our conversation on Friday I was not 100% clear if we need the mutation sync field or just the models when resolving the conflicts.

Amplify/Categories/DataStore/Model/DataStoreSchema.swift
drochetti: @jamesonwilliams @wooj2 @lawmicha I updated to `ModelSchemaProvider`. Jameson, do you think it's OK to rename `ModelProvider` to `ModelSchemaProvider` on your end?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
drochetti: done!

Amplify/Core/Support/TimeInterval+Helper.swift
```diff
@@ -0,0 +1,28 @@
...
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+extension TimeInterval {
+
+    public static func seconds(_ value: Double) -> TimeInterval {
```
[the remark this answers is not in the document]
palpatim: The return value is in seconds (because `TimeInterval` is a Double that is used as a seconds value), but the arguments are in the units specified by the function name. In any case, the caller can't specify the argument at the call site, but rather passes the value bare to the method. Thus `TimeInterval.seconds(1)` returns 1 second; `TimeInterval.minutes(30)` returns 30*60=1800 seconds, and so on. 

This is an equivalent usage to the Foundation `DispatchTime` class, which offers convenience methods like `DispatchTime.minutes()`, etc, and enables constructs like `let t = DispatchTime(.now() + .seconds(30))`.

It's possible I'm misunderstanding your point :)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
palpatim: (I inadvertently deleted a comment here from @drochetti -- I've pinged him to reinstate it. Sorry for that!)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
```diff
@@ -0,0 +1,112 @@
...
+public typealias DataStoreConflictHandler = (DataStoreConclictData, DataStoreConflictHandlerResolver) -> Void
+
+/// Callback for the `DataStoreConflictHandler`.
+public typealias DataStoreConflictHandlerResolver = (DataStoreConflictHandlerResult) -> Void
+
+/// The conflict resolution result enum.
+public enum DataStoreConflictHandlerResult {
+
+    /// Discard the local changes in favor of the remote ones.
+    case discard
```
palpatim, line 33: We should be very explicit: If this discards local changes in favor of remote ones, does that mean updating the system, at the time of the conflict resolution handler, with the state of the remote item? If so then let's document that. Better yet, maybe more appropriate names would be: `applyRemote`, `retryLocal`, and `retry(Model)`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
wooj2: Based on the most recent changes, looks like we've adopted a TimeInterval approach, and looks good to me!. resolving.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
wooj2: lgtm!, resolving.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
wooj2: Thanks for bringing this up, and all valid points.  I've opened an internal ticket against us to complete this before GA.

Amplify/Core/Support/TimeInterval+Helper.swift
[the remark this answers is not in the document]
jamesonwilliams: Ah! I got it.

This says "Give me a time interval with this many seconds." As opposed to, "Convert this quantity of time (in what units?) _into_ seconds."

My brain was stuck in the Java `TimeUnit` utility:
```
// Converts 5,000 milliseconds (the provided value) into units of seconds. Returns 5.
TimeUnit.MILLISECONDS.toSeconds(5_000);
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
jamesonwilliams: @wooj2 I think the doco could go right into this PR, no? The error events over Hub might be a good follow up task.

Amplify/Categories/DataStore/Model/DataStoreSchema.swift
jamesonwilliams: Sorry, I may have accidentally just created more confusion. :-(

On Android, the CLI code-gens an `AmplifyModelProvider implements ModelProvider`; the `ModelProvider` interface returns a `Set<Class<? extends Model>>`.

At runtime, we access that `Set<Class<? extends Model>>`. For each `modelClass` in that set, we establish a 1:1 relationship to a `ModelSchema`.  We build the `ModelSchema`s in the library at run-time, by inspecting each `Class<? extends Model>` instance via reflection. We end up with a `List<ModelSchema>`, that is held in-memory in a 
 class called `ModelSchemaRegistry`.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
```diff
@@ -0,0 +1,112 @@
...
+
+/// Error Handler function typealias
+public typealias DataStoreErrorHandler = (AmplifyError) -> Void
+
+/// Holds a reference to both the local `Model` and the remote one during a conflict
+/// resolution. Implementations of the `DataStoreConflictHandler` use this to decide
+/// what the outcome of a conflict should be.
+public struct DataStoreConclictData {
+    public let local: Model
+    public let remote: Model
```
lawmicha, line 19: these have to be `MutationSync` for the developer to check if the local/remote was deleted

Amplify/Core/Support/TimeInterval+Helper.swift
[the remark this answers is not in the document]
wooj2: i am optimistically resolving this.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
wooj2: Lets handle this in a subsequent PR

lawmicha: i agree, currently i've made changes like
```
public struct DataStoreConclictData {
    public let local: LocalModelData
    public let remote: RemoteModelData
}

public enum LocalModelData {
    case update(Model)
    case delete(Model.Identifier)
}

public enum RemoteModelData {
    case update(Model)
    case delete(Model)
}
```

wooj2:
We will keep this PR around **FOR REFERENCE ONLY** until we resolve:
* AmplifyModelRegistration vs ModelSchemaProvider vs something else
* Renaming discard vs applyRemote, etc.

drochetti:
This PR was used as a base for discussion and work was divided into other PRs. Closing this one.

PR #379, opened 2020-04-08
0.11.0 Release

iartemiev:
* Adding `sleep` for 10 mins in between dependent pods. This is a temporary measure, as the CocoaPods team now have a [PR](https://github.com/CocoaPods/cocoapods-trunk/pull/147) for allowing pushing pods straight to the git repo (circumventing the CDN, which will obviate the need for the delay between pushes)

* Bump versions to 0.11.0

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -21,6 +21,8 @@ COCOAPODS_PUSH_WAIT_TIME = 600
 
 pods = [
   {spec: "Amplify.podspec", sleep: true},
+  # AmplifyPlugins depends on AWSPluginsCore, but skipping the sleep here
+  # since there's one after CoreMLPredictionsPlugin
   {spec: "AWSPluginsCore.podspec", sleep: false},
   {spec: "CoreMLPredictionsPlugin.podspec", sleep: true},
   {spec: "AWSPredictionsPlugin.podspec", sleep: false},
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -17,27 +17,28 @@ opt_out_usage
 
 default_platform(:ios)
 
+COCOAPODS_PUSH_WAIT_TIME = 6
+
 pods = [
-  "Amplify.podspec",
-  "AWSPluginsCore.podspec",
-  "CoreMLPredictionsPlugin.podspec",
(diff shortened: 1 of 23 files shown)


fastlane/Fastfile
```diff
@@ -17,27 +17,28 @@ opt_out_usage
...
+COCOAPODS_PUSH_WAIT_TIME = 600
+
 pods = [
-  "Amplify.podspec",
-  "AWSPluginsCore.podspec",
-  "CoreMLPredictionsPlugin.podspec",
-  "AWSPredictionsPlugin.podspec",
-  "AmplifyPlugins.podspec"
+  {spec: "Amplify.podspec", sleep: true},
+  {spec: "AWSPluginsCore.podspec", sleep: false},
```
lawmicha, line 24: Should we sleep after AWSPluginsCore as well? AmplifyPlugins depends on AWSPluginsCore

palpatim, line 24: `AmplifyPlugins` relies on `AWSPluginsCore`. Doesn't that mean `AWSPluginsCore` should `sleep`?

iartemiev: We're sleeping after CoreMLPredictionsPlugin, so that should be good enough, I believe, since CoreMLPredictionsPlugin comes before AmplifyPlugins

iartemiev: I think the `sleep` after CoreMLPredictionsPlugin should cover that (since these are synchronous)

palpatim: Yeah, what he said. :)

palpatim: I see. Since we're using this to model deployment steps, rather than dependencies, I'm OK with leaving it as-is, but let's add a note since it's easy to conflate the two and we don't want to accidentally lose the sleep if we edit this prior to the merge of the CocoaPods "synchronous support" PR.

iartemiev: Sounds good. Added a comment: https://github.com/aws-amplify/amplify-ios/pull/379/commits/63042bfff539e02ca614c99b034654dff2555beb#diff-56c5021ef06ac6cea807ae53269b6f4fR24-R25

PR #388, opened 2020-04-14
feat: Datastore use configurable syncMaxRecords and syncPageSize

wooj2:

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
@@ -20,7 +20,14 @@ final class InitialSyncOperation: AsynchronousOperation {
     private let modelType: Model.Type
     private let completion: AWSInitialSyncOrchestrator.SyncOperationResultHandler
 
-    private var lastSyncTime: Int?
+    private var recordsReceived: UInt
+
+    private var syncMaxRecords: UInt {
+        return dataStoreConfiguration.syncMaxRecords
+    }
+    private var syncPageSize: UInt {
+        return dataStoreConfiguration.syncPageSize
+    }
 
     init(modelType: Model.Type,
          api: APICategoryGraphQLBehavior?,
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
```diff
@@ -141,23 +152,23 @@ final class InitialSyncOperation: AsynchronousOperation {
...
             }
         } else {
-            updateModelSyncMetadata()
+            updateModelSyncMetadata(lastSyncTime: syncQueryResult.startedAt)
         }
 
     }
 
-    private func updateModelSyncMetadata() {
+    private func updateModelSyncMetadata(lastSyncTime: Int?) {
```
palpatim, line 171: I'm not seeing that you are using `lastSyncTime` in this method?

wooj2: On line 182.  But hmm.. seems like it's not a good idea to overlap the local function variable `lastSyncTime` with the instance variable of `lastSyncTime`.

We were previously always setting the `lastSyncTime`, which is a bug, because we can have a case where we need to paginate, and when we paginate through a base query, we do not want `lastSyncTime` to be populated.

But, this has made me think....to make things more clear, it seems like we don't need to keep `lastSyncTime` as an instance variable.

wooj2: Updated to remove instance variable, and will hopefully make things clearer! thanks for the input!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
```diff
@@ -43,34 +51,31 @@ final class InitialSyncOperation: AsynchronousOperation {
...
+            return nil
         }
 
         let lastSyncMetadata = getLastSyncMetadata()
         guard let lastSync = lastSyncMetadata?.lastSync else {
-            return
+            return nil
         }
 
         //TODO: Update to use TimeInterval.milliseconds when it is pushed to master branch
```
palpatim, line 69: Consider adding an item to our backlog & linking it here so we don't lose track of this TODO

wooj2: done.
https://github.com/aws-amplify/amplify-ios/issues/398

PR #390, opened 2020-04-16
Auth - SignIn with SRPA using wrapper

royjit:
*Description of changes:*

Implemented signIn operation for username / password flow. Internally this uses AWSMobileClient.

### DIFF
--- a/Amplify/Categories/Auth/Result/AuthSignInResult.swift
+++ b/Amplify/Categories/Auth/Result/AuthSignInResult.swift
@@ -13,7 +13,9 @@ public struct AuthSignInResult {
     ///
     /// When this value is false, it means that there are more steps to follow for the signIn flow. Check `nextStep`
     /// to understand the next flow. If `isSignedIn` is true, signIn flow has been completed.
-    public var isSignedIn: Bool { nextStep == .done }
+    public var isSignedIn: Bool {
+        nextStep == .done
+    }
 
     /// Shows the next step required to complete the signIn flow.
     ///
--- a/Amplify/Categories/Auth/Result/AuthSignInResult.swift
+++ b/Amplify/Categories/Auth/Result/AuthSignInResult.swift
@@ -13,18 +13,15 @@ public struct AuthSignInResult {
     ///
     /// When this value is false, it means that there are more steps to follow for the signIn flow. Check `nextStep`
     /// to understand the next flow. If `isSignedIn` is true, signIn flow has been completed.
-    public let isSignedIn: Bool
+    public var isSignedIn: Bool { nextStep == .done }
 
     /// Shows the next step required to complete the signIn flow.
     ///
-    /// This value will be nil if `isSignedIn` is true.
-    public var nextStep: AuthSignInStep?
+    public var nextStep: AuthSignInStep
 
     //TODO: Add Code delivery #172336364
 
-    public init(isSignedIn: Bool,
-                 nextStep: AuthSignInStep? = nil) {
-        self.isSignedIn = isSignedIn
+    public init(nextStep: AuthSignInStep) {
         self.nextStep = nextStep
     }
 }
@@ -57,4 +54,8 @@ public enum AuthSignInStep {
     /// Auth step require the user to setup their MFA preferences
     ///
     case mfaSetup
+
+    /// There is no next step and the signIn flow is complete
+    ///
+    case done
 }
--- a/Amplify/Categories/Auth/Result/AuthSignInResult.swift
+++ b/Amplify/Categories/Auth/Result/AuthSignInResult.swift
@@ -9,9 +9,52 @@ import Foundation
 
 public struct AuthSignInResult {
 
-    let authInfo: AmplifyAuthInformation
+    /// Informs whether the user is signedIn or not.
+    ///
+    /// When this value is false, it means that there are more steps to follow for the signIn flow. Check `nextStep`
+    /// to understand the next flow. If `isSignedIn` is true, signIn flow has been completed.
+    public let isSignedIn: Bool
 
-    public init(authInfo: AmplifyAuthInformation) {
-        self.authInfo = authInfo
+    /// Shows the next step required to complete the signIn flow.
+    ///
+    /// This value will be nil if `isSignedIn` is true.
+    public var nextStep: AuthSignInStep?
+
+    //TODO: Add Code delivery #172336364
+
+    public init(isSignedIn: Bool,
+                 nextStep: AuthSignInStep? = nil) {
+        self.isSignedIn = isSignedIn
+        self.nextStep = nextStep
     }
+}
+
+/// Auth SignIn flow steps
+public enum AuthSignInStep {
+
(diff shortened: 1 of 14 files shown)


Amplify/Categories/Auth/Result/AuthSignInResult.swift
```diff
@@ -9,9 +9,52 @@ import Foundation
...
+    /// When this value is false, it means that there are more steps to follow for the signIn flow. Check `nextStep`
+    /// to understand the next flow. If `isSignedIn` is true, signIn flow has been completed.
+    public let isSignedIn: Bool
 
-    public init(authInfo: AmplifyAuthInformation) {
-        self.authInfo = authInfo
+    /// Shows the next step required to complete the signIn flow.
+    ///
+    /// This value will be nil if `isSignedIn` is true.
+    public var nextStep: AuthSignInStep?
```
palpatim, line 21: It would be better to model this as an enum, so that states that don't have a next step won't even need to inspect this. Something like this:

```swift
public struct AuthSignInResult {
    let authInfo: AmplifyAuthInformation
    public let disposition: AuthSignInResultDisposition
    public var isSignedIn: Bool {
        disposition == .complete // Will no next steps always mean we're signed in, or could it represent an error state?
    }
}

public enum AuthSignInResultFlow (flow? disposition? nextStep?)  {
    case complete
    case inProcess(AuthSignInStep)
}
```

There may even be a way to model this as part of the AuthSignInStep?

AmplifyPlugins/Auth/AWSAuthPlugin/AWSAuthPlugin+ClientBehavior.swift
```diff
@@ -31,18 +31,27 @@ extension AWSAuthPlugin {
...
+                                        password: password,
+                                        options: options)
+        let signInOperation = AWSAuthSignInOperation(request,
+                                                     authenticationProvider: authenticationProvider,
+                                                     listener: listener)
+        queue.addOperation(signInOperation)
+        return signInOperation
     }
 
     public func signInWithSocial(provider: AuthSocialProvider,
```
palpatim, line 49: I know we haven't done an API naming yet, but this would be Swiftier as `signIn(withSocialProvider provider: AuthSocialProvider...`

Amplify/Categories/Auth/Result/AuthSignInResult.swift
royjit: Yes, like the idea... but, it will take a lot of work from the developer to implement it. With all the nested switch cases it looks ugly.

```
 _ = Amplify.Auth.signIn(username: username, password: password) { (event) in
       switch event {
            case .completed(let result):
                switch result.disposition {
                case inProgress(let step):
                    switch step: {
                        case .sms:
                        print("SMS Step")
                    }
                case completed:
                    print("Completed singIn flow")
                }

            case .failed(let error):
                self.handleError(error: error)
            default:
                print("Default")
            }
```

AmplifyPlugins/Auth/AWSAuthPlugin/AWSAuthPlugin+ClientBehavior.swift
royjit: Yes, will update after the naming.

Amplify/Categories/Auth/Result/AuthSignInResult.swift
palpatim: I'm not seeing much difference in ugliness, and the current way leaves the ambiguous condition akin to the current pattern of returning optional results and errors, requiring customers to inspect both. Am I misunderstanding how the current proposal would be implemented by the customer?

```swift
 _ = Amplify.Auth.signIn(username: username, password: password) { (event) in
       switch event {
            case .completed(let result):
                if !isSignInCompleted and let nextStep = result.nextStep {
                    switch step: {
                        case .sms:
                        print("SMS Step")
                    }
                } else if isSignInCompleted {
                    print("Completed singIn flow")
                } else {
                    print("Unexpected condition: sign in isn't completed, but we didn't get a next step")
                }
            case .failed(let error):
                self.handleError(error: error)
            default:
                print("Default")
            }
```

royjit: Hmm.. yea, the current implementation looks like this:


```swift
 _ = Amplify.Auth.signIn(username: username, password: password) { (event) in

            switch event {
            case .completed(let result):

                if (result.isSignedIn) {
                    print("Completed signIn flow")
                } else {

                    switch result.nextStep! {
                    case .smsMFAChallenge:
                        print("SMS Step")
                    default:
                        print("")
                    }
                }

            case .failed(let error):
                self.handleError(error: error)
            default:
                print("Default")
            }
}
```

Another option that we discussed was to add a `.done` case  in the enum `AuthSignInStep` and make the nextStep in result non-optional. I guess that looks little more cleaner?

```swift

        _ = Amplify.Auth.signIn(username: username, password: password) { (event) in

            switch event {
            case .completed(let result):

                switch result.nextStep {
                    
                case .smsMFAChallenge:
                    print("SMS step")
                case .done:
                    print("SignIn complete")
                default:
                    print("Default")
                }

            case .failed(let error):
                self.handleError(error: error)
            default:
                print("Default")
            }
        }

```

palpatim: Yeah, that force-unwrap isn't going to fly.

The `done` case in `AuthSignInStep` was what I was thinking of in my original comment about using that enum. I think that looks like the right way to go, then you don't need an `isComplete` storage value. If needed for customer convenience, you can make it a computed property `result.nextStep == .done`

royjit: Cool, thanks! I will make the change.

@TrekSoft Like you said we might need to add `.done` in the enum list.

Amplify/Categories/Auth/Result/AuthSignInResult.swift
```diff
@@ -13,18 +13,15 @@ public struct AuthSignInResult {
     ///
     /// When this value is false, it means that there are more steps to follow for the signIn flow. Check `nextStep`
     /// to understand the next flow. If `isSignedIn` is true, signIn flow has been completed.
-    public let isSignedIn: Bool
+    public var isSignedIn: Bool { nextStep == .done }
```
palpatim, line 16: nit, formatting should be on separate lines

Amplify/Categories/Auth/Result/AuthSignInResult.swift
```diff
@@ -13,18 +13,15 @@ public struct AuthSignInResult {
...
 
     /// Shows the next step required to complete the signIn flow.
     ///
-    /// This value will be nil if `isSignedIn` is true.
-    public var nextStep: AuthSignInStep?
+    public var nextStep: AuthSignInStep
```
palpatim, line 20: nit: this is no longer a `nextStep`, let's make sure to cover this during API review

PR #392, opened 2020-04-18
Mark outgoing mutation as inProcess if nextEventPromise exists

wooj2:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/AWSMutationDatabaseAdapter/AWSMutationDatabaseAdapter+MutationEventIngester.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/AWSMutationDatabaseAdapter/AWSMutationDatabaseAdapter+MutationEventIngester.swift
@@ -202,7 +202,11 @@ extension AWSMutationDatabaseAdapter: MutationEventIngester {
               completionPromise: @escaping Future<MutationEvent, DataStoreError>.Promise) {
 
         log.verbose("\(#function) mutationEvent: \(mutationEvent)")
-        storageAdapter.save(mutationEvent, condition: nil) { result in
+        var eventToPersist = mutationEvent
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/AWSMutationDatabaseAdapter/AWSMutationDatabaseAdapter+MutationEventIngester.swift
```diff
@@ -202,7 +202,11 @@ extension AWSMutationDatabaseAdapter: MutationEventIngester {
               completionPromise: @escaping Future<MutationEvent, DataStoreError>.Promise) {
 
         log.verbose("\(#function) mutationEvent: \(mutationEvent)")
-        storageAdapter.save(mutationEvent, condition: nil) { result in
+        var eventToPersist = mutationEvent
+        if nextEventPromise != nil {
+            eventToPersist.inProcess = true
```
drochetti, line 207: just to clarify if I understand it correctly. This is important in case the mutation event fails so it can be retried later on, is that a correct statement?

wooj2: Unfortunately, that is not a correct statement.

To give you a little bit more context, this code should actually be refactored, and technically doesn't belong here.  nextEventPromise is assigned if the outgoingMutationQueue asks the AWSMutationDatabaseAdapter for a mutationEvent to be sent out, but the AWSMutationDatabaseAdapter doesn't have an event to offer up.  So, when this happens, the AWSMutationDatabaseAdapter saves this promise, and offers up an event when a mutationEvent is ready to be sent out (e.g. customer makes a mutation).  

We can sync offline on this, and I can show you a component diagram of various pieces at work here.

PR #402, opened 2020-04-23
feat: [DataStore] Process ConflictUnhandled errors for Conflict Resolution

lawmicha:
*Description of changes:*

The conflict resolution logic here pretains to the following logic:
1. if create -> should never happen
2. if delete, and remote is also deleted -> do nothing
3. if delete, and remote is not deleted -> call conflict handler scenario 1
4. if updated, and remote is deleted -> delete local, notify publisher
5. if updated, and remote is not deleted -> call conflict handler scenario 2

conflict handler scenario 1 returns
1. `.applyRemote` -> recreate local, notify publisher
2. `.retryLocal` -> call API.delete
3. `retry(model)` -> call API.update

conflict handler scenario 2 returns 
1. `.applyRemote` -> update local, notify publisher
2. `.retryLocal` -> call API.update
3. `retry(model)` -> call API.update

DataStore API updates
- delete API's now all call into first querying for the data, then deleting the data.
- this is the case for Delete(modelType, id), Delete(model), Delete(modelType, predicate)
- pass queried model into to mutation event, for conflict resolution scenario 3.

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/AnyModelIntegrationTests.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/AnyModelIntegrationTests.swift
@@ -20,7 +20,7 @@ class AnyModelIntegrationTests: XCTestCase {
         Amplify.reset()
         Amplify.Logging.logLevel = .verbose
 
-        let plugin = AWSAPIPlugin(modelRegistration: PostCommentModelRegistration())
+        let plugin = AWSAPIPlugin(schema: PostCommentSchema())
 
         do {
             try Amplify.add(plugin: plugin)
--- a/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/AnyModelIntegrationTests.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/AnyModelIntegrationTests.swift
@@ -20,7 +20,7 @@ class AnyModelIntegrationTests: XCTestCase {
         Amplify.reset()
         Amplify.Logging.logLevel = .verbose
 
-        let plugin = AWSAPIPlugin(schema: PostCommentSchema())
+        let plugin = AWSAPIPlugin(modelRegistration: PostCommentModelRegistration())
 
         do {
             try Amplify.add(plugin: plugin)
--- a/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
@@ -19,7 +19,7 @@ class GraphQLSyncBasedTests: XCTestCase {
 
     override func setUp() {
         Amplify.reset()
-        let plugin = AWSAPIPlugin(modelRegistration: PostCommentModelRegistration())
+        let plugin = AWSAPIPlugin(schema: PostCommentSchema())
 
         do {
             try Amplify.add(plugin: plugin)
@@ -358,7 +358,7 @@ class GraphQLSyncBasedTests: XCTestCase {
             return
         }
 
-        let conflictUnhandledError = expectation(description: "error should be conditional request failed")
+        let conflictUnhandledError = expectation(description: "error should be conflict unhandled")
         switch response {
         case .success(let mutationSync):
             XCTFail("success: \(mutationSync)")
@@ -398,6 +398,113 @@ class GraphQLSyncBasedTests: XCTestCase {
         }
 
         wait(for: [conflictUnhandledError], timeout: TestCommonConstants.networkTimeout)
+    }
+
+    // Given: A newly created post
+    // When: Call delete mutation, twice
+    // Then: The first delete mutation is successful, and second returns conflict unhandled exception due to older version.
+    func testCreatePostThenDeleteTwiceConflictUnhandledException() {
+        let uuid = UUID().uuidString
+        let testMethodName = String("\(#function)".dropLast(2))
+        let title = testMethodName + "Title"
+        let post = Post.keys
+        guard let createdPost = createPost(id: uuid, title: title) else {
+            XCTFail("Failed to create post with version 1")
+            return
+        }
+        let firstDeleteSuccess = expectation(description: "first delete mutation should be successful")
+
+        let request = GraphQLRequest<MutationSyncResult>.deleteMutation(modelName: createdPost.model.modelName,
+                                                                        id: createdPost.model.id,
+                                                                        version: 1)
+        _ = Amplify.API.mutate(request: request) { event in
+            switch event {
+            case .completed(let graphQLResponse):
+                firstDeleteSuccess.fulfill()
+            case .failed(let apiError):
+                XCTFail("\(apiError)")
+            default:
+                XCTFail("Could not get data back")
+            }
+        }
+        wait(for: [firstDeleteSuccess], timeout: TestCommonConstants.networkTimeout)
+
+        var responseFromOperation: GraphQLResponse<MutationSync<AnyModel>>?
+        let secondDeleteFailed = expectation(
+            description: "second delete mutatiion request should failed with ConflictUnhandled errorType")
+        _ = Amplify.API.mutate(request: request) { event in
+            defer {
+                secondDeleteFailed.fulfill()
+            }
+            switch event {
+            case .completed(let graphQLResponse):
+                responseFromOperation = graphQLResponse
+            case .failed(let apiError):
+                XCTFail("\(apiError)")
+            default:
+                XCTFail("Could not get data back")
+            }
+        }
+        wait(for: [secondDeleteFailed], timeout: TestCommonConstants.networkTimeout)
+
+        guard let response = responseFromOperation else {
+            XCTAssertNotNil(responseFromOperation)
+            return
+        }
+
+        let conflictUnhandledError = expectation(description: "error should be conflict unhandled")
+        switch response {
+        case .success(let mutationSync):
+            XCTFail("success: \(mutationSync)")
+        case .failure(let error):
+            switch error {
+            case .error(let errors):
+                XCTAssertEqual(errors.count, 1)
+                guard let error = errors.first,
+                    let appSyncGraphQLError = error as? AppSyncGraphQLError<MutationSync<AnyModel>>,
+                    let errorType = appSyncGraphQLError.appSyncErrorType else {
+                        XCTFail("Failed to get AppSyncGraphQLError's AppSyncErrorType and response object")
+                        return
+                }
+                XCTAssertEqual(errorType, AppSyncErrorType.conflictUnhandled)
+                guard let mutationSync = appSyncGraphQLError.data else {
+                    XCTFail("Failed to get data object for conflict unhandled error")
+                    return
+                }
+                XCTAssertEqual(mutationSync.syncMetadata.deleted, true)
+                XCTAssertEqual(mutationSync.syncMetadata.version, 2)
+                conflictUnhandledError.fulfill()
+            case .partial(let model, let errors):
+                XCTFail("partial: \(model), \(errors)")
+            case .transformationError(let rawResponse, let apiError):
+                XCTFail("transformationError: \(rawResponse), \(apiError)")
+            }
+        }
+
+        wait(for: [conflictUnhandledError], timeout: TestCommonConstants.networkTimeout)
+
+        let updateSuccess = expectation(description: "update with correct version should be successful")
+
+        let updatedTitle = title + "Updated"
+        let modifiedPost = Post(id: createdPost.model["id"] as? String ?? "",
+                                title: updatedTitle,
+                                content: createdPost.model["content"] as? String ?? "",
+                                createdAt: Date())
+
+        let updateRequest = GraphQLRequest<MutationSyncResult>.updateMutation(of: modifiedPost,
+                                                                              version: 2)
+        _ = Amplify.API.mutate(request: updateRequest) { event in
+            switch event {
+            case .completed(let graphQLResponse):
+                updateSuccess.fulfill()
+            case .failed(let apiError):
+                XCTFail("\(apiError)")
+            default:
+                XCTFail("Could not get data back")
+            }
+        }
+        wait(for: [updateSuccess], timeout: TestCommonConstants.networkTimeout)
+
     }
 
     // Given: Two newly created posts
--- a/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
@@ -19,7 +19,7 @@ class GraphQLSyncBasedTests: XCTestCase {
 
     override func setUp() {
         Amplify.reset()
-        let plugin = AWSAPIPlugin(schema: PostCommentSchema())
+        let plugin = AWSAPIPlugin(modelRegistration: PostCommentModelRegistration())
 
         do {
             try Amplify.add(plugin: plugin)
@@ -398,113 +398,6 @@ class GraphQLSyncBasedTests: XCTestCase {
         }
 
         wait(for: [conflictUnhandledError], timeout: TestCommonConstants.networkTimeout)
-    }
-
-    // Given: A newly created post
-    // When: Call delete mutation, twice
-    // Then: The first delete mutation is successful, and second returns conflict unhandled exception due to older version.
-    func testCreatePostThenDeleteTwiceConflictUnhandledException() {
-        let uuid = UUID().uuidString
-        let testMethodName = String("\(#function)".dropLast(2))
-        let title = testMethodName + "Title"
-        let post = Post.keys
-        guard let createdPost = createPost(id: uuid, title: title) else {
-            XCTFail("Failed to create post with version 1")
-            return
-        }
-        let firstDeleteSuccess = expectation(description: "first delete mutation should be successful")
-
-        let request = GraphQLRequest<MutationSyncResult>.deleteMutation(modelName: createdPost.model.modelName,
-                                                                        id: createdPost.model.id,
-                                                                        version: 1)
-        _ = Amplify.API.mutate(request: request) { event in
-            switch event {
-            case .completed(let graphQLResponse):
-                firstDeleteSuccess.fulfill()
-            case .failed(let apiError):
-                XCTFail("\(apiError)")
-            default:
-                XCTFail("Could not get data back")
-            }
-        }
-        wait(for: [firstDeleteSuccess], timeout: TestCommonConstants.networkTimeout)
-
-        var responseFromOperation: GraphQLResponse<MutationSync<AnyModel>>?
-        let secondDeleteFailed = expectation(
-            description: "second delete mutatiion request should failed with ConflictUnhandled errorType")
-        _ = Amplify.API.mutate(request: request) { event in
-            defer {
-                secondDeleteFailed.fulfill()
-            }
-            switch event {
-            case .completed(let graphQLResponse):
-                responseFromOperation = graphQLResponse
-            case .failed(let apiError):
-                XCTFail("\(apiError)")
-            default:
-                XCTFail("Could not get data back")
-            }
-        }
-        wait(for: [secondDeleteFailed], timeout: TestCommonConstants.networkTimeout)
-
-        guard let response = responseFromOperation else {
-            XCTAssertNotNil(responseFromOperation)
-            return
-        }
-
-        let conflictUnhandledError = expectation(description: "error should be conflict unhandled")
-        switch response {
-        case .success(let mutationSync):
-            XCTFail("success: \(mutationSync)")
-        case .failure(let error):
-            switch error {
-            case .error(let errors):
-                XCTAssertEqual(errors.count, 1)
-                guard let error = errors.first,
-                    let appSyncGraphQLError = error as? AppSyncGraphQLError<MutationSync<AnyModel>>,
-                    let errorType = appSyncGraphQLError.appSyncErrorType else {
-                        XCTFail("Failed to get AppSyncGraphQLError's AppSyncErrorType and response object")
-                        return
-                }
-                XCTAssertEqual(errorType, AppSyncErrorType.conflictUnhandled)
-                guard let mutationSync = appSyncGraphQLError.data else {
-                    XCTFail("Failed to get data object for conflict unhandled error")
-                    return
-                }
-                XCTAssertEqual(mutationSync.syncMetadata.deleted, true)
-                XCTAssertEqual(mutationSync.syncMetadata.version, 2)
-                conflictUnhandledError.fulfill()
-            case .partial(let model, let errors):
-                XCTFail("partial: \(model), \(errors)")
-            case .transformationError(let rawResponse, let apiError):
-                XCTFail("transformationError: \(rawResponse), \(apiError)")
-            }
-        }
-
-        wait(for: [conflictUnhandledError], timeout: TestCommonConstants.networkTimeout)
-
-        let updateSuccess = expectation(description: "update with correct version should be successful")
-
-        let updatedTitle = title + "Updated"
-        let modifiedPost = Post(id: createdPost.model["id"] as? String ?? "",
-                                title: updatedTitle,
-                                content: createdPost.model["content"] as? String ?? "",
-                                createdAt: Date())
-
-        let updateRequest = GraphQLRequest<MutationSyncResult>.updateMutation(of: modifiedPost,
-                                                                              version: 2)
-        _ = Amplify.API.mutate(request: updateRequest) { event in
-            switch event {
-            case .completed(let graphQLResponse):
-                updateSuccess.fulfill()
-            case .failed(let apiError):
-                XCTFail("\(apiError)")
-            default:
-                XCTFail("Could not get data back")
-            }
-        }
-        wait(for: [updateSuccess], timeout: TestCommonConstants.networkTimeout)
-
     }
 
     // Given: Two newly created posts
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
@@ -92,29 +92,17 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
                                  withId id: String,
                                  completion: @escaping DataStoreCallback<Void>) {
         reinitStorageEngineIfNeeded()
-        storageEngine.delete(modelType,
-                             withId: id,
-                             completion: completion)
+        storageEngine.delete(modelType, withId: id) { result in
+            self.onDeleteCompletion(result: result, completion: completion)
+        }
     }
 
     public func delete<M: Model>(_ model: M,
                                  completion: @escaping DataStoreCallback<Void>) {
         reinitStorageEngineIfNeeded()
-        let publishingCompletion: DataStoreCallback<Void> = { result in
-            switch result {
-            case .success:
-                // TODO: Handle errors from mutation event creation
-                self.publishMutationEvent(from: model, mutationType: .delete)
-            case .failure:
-                break
-            }
-
-            completion(result)
+        storageEngine.delete(type(of: model), withId: model.id) { result in
+            self.onDeleteCompletion(result: result, completion: completion)
         }
-
-        delete(type(of: model),
-               withId: model.id,
-               completion: publishingCompletion)
     }
 
     public func delete<M: Model>(_ modelType: M.Type,
@@ -155,6 +143,19 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
     }
 
     // MARK: Private
+
+    private func onDeleteCompletion<M: Model>(result: DataStoreResult<M?>,
+                                              completion: @escaping DataStoreCallback<Void>) {
+        switch result {
+        case .success(let modelOptional):
+            if let model = modelOptional {
+                publishMutationEvent(from: model, mutationType: .delete)
+            }
+            completion(.emptyResult)
+        case .failure(let error):
+            completion(.failure(error))
+        }
+    }
 
     private func publishMutationEvent<M: Model>(from model: M,
                                                 mutationType: MutationEvent.MutationType) {
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
@@ -19,8 +19,10 @@ public struct DataStoreConflictData {
     public let remote: Model
 }
 
-/// could not be resolved and requires a decision from the consumer. The consumer can call other methods (optionally
-/// asynchronous) before calling the escaping `DataStoreConflictHandlerResolver` with the resolution,
+/// The `DataStoreConflictHandler` is an asynchronous callback which allows consumers to decide how to resolve conflicts
+/// between the frontend and backend. This can be configured on the `conflictHandler` of the `DataStoreConfiguration`
+/// by implementing the body of the closure, processing `DataStoreConflictData` and resolving the conflict by calling
+/// `DataStoreConflictHandlerResolver`
 public typealias DataStoreConflictHandler = (DataStoreConflictData, @escaping DataStoreConflictHandlerResolver) -> Void
 
 /// Callback for the `DataStoreConflictHandler`.
@@ -29,10 +31,10 @@ public typealias DataStoreConflictHandlerResolver = (DataStoreConflictHandlerRes
 /// The conflict resolution result enum.
 public enum DataStoreConflictHandlerResult {
 
-    /// Discard the local changes in favor of the remote ones.
+    /// Discard the local changes in favor of the remote ones. Semantically the same as `DISCARD` on Amplify-JS
     case applyRemote
 
-    /// Keep the local changes.
+    /// Keep the local changes. (semantic shortcut to `retry(local)`).
     case retryLocal
 
     /// Return a new `Model` instance that should used instead of the local and remote changes.
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
@@ -19,9 +19,9 @@ public struct DataStoreConflictData {
     public let remote: Model
 }
 
-/// Conflict Handler function typealias. The function is used during a conflict that
-/// could not be resolved and requires a decision from the consumer.
-public typealias DataStoreConflictHandler = (DataStoreConflictData, DataStoreConflictHandlerResolver) -> Void
+/// could not be resolved and requires a decision from the consumer. The consumer can call other methods (optionally
+/// asynchronous) before calling the escaping `DataStoreConflictHandlerResolver` with the resolution,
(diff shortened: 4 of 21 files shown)


AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/AnyModelIntegrationTests.swift
```diff
@@ -20,7 +20,7 @@ class AnyModelIntegrationTests: XCTestCase {
         Amplify.reset()
         Amplify.Logging.logLevel = .verbose
 
-        let plugin = AWSAPIPlugin(modelRegistration: PostCommentModelRegistration())
+        let plugin = AWSAPIPlugin(schema: PostCommentSchema())
```
wooj2, line 23: do we need to rebase changes? in `AWSAPIPlugin.swift`, i'm still seeing:
```
 public init(modelRegistration: AmplifyModelRegistration? = nil,
                sessionFactory: URLSessionBehaviorFactory? = nil)
```

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
```diff
@@ -400,6 +400,113 @@ class GraphQLSyncBasedTests: XCTestCase {
         wait(for: [conflictUnhandledError], timeout: TestCommonConstants.networkTimeout)
     }
 
+    // Given: A newly created post
+    // When: Call delete mutation, twice
+    // Then: The first delete mutation is successful, and second returns conflict unhandled exception due to older version.
+    func testCreatePostThenDeleteTwiceConflictUnhandledException() {
```
wooj2, line 406: Is calling delete twice on an item a real use case for our software, or is this more or less a workaround because we don't have the mechanism to setup a scenario where we'd run into a conflict?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -156,6 +144,19 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
...
+                                              completion: @escaping DataStoreCallback<Void>) {
+        switch result {
+        case .success(let modelOptional):
+            if let model = modelOptional {
+                publishMutationEvent(from: model, mutationType: .delete)
+            }
+            completion(.emptyResult)
+        case .failure(let error):
+            completion(.failure(error))
+        }
```
wooj2, line 157: We were previously just calling `completion(result)` -- curious as to why we changed it to completion(.emptyResult) on the success case?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
```diff
@@ -19,9 +19,9 @@ public struct DataStoreConflictData {
     public let remote: Model
 }
 
-/// Conflict Handler function typealias. The function is used during a conflict that
-/// could not be resolved and requires a decision from the consumer.
-public typealias DataStoreConflictHandler = (DataStoreConflictData, DataStoreConflictHandlerResolver) -> Void
+/// could not be resolved and requires a decision from the consumer. The consumer can call other methods (optionally
+/// asynchronous) before calling the escaping `DataStoreConflictHandlerResolver` with the resolution,
```
wooj2, line 23: How about something like..

> The DataStoreConflictHandler is an asynchronous callback which allows consumers to decide how to resolve conflicts between the frontend and backend.  This can be configured by....

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -144,8 +144,15 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
 
     func delete<M: Model>(_ modelType: M.Type,
                           withId id: Model.Identifier,
-                          completion: (DataStoreResult<Void>) -> Void) {
-        delete(untypedModelType: modelType, withId: id, completion: completion)
+                          completion: (DataStoreResult<M?>) -> Void) {
+        delete(untypedModelType: modelType, withId: id) { result in
```
wooj2, line 148: Hmm why are we changing this?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -160,66 +160,61 @@ final class StorageEngine: StorageEngineBehavior {
...
-            guard modelType.schema.isSyncable, let syncEngine = self.syncEngine else {
-                if !modelType.schema.isSystem {
-                    self.log.error("Unable to sync modelType (\(modelType)) where isSyncable is false")
-                }
-                if self.syncEngine == nil {
-                    self.log.error("Unable to sync because syncEngine is nil")
-                }
-                completion(result)
+                          completion: @escaping (DataStoreResult<M?>) -> Void) {
+        let transactionResult = delete(modelType, predicate: field("id").eq(id))
```
wooj2, line 164: Why did we start using the predicate version for deleting withId?  If we can avoid using the predicate version, we should to avoid having to do a transaction?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -30,8 +31,21 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
 
     private weak var api: APICategoryGraphQLBehavior?
     private var subscription: Subscription?
+    private let dataStoreConfiguration: DataStoreConfiguration
+    private let storageAdapter: StorageEngineAdapter
 
-    init(_ stateMachine: StateMachine<State, Action>? = nil) {
+    private let mutationEventPublisher: PassthroughSubject<MutationEvent, Never>
```
wooj2, line 37: A better name might be something like `OutgoingMutationQueueSubject` (this thing's role is not a publisher)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -30,8 +31,21 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
...
+    public var publisher: AnyPublisher<MutationEvent, Never> {
+        return mutationEventPublisher.eraseToAnyPublisher()
+    }
+
+    private var processMutationErrorOperationSink: AnyCancellable?
```
wooj2, line 42: I'm not seeing where this is used?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -165,26 +181,38 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
...
-
-                        self?.log.verbose(
-                            "[ProcessMutationErrorFromCloudOperation] mutation error processed, result: \(result)")
-                        self?.completeProcessingEvent(mutationEvent)
-                    }
-                    self?.operationQueue.addOperation(processMutationErrorFromCloudOperation)
-                } else {
-                    self?.completeProcessingEvent(mutationEvent)
+        let syncMutationToCloudOperation = SyncMutationToCloudOperation(mutationEvent: mutationEvent,
+                                                                        api: api) { [weak self] result in
```
wooj2, line 185: Still not entirely convinced we need to do any [weak self] casts here. have we run the memory leak analyzer?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -165,26 +181,38 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
...
+                                                       error: error) { [weak self] result in
+                guard let self = self else {
+                    return
                 }
+
+                self.log.verbose("[ProcessMutationErrorFromCloudOperation] mutation error processed, result: \(result)")
+                if case let .success(mutationEventOptional) = result,
+                    let outgoingMutationEvent = mutationEventOptional {
+                    self.mutationEventPublisher.send(outgoingMutationEvent)
+                }
```
wooj2, line 212: Hmmm.. i guess it's safe to keep running if there is some error as part of this operation?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
```diff
@@ -33,68 +43,306 @@ class ProcessMutationErrorFromCloudOperation: Operation {
         log.verbose(#function)
 
         guard !isCancelled else {
-            queryOperation?.cancel()
-            let apiError = DataStoreError.unknown("Operation cancelled", "")
-            finish(result: .failure(apiError))
+            let error = DataStoreError.unknown("Operation cancelled", "")
```
wooj2, line 46: seems like we need to cancel the mutationOperation ?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -122,6 +124,14 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
                     self.respond(to: newState)
                 }
         }
+
+        self.outgoingMutationQueueSink = self.outgoingMutationQueue.publisher.sink { [weak self] mutationEvent in
+            guard let self = self else {
+                return
+            }
```
wooj2, line 131: are we sure we need to do [weak self] here? (I don't think we do?)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -122,6 +124,14 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
+            self.log.verbose("MutationEvent received from outgoing mutation queue for topic publisher \(mutationEvent)")
```
wooj2, line 132: we should probably remove this log statements since we aren't sending one when we send one in RemoteSyncEngine+IncomingEventReconciliationQueueEvent.

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/AnyModelIntegrationTests.swift
lawmicha: I believe we are sticking with modelRegistration for now. reverting this, it looks like it got pulled in due to cherry-picking my commit that was built on the `schema` change that was reverted

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
lawmicha: because the delete storage behavior now returns an optional model. the Delete first queries for the data, then deletes it. it may result a successful and nothing deleted case. the developer facing API still returns void, since the change in interface resulted in no longer a passthrough from `result` to `completion`, we take the optional model, if delete was successful, model retrieved from modelOptional, then we publish to subscribers. and return .emptyResult for the caller

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
lawmicha: the delete is no longer just a call to delete on storage but queries for the data and then deletes it. we pass the queried data back to the caller like we do with the `delete(modelType, predicate) -> [M]`  to send to the subscribers. we also take the queried data and store it in the MutationEvent for conflict resolution

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
lawmicha: this consolidated the code paths for all Delete APIs to a single delete method, though there might be a performance difference between using the predicate verse using the Id based on the SQL statement being constructed, lemme take a look at this-

sounds right about not needing a transaction for a single query+delete

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
lawmicha: thanks, updating

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
lawmicha: removing

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
lawmicha: i believe so, the operations (SyncMutationToCloudOperation/ProcessMutationErrorFromCloudOperation) should have its own retries like the retryability on API, by the time it's back here it's done all it can so i think it's better to let the system continue to operate by removing the mutationEvent and continue processing the next one

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
lawmicha: i think this is fine, the `mutationOperation` is the reference to the API.mutate call, if it's cancelled, the `cancel()` override cancelled would have been called already which calls `mutationOperation.cancel()`, and then here we call `override finish()` which does the clean up and removes the listener

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
lawmicha: i was following the same pattern in the few lines above, so i'm not sure
```
self.stateMachineSink = self.stateMachine
            .$state
            .sink { [weak self] newState in
                guard let self = self else {
                    return
                }
                self.log.verbose("New state: \(newState)")
                self.workQueue.async {
                    self.respond(to: newState)
                }
        }
```

should be safe to remove since these instances are long lived throughout a lifecycle of an app, and my understanding of using [weak self] is to prevent retain cycle when one of the two instances are being deinit, but neither of them should be deinitialized and if they are, probably both and all of DataStore would be torn down

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
lawmicha: sorry, what do u mean?

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
```diff
@@ -397,6 +397,113 @@ class GraphQLSyncBasedTests: XCTestCase {
...
+        let uuid = UUID().uuidString
+        let testMethodName = String("\(#function)".dropLast(2))
+        let title = testMethodName + "Title"
```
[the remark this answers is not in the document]
lawmicha: this test has been removed (cherry-pick issue, there is already a test that covers this scenario but instead of double delete, it's double update)

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLSyncBased/GraphQLSyncBasedTests.swift
lawmicha: this was one way i found to easily get the conflict unhandled exception for the integration tests, (test is removed though, we have another one that covers this scenario by double update). i would consider this scenario to be where both users (or same user on different device) has access to modify the model instance. does it at the same time, one wins and other fails. so even if we think of other scenarios, i think this one is still good to keep

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
lawmicha: thanks, improving this

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreConfiguration.swift
```diff
@@ -30,10 +30,10 @@ public typealias DataStoreConflictHandlerResolver = (DataStoreConflictHandlerRes
...
-    case discard
+    case applyRemote
 
-    /// Keep the local changes (semantic shortcut to `retry(local)`).
-    case keep
+    /// Keep the local changes.
+    case retryLocal
 
     /// Return a new `Model` instance that should used instead of the local and remote changes.
     case retry(Model)
```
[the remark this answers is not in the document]
lawmicha: adding

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
lawmicha: i haven't ran much else other than the one you showed me that day that said there were no issues, i think for operations that are short lived, it is stay to keep the [weak self] in case the operation goes out of scope but then doesn't get deinitialized because there's a strong reference to `self` which is the OutgoingMutationQueue

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
lawmicha: so it looks like the query(byId) and delete(byId) are both creating predicates which create conditional sql statements so there is no difference between using the predicate version.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
lawmicha: removed weak self here

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
wooj2: My concern is that it sets a precedence for the rest of the code base, and this seems like this should only be added if we know that we are actually running into a case where it's needed.

In this case, I don't think it's needed because the `OutgoingMutationQueue` is expected to be in memory until the system comes down because there is a strong reference from the RemoteSyncEngine that is expected to live for the life that the application is in memory (afaik, OutgoingMutationQueue gets created in init() and then never gets de-allocated, and similarly in StorageEngine, it creates the SyncEngine in init(), and then never deallocates it).  When I see the use of `[weak self]` here, this tells me that we are expecting that the OutgoingMutationQueue will be going out of memory/scope and if that happens the callback from SyncMutationToCloudOperation (and ProcessMutationErrorFromCloudOperation) should not continue w/ the guard statement.  If this is behavior we do want, then great-- but I don't think this is the case.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
wooj2: Remove this log statement because in the other case where we are sending a mutation event to the subject, we are do not have a log statement.  (See RemoteSyncEngine+IncomingEventReconciliationQueueEvent.)  Putting a log statement here would make mutation events logging inconsistent.

lawmicha: removing

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
lawmicha: thanks for explaining, removing weak self here since we don't expect OutgoingMutatioinQueue to be deinitialized

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
lawmicha: synced on this; the atomicity that we are looking for is to ensure an external update like DataStore.save does not happen between the entire operation of the delete API, ie. between the sql query and sql delete. If it does happen without the transaction, then the subscribers to the model will get an update that the model was deleted with the fields from the query, and not the actual field values that were modified by the update

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -261,6 +255,35 @@ final class StorageEngine: StorageEngineBehavior {
         }
     }
 
+    private func delete<M: Model>(_ modelType: M.Type, predicate: QueryPredicate) -> DataStoreResult<[M]> {
```
wooj2, line 258: if we change this private function named to `queryAndDeleteTransaction`, then it's calling code would be more readable:
```
        let transactionResult = queryAndDeleteTransaction(modelType, predicate: predicate)
```
(where currently we have:
```
        let transactionResult = delete(modelType, predicate: predicate)
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -160,66 +160,61 @@ final class StorageEngine: StorageEngineBehavior {
...
-                    self.log.error("Unable to sync modelType (\(modelType)) where isSyncable is false")
-                }
-                if self.syncEngine == nil {
-                    self.log.error("Unable to sync because syncEngine is nil")
-                }
-                completion(result)
+                          completion: @escaping (DataStoreResult<M?>) -> Void) {
+        let transactionResult = delete(modelType, predicate: field("id").eq(id))
+
+        let model: M
```
wooj2, line 166: a better name for this might be `deletedModel`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -15,6 +15,7 @@ protocol OutgoingMutationQueueBehavior: class {
     func pauseSyncingToCloud()
     func startSyncingToCloud(api: APICategoryGraphQLBehavior,
                              mutationEventPublisher: MutationEventPublisher)
+    var publisher: AnyPublisher<MutationEvent, Never> { get }
```
wooj2, line 18: We should be exposing an enum interface like `OutgoingMutationQueueEvent` for the publisher.. but this PR is getting a bit long, so we can hold off on doing this in a subsequent PR.  Helps with testing and understanding outgoingmutationqueue's internal state, etc.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -165,26 +179,31 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
...
-                            "[ProcessMutationErrorFromCloudOperation] mutation error processed, result: \(result)")
-                        self?.completeProcessingEvent(mutationEvent)
-                    }
-                    self?.operationQueue.addOperation(processMutationErrorFromCloudOperation)
-                } else {
-                    self?.completeProcessingEvent(mutationEvent)
+        let syncMutationToCloudOperation = SyncMutationToCloudOperation(mutationEvent: mutationEvent,
+                                                                        api: api) { result in
+            self.log.verbose(
+                "[SyncMutationToCloudOperation] mutationEvent finished: \(mutationEvent); result: \(result)")
```
wooj2, line 185: If this is printing the contents of the mutationEvent, we should avoid this (maybe only print the Id of the mutation event).  It has the potential of leaking sensitive data.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
```diff
@@ -12,18 +12,28 @@ import AWSPluginsCore
...
 /// 1. When there is a "conditional request failed" error, then emit to the Hub a 'conditionalSaveFailed' event.
+/// 2. When there is a "conflict unahandled" error, trigger the conflict handler and reconcile the state of the system.
 @available(iOS 13.0, *)
-class ProcessMutationErrorFromCloudOperation: Operation {
+class ProcessMutationErrorFromCloudOperation: AsynchronousOperation {
 
+    private let dataStoreConfiguration: DataStoreConfiguration
+    private let storageAdapter: StorageEngineAdapter
     private let mutationEvent: MutationEvent
     private let error: GraphQLResponseError<MutationSync<AnyModel>>
```
wooj2, line 22: We should rename this 'error' because we have local variables in functions that also go by the name of 'error' as well.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
```diff
@@ -33,68 +43,306 @@ class ProcessMutationErrorFromCloudOperation: Operation {
...
+            let localModel: Model
+            do {
+                localModel = try mutationEvent.decodeModel()
+            } catch {
+                let error = DataStoreError.unknown("Couldn't decode local model", "")
+                finish(result: .failure(error))
+                return
+            }
+            let conflictData = DataStoreConflictData(local: localModel, remote: remote.model.instance)
+            let latestVersion = remote.syncMetadata.version
```
wooj2, line 138: This feels like duplicate code, seems like we can refactor this?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
```diff
@@ -33,68 +43,306 @@ class ProcessMutationErrorFromCloudOperation: Operation {
...
                                          data: mutationEvent)
                 Amplify.Hub.dispatch(to: .dataStore, payload: payload)
+                finish(result: .success(nil))
+            case .conflictUnhandled:
+                processConflictUnhandled(extensions)
+            case .unknown(let errorType):
+                log.debug("Unhandled error with errorType \(errorType)")
+                finish(result: .success(nil))
             }
         }
```
wooj2, line 80: Pretty sure we do not have any unit tests that hit this case nor the case where we don't fall into the `if let` statement.. we should add a todo to write some unit tests

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/MutationQueue/ProcessMutationErrorFromCloudOperationTests.swift
```diff
@@ -46,19 +52,719 @@ class ProcessMutationErrorFromCloudOperationTests: XCTestCase {
...
+        let queue = OperationQueue()
+        queue.addOperation(operation)
+
+        wait(for: [expectConflicthandlerCalled, apiMutateCalled], timeout: defaultAsyncWaitTimeout)
+        guard let eventListener = eventListenerOptional else {
+            XCTFail("Listener was not called through MockAPICategoryPlugin")
+            return
+        }
+        let updatedMetadata = MutationSyncMetadata(id: remotePost.id, deleted: true, lastChangedAt: 0, version: 3)
+        let local = MutationSync(model: try localPost.eraseToAnyModel(), syncMetadata: updatedMetadata)
```
wooj2, line 246: is there anything we can do to verify the data we are sending back through the event listener?  if i change deleted:false and the version to something like 99, the unit test still passes.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/MutationQueue/ProcessMutationErrorFromCloudOperationTests.swift
```diff
@@ -46,19 +52,719 @@ class ProcessMutationErrorFromCloudOperationTests: XCTestCase {
...
+                                                               api: mockAPIPlugin,
+                                                               storageAdapter: storageAdapter,
+                                                               error: graphQLResponseError,
+                                                               completion: completion)
+
+        let queue = OperationQueue()
+        queue.addOperation(operation)
+
+        wait(for: [modelDeletedEvent, metadataSavedEvent, expectHubEvent, expectCompletion],
+             timeout: defaultAsyncWaitTimeout)
```
wooj2, line 456: It might be more readable if we be strict about the order in which expectations finish,  we can break them up into separate lines and have multiple wait statements.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
lawmicha: good idea, i didn't do this because i didn't really have a use case, only had to get the MutationEvent back to the DataStore subscribers, but i can see the events being useful to align with the OutGoingMutationQueue.State's. will keep this in mind when working on the auth directive work

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
```diff
@@ -33,68 +43,306 @@ class ProcessMutationErrorFromCloudOperation: Operation {
...
+            return
+        }
 
-    private func processConditionalRequestFailed() {
-        if case let .error(graphQLErrors) = error {
-            // TODO: Check for 'ConflictUnhandled', execute conflict handler
+        guard graphQLErrors.count == 1 else {
+            finish(result: .success(nil))
+            return
+        }
```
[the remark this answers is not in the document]
lawmicha: i was going back and forth on this as well, to be future proof, it would make sense to simply process all errors in the array, however i don't think this operation should be responsible for that (it would also make the unit tests more complicated). perhaps another operation or method which enqueues each operation with a single error. currently i didn't have an example where there would be an additional error in the array so i didn't have a meaningful test to write. I think an improvement is to log it here so that we know what the additional errors are and then we can revisit the scenario and account for it.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
lawmicha: thanks, operation needs to finish if it doesn't fall into `if let`, updating logic and adding unit tests for both cases

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/MutationQueue/ProcessMutationErrorFromCloudOperationTests.swift
lawmicha: this `local` model is just so we can pass something to the eventListener when we call it with .success(local) since we are mocking the API call to be successful. we don't really do anything with the service response after that, as long as it was successful, we finish the operation. it's the subscription reconcilation path that will take this mutation response and reconcile local store with it. a better name for this is probably `mockAPIResponse` or something and can be static data

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/MutationQueue/ProcessMutationErrorFromCloudOperationTests.swift
lawmicha: really like this idea, makes it clear which fulfill is supposed to be executed first

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
lawmicha: renamed to type name `graphQLResponseError`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
```diff
@@ -33,68 +43,306 @@ class ProcessMutationErrorFromCloudOperation: Operation {
...
+            case .conflictUnhandled:
+                processConflictUnhandled(extensions)
+            case .unknown(let errorType):
+                log.debug("Unhandled error with errorType \(errorType)")
+                finish(result: .success(nil))
             }
         }
+    }
+
+    private func processConflictUnhandled(_ extensions: [String: JSONValue]) {
```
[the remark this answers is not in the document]
lawmicha: took a refactoring pass on this

Issue #381, opened 2020-04-08
AnalyticsPinpointPlugin Updates 


lawmicha:
1. check if iOS’s analytics unregisterGlobal, add varargs
2. update BasicAnalyticsEvent to remove “_” from first parameter to have developer write out the first parameter name. BasicAnalyticsEvent(name:"myEventName")
3. Hub events alignment with Android
4. `public typelias AnalyticsProperties = [String: AnalyticsPropertyValue]`

PR #404, opened 2020-04-24
feat: analytics updates API improvements (unregisterGlobalProperties)

lawmicha:
*Description of changes:*
- Added `unregisterGlobalProperties(_ keys: String...)` at the Category level. Because Amplify.Analytics.methods will call into the plugin. A variadic parameter turns into an array so there isn't an easy way to pass it in to call the plugin's method with variadic parameter. This way, the Analytics Category provides a helper method which calls into the `unregisterGlobalProperties(Set of strings)`, and plugin only needs to implement the single unregister method.
- Using typealias for map of property values, `AnalyticsPropertyValues`


drochetti:
This PR addresses #381 right?

lawmicha:
> This PR addresses #381 right?

yes!

PR #412, opened 2020-05-03
Remove AsyncEvent

palpatim:
## Description of changes

**Summary**

- Completion and Progress reporting is no longer consolidated into a single type. Plugin implementations must dispatch two separate Hub payloads for completion vs. in-process events
- Call sites must change all event listeners to use a Foundation `Result` type for completion events. Callers that wish to receive in-process results for those APIs that support them (Storage `downloadData`, `downloadFile`, `uploadData`, and `uploadFile`; GraphQL `subscribe`) must register an additional in-process listener via the new API.
- The parameter names for the new listeners vary by API
  - For all APIs except `GraphQL.subscribe()`, the result handler is named `resultListener`.
  - For Storage upload/download APIs, the in-process handler is named `progressListener`
  - For `GraphQL.subscribe()`, the in-process handler is named `valueListener`, and the result handler is named `completionListener`.
- Renamed convenience typealias `EventListener` to `ResultListener`
- Renamed generic placeholders `Complete` and `Error` to `Success` and `Failure`, to match with Swift's standard `Result` type

**Open Questions**

- Should we refactor the `listener` parameter name for all operations to `resultListener`? If so, should we continue to name the GraphQLSubscription separately? The names `resultListener` and `inProcessListener` aren't specific enough for a subscription, even though they serve exactly the same purpose.

**Discussion**

The `AsyncEvent` type tries to do too much, and is cumbersome to use as a result. This PR removes AsyncEvent, in favor of making most AmplifyOperations return a simple `Result` type. The two cases where in-process events made sense use a new subclass of AmplifyOperation named `AmplifyInProcessReportingOperation`.

The change is quite large and far-reaching, but will result in a greatly-reduced footprint at the call sites. Whereas a previous call to `getURL` might have looked like this:

```swift
_ = Amplify.Storage.getURL(key: "foo") { event in
    switch event {
        case .unknown:
            print("As a customer, I don't really know how this can happen or what I should do about it.")
        case .notInProcess:
            print("This either means 'not started' or 'completed' I guess? I'd have to read the docs to find out.")
        case .inProcess(let value):
            print("An in-process event, which is always `void` for getURL, but I have to deal with it anyway")
        case .completed(let url):
            print("The only meaningful value case: the URL is \(url)")
        case .failed(let error):
            print("Something went wrong")
    }
}
```

The new call simply returns a result:

```swift
_ = Amplify.Storage.getURL(key: "foo") { result in
    switch result {
        case .success(let url):
            print("The only meaningful value case: the URL is \(url)")
        case .failure(let error):
            print("Something went wrong")
    }
}
```

Category API calls that have meaningful in-process states are now modeled as `AmplifyInProcessReportingOperation`s, which take an additional "in-process listener" that will be triggered when the operation reports an interim value. This might be a fractional `Progress` for a file download, or a `GraphQLSubscriptionEvent` for an operation that receives a value on a GraphQL subscription:

```swift
let downloadOperation = Amplify.Storage.downloadFile(
    "key", 
    progressListener: { print("\($0.fractionCompleted * 100)% complete") }) { result in
    switch result {
        case .success:
            print("File downloaded successfully")
        case .failure(let error):
            print("Something went wrong")
    }
}
```

This refactored set of Operation classes maintains the important features of the previous AmplifyOperation vis-a-vis Hub interoperability:

- Behind the scenes, `OperationResult` values are delivered via Hub
- Customers can register for notifications about an operation by either passing a listener during operation creation (that is, they can pass listeners at the time they make the category API call like `Storage.downloadFile()`), or after the API call by passing the operation to `Hub.listen(to:)`
- Operation listeners are automatically unsubscribed, and progress listeners are released, when a completion result is received.

## Next steps

This PR modifies the core types for the new async event handling model. Category plugins must be modified to use the new model, but that should be done on separate PRs for easier review.


Amplify/Core/Support/AmplifyInProcessReportingOperation.swift
```diff
@@ -0,0 +1,96 @@
...
+    /// automatically be removed.
+    ///
+    /// - Parameter inProcessListener: The listener for in-process events
+    /// - Returns: an UnsubscribeToken that can be used to remove the listener from Hub
+    func subscribe(inProcessListener: @escaping InProcessListener) -> UnsubscribeToken {
+        let channel = HubChannel(from: categoryType)
+        let filterById = HubFilters.forOperation(self)
+
+        var progressListenerToken: UnsubscribeToken!
+        let progressHubListener: HubListener = { payload in
```
lawmicha, line 51: `inProcess` vs `progress`, I was thinking whether this should called `AmplifyProgressOperation`, i don't have a strong opinion on this, just that I noticed here we define a local variable `progressListenerToken` and not something like `inProcessListenerToken`. At the same time, InProcess makes more sense for Subscriptions values. but end of the day all of it is all typealias'd

Amplify/Core/Support/AmplifyInProcessReportingOperation.swift
```diff
@@ -0,0 +1,96 @@
...
+import Foundation
+
+/// An AmplifyOperation that emits InProcess values intermittently during the operation.
+///
+/// Unlike a regular `AmplifyOperation`, which emits a single Result at the completion of the operation's work, an
+/// `AmplifyInProcessReportingOperation` may emit intermediate values while its work is ongoing. These values could be
+/// incidental to the operation (such as a `Storage.downloadFile` operation reporting Progress values periodically as
+/// the download proceeds), or they could be the primary delivery mechanism for an operation (such as a
+/// `GraphQLSubscriptionOperation`'s emitting new subscription values).
+open class AmplifyInProcessReportingOperation<
```
lawmicha, line 17: `AmplifyInProcessOperation`? would reduce size of name and still represent the same thing

palpatim: The part of me that values terseness approves of that idea, but "InProcessOperation" isn't the same as "InProcessReportingOperation". The first describes an operation that is in the middle of performing work, which actually applies to all of our operations. The second describes an operation that *tells you* when it is in the middle of performing work, which is the critical feature of this type.

Amplify/Core/Support/AmplifyInProcessReportingOperation.swift
palpatim: I considered "progress" as a generic term, but it's somewhat overloaded because of the Foundation type, and the fact that we explicitly use Progress as an "in-process" value. That made me settle on the slightly more awkward but more generically applicable "in-process". I can be argued out of "in-process" to something else, but I feel like "progress" or variations thereof are going to cause more confusion than they'll solve.

palpatim:
Closing in favor of #462

PR #415, opened 2020-05-04
feat(Auth): Implementation of Device operations

royjit:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Amplify/Categories/Auth/Models/AuthDevice.swift
+++ b/Amplify/Categories/Auth/Models/AuthDevice.swift
@@ -1,0 +1,15 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+public protocol AuthDevice {
+
+    var id: String { get }
+
+    var name: String { get }
+}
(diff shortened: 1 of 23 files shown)


Amplify/Categories/Auth/Models/AuthDevice.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+
+import Foundation
+
+public protocol AuthDevice {
+
+    var id: String { get }
+
+    var name: String { get }
+}
```
lawmicha, line 15: should this be a struct/class? this means that I would have to fetchDevices before I can call remember device or forget device. Amplify could provide a struct/class to allow the developer to instantiate a device after retrieving the device id from their own persistance layer to call forget/remember device. by having a protocol, this also forces the plugin to have to create their own conformance to it like the `AWSAuthDevice`. If we needed to store more information, we can still create AWSAuthDevice as a subclass of AuthDevice instead of conformance. 

or perhaps AWSAuthDevice, which is currently a struct, needs a public constructor since it is internal by default.

lawmicha: technically, the developer could still have their own persistance of device information, and call fetch devices, then match the id with the response data to find the device they want to apply further actions on

royjit: I agree in having  a public structure that the developer can create. I feel strongly in having  a struct for AuthDevice since its basically like a model object and doesnot need any of the class features. I guess the best approach here is to make `AWSAuthDevice` public.

lawmicha: if AuthDevice is a struct, then AWSAuthDevice won't be able to subclass it (i think), so might need to be a class to extend it

PR #423, opened 2020-05-07
feat: AuthRuleDecorator for constructing GraphQL request

lawmicha:
*Issue #, if available:*

*Description of changes:*
This introduces the AuthRuleDecorator which is based on current provisioning logic, until https://github.com/aws-amplify/amplify-cli/issues/4182 is fixed. The outcome of https://github.com/aws-amplify/amplify-cli/issues/4182 will change the logic relating to subscription operation inputs, and shouldn't impact the selection set generation.

Current logic to match provisioned backend
- The focus is to get `.owner` auth strategy working
- static `.group` auth should just works as it doesn't add anything to the graphQL document.
- dynamic `.group` auth strategy requires adding additional inputs to the document, not implemented here.
- the `ownerField` which is "owner" or specified by the AuthRule.ownerField is added to the selection set if it does not exist for the model.
- subscription operations require the owner value when the AuthRule.operation exists on the AuthRule, ie. AuthRule(operation = .create) means onCreateSubscription(owner:String!). This comes from the caller, which will be something along the lines of `Amplify.Auth.fetchUserSession as AuthIdentityProvider, then get the UserSub()`
- ModelDecorator will take the instance of the model and use that as the variables input for the graphQL request. the ownerField on the model that is set to nil will be added to the request as `"owner": nil`. This will be treated as `null` when serialized and treated as a value for the service. Service will not overwrite it with the owner in the auth token, causing an Unauthorized request. We intercept this in ModelDecorator to filter out ownerFields which have nil values.

Next up will be to provision the sample Blog model in AmpifyTestCommon and run integration tests in the API integration test target with conflict resolution enabled, to replicate DataStore's usage of API Plugin.


AmplifyPlugins/Core/AWSPluginsCoreTests/Model/Decorator/AuthRuleDecorator/ModelMultipleOwnerAuthRuleTest.swift
```diff
@@ -0,0 +1,282 @@
...
+        ]
+        model.fields(
+            .id(),
+            .field(modelMultipleOwner.content, is: .required, ofType: .string),
+            .field(modelMultipleOwner.editors, is: .optional, ofType: .customType([String].self))
+        )
+    }
+}
+
+class ModelMultipleOwnerAuthRuleTest: XCTestCase {
```
drochetti, line 59: Test**s** (filename as well)

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
```diff
@@ -0,0 +1,105 @@
...
+                inputs[ownerField] = GraphQLDocumentInput(type: "String!", value: .scalar(ownerId))
+                return document.copy(inputs: inputs, selectionSet: selectionSet)
+            }
+        case .onDelete:
+            if operations.contains(.delete), let ownerId = ownerId {
+                var inputs = document.inputs
+                inputs[ownerField] = GraphQLDocumentInput(type: "String!", value: .scalar(ownerId))
+                return document.copy(inputs: inputs, selectionSet: selectionSet)
+            }
+        }
```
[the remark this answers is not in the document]
lawmicha: refactored this a bit

lawmicha:
thanks for taking a look, I discovered a scenario in the integration test where the `owner` was being passed in the variables as nil which was causing Unauthorized response. Added a check to see if there is a `nil` value and removing it completely from being sent to the service for the ownerFields

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
```diff
@@ -0,0 +1,106 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+// https://github.com/aws-amplify/amplify-cli/issues/4182
+// If the owner authorization does not contain `.read` operation, then others can read owners data,
+// do not have to apply any owner field to the subscription document input
+/// Decorate the document with auth related fields such as the selection set and owner input to subscription documents
```
wooj2, line 14: It's better to be precise, for example
```
Decorate the document with auth related fields.  Fields include:
* selection set
* owner field for subscription documents
* Something else?
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
```diff
@@ -0,0 +1,106 @@
...
+public struct AuthRuleDecorator: ModelBasedGraphQLDocumentDecorator {
+
+    private let subscriptionType: GraphQLSubscriptionType?
+    private let ownerId: String?
+
+    public init(subscriptionType: GraphQLSubscriptionType? = nil, ownerId: String? = nil) {
+        self.subscriptionType = subscriptionType
```
wooj2, line 21: Is this optional because we are using this decorator for both subscriptions out mutations?  If so, maybe it would be better to introduce a new type/enum which describes what we are decorating?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/AuthRule+Extension.swift
```diff
@@ -0,0 +1,18 @@
...
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+
+extension AuthRule {
+    func getOwnerFieldOrDefault() -> String {
+        return (ownerField != nil) ? ownerField!.stringValue : "owner"
```
wooj2, line 12: I think it would be better to use guard/let over force unwrap
```
        guard let ownerField = ownerField else {
            return "owner"
        }
        return ownerField.stringValue
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
```diff
@@ -0,0 +1,106 @@
...
+        if requiredOwnerInput, let ownerId = ownerId {
+            var inputs = document.inputs
+            inputs[ownerField] = GraphQLDocumentInput(type: "String!", value: .scalar(ownerId))
+            return document.copy(inputs: inputs, selectionSet: selectionSet)
+        }
+        return document.copy(selectionSet: selectionSet)
+    }
+
+    /// First finds the first `model` SelectionSet. Then, only append it when the `ownerField` does not exist.
+    func withOwnerField(selectionSet: SelectionSet, ownerField: String) -> SelectionSet {
```
wooj2, line 82: a name like `appendOwnerFieldToSelectionSetIfNeeded`

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
```diff
@@ -0,0 +1,106 @@
...
+            }
+        case .onUpdate:
+            if operations.contains(.update) {
+                requiredOwnerInput = true
+            }
+        case .onDelete:
+            if operations.contains(.delete) {
+                requiredOwnerInput = true
+            }
+        }
```
wooj2, line 71: Seems like we should capture this in a private function?

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
```diff
@@ -0,0 +1,106 @@
...
+
+        if requiredOwnerInput, let ownerId = ownerId {
```
wooj2, line 73: I'm trying to think if we should error here if ownerId is `nil`.  Correct me if i'm wrong here, but I think the answer is "no": we should not have a warning here because if ownerId is nil, we are hitting a case where users have been logged out, and we should be able to still create a document without the ownerId and attempt to make the request, correct?

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+AnyModelWithSync.swift
```diff
@@ -20,10 +20,11 @@ extension GraphQLRequest {
         documentBuilder.add(decorator: DirectiveNameDecorator(type: .get))
         documentBuilder.add(decorator: ModelIdDecorator(id: id))
         documentBuilder.add(decorator: ConflictResolutionDecorator())
+        documentBuilder.add(decorator: AuthRuleDecorator())
         let document = documentBuilder.build()
 
         return GraphQLRequest<MutationSyncResult?>(document: document.stringValue,
-                                                   variables: document.variables,
+                                                   variables: document.variables.isEmpty ? nil : document.variables,
```
wooj2, line 27: hmm why are we having to do this?

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
```diff
@@ -0,0 +1,106 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+// https://github.com/aws-amplify/amplify-cli/issues/4182
+// If the owner authorization does not contain `.read` operation, then others can read owners data,
+// do not have to apply any owner field to the subscription document input
```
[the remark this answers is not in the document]
lawmicha: for 1. this comment in particular is regarding the tracking issue, once the tracking issue is fixed, ideally the subscription operations provided require no additional inputs for all cases and the service simply decides whether or not to send subscription events to subscribers based on the `operations` defined. which is what i mean by "do noot have to apply any owner field to the subscription document input" - once issue is resolved. will expand on this

for 2. thanks, improving this comment

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
lawmicha: yeah it's used for both, adding enum like
```
public enum AuthRuleDecoratorInput {
    public typealias OwnerId = String
    case subscription(GraphQLSubscriptionType, OwnerId?)
    case mutation
}
```
to differentiate the two

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
lawmicha: this has got me thinking more and think it makes more sense that ownerId not optional in the AuthRuleDecorator because the decorators are supposed to be dumb and all it should do is to take the schema information and required inputs to decorate the document. This means it will only decorate the document input when 
1. we are in the context of providing a subscription document (input == .subscription)
2. ownerInput is required for this subscriptionType based on the auth rule's operation

By making ownerId required, the caller (DataStore) only has to check what type of backend is provisioned for that particular model. The backend is a conflict resolution enabled service by default, with a bunch of types with model directives. At runtime, it needs to check if the model.schema has authRules with owner auth strategy, then it will be responsible for checking if the user is signed in and retrieving the ownerId value. It will leave it up to the auth decorator, as part of building the GraphQLRequest, to know whether or not that ownerId is even needed.

lawmicha: If the user has logged out and the backend is provisioned with authRules containing owner strategy, then the caller would be responsible for waiting for the user to sign back in to retrieve the ownerId. subscription graphQLRequests cannot be made, now that we have enforced ownerId to be non-optional. 

If user is signed in already, and the graphQLRequest is created for the subscription operation, then before Amplify.API.subscription is called, the user is signed out. the call to Amplify.APi.Subscription should then fail with unauthenticated since API will also try to intercept request header with the user pool token and fail

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+AnyModelWithSync.swift
lawmicha: this is just because the `document.variables` returns [String:Any] that may be empty, and GraphQLRequest contains an optional [String:Any], so to be more accurate the GraphQLRequest should be nil when `document.variables` is empty. There's no real effect on the service by passing in empty variables payload. One thing we can do is to update the variables to be `[String:Any]?` since there is no requirement for `document.variables` to be non-optional, then we can simply pass `document.varibles` to the `GraphQLRequest.variables`

lawmicha: i didn't do this becuase it updates all the tests (bad reason lol) but it will be better since it will be aligned with the GraphQLRequest variables

lawmicha: made variables optional and updated this to just `variables: document.variables`

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
lawmicha: made ownerId required

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
```diff
@@ -0,0 +1,124 @@
...
+// @auth(rules: [ { allow: owner, operations: [create, delete] } ])
+// contains create and delete, therefore the onCreate and onDelete subscriptions require the owner field, but not the
+// onUpdate subscription.
+
+/// Decorate the document with auth rule related fields such as automatically adding owner to the the selection set and
+/// adding the owner input to subscription document input
+///
+/// Decorate the document with auth related fields. For `owner` strategy, fields include:
+/// * add the value of `ownerField` to the model selection set, defaults "owner" when `ownerField` is not specified
+/// * owner field value for subscription document inputs for the corresponding auth rule `operation`
```
wooj2, line 31: I think we can remove all of this?

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/Decorator/AuthRuleDecorator/ModelMultipleOwnerAuthRuleTests.swift
```diff
@@ -0,0 +1,298 @@
...
+@testable import AmplifyTestCommon
+@testable import AWSPluginsCore
+/*
+ type ModelMultipleOwner @model
+   @auth(rules: [
+
+     # Defaults to use the "owner" field.
+     { allow: owner },
+
+     # Authorize the update mutation and both queries. Use `queries: null` to disable auth for queries.
```
wooj2, line 19: both queries? do you mean read queries?

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
lawmicha: woops, forgot to remove line 26-27

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/Decorator/AuthRuleDecorator/ModelMultipleOwnerAuthRuleTests.swift
lawmicha: good question, i copied that example from https://docs.amplify.aws/cli/graphql-transformer/directives#owner-authorization .. going to remove it from our comment for now.. until i re-read the documentation about that

PR #425, opened 2020-05-07
feat: [API] Mutate with condition

lawmicha:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+GraphQLBehavior.swift
+++ b/Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+GraphQLBehavior.swift
@@ -22,9 +22,10 @@ extension AmplifyAPICategory: APICategoryGraphQLBehavior {
     }
 
     public func mutate<M: Model>(of model: M,
+                                 where condition: QueryPredicate? = nil,
                                  type: GraphQLMutationType,
                                  listener: GraphQLOperation<M>.EventListener?) -> GraphQLOperation<M> {
-        plugin.mutate(of: model, type: type, listener: listener)
+        plugin.mutate(of: model, where: condition, type: type, listener: listener)
     }
 
     public func subscribe<M: Model>(from modelType: M.Type,
--- a/Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+GraphQLBehavior.swift
+++ b/Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+GraphQLBehavior.swift
@@ -28,6 +28,12 @@ extension AmplifyAPICategory: APICategoryGraphQLBehavior {
         plugin.mutate(of: model, where: condition, type: type, listener: listener)
     }
 
+    public func mutate<M: Model>(of model: M,
+                                 type: GraphQLMutationType,
+                                 listener: GraphQLOperation<M>.EventListener?) -> GraphQLOperation<M> {
+        plugin.mutate(of: model, type: type, listener: listener)
+    }
+
     public func subscribe<M: Model>(from modelType: M.Type,
                                     type: GraphQLSubscriptionType,
                                     listener: GraphQLSubscriptionOperation<M>.EventListener?)
--- a/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
@@ -295,7 +295,7 @@ class GraphQLModelBasedTests: XCTestCase {
         wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
     }
 
-    func testUpdatePostWithInvalidCondition() {
+    func testUpdatePostWithInvalidConditionShouldReturnConditionalCheckFailed() {
         let uuid = UUID().uuidString
         let testMethodName = String("\(#function)".dropLast(2))
         let title = testMethodName + "Title"
@@ -303,13 +303,12 @@ class GraphQLModelBasedTests: XCTestCase {
             XCTFail("Failed to ensure at least one Post to be retrieved on the listQuery")
             return
         }
-        let conditionTitle = "IdontCareUpdated"
         let failureInvoked = expectation(description: "request failed")
 
         let post = Post.keys
-        let condition = post.title == conditionTitle
-
-        _ = Amplify.API.mutate(of: createPost, where: condition, type: .update, listener: { event in
+        let invalidCondition = post.title == "invalidTitle"
+        
+        _ = Amplify.API.mutate(of: createPost, where: invalidCondition, type: .update, listener: { event in
             switch event {
             case .completed(let data):
                 switch data {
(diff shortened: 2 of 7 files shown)


Amplify/Categories/API/ClientBehavior/AmplifyAPICategory+GraphQLBehavior.swift
```diff
@@ -21,6 +21,13 @@ extension AmplifyAPICategory: APICategoryGraphQLBehavior {
         plugin.query(from: modelType, where: predicate, listener: listener)
     }
 
+    public func mutate<M: Model>(of model: M,
+                                 where condition: QueryPredicate? = nil,
```
lawmicha, line 25: we should confirm that this is the case, before finalizing on adding the method overload, that swift does not support omitting the parameter where there is a default parameter.

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
```diff
@@ -294,6 +295,54 @@ class GraphQLModelBasedTests: XCTestCase {
...
+                        XCTAssertEqual(errorType, .conditionalCheck)
+                        failureInvoked.fulfill()
+                    case .partial(let model, let errors):
+                        XCTFail("partial: \(model), \(errors)")
+                    case .transformationError(let rawResponse, let apiError):
+                        XCTFail("transformationError: \(rawResponse), \(apiError)")
+                    }
+                }
+            case .failed(let error):
+                print(error)
```
lawmicha, line 338: can we change this to XCTFail as well?

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
```diff
@@ -294,6 +295,54 @@ class GraphQLModelBasedTests: XCTestCase {
...
+
+        let post = Post.keys
+        let condition = post.title == conditionTitle
+
+        _ = Amplify.API.mutate(of: createPost, where: condition, type: .update, listener: { event in
+            switch event {
+            case .completed(let data):
+                switch data {
+                case .success:
+                    XCTFail("This test should fail since we are expecting the condition check to fail on backend")
```
lawmicha, line 317: i think we can reduce this message without losing any the information, making it more concise:
"Received successful response data but failure with conditional check error"
- No need to say "this test should fail" since it is in the context of running the test
- "condtion[al] check" 
- "on backend" also seems like extra info since this is only an integration with the one service

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
```diff
@@ -294,6 +295,54 @@ class GraphQLModelBasedTests: XCTestCase {
...
 
+    func testUpdatePostWithInvalidCondition() {
+        let uuid = UUID().uuidString
+        let testMethodName = String("\(#function)".dropLast(2))
+        let title = testMethodName + "Title"
+        guard let createPost = createPost(id: uuid, title: title) else {
+            XCTFail("Failed to ensure at least one Post to be retrieved on the listQuery")
+            return
+        }
+        let conditionTitle = "IdontCareUpdated"
```
lawmicha, line 306: the test is "testUpdatePostWithInvalidCondition", perfer the naming of the variable match closely ie.
```
let invalidCondition = post.title == "invalidTitle"
```
also there's some verbosity with setting up the test with one-time use variables, but in this case, creating the condition is part of setting the test up, and i don't think we need an extra variable like "let conditionTitle"

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
```diff
@@ -294,6 +295,54 @@ class GraphQLModelBasedTests: XCTestCase {
         wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
     }
 
+    func testUpdatePostWithInvalidCondition() {
```
lawmicha, line 298: `testUpdatePostWithInvalidConditionShouldReturnConditionalCheckFailed()`

AmplifyPlugins/Core/AWSPluginsCore/API/AppSyncErrorType.swift
```diff
@@ -12,7 +12,7 @@ public enum AppSyncErrorType: Equatable {
 
     private static let conditionalCheckFailedErrorString = "ConditionalCheckFailedException"
     private static let conflictUnhandledErrorString = "ConflictUnhandled"
-
+    private static let dynomoDBConditionalCheckFailed = "DynamoDB:ConditionalCheckFailedException"
```
lawmicha, line 15: move this line above line 14, to group the conditional check failed strings together

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
DongQuanRui: Well, this is a long name😂

lawmicha:
let's add a unit test for this

AmplifyPlugins/Core/AWSPluginsCore/API/AppSyncErrorType.swift
DongQuanRui: moved

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
DongQuanRui: true. Changed

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
DongQuanRui: Changed to "XCTFail("Received successful response data but expecting conditional check failed")"

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLModelBasedTests.swift
DongQuanRui: I don't know. I noticed that some test simply use "print(error)", and some use "XCTFail("Unexpected .failed event: \(error)")". But if we want the test to stop there, we can use XCTFail.

lawmicha:
this can be closed for https://github.com/aws-amplify/amplify-ios/pull/470, however tests can be carried forward in a new PR if rebasing from that is more work than just grabbing the latest master branch (after PR is merged)

PR #433, opened 2020-05-11
feat: [API] auth directive integration tests

lawmicha:
*Issue #, if available:*

*Description of changes:*
This builds on top of https://github.com/aws-amplify/amplify-ios/pull/423 which provides the graphQL document builders to make the requests.

The integration tests run against a backend provisioined
- with a schema containing a model with auth directive
- cognito user pool as the auth mode.
- conflict resolution enabled

The schema is a read-only model by others other than owner.
```
type SocialNote
    @model
    @auth(rules: [
        { allow: owner, ownerField: "owner", operations: [create, update, delete] },
    ]) {
    id: ID!
    content: String!
    owner: String
}
```

This test mainly executes methods relating to DataStore usage, ensuring expected service behavior and validating how the caller handles the API response, especially when the caller is unauthorized:
- Mutations by owner/other
- syncQuery, when authorized and not authorized
- subscription operation with ownerId in document input

### DIFF
--- a/AmplifyPlugins/API/APICategoryPlugin.xcodeproj/xcshareddata/xcschemes/GraphQLWithUserPoolIntegrationTests.xcscheme
+++ b/AmplifyPlugins/API/APICategoryPlugin.xcodeproj/xcshareddata/xcschemes/GraphQLWithUserPoolIntegrationTests.xcscheme
@@ -23,6 +23,9 @@
             </BuildableReference>
             <SkippedTests>
                <Test
+                  Identifier = "GraphQLWithUserPoolIntegrationTests/testCreateTodoMutationWithMissingInputFromVariables()">
+               </Test>
+               <Test
                   Identifier = "GraphQLWithUserPoolIntegrationTests/testSetUpOnce()">
                </Test>
             </SkippedTests>
--- a/AmplifyPlugins/API/APICategoryPlugin.xcodeproj/xcshareddata/xcschemes/GraphQLWithUserPoolIntegrationTests.xcscheme
+++ b/AmplifyPlugins/API/APICategoryPlugin.xcodeproj/xcshareddata/xcschemes/GraphQLWithUserPoolIntegrationTests.xcscheme
@@ -23,9 +23,6 @@
             </BuildableReference>
             <SkippedTests>
                <Test
-                  Identifier = "GraphQLWithUserPoolIntegrationTests/testCreateTodoMutationWithMissingInputFromVariables()">
-               </Test>
-               <Test
                   Identifier = "GraphQLWithUserPoolIntegrationTests/testSetUpOnce()">
                </Test>
             </SkippedTests>
--- a/AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
@@ -1,0 +1,174 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+@testable import Amplify
+import AWSMobileClient
+import AWSAPICategoryPlugin
+@testable import AWSAPICategoryPluginTestCommon
+@testable import AmplifyTestCommon
+import AWSPluginsCore
+
+extension GraphQLAuthDirectiveIntegrationTests {
+    func createNote(_ id: String = UUID().uuidString,
+                    content: String) -> Result<MutationSyncResult, GraphQLResponseError<MutationSyncResult>> {
+        let createdNoteInvoked = expectation(description: "note was created")
+        var result: Result<MutationSyncResult, GraphQLResponseError<MutationSyncResult>>?
+        let note = SocialNote(id: id, content: content, owner: nil)
+        let request = GraphQLRequest<MutationSyncResult>.createMutation(of: note)
+        _ = Amplify.API.mutate(request: request, listener: { event in
+            switch event {
+            case .completed(let data):
+                switch data {
+                case .success(let note):
+                    result = .success(note)
+                case .failure(let error):
+                    result = .failure(error)
+                }
+                createdNoteInvoked.fulfill()
+            case .failed(let error):
+                XCTFail("Got failed, error: \(error)")
+            default:
+                XCTFail("Unexpected event: \(event)")
+            }
+        })
+
+        wait(for: [createdNoteInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result!
+    }
+
+    func updateNote(_ note: SocialNote,
+                    version: Int) -> Result<MutationSyncResult, GraphQLResponseError<MutationSyncResult>> {
+        let updateNoteInvoked = expectation(description: "note was update")
+        var result: Result<MutationSyncResult, GraphQLResponseError<MutationSyncResult>>?
+        let request = GraphQLRequest<MutationSyncResult>.updateMutation(of: note, version: version)
+        _ = Amplify.API.mutate(request: request, listener: { event in
+            switch event {
+            case .completed(let data):
+                switch data {
+                case .success(let note):
+                    result = .success(note)
+                case .failure(let error):
+                    result = .failure(error)
+                }
+                updateNoteInvoked.fulfill()
+            case .failed(let error):
+                XCTFail("Got failed, error: \(error)")
+            default:
+                XCTFail("Unexpected event: \(event)")
+            }
+        })
+
+        wait(for: [updateNoteInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result!
+    }
+
+    func deleteNote(byId id: String,
+                    version: Int) -> Result<MutationSyncResult, GraphQLResponseError<MutationSyncResult>> {
+        let deleteNoteInvoked = expectation(description: "note was deleted")
+        var result: Result<MutationSyncResult, GraphQLResponseError<MutationSyncResult>>?
+        let request = GraphQLRequest<MutationSyncResult>.deleteMutation(modelName: SocialNote.modelName,
+                                                                        id: id,
+                                                                        version: version)
+        _ = Amplify.API.mutate(request: request, listener: { event in
+            switch event {
+            case .completed(let data):
+                switch data {
+                case .success(let note):
+                    result = .success(note)
+                case .failure(let error):
+                    result = .failure(error)
+                }
+                deleteNoteInvoked.fulfill()
+            case .failed(let error):
+                XCTFail("Got failed, error: \(error)")
+            default:
+                XCTFail("Unexpected event: \(event)")
+            }
+        })
+
+        wait(for: [deleteNoteInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result!
+    }
+
+    func queryNote(byId id: String) -> Result<MutationSyncResult?, GraphQLResponseError<MutationSyncResult?>> {
+        let queryNoteInvoked = expectation(description: "note was queried successfully")
+        var result: Result<MutationSyncResult?, GraphQLResponseError<MutationSyncResult?>>?
+        let request = GraphQLRequest<MutationSyncResult?>.query(modelName: SocialNote.modelName, byId: id)
+        _ = Amplify.API.query(request: request) { event in
+            switch event {
+            case .completed(let data):
+                switch data {
+                case .success(let note):
+                    result = .success(note)
+                case .failure(let error):
+                    result = .failure(error)
+                }
+                queryNoteInvoked.fulfill()
+            case .failed(let error):
+                XCTFail("Got failed, error: \(error)")
+            default:
+                XCTFail("Unexpected event: \(event)")
+            }
+        }
+
+        wait(for: [queryNoteInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result!
+    }
+
+    func syncQuery() -> Result<SyncQueryResult, GraphQLResponseError<SyncQueryResult>> {
+        let syncQueryInvoked = expectation(description: "sync query successfully")
+        var result: Result<SyncQueryResult, GraphQLResponseError<SyncQueryResult>>?
+        let request = GraphQLRequest<SyncQueryResult>.syncQuery(modelType: SocialNote.self, limit: 1)
+        _ = Amplify.API.query(request: request) { event in
+            switch event {
+            case .completed(let data):
+                switch data {
+                case .success(let paginatedNotes):
+                    result = .success(paginatedNotes)
+                case .failure(let error):
+                    result = .failure(error)
+                }
+                syncQueryInvoked.fulfill()
+            case .failed(let error):
+                XCTFail("Got failed, error: \(error)")
+            default:
+                XCTFail("Unexpected event: \(event)")
+            }
+        }
+        wait(for: [syncQueryInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result!
+    }
+
+    func getAppSyncError(_ graphQLResponseError: GraphQLResponseError<MutationSyncResult>) -> AppSyncErrorType? {
+        guard case let .error(errors) = graphQLResponseError,
+            let error = errors.first,
+            let extensions = error.extensions,
+            case let .string(errorTypeValue) = extensions["errorType"] else {
+                XCTFail("Other should not be able to update owner's note")
+            return nil
+        }
+        return AppSyncErrorType(errorTypeValue)
+    }
+
+    func assertNotAuthenticated(_ error: APIError) {
+        guard case let .operationError(_, _, underlyingError) = error else {
+            XCTFail("Error should be operationError")
+            return
+        }
+
+        guard let authError = underlyingError as? AuthError else {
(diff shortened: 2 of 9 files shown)


AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
```diff
@@ -0,0 +1,174 @@
...
+            }
+        })
+
+        wait(for: [createdNoteInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result!
+    }
+
+    func updateNote(_ note: SocialNote,
+                    version: Int) -> Result<MutationSyncResult, GraphQLResponseError<MutationSyncResult>> {
+        let updateNoteInvoked = expectation(description: "note was update")
```
wooj2, line 46: s/update/updated/ ?

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
```diff
@@ -0,0 +1,174 @@
...
+                createdNoteInvoked.fulfill()
+            case .failed(let error):
+                XCTFail("Got failed, error: \(error)")
+            default:
+                XCTFail("Unexpected event: \(event)")
+            }
+        })
+
+        wait(for: [createdNoteInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result!
```
wooj2, line 41: Instead of using force unwrap how about something like..
```
guard let result = result else {
    XCTFail("Failed on creating note")
}
return result
```
Seems like it would be a friendlier error in the event that we couldn't get a result? And of course, if we do this here, we can apply this to the rest of the functions in this file.

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
```diff
@@ -0,0 +1,174 @@
...
+                XCTFail("Unexpected event: \(event)")
+            }
+        })
+
+        wait(for: [deleteNoteInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result!
+    }
+
+    func queryNote(byId id: String) -> Result<MutationSyncResult?, GraphQLResponseError<MutationSyncResult?>> {
+        let queryNoteInvoked = expectation(description: "note was queried successfully")
```
wooj2, line 99: remove `successfully` to be consistent with other functions

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
```diff
@@ -0,0 +1,174 @@
...
+                case .failure(let error):
+                    result = .failure(error)
+                }
+                queryNoteInvoked.fulfill()
+            case .failed(let error):
+                XCTFail("Got failed, error: \(error)")
+            default:
+                XCTFail("Unexpected event: \(event)")
+            }
+        }
```
wooj2, line 117: Seems like there should be a way to consolidate this handler into a single function across all these functions?

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
```diff
@@ -0,0 +1,174 @@
...
+        wait(for: [syncQueryInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result!
+    }
+
+    func getAppSyncError(_ graphQLResponseError: GraphQLResponseError<MutationSyncResult>) -> AppSyncErrorType? {
+        guard case let .error(errors) = graphQLResponseError,
+            let error = errors.first,
+            let extensions = error.extensions,
+            case let .string(errorTypeValue) = extensions["errorType"] else {
+                XCTFail("Other should not be able to update owner's note")
```
wooj2, line 152: This error of others not being able to update an owners note seems super specific...  we either need to update the function name, or make this function more generic.

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
```diff
@@ -0,0 +1,244 @@
...
+import AWSPluginsCore
+
+class GraphQLAuthDirectiveIntegrationTests: XCTestCase {
+
+    static let amplifyConfiguration = "GraphQLAuthDirectiveIntegrationTests-amplifyconfiguration"
+    static let awsconfiguration = "GraphQLAuthDirectiveIntegrationTests-awsconfiguration"
+    static let credentials = "GraphQLAuthDirectiveIntegrationTests-credentials"
+    static var user1: String!
+    static var user2: String!
+    static var password: String!
```
wooj2, line 23: Is there a strong argument/reason to make these static?  Reading this make me believe that the tests in this file depend on each other and, I don't think that we want that.... Instead, i think we want independent test that we can run in isolation.

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
```diff
@@ -0,0 +1,244 @@
...
+        }
+    }
+
+    override func tearDown() {
+        Amplify.reset()
+        AuthHelper.signOut()
+    }
+
+    /// A model with operations [create, update, delete] allows the model to be read by others but not updated or
+    /// deleted. This creates a read-only model.
```
wooj2, line 74: I think we should adopt the following ascii table, and put something like this directly in our code.
```
Models created with:
   @auth(rules: [ { allow: owner, operations: [create, update, delete] } ])

Yield the following permissions model:
            Create   Read   Update   Delete
   Owner    allow   allow    allow    allow
   Others    x      allow    deny     deny

```

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
```diff
@@ -0,0 +1,244 @@
...
+            guard let user1 = credentials["user1"],
+                let user2 = credentials["user2"],
+                let password = credentials["password"] else {
+                    XCTFail("Missing credentials.json data")
+                    return
+            }
+
+            GraphQLAuthDirectiveIntegrationTests.user1 = user1
+            GraphQLAuthDirectiveIntegrationTests.user2 = user2
+            GraphQLAuthDirectiveIntegrationTests.password = password
```
wooj2, line 40: ideally, user1 and user2 would have different passwords to avoid us thinking we are logging into user1 when we are really logging in with user2.

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
```diff
@@ -0,0 +1,244 @@
...
+        }
+        let ownerUpdateNote = SocialNote(id: ownerReadNote.model.id, content: "owner updated content", owner: nil)
+        let ownerUpdatedNoteResult = updateNote(ownerUpdateNote, version: ownerReadNote.syncMetadata.version)
+        guard case let .success(ownerUpdatedNote) = ownerUpdatedNoteResult else {
+            XCTFail("Owner should be able to update own note")
+            return
+        }
+        AuthHelper.signOut()
+        AuthHelper.signIn(username: GraphQLAuthDirectiveIntegrationTests.user2,
+                          password: GraphQLAuthDirectiveIntegrationTests.password)
```
wooj2, line 107: Is there any way to validate that we are we are actually logged in as a different user? e.g. validate a change in the userSub or something like that (and fail otherwise)?

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
```diff
@@ -0,0 +1,244 @@
...
+        let otherReadNoteResult = queryNote(byId: id)
```
wooj2, line 108: Technically this is cheating since we had the Id -- is it a ton more work to query by some other means?

AmplifyPlugins/API/APICategoryPlugin.xcodeproj/xcshareddata/xcschemes/GraphQLWithUserPoolIntegrationTests.xcscheme
```diff
@@ -22,6 +22,9 @@
                ReferencedContainer = "container:APICategoryPlugin.xcodeproj">
             </BuildableReference>
             <SkippedTests>
+               <Test
+                  Identifier = "GraphQLWithUserPoolIntegrationTests/testCreateTodoMutationWithMissingInputFromVariables()">
+               </Test>
```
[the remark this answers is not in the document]
lawmicha: no, reverted

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
lawmicha: updated, only the description, i think the variable ends with "Invoked" is already past tense

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
lawmicha: i wanted the return value to be easy to manage for the tests to use, which makes it hard to return something in the guard that is not coming from asyncEvent's completed case. I can use guard here and throw a fatalError because it's unexpected, which is equivalent of the force cast

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
lawmicha: I couldn't find a way to do this, the problem is that return value is different for each case
createMutation -> Note
updateMutation -> Note
deleteMutation -> Note
query -> Note?
syncQuery -> SyncQueryResult

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
lawmicha: no strong reason, making these instances and removing static init made accessing these variables much easier, without having to put the class name in front. having them static meant we only read the credentials file once

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
lawmicha: updating

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
lawmicha: thanks, adding this

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
lawmicha: good idea, this would be good to do when we go to migrate the AuthHelper to use Amplify.Auth, we can improve the method to check 
- if signed in 
- if yes, grab existing userSub then sign out
- and sign in with user2, and check if userSub is updated. 
I would prefer to keep this in mind and do this when we migrate each category to use Amplify.Auth

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
lawmicha: in a real scenario, the other would probably listQuery with a reduced selection set to display minimal info and then when someone taps on the item, they will then load the details with the id using getQuery. so someone has to choose which item they want to load. in our case, we're testing permissions to be able to get that data back, so whether we're doing a `listQuery(filter: SocialNote.keys.id == id)` or `queryNote(byId: id)`, we would still need the id to begin with

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests.swift
wooj2: I generally like to avoid static variables for unit tests.  variables can be accidentally overwritten and then impact other tests..  If it's not a lot of work can we switch this over to non-static?

lawmicha: ah yes, should have been more clear, i have already updated to use instance variables

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective/GraphQLAuthDirectiveIntegrationTests+Support.swift
```diff
@@ -0,0 +1,174 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+@testable import Amplify
+import AWSMobileClient
+import AWSAPICategoryPlugin
+@testable import AWSAPICategoryPluginTestCommon
+@testable import AmplifyTestCommon
+import AWSPluginsCore
```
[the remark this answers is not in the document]
lawmicha: missing one of the files, updating

PR #434, opened 2020-05-12
refactor: Renames AmplifyAuthError to AuthError

royjit:
BREAKING CHANGE: Auth category error type is changed to AuthError, current implementation that make use of AmplifyAuthError
will break with build time error.

Also deleted AuthError in storage category.


AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthService.swift
```diff
@@ -81,31 +81,8 @@ public class AWSAuthService: AWSAuthServiceBehavior {
...
-                                              "The user needs be authenticated to make this request",
-                                              error)
-        case .notAuthorized(let message):
-            return AuthError.notAuthorized(message,
-                                           "The user does not have the correct permissions to make this request",
-                                           error)
-        default:
-            return AuthError.unknown(error.message)
-        }
+        // TODO: Use main auth error from Auth plugin
```
lawmicha, line 84: what does this mean?

royjit: My next PR will handle this, basically to use the error coming from Auth plugin. We will not need this mapping method since we will be removing AWSMobileClient from AWSPluginsCore

lawmicha: these look like they were specific to when using storage, need to verify Storage Integration tests negative paths

lawmicha: https://github.com/aws-amplify/amplify-ios/issues/435

PR #438, opened 2020-05-12
feat: [DataStore][AuthDirective] SyncToCloud not authenticated error handling

lawmicha:
*Issue #, if available:*

*Description of changes:*
Scenario: 'Owner' is not signed in, persists to local store via DataStore.save()/DataStore.delete(), the mutationEvent is then attempted to call API.mutate. the mutate fails with `notAuthenticated` error.

This change checks for unauthenticated calls to API and propagates the error and corresponding MutationEvent to the developer using `DataStoreConfiguration.errorHandler`.

The error provided to the developer is
```
public enum DataStoreError: Error {
    case api(AmplifyError, MutationEvent? = nil)
```
With the mutation event, the developer can see that the API call failed for some particular DataStore call. The developer can decide what to do. For example, 
- delete the data using the mutation event's model identifier, if it were a `.create` mutation type.
- prompt the user for sign in the user, and oncall back, reattempt to the DataStore.save()/DataStore.delete()

Another Scenario to consider: 'Other' retrieves data from local store, and attempts to update it. The update is successful on local store, however the mutationEvent is attempted fails with conditional save failed. This is already propagated to the Hub as a `HubPayload.EventName.DataStore.conditionalSaveFailed` event

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
@@ -8,6 +8,7 @@
 import Amplify
 import Combine
 import Foundation
+import AWSPluginsCore
 
 /// Submits outgoing mutation events to the provisioned API
 @available(iOS 13.0, *)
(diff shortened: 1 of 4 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -179,36 +180,51 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
...
+                dataStoreConfiguration: self.dataStoreConfiguration,
+                mutationEvent: mutationEvent,
+                api: api,
+                storageAdapter: self.storageAdapter,
+                apiError: apiError) { result in
+                    if case let .success(mutationEventOptional) = result,
+                        let outgoingMutationEvent = mutationEventOptional {
+                        self.outgoingMutationQueueSubject.send(outgoingMutationEvent)
+                    }
+                    self.completeProcessingEvent(mutationEvent)
```
wooj2, line 222: rather than repeat this twice, maybe it would be better to have:
```
    private func processSyncMutationToCloudResult(_ result: GraphQLOperation<MutationSync<AnyModel>>.Event,
                                                  mutationEvent: MutationEvent,
                                                  api: APICategoryGraphQLBehavior) {
        if case let .completed(response) = result, case let .failure(graphQLResponseError) = response {
            processMutationErrorFromCloudOperation(mutationEvent: mutationEvent,
                                                   api: api,
                                                   apiError: nil,
                                                   graphQLResponseError: graphQLResponseError)
        } else if case let .failed(apiError) = result {
            processMutationErrorFromCloudOperation(mutationEvent: mutationEvent,
                                                   api: api,
                                                   apiError: apiError,
                                                   graphQLResponseError: nil)

        } else {
            completeProcessingEvent(mutationEvent)
        }
    }

    func processMutationErrorFromCloudOperation(mutationEvent: MutationEvent,
                                                api: APICategoryGraphQLBehavior,
                                                apiError: APIError?,
                                                graphQLResponseError: GraphQLResponseError<MutationSync<AnyModel>>?) {
        let processMutationErrorFromCloudOperation = ProcessMutationErrorFromCloudOperation(
            dataStoreConfiguration: dataStoreConfiguration,
            mutationEvent: mutationEvent,
            api: api,
            storageAdapter: storageAdapter,
            graphQLResponseError: graphQLResponseError,
            apiError: apiError) { result in
                if case let .success(mutationEventOptional) = result,
                    let outgoingMutationEvent = mutationEventOptional {
                    self.outgoingMutationQueueSubject.send(outgoingMutationEvent)
                }
                self.completeProcessingEvent(mutationEvent)
        }
        operationQueue.addOperation(processMutationErrorFromCloudOperation)
    }
```

lawmicha: thanks for the code snippet!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -179,36 +180,51 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
...
+                                                  mutationEvent: MutationEvent,
+                                                  api: APICategoryGraphQLBehavior) {
+        if case let .completed(response) = result, case let .failure(graphQLResponseError) = response {
+            let processMutationErrorFromCloudOperation = ProcessMutationErrorFromCloudOperation(
+                dataStoreConfiguration: self.dataStoreConfiguration,
+                mutationEvent: mutationEvent,
+                api: api,
+                storageAdapter: self.storageAdapter,
+                graphQLResponseError: graphQLResponseError) { result in
+                    if case let .success(mutationEventOptional) = result,
```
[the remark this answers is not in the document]
lawmicha: adding back

lawmicha:
just an update on this, the test was failing unit test after rebasing with the Amplify.Auth changes. Amplify.Auth returns `.service` with `AWSCognitoError.notSignedIn`. for developers, they would then check the underlying `AWSCognitoError`. For DataStore this is a bit difficult because AWSCognitoError is defined in AWSAuthPlugin. 

it would be convenient for DataStore logic and developers if they could just check something like .notSignedIn or .notAuthenticated from the top level as an AuthError case. thoughts @royjit ?

royjit:
> just an update on this, the test was failing unit test after rebasing with the Amplify.Auth changes. Amplify.Auth returns `.service` with `AWSCognitoError.notSignedIn`. for developers, they would then check the underlying `AWSCognitoError`. For DataStore this is a bit difficult because AWSCognitoError is defined in AWSAuthPlugin.
> 
> it would be convenient for DataStore logic and developers if they could just check something like .notSignedIn or .notAuthenticated from the top level as an AuthError case. thoughts @royjit ?

Added signedOut and sessionExpiry error - https://github.com/aws-amplify/amplify-ios/pull/459

Issue #419, opened 2020-05-06
Auth - Fix attribute string


royjit:
Fix the auth attribute structure as per the PR comment - https://github.com/aws-amplify/amplify-ios/pull/407/files#r420521508

royjit:
PR out for handling this issue

royjit:
PR merged https://github.com/aws-amplify/amplify-ios/pull/439

PR #439, opened 2020-05-12
refactor(Auth): Refactored AuthUserAttributeKey to be RawRepresentable

royjit:
- AuthUserAttributeKey to [RawRepresentable](https://developer.apple.com/documentation/swift/rawrepresentable), handles #419 
- Changed String attributeName in CodeDelivery to AuthUserAttributeKey

Update 1

- Added a new case `unknown` to the enum list to track an attribute unknown to the system and not a custom attribute.
- Raw value given in init will always be returned when we call `Enum.rawValue`
- Added missing userService and deviceService to the configuration

### DIFF
--- a/AmplifyPlugins/Auth/AWSAuthPlugin/Support/Utils/AuthUserAttributeKey+Extension.swift
+++ b/AmplifyPlugins/Auth/AWSAuthPlugin/Support/Utils/AuthUserAttributeKey+Extension.swift
@@ -8,75 +8,97 @@
 import Amplify
 
 // swiftlint:disable cyclomatic_complexity
-extension AuthUserAttributeKey {
+extension AuthUserAttributeKey: RawRepresentable {
 
-    func toString() -> String {
-        switch self {
-        case .address:
-            return "address"
-        case .birthDate:
-            return "birthdate"
-        case .email:
-            return "email"
-        case .familyName:
-            return "family_name"
-        case .gender:
-            return "gender"
-        case .givenName:
-            return "given_name"
-        case .locale:
-            return "locale"
-        case .middleName:
-            return "middle_name"
-        case .name:
-            return "name"
-        case .nickname:
-            return "nickname"
-        case .phoneNumber:
-            return "phone_number"
-        case .picture:
-            return "picture"
-        case .preferredUsername:
-            return "preferred_username"
-        case .custom(let attribute):
-            return "custom: \(attribute)"
+    public typealias RawValue = String
+
+    // Values are taken from:
+    // https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html
+    private static let addressRawValue = "address"
+    private static let birthDateRawValue = "birthdate"
+    private static let emailRawValue = "email"
+    private static let familyNameRawValue = "family_name"
+    private static let genderRawValue = "gender"
+    private static let givenNameRawValue = "given_name"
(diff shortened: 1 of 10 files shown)


AmplifyPlugins/Auth/AWSAuthPlugin/Support/Utils/AuthUserAttributeKey+Extension.swift
```diff
@@ -8,75 +8,89 @@
...
-            return "phone_number"
+            return AuthUserAttributeKey.phoneNumberRawValue
         case .picture:
-            return "picture"
+            return AuthUserAttributeKey.pictureRawValue
         case .preferredUsername:
-            return "preferred_username"
+            return AuthUserAttributeKey.preferredUsernameRawValue
         case .custom(let attribute):
             return "custom: \(attribute)"
```
lawmicha, line 93: i think this should just be `return attribute`, since what was passed in was the `attribute` string and stored in `.custom` and then converting it back should be lossy

AmplifyPlugins/Auth/AWSAuthPluginTests/Utils/AuthUserAttributeKeyTests.swift
```diff
@@ -0,0 +1,73 @@
...
+    /// - Given: A custom attribute from AuthUserAttributeKey
+    /// - When:
+    ///    - I get rawValue of the enum
+    /// - Then:
+    ///    - The value should be equal to "custom: <value given>"
+    ///
+    func testSuccessfulCustomRawValue() {
+        let customAttributeKey = "someattribute2"
+        let attribute = AuthUserAttributeKey.custom(customAttributeKey)
+        XCTAssertEqual(attribute.rawValue, "custom: \(customAttributeKey)")
```
lawmicha, line 70: i'm not entirely sure when there is a custom attribute key, maybe when the service updates and library is not yet handling the new key, then it would get created as `.custom(key)`. then if it's forward compatible, i would expect the library to just get the attributeKey as is as a string `<key>`. Using `custom: <key>` will break since the service doesn't know what to do with it

AmplifyPlugins/Auth/AWSAuthPlugin/Support/Utils/AuthUserAttributeKey+Extension.swift
royjit: Changed the logic to return what was passed in during init. Custom attribute require to be init with "custom:" as the prefix. Here are some examples:
```
let x = AuthUserAttributeKey(rawValue: "someUnknownattribute")
x.rawValue // returns "someUnknownattribute"
x == .unknown("someUnknownattribute")

let x = AuthUserAttributeKey(rawValue: "custom:someattribute")
x.rawValue // returns "custom:someattribute"
x == .custom("someattribute")

let x = AuthUserAttributeKey(rawValue: "middle_name")
x.rawValue // returns "middle_name"
x == .middleName
```

This way the logic doesnot break if the service adds new attributes. It will just fall back to unknown enum meaning that the attribute is not yet known to the Amplify.Auth.

AmplifyPlugins/Auth/AWSAuthPluginTests/Utils/AuthUserAttributeKeyTests.swift
royjit: Explained above

PR #445, opened 2020-05-14
fix(tools): change the tools script to comply with amplify-app changes

drochetti:
See https://github.com/aws-amplify/amplify-cli/pull/4268 for details

### DIFF
--- a/AmplifyTools/amplify-tools.sh
+++ b/AmplifyTools/amplify-tools.sh
@@ -11,11 +11,11 @@ export PATH=$PATH:$(npm bin -g)
 if ! which node >/dev/null; then
   echo "warning: Node is not installed. Vist https://nodejs.org/en/download/ to install it"
   exit 1
-elif ! test -f ./amplifyxc.config; then
+elif ! test -f ./amplifytools.xcconfig; then
   npx amplify-app --platform ios
 fi
 
-. amplifyxc.config
+. amplifytools.xcconfig
 amplifyPush=$push
 amplifyModelgen=$modelgen
 amplifyProfile=$profile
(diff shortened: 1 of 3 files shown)


Podfile.lock
```diff
@@ -24,7 +24,7 @@ DEPENDENCIES:
   - SwiftLint
 
 SPEC REPOS:
-  trunk:
+  https://cdn.cocoapods.org/:
```
drochetti, line 27: @palpatim is this expected/correct?

AmplifyTools/amplify-tools.sh
```diff
@@ -25,7 +25,10 @@ amplifyRegion=$region
 amplifyEnvName=$envName
 
 if $amplifyModelgen; then
```
wooj2, line 27: not familiar with this syntax, i would have expected to check if amplifyModelgen is not set by using:
```
if [[ ! -z $amplifyModelgen ]]; then
```

Pods/Manifest.lock
```diff
@@ -24,7 +24,7 @@ DEPENDENCIES:
   - SwiftLint
 
 SPEC REPOS:
-  trunk:
+  https://cdn.cocoapods.org/:
```
wooj2, line 27: per tim's comment above -- same with this.. is this expected? i would have probably reverted this change.

AmplifyTools/amplify-tools.sh
```diff
@@ -25,7 +25,10 @@ amplifyRegion=$region
...
+  echo "modelgen is set to true, generating Swift models from schema.graphql..."
   amplify codegen model
+  # calls amplify-app again so the Xcode project is updated with the generated models
+  amplify-app --platform ios
```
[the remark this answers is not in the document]
drochetti: no because at this point it is expected for amplify-app to be available

AmplifyTools/amplify-tools.sh
drochetti: I'm not familiar either tbh

Pods/Manifest.lock
drochetti: me too, but why did my cocoapods make that change? If that's not supposed to happen then I'm very confused 😕

Podfile.lock
palpatim: Yep!

Pods/Manifest.lock
palpatim: https://github.com/CocoaPods/CocoaPods/issues/9190

PR #458, opened 2020-05-18
feat: AuthRule alignment with codegen

lawmicha:
*Issue #, if available:*

*Description of changes:*
- this removes `provider` as that is not needed in the schema, only the API needs to configured with the correct provider at provisioning time.
- moves CodingKey to a String, for the case where developer does not provide an owner field in the model, we need to pass the default values over to the schema (less logic stored in the library), allows CLI to be source of truth for default values.

### DIFF
--- a/Amplify/Categories/DataStore/Model/Schema/AuthRule.swift
+++ b/Amplify/Categories/DataStore/Model/Schema/AuthRule.swift
@@ -23,7 +23,6 @@ public typealias AuthRules = [AuthRule]
 
 public struct AuthRule {
     public let allow: AuthStrategy
-    public let field: CodingKey?
     public let ownerField: CodingKey?
     public let identityClaim: String?
     public let groupClaim: String?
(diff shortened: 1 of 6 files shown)


Amplify/Categories/DataStore/Model/Schema/AuthRule.swift
```diff
@@ -31,31 +24,25 @@ public typealias AuthRules = [AuthRule]
 public struct AuthRule {
     public let allow: AuthStrategy
     public let field: CodingKey?
-    public let provider: AuthProvider?
     public let ownerField: CodingKey?
     public let identityClaim: String?
     public let groupClaim: String?
     public let groups: [String]
-    public let groupsField: CodingKey?
```
drochetti, line 39: if the group field is customized by the developer isn't this needed? (genuine question, I'm not sure how it works)

lawmicha: this is a valid question and i removed it because it was relating to "dynamic" groups, and thought we didn't really support this yet. but really what dynamic groups is is to specific the field on the model which is to contain the groups, just like how developer can override the ownerField to something other than "owner". 

Adding this back in to better have this info and follow up with testing expected functionality

PR #459, opened 2020-05-19
refactor(Auth): Moved signedOut and sessionExpiry error to core Auth

royjit:
*Description of changes:*

- Moved signedOut and sessionExpiry error to core Auth. This makes it easier for checking the error in fetchAuthSession callback.
- Integration test for signedOut case


Amplify/Categories/Auth/Error/AuthError.swift
```diff
@@ -14,6 +14,8 @@ public enum AuthError {
     case validation(Field, ErrorDescription, RecoverySuggestion, Error? = nil)
     case notAuthorized(ErrorDescription, RecoverySuggestion, Error? = nil)
     case invalidState(ErrorDescription, RecoverySuggestion, Error? = nil)
+    case signedOut(ErrorDescription, RecoverySuggestion, Error? = nil)
+    case sessionExpired(ErrorDescription, RecoverySuggestion, Error? = nil)
```
palpatim, line 18: 1. Add doc comments for all cases
1. Are these error cases common across platforms?

AmplifyTestCommon.podspec
```diff
@@ -23,7 +23,8 @@ Pod::Spec.new do |spec|
 
   spec.source_files = 'AmplifyTestCommon/**/*.swift'
   spec.dependency 'Amplify', '0.11.0'
-
+  spec.dependency 'AWSCore'
+
```
palpatim, line 27: It's reasonable that we'd need this in AWSPluginsTestCommon, but we shouldn't put an AWSCore dep at the top level AmplifyTestCommon.

royjit: Moved to AWSPluginsTestCommon

Issue #456, opened 2020-05-18
The App does not run as expected with multiple sign in and sign out.


DongQuanRui:
**Describe the bug**
One user is signed in in two iOS simulators. and when signing out from one of the two simulators, the other one cannot perform sign in or sign out as expected (The App is somehow crashed, need to go to "Device" -> "Erase all content and settings", and reinstall this app then it become normal).

**To Reproduce**
1. Start two simulators (call 1st simulator and 2nd simulator respectively) with the same iOS project (I think with signIn(), signOut() and signUp() are enough to reproduce this issue). 
2. Should have a signed up user (this user I would call "test" here) in Cognito user pool.
3. Log in "test" user in 1st simulator, and then log in this "test" user in the 2nd simulator. Now this "test" user is signed in in both simulators.
4. Since both simulators are signed in with the same user, now sign out from 2nd simulator (it doesn't matter sign out from which simulator).
5. It shows error when trying to sign out from 1st simulator. The error can be seen in the screenshot I attach below.

**Expected behavior**
I think no matter how many devices one user signs in with, this user should be able to sign out from all of them one by one using "Amplify.Auth.signOut()".

**Screenshots**
<img width="1202" alt="Screen Shot 2020-05-18 at 09 52 11" src="https://user-images.githubusercontent.com/48600426/82256438-302a7400-990b-11ea-984e-df0c98f6121e.png">

**Environment(please complete the following information):**
 - Xcode Version: 11.4.1 
 - Amplify Framework Version: The latest one in aws-amplify/amplify-ios, clone it locally and use local pods.
 - Dependency Manager: Cocoapods
 - Swift Version : [e.g. 5.0]

**Device Information (please complete the following information):**
 - Device: iPhone SE (2nd generation) simulator, iPhone 11 simulator
 - iOS Version: iOS 13.4.1

royjit:
Fix is merged in to master. Please verify

DongQuanRui:
The sign out works very well now, closing this issue

PR #461, opened 2020-05-19
fix(Auth): Locally signOut the user if globaly signout fail with notAuthorized

royjit:
Description:

- Made local signOut as the default behavior
- Added an option to signOutRequest for globalSignOut
- If global signOut fails with notAuthorized error, we do a localSignOut, since it is not a recoverable error.

```swift

      // Global signout
        let options = AuthSignOutRequest.Options(globalSignOut: true)
        Amplify.Auth.signOut(options: options) { (event) in
            switch event {
            case .failure(let error):
                print(error)
            case .success:
                print("success")
            }
        }
        
       // Local signOut
      Amplify.Auth.signOut { (event) in
            switch event {
            case .failure(let error):
                print(error)
            case .success:
                print("success")
            }
        }
```

*Issue #456 

### DIFF
--- a/AmplifyPlugins/Auth/AWSAuthPlugin/Dependency/AuthenticationProviderAdapter+SignOut.swift
+++ b/AmplifyPlugins/Auth/AWSAuthPlugin/Dependency/AuthenticationProviderAdapter+SignOut.swift
@@ -27,11 +27,19 @@ extension AuthenticationProviderAdapter {
         // There is no point on returning an error back to the developer, because
         // they do not control how the UI is presented.
         dispatchPrecondition(condition: .onQueue(DispatchQueue.main))
+
         let signOutOptions = SignOutOptions(signOutGlobally: true, invalidateTokens: true)
-        awsMobileClient.signOut(options: signOutOptions) { error in
+        awsMobileClient.signOut(options: signOutOptions) { [weak self] error in
             guard error == nil else {
                 let authError = AuthErrorHelper.toAuthError(error!)
-                completionHandler(.failure(authError))
+                if case .notAuthorized(_, _, _) = authError {
+                    // signOut globally might return notAuthorized when the current token is expired or invalidated
+                    // In this case, we just signOut the user locally and return a success result back.
+                    self?.awsMobileClient.signOut()
+                    completionHandler(.success(()))
+                } else {
+                    completionHandler(.failure(authError))
+                }
                 return
             }
             completionHandler(.success(()))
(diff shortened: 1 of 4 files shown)


AmplifyPlugins/Auth/AWSAuthPlugin/Service/AWSMobileClient/AWSMobileClientAdapter.swift
```diff
@@ -103,6 +103,10 @@ class AWSMobileClientAdapter: AWSMobileClientBehavior {
         awsMobileClient.signOut(options: options, completionHandler: completionHandler)
     }
 
+    func signOut() {
```
palpatim, line 106: as discussed, change this to `signOutLocally`

AmplifyPlugins/Auth/AWSAuthPlugin/Dependency/AuthenticationProviderAdapter+SignOut.swift
```diff
@@ -27,11 +27,19 @@ extension AuthenticationProviderAdapter {
...
+
         let signOutOptions = SignOutOptions(signOutGlobally: true, invalidateTokens: true)
-        awsMobileClient.signOut(options: signOutOptions) { error in
+        awsMobileClient.signOut(options: signOutOptions) { [weak self] error in
             guard error == nil else {
                 let authError = AuthErrorHelper.toAuthError(error!)
-                completionHandler(.failure(authError))
+                if case .notAuthorized(_, _, _) = authError {
+                    // signOut globally might return notAuthorized when the current token is expired or invalidated
+                    // In this case, we just signOut the user locally and return a success result back.
```
palpatim, line 37: This is misleading if the contract is that `signOut` does a global signout. Let's document this, ensure the behavior is consistent across platforms, and emit a Hub event if we have to fall back to "local sign out only"

AmplifyPlugins/Auth/AWSAuthPlugin/Service/AWSMobileClient/AWSMobileClientAdapter.swift
royjit: Changed

AmplifyPlugins/Auth/AWSAuthPlugin/Dependency/AuthenticationProviderAdapter+SignOut.swift
royjit: Made the default behavior of signOut to local signOut and added an option to globalSignOut.
We are handling `notAuthorized` since its not a recoverable error.

PR #470, opened 2020-05-20
refactor(API): move the model-based functions to helpers

drochetti:
**Notes:**

Following the decision of not having the model-based calls as part of the API Category contract, I moved them out and created semantic factories to build the `GraphQLRequest` from the models, so the functionality is preserved.

**Examples:**

Mutation of type create:

```swift
// previous
Amplify.API.mutate(of: post, type: .create) { response in
}

// new (mimics the underlying GraphQL statement: createPost)
Amplify.API.mutate(request: .create(post)) { response in
}
```

Queries:

```swift
// previous
Amplify.API.query(from: Post.self, where: nil, listener: { response in
}

// new  (mimics the underlying GraphQL statement: listPosts)
Amplify.API.query(request: .list(Post.self)) { response in
}

// ---

// previous
Amplify.API.query(from: Post.self, byId: "id", listener: { response in
}

// new  (mimics the underlying GraphQL statement: getPost)
Amplify.API.query(request: .get(Post.self, byId: "id")) { response in
}
```

See the diff for more examples.


AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
```diff
@@ -8,18 +8,123 @@
...
+    /// Creates a `GraphQLRequest` that represents an update mutation
+    /// for a given `model` instance.
+    ///
+    /// - Parameters:
+    ///   - model: the model instance populated with values
+    ///   - predicate: a predicate passed as the condition to apply the mutation
+    /// - Returns: a valid `GraphQLRequest` instance
+    /// - seealso: `GraphQLRequest.mutation(of:where:type:)`
+    static func update<M: Model>(_ model: M,
+                                 where predicate: QueryPredicate?) -> GraphQLRequest<M>
```
lawmicha, line 77: ah nicee, thanks for the mutation with condition, this was missing before

drochetti:
> so does that mean with this change we are keep the APIs for DataStore the same but we can follow up by refactoring those over to something simiilar? (syncQuery, mutations, query) not a priority since those are not public facing

I believe they already follow the same pattern, those queries and mutations are added as an extension to the `GraphQLRequest` on https://github.com/aws-amplify/amplify-ios/blob/master/AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+AnyModelWithSync.swift

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
```diff
@@ -8,18 +8,123 @@
...
     /// - Parameters:
     ///   - model: the model instance populated with values
     ///   - predicate: a predicate passed as the condition to apply the mutation
     ///   - type: the mutation type, either `.create`, `.update`, or `.delete`
-    /// - Returns: the `GraphQLRequest` ready to be used
+    /// - Returns: a valid `GraphQLRequest` instance
+    static func mutation<M: Model>(of model: M,
+                                   where predicate: QueryPredicate?,
+                                   type: GraphQLMutationType) -> GraphQLRequest<M>
+
```
palpatim, line 58: Do we need this if we're exposing create/update/delete individual factory methods?

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
```diff
@@ -101,17 +186,8 @@ extension GraphQLRequest {
...
-    ///
-    /// - Parameters:
-    ///   - modelType: the metatype of the model
-    ///   - type: the subscription type, either `.onCreate`, `.onUpdate` or `.onDelete`
-    /// - Returns: the `GraphQLRequest` ready to be used
-    ///
-    /// - seealso: `GraphQLSubscription`, `GraphQLSubscriptionType`
     public static func subscription<M: Model>(of modelType: M.Type,
-                                              type: GraphQLSubscriptionType) -> GraphQLRequest<M> {
+                                              _ type: GraphQLSubscriptionType) -> GraphQLRequest<M> {
```
palpatim, line 190: It's a bit unusual to see an API method with a named first parameter and unnamed second one (that's not a trailing closure). That means the call site would look something like `.subscription(of: Post.self, .create)`? I appreciate the brevity, but it feels a bit wonky.

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
drochetti: This is useful when creating a mutation based on some logic (i.e. should it be a `create` or `update`?). We use this internally in the sync engine, so I kept it to reduce the impact of the changes.

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
drochetti: fair point, I can definitely change it back.

The call looks like this:

```swift
.subscription(of: Post.self, .onCreate)
```

The `on` prefix in the enum motivated me to make the change, but I think I prefer to have it named as well.

drochetti: changed back to the named `type`

PR #471, opened 2020-05-20
feat(datastore): StartSync with Auth 

lawmicha:
*Issue #, if available:*

*Description of changes:*
This adds support for auth enabled models.
- dependency on Amplify.Auth's cognito auth plugin, ie. "awsCognitoAuthPlugin"
- moved StorageEngine's startSync to `tryStartSync` that does much more: checks if auth is required, check if signed in, if not then wait for sign in. if signed in, start waiting for signed out
- on sign-in, it will `remoteSyncEngine.start()`
- on sign-out, it will `clear()`. Added new `clearCompleted` event for tests to listen on
- pass the auth plugin down to the actual creation of the subscriptions, if auth exists, it will retrieve necessary identityClaim (initial support for this defaults to username at the moment, could be sub, or something else) to pass as ownerId to the subscription APIs

*Misc updates*
- change AuthError.service check to AuthError.signedOut check, to match @royjit 's expected response for unauthenticated users when processing errors in `ProcessMutationErrorFromCloudOperation`

*Acceptance tests*
1. `testUnauthenticatedSavesToLocalStoreIsReconciledWithCloudStoreAfterAuthentication`
2. `testOwnerCreatedDataCanBeReadByOtherUsersForReadableModel`

### DIFF
--- a/Amplify/Categories/Auth/AuthCategoryDeviceBehavior.swift
+++ b/Amplify/Categories/Auth/AuthCategoryDeviceBehavior.swift
@@ -7,7 +7,7 @@
 
 import Foundation
 
-public protocol AuthCategoryDeviceBehavior {
+public protocol AuthCategoryDeviceBehavior: class {
 
     /// Fetch devices assigned to the current device
     /// - Parameters:
--- a/Amplify/Categories/DataStore/Model/Schema/ModelSchema+Attributes.swift
+++ b/Amplify/Categories/DataStore/Model/Schema/ModelSchema+Attributes.swift
@@ -16,4 +16,7 @@ public extension ModelSchema {
         attributes.contains(.isSystem)
     }
 
+    var isAuthEnabled: Bool {
+        return !authRules.isEmpty
+    }
 }
--- a/Amplify/Categories/DataStore/Model/Schema/ModelSchema+Attributes.swift
+++ b/Amplify/Categories/DataStore/Model/Schema/ModelSchema+Attributes.swift
@@ -16,7 +16,7 @@ public extension ModelSchema {
         attributes.contains(.isSystem)
     }
 
-    var hasAuth: Bool {
+    var hasAuthenticationRules: Bool {
         return !authRules.isEmpty
     }
 }
--- a/Amplify/Categories/DataStore/Model/Schema/ModelSchema+Attributes.swift
+++ b/Amplify/Categories/DataStore/Model/Schema/ModelSchema+Attributes.swift
@@ -16,7 +16,7 @@ public extension ModelSchema {
         attributes.contains(.isSystem)
     }
 
-    var isAuthEnabled: Bool {
+    var hasAuth: Bool {
         return !authRules.isEmpty
     }
 }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -88,56 +88,10 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         let filter = HubFilters.forEventName(HubPayload.EventName.Amplify.configured)
         var token: UnsubscribeToken?
         token = Amplify.Hub.listen(to: .dataStore, isIncluded: filter) { _ in
-            if self.hasRequiredPlugins() {
-                self.storageEngine.startSync()
-            } else {
-                self.log.info("Unable to find suitable plugin for syncEngine.  syncEngine will not be started")
-            }
-
+            self.storageEngine.tryStartSync()
             if let token = token {
                 Amplify.Hub.removeListener(token)
             }
-        }
-    }
-
-    func hasRequiredPlugins() -> Bool {
-        guard hasValidAPIPlugin() else {
-            return false
-        }
-        if !requireAuthPlugin() {
-            return true
-        }
-
-        if requireAuthPlugin() && hasValidAuthPlugin() {
-            return true
-        }
-
-        return false
-    }
-
-    func hasValidAPIPlugin() -> Bool {
-        do {
-            _ = try Amplify.API.getPlugin(for: validAPIPluginKey)
-            return true
-        } catch {
-            return false
-        }
-    }
-
-    func requireAuthPlugin() -> Bool {
-        let containsAuthEnabledSyncableModels = ModelRegistry.models.contains {
-            $0.schema.isSyncable && $0.schema.isAuthEnabled
-        }
-
-        return containsAuthEnabledSyncableModels
-    }
-
-    func hasValidAuthPlugin() -> Bool {
-        do {
-            _ = try Amplify.Auth.getPlugin(for: validAuthPluginKey)
-            return true
-        } catch {
-            return false
         }
     }
 
@@ -151,7 +105,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
             }
             try resolveStorageEngine(dataStoreConfiguration: dataStoreConfiguration)
             try storageEngine.setUp(models: ModelRegistry.models)
-            storageEngine.startSync()
+            storageEngine.tryStartSync()
         } catch {
             log.error(error: error)
         }
@@ -162,7 +116,10 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
             return
         }
 
-        storageEngine = try StorageEngine(isSyncEnabled: isSyncEnabled, dataStoreConfiguration: dataStoreConfiguration)
+        storageEngine = try StorageEngine(isSyncEnabled: isSyncEnabled,
+                                          dataStoreConfiguration: dataStoreConfiguration,
+                                          validAPIPluginKey: validAPIPluginKey,
+                                          validAuthPluginKey: validAuthPluginKey)
         if #available(iOS 13.0, *) {
             setupStorageSink()
         }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -24,6 +24,9 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
     let dataStoreConfiguration: DataStoreConfiguration
 
     let validAPIPluginKey: String
+
+    let validAuthPluginKey: String
+
     /// The local storage provider. Resolved during configuration phase
     var storageEngine: StorageEngineBehavior!
 
@@ -48,6 +51,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         self.dataStoreConfiguration = dataStoreConfiguration
         self.isSyncEnabled = false
         self.validAPIPluginKey =  "awsAPIPlugin"
+        self.validAuthPluginKey = "awsCognitoAuthPlugin"
         if #available(iOS 13.0, *) {
             self.dataStorePublisher = DataStorePublisher()
         } else {
@@ -60,13 +64,15 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
          configuration dataStoreConfiguration: DataStoreConfiguration = .default,
          storageEngine: StorageEngineBehavior,
          dataStorePublisher: DataStoreSubscribeBehavior,
-         validAPIPluginKey: String) {
+         validAPIPluginKey: String,
+         validAuthPluginKey: String) {
         self.modelRegistration = modelRegistration
         self.dataStoreConfiguration = dataStoreConfiguration
         self.isSyncEnabled = false
         self.storageEngine = storageEngine
         self.dataStorePublisher = dataStorePublisher
         self.validAPIPluginKey = validAPIPluginKey
+        self.validAuthPluginKey = validAuthPluginKey
     }
 
     /// By the time this method gets called, DataStore will already have invoked
@@ -82,7 +88,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         let filter = HubFilters.forEventName(HubPayload.EventName.Amplify.configured)
         var token: UnsubscribeToken?
         token = Amplify.Hub.listen(to: .dataStore, isIncluded: filter) { _ in
-            if self.hasValidAPIPlugin() {
+            if self.hasRequiredPlugins() {
                 self.storageEngine.startSync()
             } else {
                 self.log.info("Unable to find suitable plugin for syncEngine.  syncEngine will not be started")
@@ -94,9 +100,41 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         }
     }
 
+    func hasRequiredPlugins() -> Bool {
+        guard hasValidAPIPlugin() else {
+            return false
+        }
+        if !requireAuthPlugin() {
+            return true
+        }
+
+        if requireAuthPlugin() && hasValidAuthPlugin() {
+            return true
+        }
+
+        return false
+    }
+
     func hasValidAPIPlugin() -> Bool {
         do {
             _ = try Amplify.API.getPlugin(for: validAPIPluginKey)
+            return true
+        } catch {
+            return false
+        }
+    }
+
+    func requireAuthPlugin() -> Bool {
+        let containsAuthEnabledSyncableModels = ModelRegistry.models.contains {
+            $0.schema.isSyncable && $0.schema.isAuthEnabled
+        }
+
+        return containsAuthEnabledSyncableModels
+    }
+
+    func hasValidAuthPlugin() -> Bool {
+        do {
+            _ = try Amplify.Auth.getPlugin(for: validAuthPluginKey)
             return true
         } catch {
             return false
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -88,7 +88,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         let filter = HubFilters.forEventName(HubPayload.EventName.Amplify.configured)
         var token: UnsubscribeToken?
         token = Amplify.Hub.listen(to: .dataStore, isIncluded: filter) { _ in
-            self.storageEngine.tryStartSync()
+            self.storageEngine.startSync()
             if let token = token {
                 Amplify.Hub.removeListener(token)
             }
@@ -105,7 +105,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
             }
             try resolveStorageEngine(dataStoreConfiguration: dataStoreConfiguration)
             try storageEngine.setUp(models: ModelRegistry.models)
-            storageEngine.tryStartSync()
+            storageEngine.startSync()
         } catch {
             log.error(error: error)
         }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
@@ -23,23 +23,17 @@ extension RemoteSyncEngine {
             case (.pausingMutationQueue, .pausedMutationQueue(let storageEngineAdapter)):
                 return .clearingStateOutgoingMutations(storageEngineAdapter)
 
-            case (.clearingStateOutgoingMutations, .clearedStateOutgoingMutations):
-                return .checkIfAuthenticationRequired
-            case (.checkIfAuthenticationRequired, .requireAuthenticatedUser):
-                return .waitForAuthenticatedUser
-            case (.waitForAuthenticatedUser, .authenticatedUser(let api,
-                                                                let storageEngineAdapter,
-                                                                let userId)):
-                return .initializingSubscriptions(api, storageEngineAdapter, userId)
-            case (.checkIfAuthenticationRequired, .authenticatedUser(let api,
-                                                                     let storageEngineAdapter,
-                                                                     let userId)):
-                return .initializingSubscriptions(api, storageEngineAdapter, userId)
-            case (.checkIfAuthenticationRequired, .authenticationNotRequired(let api, let storageEngineAdapter)):
+            case (.clearingStateOutgoingMutations, .clearedStateOutgoingMutations(let api, let storageEngineAdapter)):
                 return .initializingSubscriptions(api, storageEngineAdapter)
+
             case (.initializingSubscriptions, .initializedSubscriptions):
                 return .performingInitialSync
             case (.initializingSubscriptions, .errored(let error)):
+                return .cleaningUp(error)
+
+            case (.performingInitialSync, .performedInitialSync):
+                return .activatingCloudSubscriptions
+            case (.performingInitialSync, .errored(let error)):
                 return .cleaningUp(error)
 
             case (.performingInitialSync, .performedInitialSync):
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
@@ -36,11 +36,6 @@ extension RemoteSyncEngine {
             case (.performingInitialSync, .errored(let error)):
                 return .cleaningUp(error)
 
-            case (.performingInitialSync, .performedInitialSync):
-                return .activatingCloudSubscriptions
-            case (.performingInitialSync, .errored(let error)):
-                return .cleaningUp(error)
-
             case (.activatingCloudSubscriptions, .activatedCloudSubscriptions(let api,
                                                                               let mutationEventPublisher)):
                 return .activatingMutationQueue(api, mutationEventPublisher)
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
@@ -23,9 +23,20 @@ extension RemoteSyncEngine {
             case (.pausingMutationQueue, .pausedMutationQueue(let storageEngineAdapter)):
                 return .clearingStateOutgoingMutations(storageEngineAdapter)
 
-            case (.clearingStateOutgoingMutations, .clearedStateOutgoingMutations(let api, let storageEngineAdapter)):
+            case (.clearingStateOutgoingMutations, .clearedStateOutgoingMutations):
+                return .checkIfAuthenticationRequired
+            case (.checkIfAuthenticationRequired, .requireAuthenticatedUser):
+                return .waitForAuthenticatedUser
+            case (.waitForAuthenticatedUser, .authenticatedUser(let api,
+                                                                let storageEngineAdapter,
+                                                                let userId)):
+                return .initializingSubscriptions(api, storageEngineAdapter, userId)
+            case (.checkIfAuthenticationRequired, .authenticatedUser(let api,
+                                                                     let storageEngineAdapter,
+                                                                     let userId)):
+                return .initializingSubscriptions(api, storageEngineAdapter, userId)
+            case (.checkIfAuthenticationRequired, .authenticationNotRequired(let api, let storageEngineAdapter)):
                 return .initializingSubscriptions(api, storageEngineAdapter)
-
             case (.initializingSubscriptions, .initializedSubscriptions):
                 return .performingInitialSync
             case (.initializingSubscriptions, .errored(let error)):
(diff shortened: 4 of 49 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -82,7 +88,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         let filter = HubFilters.forEventName(HubPayload.EventName.Amplify.configured)
         var token: UnsubscribeToken?
         token = Amplify.Hub.listen(to: .dataStore, isIncluded: filter) { _ in
-            if self.hasValidAPIPlugin() {
+            if self.hasRequiredPlugins() {
                 self.storageEngine.startSync()
             } else {
                 self.log.info("Unable to find suitable plugin for syncEngine.  syncEngine will not be started")
```
drochetti, line 94: I think we could be more informative here. Since the sync engine could not start for different reasons it would be cool if we provided a more detailed message of the actual reason it didn't start (and maybe use `log.warn`?)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -228,18 +236,100 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
-            if let api = self.api {
-                self.remoteSyncTopicPublisher.send(.clearedStateOutgoingMutations)
-                self.stateMachine.notify(action: .clearedStateOutgoingMutations(api, storageAdapter))
+            self.remoteSyncTopicPublisher.send(.clearedStateOutgoingMutations)
+            self.stateMachine.notify(action: .clearedStateOutgoingMutations)
+        }
+    }
+
+    private func checkIfAuthenticationRequired() {
+        log.debug(#function)
```
drochetti, line 245: what will this print? We could have a more informative message here, so it's easier to search in the logs

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -228,18 +236,100 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
+
+    private func isSignedIn(auth: AuthCategoryBehavior, onComplete: @escaping (Result<Bool, AuthError>) -> Void) {
+        _ = auth.fetchAuthSession(options: nil) { event in
+            switch event {
+            case .success(let authSession):
+                if authSession.isSignedIn {
+                    onComplete(.success(true))
+                } else {
+                    onComplete(.success(false))
+                }
```
drochetti, line 320: could be simplified to `onComplete(.success(authSession.isSignedIn))`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -228,18 +236,100 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
+    }
+
+    private func removeSignInListener() {
+        if let listener = signInListener {
+            Amplify.Hub.removeListener(listener)
+        }
+    }
+
+    private func isSignedIn(auth: AuthCategoryBehavior, onComplete: @escaping (Result<Bool, AuthError>) -> Void) {
+        _ = auth.fetchAuthSession(options: nil) { event in
```
drochetti, line 313: we should have a way of omitting options if it can be `nil` @royjit

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/MutationQueue/ProcessMutationErrorFromCloudOperationTests.swift
```diff
@@ -56,14 +56,14 @@ class ProcessMutationErrorFromCloudOperationTests: XCTestCase {
         let configuration = DataStoreConfiguration.custom(errorHandler: { error in
             guard let dataStoreError = error as? DataStoreError,
                 case let .api(amplifyError, mutationEventOptional) = dataStoreError else {
-                    XCTFail("Expected API error with mutationEvent")
+                    XCTFail("Expected API error wpith mutationEvent")
```
drochetti, line 59: typo *wpith*

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -228,18 +236,100 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
+            }
+        }
+    }
+
+    private func waitForAuthenticatedUser() {
+        log.debug(#function)
+        signInListener = Amplify.Hub.listen(to: .auth) { payload in
+            guard payload.eventName == HubPayload.EventName.Auth.signIn else {
+                return
+            }
```
drochetti, line 283: you could create a `HubFilter` instead and pass to `.listen()`

Amplify/Categories/DataStore/Model/Schema/ModelSchema+Attributes.swift
```diff
@@ -16,4 +16,7 @@ public extension ModelSchema {
         attributes.contains(.isSystem)
     }
 
+    var isAuthEnabled: Bool {
+        return !authRules.isEmpty
+    }
```
wooj2, line 21: I'm curious as to what `isAuthEnabled` refers to.  Is it simply to say that the models use an `@auth` directive?  If so, then using "Enabled" might be misleading.  Perhaps a better name would be `isAuth`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
```diff
@@ -23,9 +23,20 @@ extension RemoteSyncEngine {
...
+                return .waitForAuthenticatedUser
+            case (.waitForAuthenticatedUser, .authenticatedUser(let api,
+                                                                let storageEngineAdapter,
+                                                                let userId)):
+                return .initializingSubscriptions(api, storageEngineAdapter, userId)
+            case (.checkIfAuthenticationRequired, .authenticatedUser(let api,
+                                                                     let storageEngineAdapter,
+                                                                     let userId)):
+                return .initializingSubscriptions(api, storageEngineAdapter, userId)
+            case (.checkIfAuthenticationRequired, .authenticationNotRequired(let api, let storageEngineAdapter)):
```
wooj2, line 38: Seems like we've adopted a strategy where we handle listening to auth state within the sync engine.  Did we consider another approach like listening to auth state in StorageEngine and determine whether or not we should start/stop the sync engine from there?

lawmicha: Been chatting with @elorzafe regarding the auth strategy and we agree that DataStore should listen to auth state changes in general 

JS current requires the developers to listen to auth state changes themselves:
@elorzafe, please correct me if i'm wrong here:
If sign-in event: no public facing API to start the sync engine, but can call `DataStore.clear()` which re-iniitializes the syncEngine as a secondary behavior
if sign out event: call DataStore.clear() to clear local data

JS customers have to make a call to DataStore APIs like DataStore.query(). this wll try to
- start to start the sync engine (fail if not authenticated, fall back to IAM credentials for sigv4 signing, different architecture from iOS intercategory plugin communication - IOS cannot fall back to IAM, but utimately there is an eventual fail case). 
- assuming this is async, the data returned to the customer may not be up to date since sync engine's reconcilation path may not have caught up.

The most efficient time to start the sync engine is by listening to auth sign in event. if all customers which use auth enabled models require the user to be signed in before making API calls (subscriptions and sync query) then DataStore should hold this listening logic. it should be waiting for Auth events like sign-in and sign-out.

The different places this auth state check can live:
1. At the very top level, DataStorePlugin will wait for DataStore configured event, check for API plugin before starting the sync engine. the check if there are required API optionally Auth plugin can be expanded to auth state change listener here
2. As you mentioned, in the StorageEngine, this looks like a possible candidate and be a good place when we address the auth state listener for sign-out because the `clear()` API is there. (we may then refactor the auth state lisnter code for sign-in event to the same place)
3. where it is currently in syncEngine: this allows the sync engine to do some of the non-auth related house keeping work that can be done without the API calls (subscription and sync query). leaving the wait for auth state change in the most optimal spot. and SyncEngine already exists a state machine like `schedulingRestart` and `errored` which will allow us to easily add "expoential retry if user is not signed in, then give up and listen for auth state change"

Amplify/Categories/DataStore/Model/Schema/ModelSchema+Attributes.swift
drochetti: Interesting feedback, maybe `hasAuth`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
lawmicha: i consolidated this logic and the logic from RemoteSyncEngine over to StorageEngine and added more logging. I have some `warn` and some `error` as well now. the initial check (in the new code) checks for APIPlugin and that is an `info` because this is to account for developers using local store first without API. all other logging are `warn` afterwards where there is API but something else is unexpected

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
lawmicha: i believe `checkIfAuthenticationRequired()`. added more to all
```
log.debug("\(#function) <message>")
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
lawmicha: the problem here is that it is using the protocol `AuthCategoryBehavior`, not the Category instance `Amplify.Auth`.

Amplify/Categories/DataStore/Model/Schema/ModelSchema+Attributes.swift
lawmicha: i was thinking it means these "models are enabled with auth rules for finer grain control", which means that they require auth, or an authenticated signed request when syncing to the cloud. By having auth directive on the model, they can control how the service checks for access, whether that is checking the token or signed request. 

from the two `isAuth` and `hasAuth`, i think isAuth makes me think of "the model is authenticated" which doesn't make a lot of sense, maybe "hasAuth" is "hasAuthentication" or "hasAuthenticationRules" which is closer to the check `!authRules.isEmpty`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
lawmicha: nice, thanks!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -94,6 +100,21 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
...
+        }
+        if !requireAuthPlugin() {
+            return true
+        }
+
+        if requireAuthPlugin() && hasValidAuthPlugin() {
+            return true
+        }
+
+        return false
```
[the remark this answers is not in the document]
lawmicha: nice, thanks! i ended up adding a bunch of guard statements to expand the checks and add logging in between

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
lawmicha: - refactored code to put it in StorageEngine
- also added the signOut listener

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
drochetti: ah, I see

Amplify/Categories/DataStore/Model/Schema/ModelSchema+Attributes.swift
drochetti: `hasAuthenticationRules` lgtm, I in favor of verbosity to communicate clear intent

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Resolver.swift
lawmicha: removed auth listener code (removed sign in/sign out listener code) for less impactful behavior change

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
```diff
@@ -147,12 +150,21 @@ final class IncomingAsyncSubscriptionEventPublisher: Cancellable {
...
+        auth: AuthCategoryBehavior?,
         valueListener: @escaping GraphQLSubscriptionOperation<Payload>.InProcessListener,
         completionListener: @escaping GraphQLSubscriptionOperation<Payload>.ResultListener
     ) -> GraphQLSubscriptionOperation<Payload> {
 
-        let request = GraphQLRequest<Payload>.subscription(to: modelType,
-                                                           subscriptionType: subscriptionType)
+        let request: GraphQLRequest<Payload>
+        if let auth = auth, modelType.schema.hasAuthenticationRules, let user = auth.getCurrentUser() {
+            // TODO: check model schema for identityClaim to figure out which is the ownerId field coming from
```
wooj2, line 160: Seems like we should open an issue around this.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/README.md
```diff
@@ -0,0 +1,74 @@
+## DataStore with Auth Integration Tests
+
+The following steps demonstrate how to set up an GraphQL endpoint with AppSync. 
+The auth configured will be Cognito UserPools.
```
wooj2, line 4: We can probably combine these two sentence into one:

The following steps demonstrate how to setup a GraphQL endpoint with AppSync and Cognito User Pools.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/README.md
```diff
@@ -0,0 +1,74 @@
+## DataStore with Auth Integration Tests
+
+The following steps demonstrate how to set up an GraphQL endpoint with AppSync. 
+The auth configured will be Cognito UserPools. 
+This set up is used to run the tests in `AWSDataStoreCategoryPluginAuthIntegrationTests.swift`.
```
wooj2, line 5: s/set up/configuration

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/README.md
```diff
@@ -0,0 +1,74 @@
...
+? Please select from one of the below mentioned services: `GraphQL`
+? Select from the options below `Enable DataStore for entire API`
+
+4. `amplify push`
+```perl
+? Are you sure you want to continue? `Yes`
+? Do you want to generate code for your newly created GraphQL API `No`
+```
+
+5. Copy `amplifyconfiguration.json` over as `AWSDataStoreCategoryPluginAuthIntegrationTests-amplifyconfiguration.json`
```
wooj2, line 49: Are you saying to change the file name from amplifyconfiguration.json to AWSDataStoreCategoryPluginAuthIntegrationTests-amplifyconfiguration.json ?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+8. Click on `Users and groups`, Sign up the two new users with the email and a temporary password. 
+
+9. Click on App clients, and keep note of the app client web's `App client id`. This can be used the AWS AppSync console Queries.
+
+10. `amplify console api`
+Click on Queries tab, and click on Log in. This will prompt you to enter the app client id, username, and temporary password. After logging in successfully, it will ask you to enter a new password. Make sure those are the same as the one specified in the credentials json file from step 5. Do this for both users.
+
+You can now run the tests!
```
wooj2, line 74: Maybe we should be more specific about what tests we can run:
Now you can run the AWSDataStoreCategoryPluginAuthIntegrationTests.
or
only specific tests?

Amplify/Categories/Auth/AuthCategoryDeviceBehavior.swift
```diff
@@ -7,7 +7,7 @@
 
 import Foundation
 
-public protocol AuthCategoryDeviceBehavior {
+public protocol AuthCategoryDeviceBehavior: class {
```
wooj2, line 10: Hmm what was the need for adding this?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOperationTests.swift
```diff
@@ -259,6 +259,65 @@ class InitialSyncOperationTests: XCTestCase {
...
+            guard let dataStoreError = error as? DataStoreError,
+                case let .api(amplifyError, mutationEventOptional) = dataStoreError else {
+                    XCTFail("Expected API error with mutationEvent")
+                    return
+            }
+            guard let actualAPIError = amplifyError as? APIError,
+                case let .operationError(_, _, underlyingError) = actualAPIError,
+                let authError = underlyingError as? AuthError,
+                case .signedOut = authError else {
+                    XCTFail("Should be `signedOut` error")
```
wooj2, line 294: might be helpful to print out `actualAPIError` ?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests.swift
```diff
@@ -0,0 +1,230 @@
...
+            try Amplify.add(plugin: AWSCognitoAuthPlugin())
+            let amplifyConfig = try TestConfigHelper.retrieveAmplifyConfiguration(forResource: amplifyConfigurationFile)
+            try Amplify.configure(amplifyConfig)
+
+        } catch {
+            XCTFail("Error during setup: \(error)")
+        }
+        if isSignedIn() {
+            signOut()
+        }
```
wooj2, line 58: is there any reason for this... because we are calling tearDown() which has signOut and reset?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests.swift
```diff
@@ -0,0 +1,230 @@
...
+    ///    - User signs in, retrieves note, sync engine is started and reconciles local store with the ownerId
+    ///    - The note now it contains the ownerId
+    func testUnauthenticatedSavesToLocalStoreIsReconciledWithCloudStoreAfterAuthentication() throws {
+        let savedLocalNote = saveNote(content: "owner saved note")
+        let queriedNoteOptional = queryNote(byId: savedLocalNote.id)
+        guard let note = queriedNoteOptional else {
+            XCTFail("Faled to query local note")
+            return
+        }
+        XCTAssertNil(note.owner)
```
wooj2, line 83: this seems more or less like a sanity case check where we just want to make sure `note.owner`  hasn't been set?  if that is the case.. maybe we should have something like:
```
guard let note.owner == nil else {
   XCTFail("heres what happend..etc.."
    return
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests.swift
```diff
@@ -0,0 +1,230 @@
...
+    /// - When:
+    ///    - User is not signed in, then user can successfully save a note to local store
+    ///    - User remains signed out, then user can successfully retrieve the saved note, with empty owner field
+    ///    - User signs in, retrieves note, sync engine is started and reconciles local store with the ownerId
+    ///    - The note now it contains the ownerId
+    func testUnauthenticatedSavesToLocalStoreIsReconciledWithCloudStoreAfterAuthentication() throws {
+        let savedLocalNote = saveNote(content: "owner saved note")
+        let queriedNoteOptional = queryNote(byId: savedLocalNote.id)
+        guard let note = queriedNoteOptional else {
+            XCTFail("Faled to query local note")
```
wooj2, line 80: s/Faled/Failed/

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests.swift
```diff
@@ -0,0 +1,230 @@
...
+
+        signIn(username: user1.username, password: user1.password)
+
+        wait(for: [syncReceivedInvoked], timeout: TestCommonConstants.networkTimeout)
+        Amplify.Hub.removeListener(syncReceivedListener)
+        guard let remoteNote = remoteNoteOptional else {
+            XCTFail("Should have received a SyncReceived event with the remote note reconciled to local store")
+            return
+        }
+        XCTAssertNotNil(remoteNote.owner)
```
wooj2, line 111: instead of checking for not nil..seems like we should check to make sure the owner is a specific value?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests.swift
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /// A signed in user (the owner) creates some data in local store will be synced to cloud. After `DataStore.clear`,
+    /// the data can no longer be retrieved. Signing back in with another user will update the local store with all
+    /// the data that can be read by that the user in the sync process. Then other user can read the owner's data.
```
wooj2, line 116: User1, while signed in, creates some data, called "NewUser1Data" in a local store.  We wait for this data to be synced to the cloud and then call DataStore.clear() so that data will be destroyed in the local data store and the remote sync engine will be halted.  After this finishes, we sign out of user1 and sign in with user2, which will restart the sync engine and read all of data from the backend including "NewUser1Data".

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift
lawmicha: good idea, adding issue https://github.com/aws-amplify/amplify-ios/issues/485, impact, and linked in comment

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests+Auth.swift
```diff
@@ -0,0 +1,121 @@
...
+    func signOut() {
+        let signOutCompleted = expectation(description: "sign out completed")
+        _ = Amplify.Auth.signOut { event in
+            switch event {
+            case .success:
+                signOutCompleted.fulfill()
+            case .failure(let error):
+                print("Could not sign out user \(error)")
+                signOutCompleted.fulfill()
+            }
```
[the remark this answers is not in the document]
lawmicha: updating to XCTFail

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests+Auth.swift
```diff
@@ -0,0 +1,121 @@
...
+                case .failure(let error):
+                    XCTFail("Failed to get auth session \(error)")
+                }
+            case .failure(let error):
+                XCTFail("Failed to get auth session \(error)")
+            }
+        })
+        wait(for: [retrieveUserSubCompleted], timeout: TestCommonConstants.networkTimeout)
+        guard let result = resultOptional else {
+            fatalError("Could not get userSub for user")
```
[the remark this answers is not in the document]
lawmicha: i've updated to XCTFail because in these cases, the return type is just String or Bool which makes it easy to return "" and false

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/README.md
lawmicha: yes, currently they're added to the HostApp's bundle. If we have multiple integration test targets, we need to prefix the configuration so each target reads their own configuration file

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/README.md
lawmicha: `Now you can run the AWSDataStoreCategoryPluginAuthIntegrationTests.`

Amplify/Categories/Auth/AuthCategoryDeviceBehavior.swift
lawmicha: this was done for API awhile back (https://github.com/aws-amplify/amplify-ios/blob/master/Amplify/Categories/API/ClientBehavior/APICategoryGraphQLBehavior.swift#L9)

i was getting the error for referencing the AuthCategory protocol without the protocol inheriting from `class`. let me pull the error up again

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests.swift
lawmicha: this was to prevent previous test executions failures so tearDown isn't called, or tearDown() is called but failed for some reason. i think this is applicable for a single test run as well since one test method may fail while the next one continues

Amplify/Categories/Auth/AuthCategoryDeviceBehavior.swift
lawmicha:
```
'weak' must not be applied to non-class-bound 'AuthCategoryBehavior'; consider adding a protocol conformance that has a class bound
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests.swift
lawmicha: adding check against `Auth.getCurrentUser().username`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests.swift
lawmicha: thanks for the improved wording

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests.swift
wooj2: hmmm.. if teardown isnt' being called on a single test run, then should we just add signout to each one of the integration tests?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/README.md
wooj2: I see, we can word this to:
Copy `amplifyconfiguration.json` to a new file named `AWSDataStoreCategoryPluginAuthIntegrationTests-amplifyconfiguration.json`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginAuthIntegrationTests/AWSDataStoreCategoryPluginAuthIntegrationTests.swift
lawmicha: teardown should be called on a single test method run. i was referring to more of a scenario where test is stopped, say i manually terminate it by pressing Stop, then tearDown is never run

PR #476, opened 2020-05-21
chore: 1.0.0-rc.1 release prep

palpatim:
- ci: Enable local lib linting of monorepo
- chore: Bump release to 1.0.0-rc.1
- chore: Update minimum SDK Dependency to 2.13.4
- chore: Podfile cleanup
- chore: Remove unnecessary requires_arc build setting

*Description of changes:*

The newly-added `CircleciScripts/setup_private_specs.sh` script also makes it easier to do app development against unreleased code. We'll add something to the README later. Someday we'll do nightly builds and it won't really be necessary to jump through these hoops for app development with bleeding-edge versions.

The `setup_private_specs.sh` script also makes it possible to lint the pods before the release. It modifies the specs in the private repo so dependencies resolve correctly, but the base file itself is linted directly from the local git working directory. I have validated that all of `Amplify.podspec`, `AmplifyPlugins.podspec`, `AWSPredictionsPlugin.podspec`, and `CoreMLPredictionsPlugin.podspec` pass with `pod lib lint --sources=amplify-ios-local-specs,trunk <podspec>`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -189,8 +189,8 @@ workflows:
       - plugin_unit_test:
           name: unit_test_auth
           path: Auth
-          workspace: AWSAuthPlugin
-          scheme: AWSAuthPlugin
+          workspace: AWSCognitoAuthPlugin
+          scheme: AWSCognitoAuthPlugin
           requires:
             - install_gems
       - plugin_unit_test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,13 +173,6 @@ workflows:
           requires:
             - install_gems
       - plugin_unit_test:
-          name: unit_test_api
-          path: API
-          workspace: APICategoryPlugin
-          scheme: AWSAPICategoryPlugin
-          requires:
-            - install_gems
-      - plugin_unit_test:
           name: unit_test_analytics
           path: Analytics
           workspace: AnalyticsCategoryPlugin
@@ -187,17 +180,24 @@ workflows:
           requires:
             - install_gems
       - plugin_unit_test:
-          name: unit_test_storage
-          path: Storage
-          workspace: StoragePlugin
-          scheme: AWSS3StoragePlugin
+          name: unit_test_api
+          path: API
+          workspace: APICategoryPlugin
+          scheme: AWSAPICategoryPlugin
           requires:
             - install_gems
       - plugin_unit_test:
-          name: unit_test_predictions
-          path: Predictions
-          workspace: PredictionsCategoryPlugin
-          scheme: AWSPredictionsPlugin
+          name: unit_test_auth
+          path: Auth
+          workspace: AWSAuthPlugin
+          scheme: AWSAuthPlugin
+          requires:
+            - install_gems
+      - plugin_unit_test:
+          name: unit_test_datastore
+          path: DataStore
+          workspace: DataStoreCategoryPlugin
+          scheme: AWSDataStoreCategoryPlugin
           requires:
             - install_gems
       - plugin_unit_test:
@@ -208,10 +208,17 @@ workflows:
           requires:
             - install_gems
       - plugin_unit_test:
-          name: unit_test_datastore
-          path: DataStore
-          workspace: DataStoreCategoryPlugin
-          scheme: AWSDataStoreCategoryPlugin
+          name: unit_test_predictions
+          path: Predictions
+          workspace: PredictionsCategoryPlugin
+          scheme: AWSPredictionsPlugin
+          requires:
+            - install_gems
+      - plugin_unit_test:
+          name: unit_test_storage
+          path: Storage
+          workspace: StoragePlugin
+          scheme: AWSS3StoragePlugin
           requires:
             - install_gems
       - deploy:
@@ -221,8 +228,10 @@ workflows:
                 - release
           requires:
             - build_test_amplify
-            - unit_test_api
             - unit_test_analytics
-            - unit_test_storage
-            - unit_test_predictions
+            - unit_test_api
+            - unit_test_auth
+            - unit_test_core_ml
             - unit_test_datastore
+            - unit_test_predictions
+            - unit_test_storage
--- a/AmplifyPlugins.podspec
+++ b/AmplifyPlugins.podspec
@@ -9,27 +9,33 @@
 Pod::Spec.new do |s|
 
   s.name         = 'AmplifyPlugins'
-  s.version      = '0.11.0'
+  s.version      = '1.0.0-rc.1'
   s.summary      = 'Amazon Web Services Amplify for iOS.'
 
   s.description  = 'AWS Amplify for iOS provides a declarative library for application development using cloud services'
-  
-  s.homepage     = 'http://aws.amazon.com/mobile/sdk'
+
+  s.homepage     = 'https://aws.amazon.com/amplify/'
   s.license      = 'Apache License, Version 2.0'
   s.author       = { 'Amazon Web Services' => 'amazonwebservices' }
-  s.platform     = :ios, '11.0'
   s.source       = { :git => 'https://github.com/aws-amplify/amplify-ios.git', :tag => s.version}
-  
-  s.requires_arc = true 
 
-  AWS_SDK_VERSION = '~> 2.13.0'
-  AMPLIFY_VERSION = '0.11.0'
-  
+  s.platform = :ios, '11.0'
+  s.swift_version = '5.0'
+
+  AWS_SDK_VERSION = '~> 2.13.4'
+  AMPLIFY_VERSION = '1.0.0-rc.1'
+
   s.subspec 'AWSAPIPlugin' do |ss|
     ss.source_files = 'AmplifyPlugins/API/AWSAPICategoryPlugin/**/*.swift'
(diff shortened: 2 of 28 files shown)


AmplifyPlugins.podspec
```diff
@@ -9,29 +9,35 @@
...
   s.name         = 'AmplifyPlugins'
-  s.version      = '0.11.0'
+  s.version      = '1.0.0-rc.1'
   s.summary      = 'Amazon Web Services Amplify for iOS.'
 
   s.description  = 'AWS Amplify for iOS provides a declarative library for application development using cloud services'
-  
-  s.homepage     = 'http://aws.amazon.com/mobile/sdk'
+
+  s.homepage     = 'https://aws.amazon.com/amplify/'
```
wooj2, line 17: nice catch!!!

CircleciScripts/setup_private_specs.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
```
wooj2, line 2: Not exactly sure what this script is doing or when it is suppose to be run.  Would it be possible to get a high level description at the top of the file of it's purpose?

palpatim: Noted in description: The setup_private_specs.sh script also makes it possible to lint the pods before the release. It modifies the specs in the private repo so dependencies resolve correctly, but the base file itself is linted directly from the local git working directory. I have validated that all of Amplify.podspec, AmplifyPlugins.podspec, AWSPredictionsPlugin.podspec, and CoreMLPredictionsPlugin.podspec pass with pod lib lint --sources=amplify-ios-local-specs,trunk <podspec>.

This will be part of official documentation when I have time to put it all together, but that will be a post-RC task

.circleci/config.yml
```diff
@@ -180,38 +187,38 @@ workflows:
...
             - install_gems
       - plugin_unit_test:
-          name: unit_test_analytics
-          path: Analytics
-          workspace: AnalyticsCategoryPlugin
-          scheme: AWSPinpointAnalyticsPlugin
+          name: unit_test_auth
+          path: Auth
+          workspace: AWSAuthPlugin
+          scheme: AWSAuthPlugin
```
royjit, line 193: AWSCognitoAuthPlugin

palpatim: Thanks for catching--git didn't remove the old files when you changed the names, and I didn't do a Pod update to generate the new workspace files

CircleciScripts/setup_private_specs.sh
```diff
@@ -0,0 +1,109 @@
...
+# This script generates a private spec repo in your $HOME/.aws-amplify
+# directory, to enable `pod lib lint`. It also makes it easier to do
+# development against unreleased versions of Amplify, at least until we start
+# releasing nightly builds.
+
+set -e
+
+declare -r LOCAL_SPEC_GIT_ROOT="$HOME/.aws-amplify/amplify-ios/amplify-ios-podspecs.git"
```
wooj2, line 10: Since we are using `${HOME}` on line 10, we could probably use the same here.

CircleciScripts/setup_private_specs.sh
wooj2: thanks!

CircleciScripts/setup_private_specs.sh
```diff
@@ -0,0 +1,109 @@
...
+  # repairs or validation
+  if [[ -d "${LOCAL_SPEC_GIT_ROOT}" ]] ; then
+    return 0
+  fi
+
+  mkdir -p "${LOCAL_SPEC_GIT_ROOT}"
+  old_pwd="${PWD}"
+  cd "${LOCAL_SPEC_GIT_ROOT}"
+  git init --bare
+  cd "${old_pwd}"
```
wooj2, line 25: Not required, but i prefer using pushd/popd.  For example this could be:
```
mkdir -p ${LOCAL_SPEC_GIT_ROOT}
pushd ${LOCAL_SPEC_GIT_ROOT}
git init --bare
retval=$?
popd
return ${retval}
```

CircleciScripts/setup_private_specs.sh
```diff
@@ -0,0 +1,109 @@
...
+
+function init_spec_repo {
+  # If repo exists, assume it's correctly set up; don't attempt any
+  # repairs or validation
+  if [[ -n "$(pod repo list | grep ${LOCAL_SPEC_REPO_NAME})" ]] ; then
+    return 0
+  fi
+
+  init_spec_git_root
+
```
wooj2, line 36: if we return a value from init_spec_git_root, we can:
```
init_spec_git_root
retval=$?
if [[ $retval -ne 0 ]]; then
    return ${retval}
fi
```

CircleciScripts/setup_private_specs.sh
```diff
@@ -0,0 +1,109 @@
...
+  pod repo add --silent "${LOCAL_SPEC_REPO_NAME}" "${LOCAL_SPEC_GIT_ROOT}"
+
+  old_pwd="${PWD}"
+  cd "${COCOAPODS_REPO_DIR}"
+  git commit --allow-empty -m "Empty commit"
+  git push
+  cd "${old_pwd}"
```
wooj2, line 43: Again, prefer
```
pushd "${COCOAPODS_REPO_DIR}"
git commit --allow-empty -m "Empty Commit"
git push
popd
```

CircleciScripts/setup_private_specs.sh
```diff
@@ -0,0 +1,109 @@
...
+declare -r src_dir="$PWD"
+
+podspec_file_names=()
+
+while read -r podspec_file ; do
+  podspec_file_name=$( basename "$podspec_file" )
+  podspec_file_names+=("$podspec_file_name")
+  pod_name=$( basename "$podspec_file_name" .podspec )
+
+  dst_dir="${COCOAPODS_REPO_DIR}/${pod_name}/${new_version}"
```
wooj2, line 94: Seems like we check to make sure COCOAPODS_REPO_DIR is defined to something sane before running this script? (e.g. make sure it's not nil and also not '/')

CircleciScripts/setup_private_specs.sh
```diff
@@ -0,0 +1,109 @@
...
+  dst_file="${dst_dir}/${podspec_file_name}"
+
+  mkdir -p "${dst_dir}"
```
wooj2, line 97: should we do some sanity checking around what dst_dir is expected to be, otherwise bail?

CircleciScripts/setup_private_specs.sh
```diff
@@ -0,0 +1,109 @@
...
+
+  write_munged_podspec "$podspec_file" "$dst_file"
+done < <( find "${src_dir}" -maxdepth 1 -mindepth 1 -name "*.podspec" | sort --ignore-case )
```
wooj2, line 100: Not super familiar with this syntax of using done with < <....

So... feel free to take me to school but want to make sure two things
1.  If something in the while loop, do we continue to process the rest of the items?
2. If something fails in the while loop, it seems that we will execute update_sepc_repo -- is this the right thing to do?

CircleciScripts/setup_private_specs.sh
palpatim: Thanks, I'll update in the next rev, but after RC release, just to make sure I don't have to go through another round of testing

CircleciScripts/setup_private_specs.sh
palpatim: bash functions don't return values other than their exit status unfortunately :(

CircleciScripts/setup_private_specs.sh
palpatim: post RC

CircleciScripts/setup_private_specs.sh
palpatim: I'm inclined to leave this dead simple. Cocoapods by convention (possibly hardcods) it to $HOME/.cocoapods, and validation of the "actual" directory path is tricky when you start thinking about things like symlinks, ".." directory navigations, etc. Since this is a utility intended to be run as a build validator, I'd prefer not to be too clever with any of the checks until we encounter a situation where it's actually causing an issue.

CircleciScripts/setup_private_specs.sh
palpatim: We `set -e` in the top of the script--the script exits on the first error

CircleciScripts/setup_private_specs.sh
palpatim: As above, validity checking on the directory is going to be a problem. With inputs of "pod name" and "version", the only meaningful validation I'd feel comfortable with is a check that the version looks something like a version. But again, I'm not inclined to go much further with it since validation is a big old can of worms and this utility script has a very narrow focus. If/when we promote this to a customer facing script, we'll consider more robust validation.

CircleciScripts/setup_private_specs.sh
wooj2: Hmmm on line 18 (which is part of init_spec_get_root), we are doing a `return 0`... that being said.. i think you're right.. with regards to capturing the return value I think i usually set a global variable and check it.. in any case, we can handle this later.

PR #479, opened 2020-05-22
test: Auth integration test for signUp, signIn

royjit:
*Description of changes:*

- Added basic integration test for signIn, signUp and session apis
- Fixed operation that does not cancel when cancel is called
- Changed the integration test target from UITest to UnitTest so that keychain access works correctly.

### DIFF
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/AWSAuthBaseTest.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/AWSAuthBaseTest.swift
@@ -11,7 +11,7 @@ import AWSCognitoAuthPlugin
 
 class AWSAuthBaseTest: XCTestCase {
 
-    let networkTimeout = TimeInterval(10) // 180 seconds to wait before network timeouts
+    let networkTimeout = TimeInterval(10)
 
      func initializeAmplify() {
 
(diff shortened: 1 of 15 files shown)


AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/AWSAuthBaseTest.swift
```diff
@@ -11,7 +11,7 @@ import AWSCognitoAuthPlugin
 
 class AWSAuthBaseTest: XCTestCase {
 
-    let networkTimeout = TimeInterval(180) // 180 seconds to wait before network timeouts
+    let networkTimeout = TimeInterval(10) // 180 seconds to wait before network timeouts
```
palpatim, line 14: nit: remove comment, it adds nothing to comprehension and is now actively misleading

AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/README.md
```diff
@@ -0,0 +1,81 @@
...
+ Do you want to specify the user attributes this app can read and write? 
+    No
+ Do you want to enable any of the following capabilities?
+    NA
+ Do you want to use an OAuth flow? 
+    No
+? Do you want to configure Lambda Triggers for Cognito? 
+    Yes
+? Which triggers do you want to enable for Cognito 
+    Pre Sign-up
```
palpatim, line 60: Only 1 trigger? Don't we need all of them for custom auth tests, or have those not yet been added?

royjit: Custom auth is not handled in this target. Since we are using AWSMC, we cannot have two auth flows under the same target. 
The lambda trigger is to avoid `confirmSignUp` step which require receiving a code to an email.

AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/AWSAuthBaseTest.swift
royjit: Removed

PR #497, opened 2020-05-27
test(Auth): Integration test related to apis for session, signOut

royjit:

### DIFF
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/Support/AuthSessionHelper.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/Support/AuthSessionHelper.swift
@@ -1,0 +1,21 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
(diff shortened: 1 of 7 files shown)


AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/Support/AuthSessionHelper.swift
```diff
@@ -0,0 +1,21 @@
...
+@testable import Amplify
+import AWSCognitoAuthPlugin
+
+struct AuthSessionHelper {
+
+    static func invalidateSession(username: String) {
+        let bundleID = Bundle.main.bundleIdentifier
+        let keychain = AWSUICKeyChainStore(service: "\(bundleID!).\(AWSCognitoIdentityUserPool.self)")
+        let namespace = "\(AWSMobileClient.default().userPoolClient!.userPoolConfiguration.clientId).\(username)"
+        let key = "\(namespace).tokenExpiration"
```
palpatim, line 18: I don't like using production code to set up tests, but in this case, it seems like we should be using the exact same logic to make sure all session data is being removed. If we change the way session data is keyed and forget to update this method, we run the risk of hard-to-diagnose test failures.

PR #511, opened 2020-06-03
fix(tools): require min version or use @latest

wooj2:
In attempts of addressing:
https://github.com/aws-amplify/amplify-ios/issues/502

If customers have an older version of amplify-app installed on their machines, we will exec the command:
```
npx amplify-app@latest
```

Otherwise, we will execute:
```
npx amplify-app
```

Forcing latest did not seem like the right thing to do.  Also, when running npx, there isn't a good way of status-ing where the command (amplify-app) is being pulled from ( a local repo, or... online. etc..)

In addition to checking the version of `amplify-app`, I extended this to check the version of amplify-cli per the request of @drochetti.  We will fail a build if customers do not have a minimum version required for `amplify-cli`.  Upon failing a build, they will be greeted with this message in Xcode:
Screen Shot 2020-06-03 at 1 50 33 PM

Fully open to suggestions on this one!

### DIFF
--- a/AmplifyTools/amplify-tools.sh
+++ b/AmplifyTools/amplify-tools.sh
@@ -6,46 +6,75 @@
 # SPDX-License-Identifier: Apache-2.0
 
 set -e
+
+if ! which node >/dev/null; then
+  echo "warning: Node is not installed. Visit https://nodejs.org/en/download/ to install it"
+  exit 1
+fi
+
 export PATH=$PATH:$(npm bin -g)
 
-MIN_SUPPORTED_VERSION_MAJOR=2
-MIN_SUPPORTED_VERSION_MINOR=17
-MIN_SUPPORTED_VERSION_PATCH=1
+AMP_APP_VERSION_MINIMUM="2.17.1"
+AMP_APP_VERSION_CURRENT=`npx -q amplify-app --version`
+AMP_APP_VERSION_INVALID=0
 
-CURR_VERSION=`npx -q amplify-app --version`
-CURR_VERSION_MAJOR=`echo ${CURR_VERSION} | tr  '.' ' ' | awk '{print $1}'`
-CURR_VERSION_MINOR=`echo ${CURR_VERSION} | tr  '.' ' ' | awk '{print $2}'`
-CURR_VERSION_PATCH=`echo ${CURR_VERSION} | tr  '.' ' ' | awk '{print $3}'`
-LOCAL_AMPAPP_VERSION_INVALID=0
+#TODO: This is vending unicode, need to strip these characters our safely
+#      and fix whatever is outputting these characters
+#AMP_CLI_VERSION_MINIMUM="4.21.0"
+#AMP_CLI_VERSION_CURRENT=`npx -q amplify --version |grep "\."`
+#AMP_CLI_VERSION_INVALID=0
 
-checkVersionAmplifyApp() {
-    if [ ${CURR_VERSION_MAJOR} -lt ${MIN_SUPPORTED_VERSION_MAJOR} ]; then
-	LOCAL_AMPAPP_VERSION_INVALID=1
+VERSION_INVALID=
+checkMinVersionCompatibility() {
+    VERSION_INVALID=0
+    CURR_VERSION_MAJOR=`echo "${1}" | tr  '.' ' ' | awk '{print $1}'`
+    CURR_VERSION_MINOR=`echo "${1}" | tr  '.' ' ' | awk '{print $2}'`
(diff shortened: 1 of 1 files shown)


AmplifyTools/amplify-tools.sh
```diff
@@ -8,11 +8,45 @@
...
+	    return
+	else
+	    if [ ${CURR_VERSION_PATCH} -lt ${MIN_SUPPORTED_VERSION_PATCH} ]; then
+		LOCAL_AMPAPP_VERSION_INVALID=1
+		return
+	    fi
+	fi
+    fi
+}
+checkVersionAmplifyApp
```
wooj2, line 37: Need to do some sanity checking around inputs before calling this function... otherwise, we can run into errors.

AmplifyTools/amplify-tools.sh
```diff
@@ -6,13 +6,121 @@
...
+AMP_APP_VERSION_MINIMUM="2.17.1"
+AMP_APP_VERSION_INVALID=0
+
+AMP_CLI_VERSION_CURRENT=`npx -q amplify --version |tail -1`
+AMP_CLI_VERSION_MINIMUM="4.21.0"
+AMP_CLI_VERSION_INVALID=0
+
+
+STRIP_ESCAPE_RESULT=
+stripEscapeUtil() {
```
palpatim, line 34: Rather than using global vars as "return values", consider echoing the result. The caller would then invoke as `result=$(function "arg")`

AmplifyTools/amplify-tools.sh
```diff
@@ -6,13 +6,121 @@
...
+
+    if [ ${HAS_RUBY} -eq 1 ]; then
+	STRIP_ESCAPE_RESULT=`echo "${1}" | ruby -pe 'gsub(/\e\[[0-9;]*m/s, "")'`
+    else
+	STRIP_ESCAPE_RESULT=`echo "${1}" | npx -q strip-ansi-cli`
+    fi
+}
+
+VERSION_INVALID=
+checkMinVersionCompatibility() {
```
palpatim, line 53: Bash functions can return exit codes; consider returning 0 on success, 1 otherwise

AmplifyTools/amplify-tools.sh
```diff
@@ -6,13 +6,121 @@
...
+    VERSION_INVALID=0
+
+    stripEscapeUtil "${1}"
```
palpatim, line 56: nit: It seems reasonable to make the caller responsible for passing a well-formed argument, rather than the function having to deal with escaped input

AmplifyTools/amplify-tools.sh
```diff
@@ -6,13 +6,121 @@
...
+}
+
+checkMinVersionCompatibility "${AMP_APP_VERSION_CURRENT}" "${AMP_APP_VERSION_MINIMUM}"
+AMP_APP_VERSION_INVALID=${VERSION_INVALID}
+
+checkMinVersionCompatibility "${AMP_CLI_VERSION_CURRENT}" "${AMP_CLI_VERSION_MINIMUM}"
+AMP_CLI_VERSION_INVALID=${VERSION_INVALID}
+
+if [ ${AMP_CLI_VERSION_INVALID} -eq 1 ]; then
+    echo "ERROR: Minimum version required of Amplify CLI is not installed."
```
palpatim, line 101: Make this lowercase (`echo "error: ..."`) so it will show up in the main interface Issue Navigator, instead of making the customer go to the build logs to inspect an arcane "Run script build phase exited with nonzero exit status" message.

AmplifyTools/amplify-tools.sh
wooj2: I have mixed feelings on this, and I think i'm going to pass on this for the reason that we have a number of other commands in this functions that can easily emit data to stdout/stderr.

AmplifyTools/amplify-tools.sh
wooj2: Yeah, I was going to do this but we have set -e at the top, and returning 1 is halting execution of the script... I wanted to avoid turning on/off this flag as much as possible....

AmplifyTools/amplify-tools.sh
wooj2: Going to keep this as is for this reason in example B.  That is, I want to make sure I am preserving the argument I am sending to another function.

Example A:
```
func2() {
    echo "Number of args is: $#"
}
func1() {
    func2 ${1}
}
func1 "hello wooj2"
```
Outputs: 2


Example B:
```
func2() {
    echo "Number of args is: $#"
}
func1() {
    func2 "${1}"
}
func1 "hello wooj2"
```
Outputs: 1

PR #519, opened 2020-06-04
[Bug Fix][Amplify API] Fix the bug which fails to update the value to nil

DongQuanRui:
*Description of changes:*
This PR contains one line of code change, but fix the bug which fails to update the value to nil.

`dict[key] = nil` removes the key from the dict
https://developer.apple.com/documentation/swift/dictionary/2885650-subscript

`dict.updateValue(nil, forKey: key)` updates the value to nil without removing the key
https://developer.apple.com/documentation/swift/dictionary/3127179-updatevalue


Podfile.lock
```diff
@@ -61,4 +61,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: e5c47d206a7ea4550dad818207f3a4c8866cf9c0
 
-COCOAPODS: 1.9.3
+COCOAPODS: 1.9.2
```
lawmicha, line 64: if 1.9.3 is already checked in, let's stick with that and upgrade ur local cocoapods version to 1.9.3 to avoid downgrades every time

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/Support/ModelGraphQLTests.swift
```diff
@@ -0,0 +1,46 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+@testable import Amplify
```
lawmicha, line 9: not sure if Amplify is required in the tests below

DongQuanRui: I just checked, Temporal isn't working without Amplify

Podfile.lock
DongQuanRui: Changed

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/Support/ModelGraphQLTests.swift
```diff
@@ -0,0 +1,46 @@
...
+
+        XCTAssertEqual(graphQLInput["title"] as? String, post.title)
+        XCTAssertEqual(graphQLInput["content"] as? String, post.content)
+        XCTAssertEqual(graphQLInput["createdAt"] as? String, post.createdAt.iso8601String)
+        XCTAssertEqual(graphQLInput["draft"] as? Bool, post.draft)
+        XCTAssertEqual(graphQLInput["rating"] as? Double, post.rating)
+        XCTAssertEqual(graphQLInput["status"] as? String, status.rawValue)
+
+        XCTAssertTrue(graphQLInput.keys.contains("updatedAt"))
+        XCTAssertNil(graphQLInput["updatedAt"]!)
```
drochetti, line 44: do you need the force unwrap here?

DongQuanRui: Ya, graphQLInput["updatedAt"] is an `Any??`. But XCTAssertNil() accepts `Any?`. So I need to unwrap a layer.

Issue #503, opened 2020-06-01
Can't save object to DataStore if parent object has two connections to objects of the same type


anthonymoretti:
**Describe the bug**
Can't save object to DataStore if parent object has two connections to objects of the same type.

**To Reproduce**
1. Create type with two connections to objects of the same type.
2. Create type that is a child of the first type.
3. Save the parent object to DataStore
4. Try to save the child object to DataStore

**Expected behavior**
The parent object is saved and then the child object is saved.

**Environment:**
 - Amplify Framework Version: 4.21.0
 - Dependency Manager: Cocoapods
 - Swift Version : 5.0

**Device Information:**
 - Device: Simulator
 - iOS Version: iOS 13.5

**Additional context**

Schema and save code:

```
type Person @model {
  id: ID!
  name: String!
  callerOf: [PhoneCall!] @connection(keyName: "byCaller", fields: ["id"])
  calleeOf: [PhoneCall!] @connection(keyName: "byCallee", fields: ["id"])
}

type PhoneCall
  @model
  @key(name: "byCaller", fields: ["callerID"])
  @key(name: "byCallee", fields: ["calleeID"]) {
  id: ID!
  callerID: ID!
  caller: Person! @connection(fields: ["callerID"])
  calleeID: ID!
  callee: Person! @connection(fields: ["calleeID"])
  transcripts: [Transcript!] @connection(keyName: "byPhoneCall", fields: ["id"])
}

type Transcript @model @key(name: "byPhoneCall", fields: ["phoneCallID"]) {
  id: ID!
  text: String!
  language: String
  phoneCallID: ID!
  phoneCall: PhoneCall! @connection(fields: ["phoneCallID"])
}




let alice = Person(name: "Alice")
let bob = Person(name: "Bob")
let phoneCall = PhoneCall(caller: alice, callee: bob)
let englishTranscript = Transcript(text: "Hi!", language: "en-US", phoneCall: phoneCall)
let spanishTranscript = Transcript(text: "¡Hola!", language: "es", phoneCall: phoneCall)

Amplify.DataStore.save(alice) { result in
	switch result {
	case .success:
		print("Added Alice")
		Amplify.DataStore.save(bob) { result in
			switch result {
			case .success:
				print("Added Bob")
				Amplify.DataStore.save(phoneCall) { result in
					switch result {
					case .success:
						print("Added phone call")
						Amplify.DataStore.save(englishTranscript) { result in
							switch result {
							case .success:
[...]

PR #520, opened 2020-06-04
fix(datastore): eager loaded associations depth

drochetti:
*Issue* #503 

The associations fetching code (inner/left joins) was only fetch one level deep, causing issues when decoding into Swift structs that had required properties that were missing from the SQL result set.

This change not only simplifies *a lot* the result set parsing, but also adds recursion to find all required associations of a given model to make sure data contracts will be fulfilled when using Swift decode.


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
```diff
@@ -101,12 +72,81 @@ struct SelectStatement: SQLStatement {
...
+                                       columnMapping: columnMapping,
+                                       bindings: bindings)
+    }
+
+    /// Tuple that represents a pair of `columns` and `statements`
+    private typealias JoinStatements = ([String], [String], ColumnMapping)
+
+    /// Walk through the associations recursively to generate join statements.
+    ///
+    /// Implementation note: this should be revisited once we define support
```
royjit, line 85: We are doing eager loading now right? Let us add that to the comment.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
```diff
@@ -9,78 +9,49 @@ import Amplify
...
+struct SelectStatementMetadata {
+
+    typealias ColumnMapping = [String: (ModelSchema, ModelField)]
+
+    let statement: String
+    let columnMapping: ColumnMapping
+    let bindings: [Binding?]
+
+    // TODO remove additionalStatements once sorting support is added to DataStore
+    static func get(from modelType: Model.Type,
```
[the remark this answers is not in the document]
drochetti: sure thing, suggestions?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -35,99 +33,159 @@ protocol StatementModelConvertible {
...
+            "associatedField": associatedWith,
+            "elements": []
+        ]
+    }
+}
+
+internal extension Dictionary where Key == String, Value == Any? {
+
+    /// Utility to create a `NSMutableDictionary` from a Swift `Dictionary<String, Any?>`.
+    func mutableCopy() -> NSMutableDictionary {
```
wooj2, line 127: I think this name should be something like nsMutableCopy... because we are converting a data structure (dictionary) to an NS data structure?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -35,99 +33,159 @@ protocol StatementModelConvertible {
...
+internal extension NSMutableDictionary {
+
+    /// Utility to allows Swift standard types to be used in `setObject`
+    /// of the `NSMutableDictionary`.
+    ///
+    /// - Parameters:
+    ///   - value: the value to be set
+    ///   - key: the key as a `String`
+    func updateValue(_ value: Value?, forKey key: String) {
+        setObject(value as Any, forKey: key as NSString)
```
wooj2, line 144: If value is nil,  i'm not sure if this works, because I don't think NSDictionaries can hold `nil` (i believe in obj-c we'd have to set it to `[NSNull null]`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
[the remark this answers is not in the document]
lawmicha: can this just be the constructor for `SelectStatementMetadata`?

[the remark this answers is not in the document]
royjit: make `metaData(from:)`, `statementMetaData(from:)` or why not an init method?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -35,99 +33,159 @@ protocol StatementModelConvertible {
...
+    ///             "value": 1
+    ///         ]
+    ///     ]
+    /// ]
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - value: the value to be set
+    ///   - keyPath: the key path as a `String` (e.g. "nested.key")
+    func updateValue(_ value: Value?, forKeyPath keyPath: String) {
```
wooj2, line 171: Hmm stupid question: what if we have something like `nested..key`. will `keyComponents` be 2 or 3?  Maybe there's some extra sanity checks on the data that is being passed in here?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
```diff
@@ -101,12 +72,81 @@ struct SelectStatement: SQLStatement {
...
+
+    let modelType: Model.Type
+    let metadata: SelectStatementMetadata
+
+    init(from modelType: Model.Type,
+         predicate: QueryPredicate? = nil,
+         paginationInput: QueryPaginationInput? = nil,
+         additionalStatements: String? = nil) {
+        self.modelType = modelType
+        self.metadata = .get(from: modelType,
```
royjit, line 138: A lot happens here in the init method, and I think this code path is sync with the developers code `Amplify.DataStore.query`. Is this an expensive statement that can affect the query api?

In future we should add some performance test around this.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
```diff
@@ -101,12 +72,81 @@ struct SelectStatement: SQLStatement {
...
+                    columnMapping.updateValue((associatedSchema, field), forKey: "\(alias).\(field.name)")
+                    return field.columnName(forNamespace: alias) + " as " + field.columnAlias(forNamespace: alias)
+                }
+
+                let joinType = foreignKey.isRequired ? "inner" : "left outer"
+
+                joinStatements.append("""
+                \(joinType) join \(associatedTableName) as "\(alias)"
+                  on \(associatedColumn) = \(foreignKeyName)
+                """)
```
wooj2, line 114: Being ultra paranoid, do we care to filter out names of tables that might impact these statements.  Like... it would be bad if someone had a table called `Null` or `as`... maybe something to do for another PR.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
```diff
@@ -101,12 +72,81 @@ struct SelectStatement: SQLStatement {
...
+    }
+
+    /// Tuple that represents a pair of `columns` and `statements`
+    private typealias JoinStatements = ([String], [String], ColumnMapping)
+
+    /// Walk through the associations recursively to generate join statements.
+    ///
+    /// Implementation note: this should be revisited once we define support
+    /// for explicit `eager` vs `lazy` associations.
+    private static func joins(from schema: ModelSchema) -> JoinStatements {
```
lawmicha, line 87: how about `generate(from: foreignKeys: [ModelField] -> JoinStatement?` ? JoinStatement in this case would be a struct like
```
struct JoinStatement {
        let columns: [String]
        let statements: [String]
        let columnMapping: ColumnMapping
    }
```
and would return nil if there were no foreign keys to generate a JoinStatement from

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -35,99 +33,159 @@ protocol StatementModelConvertible {
...
     /// Converts all the rows in the current executed `Statement` to the given `Model` type.
     ///
-    /// - Parameter modelType the target `Model` type
+    /// - Parameters:
+    ///   - modelType - the target `Model` type
+    ///   - statement - the query executed that generated this result
     /// - Returns: an array of `Model` of the specified type
-    func convert<M: Model>(to modelType: M.Type) throws -> [M]
+    func convert<M: Model>(to modelType: M.Type,
+                           using statement: SelectStatement) throws -> [M]
```
lawmicha, line 41: can just pass over `ColumnMapping` instead of `SelectStatement`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -35,99 +33,159 @@ protocol StatementModelConvertible {
...
+    /// Utility to create a `NSMutableDictionary` from a Swift `Dictionary<String, Any?>`.
+    func mutableCopy() -> NSMutableDictionary {
+        // swiftlint:disable:next force_cast
+        return (self as NSDictionary).mutableCopy() as! NSMutableDictionary
+    }
+}
+
+/// Extension that adds utilities to created nested values in a dictionary
+/// from a `keyPath` notation (e.g. `root.with.some.nested.prop`.
+internal extension NSMutableDictionary {
```
lawmicha, line 135: seems all related and can be in `NSMutableDictionary+Extensions.swift`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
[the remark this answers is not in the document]
drochetti: > can this just be the constructor for `SelectStatementMetadata`?

using the constructor is tricky because I have to execute the entire logic inside the constructor (no other functions, etc) because I need to initialize any struct property.

That's why I went with the static factory function, otherwise the constructor becomes a mess.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
drochetti: sure... but doesn't the return type make that clear?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -35,99 +33,159 @@ protocol StatementModelConvertible {
...
+                fieldType: field.type
+            )
+            modelDictionary.updateValue(modelValue, forKeyPath: column)
+
+            // create lazy list for "many" relationships
+            if let id = modelValue as? String, field.isPrimaryKey {
+                let associations = schema.fields.values.filter {
+                    $0.isArray && $0.hasAssociation
+                }
+                let prefix = field.name.replacingOccurrences(of: field.name, with: "")
```
lawmicha, line 93: what is the prefix here? isn't it just ""?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
drochetti: they are escaped with double quotes, so you can even pass a SQL statement there and nothing would happen.

that said, the names here are a result of the models, so only valid Swift identifiers can be passed here.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
drochetti: that's an invalid `keyPath`. We could add sanity checks, but it wouldn't make a difference because this is used in a controlled environment. The `keyPath` passed here is created by `SelectStatement`, is select statement creates an invalid key, those tests will fail. If tests don't catch it, the SQL operation will fail before it gets here.

but good question, I don't know how https://developer.apple.com/documentation/objectivec/nsobject/1418139-setvalue behaves in those cases. It would be good to mimic its behavior

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
drochetti: +1 to performance benchmarks

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
drochetti: it depends on how many levels deep you are in the object graph. It starts with `""`, it becomes `"post."` then `"post.blog."`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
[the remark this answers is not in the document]
lawmicha: I was also thinking `.selectStatementMetadata(from:)`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
drochetti: I want to actually make these `fileprivate`, I don't want to make an extension that can be used elsewhere.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
drochetti: Good call, fixing it!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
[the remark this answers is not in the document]
drochetti: it's an internal API, so I renamed to `.metadata()`, feel free to change that later as it has no impact on customers.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
drochetti: done

PR #524, opened 2020-06-04
feat(DataStore): Combine support for all APIs

palpatim:
Adds a new AmplifyCombineSupport pod that provides support for Combine flavors of Amplify APIs. Currently only supports DataStore.


AmplifyCombineSupport/AmplifyCombineSupport.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,612 @@
...
+					"DEBUG=1",
+					"$(inherited)",
+				);
+				GCC_WARN_64_TO_32_BIT_CONVERSION = YES;
+				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
+				GCC_WARN_UNDECLARED_SELECTOR = YES;
+				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
+				GCC_WARN_UNUSED_FUNCTION = YES;
+				GCC_WARN_UNUSED_VARIABLE = YES;
+				IPHONEOS_DEPLOYMENT_TARGET = 13.5;
```
lawmicha, line 417: i had troubles running this on 13.5 since i only have 13.2 simulators, so i ended up setting this to `13.0` and had to update the test methods to use `override func setUp() {` instead of `override func setUpWithError() throws {` (probably a sign for me to upgrade Xcode)

AmplifyCombineSupport/AmplifyCombineSupportTests/DataStoreCombineSupportChainTests.swift
```diff
@@ -0,0 +1,236 @@
...
+                ),
+                Amplify.DataStore.save(
+                    Transcript(text: "transcripción de la llamada en español",
+                               language: "es",
+                               phoneCall: phoneCall)
+                )
+            )
+        }.sink(receiveCompletion: { completion in
+            if case .failure = completion {
+                receivedError.fulfill()
```
lawmicha, line 106: is there a way to know that it was when saving the PhoneCall that failed?

AmplifyCombineSupport/AmplifyCombineSupport/AmplifyCombineSupport.h
```diff
@@ -0,0 +1,19 @@
+//
+//  AmplifyCombineSupport.h
+//  AmplifyCombineSupport
+//
+//  Created by Schmelter, Tim on 6/4/20.
+//  Copyright © 2020 Amazon Web Services. All rights reserved.
```
lawmicha, line 6: what's this file used for? can we just remove it? also, the swiftformat didn't pick up the file to replace the header with the
```
//
// Copyright 2018-2020 Amazon.com,
// Inc. or its affiliates. All Rights Reserved.
//
//
```

AmplifyCombineSupport/AmplifyCombineSupport.xcodeproj/project.pbxproj
lawmicha: it was at the project level
image

AmplifyCombineSupport/AmplifyCombineSupport/DataStore.swift
```diff
@@ -0,0 +1,75 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+import Combine
+
+public typealias DataStorePublisher<Output> = AnyPublisher<Output, DataStoreError>
+
+public extension DataStoreBaseBehavior {
```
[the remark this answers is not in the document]
palpatim: Will investigate to see if it's possible to automatically import this as part of `import Amplify; import Combine` as long as the support library is installed. Considerations:

1. Can the solution apply to non-CocoaPods installations (e.g., SPM, Carthage, framework)?
2. Does the solution feel not-too-surprising? E.g., I would rather be explicit and require an `import AmplifyCombineSupport` until we officially roll it into Amplify Core in a couple of years, than make customers wonder why they have to `import Combine` to use the Combine-flavored APIs, even though they’re using any core Combine APIs in that file. Realistically this might not be a problem, since you'll often have to store a reference to an `AnyCancellable` or `Publisher` of some kind, which requires a Combine import.

AmplifyCombineSupport.podspec
```diff
@@ -0,0 +1,30 @@
...
+  s.name         = 'AmplifyCombineSupport'
+  s.version      = AMPLIFY_VERSION
+  s.summary      = 'Amazon Web Services Amplify for iOS.'
+
+  s.description  = 'AWS Amplify for iOS provides a declarative library for application development using cloud services'
+
+  s.homepage     = 'https://aws.amazon.com/amplify/'
+  s.license      = 'Apache License, Version 2.0'
+  s.author       = { 'Amazon Web Services' => 'amazonwebservices' }
+  s.source       = { :git => 'https://github.com/aws-amplify/amplify-ios.git', :tag => s.version}
```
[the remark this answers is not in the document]
palpatim: Thanks for this callout, will fix.

AmplifyCombineSupport/AmplifyCombineSupportTests/DataStoreCombineSupportChainTests.swift
palpatim: Not unless the failing operation enhances the error in some way that makes it obvious. In that way, this pattern is similar to a `do/catch` block, or a Promise chain in the async world--error handling is consolidated in one place.

palpatim:
> I think this is a great add to target for `1.1.0`, so my question comes down to how should we manage milestones in parallel. Currently with our branching strategy, once this is merged to `master` it means it will be in the next release. How can we manage parallel minor/path releases? I have created an item in our backlog to discuss our Git branching model, I would like to hear your thoughts on it, I think this PR is a precedent to that discussion.

My initial instinct is to merge it in when we're ready for it, where "ready" is a combination of: "the code is ready", and "we're ready to take it." Adding additional processes (long-lived, release-targeted branches) feels like the wrong way to go, especially as we try to move releases to be more automated, low-touch events.

> It seems they will be a lot of boilerplate involved and effort maintaining as we change/add APIs. I would love to discuss the possibility of automating extensions like this, maybe using something like https://github.com/krzysztofzablocki/Sourcery to help us automate the boilerplate maintenance?

I don't object to bringing in dependencies to help with builds and tests--see SwiftFormat, SwiftLint, CwlWithLove*, etc. Sourcery seems like a nice win if we can get it to take care of the boilerplate for us. We'd have to experiment to make sure it handles transformations as we want it to, especially in the progress-reporting APIs, which we haven't yet addressed with the Combine Support.

Issue #488, opened 2020-05-26
Run Script for Amplify tools can't find node when nvm is used


attilah:
**Describe the bug**
Following the Todo guide when I try to build the project, the Run Script phase fails with the error:

```
env: node: Not a directory
warning: Node is not installed. Vist https://nodejs.org/en/download/ to install it
Command PhaseScriptExecution failed with a nonzero exit code
```

I'm using node through nvm and does work from `sh`, `bash`, `zsh`, but not from xcode script.

I even tried to add this to the script, but did not help:
```
export PATH="$PATH:/Users/<user>/.nvm/versions/node/v13.12.0/bin/node"
```

drochetti:
A temporary mitigation for this issue is to prepend to your Amplify Build Phase Script the NVM initialization: `. ~/.nvm/nvm.sh;`

A complete example of the Build Phase script:

```sh
. ~/.nvm/nvm.sh;
${PODS_ROOT}/AmplifyTools/amplify-tools.sh;
```

drochetti:
This was fixed as part of [1.0.2](https://github.com/aws-amplify/amplify-ios/releases/tag/v1.0.2). Please re-open if you're still experiencing the same issue.

roxberry:
Started on Xcode 12 beta 6, adding . ~/.nvm/nvm.sh; resolves it. Using amplify 4.29.1

PR #535, opened 2020-06-09
fix(tools): nvm support added so node is correctly resolved

drochetti:
*Issue #* #488 

*Description of changes:* When NVM is installed it needs to be manually initialized inside the Xcode script runtime, otherwise a different NodeJS will be used.

ps: also did some formatting on the file (actually VSCode did).

Diff excluding the formatting changes: https://github.com/aws-amplify/amplify-ios/pull/535/commits/0b4f9bc20fd06318334e2c0884b6493ee114145b

---

### DIFF
--- a/AmplifyTools/amplify-tools.sh
+++ b/AmplifyTools/amplify-tools.sh
@@ -18,9 +18,9 @@ if [ -f "${NVM_PATH}" ]; then
   source "${NVM_PATH}"
 fi
 
-export PATH=$PATH:$(npm bin -g)
-
 set -e
+
+export PATH=$PATH:$(npm bin -g)
 
 # Note the use of tail -1 is important here because when upgrading between versions
 # the first time that you run these commands, we have seen this variable take on the value of:
--- a/AmplifyTools/amplify-tools.sh
+++ b/AmplifyTools/amplify-tools.sh
@@ -12,6 +12,15 @@ if ! which node >/dev/null; then
   exit 1
 fi
 
+# Check for NVM and make sure it's initialized
+NVM_PATH = "~/.nvm/nvm.sh"
+if [ -f $NVM_PATH ]; then
+  echo "NVM found, initializing it..."
+  . $NVM_PATH
+fi
+
+export PATH=$PATH:$(npm bin -g)
+
 # Note the use of tail -1 is important here because when upgrading between versions
 # the first time that you run these commands, we have seen this variable take on the value of:
 # """
--- a/AmplifyTools/amplify-tools.sh
+++ b/AmplifyTools/amplify-tools.sh
@@ -5,7 +5,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-set -e
+# Check NodeJS installation
 
 if ! which node >/dev/null; then
   echo "warning: Node is not installed. Visit https://nodejs.org/en/download/ to install it"
@@ -13,13 +13,15 @@ if ! which node >/dev/null; then
 fi
 
 # Check for NVM and make sure it's initialized
-NVM_PATH="~/.nvm/nvm.sh"
+NVM_PATH=~/.nvm/nvm.sh
 if [ -f "${NVM_PATH}" ]; then
   echo "NVM found, initializing it..."
-  . ${NVM_PATH}
+  source $NVM_PATH
 fi
 
 export PATH=$PATH:$(npm bin -g)
+
+set -e
 
 # Note the use of tail -1 is important here because when upgrading between versions
 # the first time that you run these commands, we have seen this variable take on the value of:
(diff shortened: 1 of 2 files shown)


AmplifyTools/amplify-tools.sh
```diff
@@ -29,66 +38,65 @@ AMP_CLI_VERSION_CURRENT=`npx -q amplify --version |tail -1`
...
-    if [ ${HAS_RUBY} -eq 1 ]; then
-	STRIP_ESCAPE_RESULT=`echo "${1}" | ruby -pe 'gsub(/\e\[[0-9;]*m/s, "")'`
-    else
-	STRIP_ESCAPE_RESULT=`echo "${1}" | npx -q strip-ansi-cli`
-    fi
+  if [ ${HAS_RUBY} -eq 1 ]; then
+    STRIP_ESCAPE_RESULT=`echo "${1}" | ruby -pe 'gsub(/\e\[[0-9;]*m/s, "")'`
+  else
+	  STRIP_ESCAPE_RESULT=`echo "${1}" | npx -q strip-ansi-cli`
+  fi
```
wooj2, line 57: it would be nice if we could revert these space changes?

AmplifyTools/amplify-tools.sh
```diff
@@ -29,66 +38,65 @@ AMP_CLI_VERSION_CURRENT=`npx -q amplify --version |tail -1`
...
-	[ -z "${CURR_VERSION_PATCH}" ] ||
-	[ -z "${REQU_VERSION_MAJOR}" ] ||
-	[ -z "${REQU_VERSION_MINOR}" ] ||
-	[ -z "${REQU_VERSION_PATCH}" ]; then
-	VERSION_INVALID=1
-	return
-    fi
+  VERSION_INVALID=0
+
+  stripEscapeUtil "${1}"
```
wooj2, line 64: Is it possible to revert the space changes in this file?

AmplifyTools/amplify-tools.sh
```diff
@@ -5,15 +5,24 @@
...
 if ! which node >/dev/null; then
   echo "warning: Node is not installed. Visit https://nodejs.org/en/download/ to install it"
   exit 1
 fi
 
+# Check for NVM and make sure it's initialized
+NVM_PATH=~/.nvm/nvm.sh
+if [ -f "${NVM_PATH}" ]; then
+  echo "NVM found, initializing it..."
+  source $NVM_PATH
```
wooj2, line 19: 1.  If we want to support directories with spaces, this needs to be:
```
NVM_PATH="${HOME}/.nvm/nvm.sh"
```
(note that ~ can not be used in a string)

2.  Then if we assume there is a space in the path, we should change the `source` command on line 19 to have quotes like:
```
source "${NVM_PATH}"
```

3.  Btw, what does `nvm.sh` expect to have?  i'm guessing it updates some path variables?

AmplifyTools/amplify-tools.sh
drochetti: it is, let me do it. Sorry, only realized to late that I had "format on save" for `.sh` files

drochetti: done

AmplifyTools/amplify-tools.sh
wooj2: I don't fully understand what is inside nvm.sh,  but going to assume it is the right thing to do..

AmplifyTools/amplify-tools.sh
```diff
@@ -5,15 +5,23 @@
...
+NVM_PATH="${HOME}/.nvm/nvm.sh"
+if [ -f "${NVM_PATH}" ]; then
+  echo "NVM found, initializing it..."
+  source "${NVM_PATH}"
+fi
+
 export PATH=$PATH:$(npm bin -g)
 
+set -e
+
```
wooj2, line 24: Sorry, I should have caught this before....

`set -e` should probably come before `export PATH=$PATH:$(npm bin -g)` in case `npm bin` fails.

AmplifyTools/amplify-tools.sh
drochetti: me neither, it's a fairly long file (~3500 lines) that handles both NVM and NodeJS version initialization. Based on my research, that safe to call and there are NVM+Xcode issues out there that suggest the same thing to fix it.

I was able to repro the issue, Xcode - for some reason - doesn't pick my default NodeJS version, which is controlled by NVM. That results in the wrong version of amplify CLI being used by our script, so even when I do `npm -g install @aws-amplify/cli` in my terminal Xcode still picks the wrong version. By forcing the NVM initialization inside the Xcode Build Phase the problem is gone.

PR #538, opened 2020-06-10
fix(build): podspec source version tag and url

drochetti:
**Notes:**

- added `dependencies.rb` which contains common dependency configuration for all modules
- added `Podfile.lock` to standardize the dependencies across environments for all modules (previously it was only for Core, for unknown reasons)
- changed the versions to global constants (prefixed with `$`) so Cocoapods stop emitting warnings for redefined values

ps: `Podfile.lock` files represent most of the diff, it is safe to ignore them during review


AWSPluginsCore.podspec
```diff
@@ -8,7 +8,7 @@
 
 Pod::Spec.new do |s|
   AMPLIFY_VERSION = '1.0.1'
-  AWS_SDK_VERSION = "~> 2.13.4"
+  AWS_SDK_VERSION = '~> 2.13.4'
```
palpatim, line 11: Can we take this opportunity to fix my mistake of including the "compatibility squiggle" in the version declaration as well? That is, let's make both version declarations be just bare semvers, and decorate them as needed (with `"~>"` compatibility prefix, or the `"v"` prefix) at the call sites.

dependencies.rb
```diff
@@ -0,0 +1,29 @@
...
+$AWS_SDK_VERSION = '2.13.4'
+
+# http://guides.cocoapods.org/using/the-podfile.html#specifying-pod-versions
+def optimistic_version(pod_version)
+  "~> #{pod_version}"
+end
+
+# GitHub tag name for Amplify releases
+def release_tag
+  "v#{$AMPLIFY_VERSION}"
```
lawmicha, line 12: @royjit  @palpatim  i recall we want to keep this 'v' here and update the circleCI workflows scripts to account for this, right?

palpatim: @lawmicha Correct

royjit: Yes, the tag name should start with v

PR #539, opened 2020-06-10
Add Embedded type to store schema info for custom types

lawmicha:
*Issue #, if available:*

*Description of changes:*
- This adds `Embbeded` protocol to return the schema information for embedded types. This means the code generated schema for an embedded type should look something like this:
```
class Category: Embedded {
...
}
extension Category {
    public enum CodingKeys: CodingKey {
        case name
        case color
    }
    public static let keys = CodingKeys.self

    // store info about the field's name, nullability, and their types.
    public static let schema = defineSchema { embedded in
        let category = Category.keys
        embedded.fields(.field(category.name, is: .required, ofType: .string),
                       .field(category.color, is: .required, ofType: .embedded(type: Color.self)))
    }
}
```

- Also introduced a `.embeddedCollection` case to hold embbeded types which are arrays. This was required when checking type at runtime. 
```
    public var embeddedType: Embedded.Type? {
        switch type {
        case .embedded(let type), .embeddedCollectiion(let type):
           // the `type` here is always castable to the Embedded.Type. 
            if let embeddedType = type as? Embedded.Type {
                return embeddedType
            }
```

Testing done so far 
- selection set verification
- graphql query variables input, `graphQLInput`
- sample app without conflict resolution enabled and making calls to `API.create` 
- sample app with conflict resolution enabled and `DataStore.save` is successfully and is synced to the cloud

### DIFF
--- a/Amplify/Categories/DataStore/Model/Embedded.swift
+++ b/Amplify/Categories/DataStore/Model/Embedded.swift
@@ -1,0 +1,28 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+// MARK: - CustomCodable
+
+/// All persistent custom types should conform to the CustomCodable protocol.
+public protocol Embedded: Codable {
+
+    /// A reference to the `ModelSchema` associated with this embedded type.
+    static var schema: ModelSchema { get }
+}
+
+extension Embedded {
+    public static func defineSchema(name: String? = nil,
+                                    attributes: ModelAttribute...,
+                                    define: (inout ModelSchemaDefinition) -> Void) -> ModelSchema {
+        var definition = ModelSchemaDefinition(name: name ?? "",
+                                               attributes: attributes)
+        define(&definition)
+        return definition.build()
+    }
+}
--- a/Amplify/Categories/DataStore/Model/Embedded.swift
+++ b/Amplify/Categories/DataStore/Model/Embedded.swift
@@ -7,16 +7,19 @@
 
 import Foundation
 
-// MARK: - CustomCodable
+// MARK: - Embeddable
 
-/// All persistent custom types should conform to the CustomCodable protocol.
-public protocol Embedded: Codable {
+/// A `Embeddable` type can be used in a `Model` as an embedded type. All types embedded in a `Model` as an
+/// `embedded(type:)` or `embeddedCollection(of:)` must comform to the `Embeddable` protocol except for Swift's Basic
+/// types embedded as a collection. A collection of String can be embedded in the `Model` as
+/// `embeddedCollection(of: String.self)` without needing to conform to Embeddable. 
+public protocol Embeddable: Codable {
 
     /// A reference to the `ModelSchema` associated with this embedded type.
     static var schema: ModelSchema { get }
 }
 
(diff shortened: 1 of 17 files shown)


Amplify/Categories/DataStore/CustomCodable/CustomCodable.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+// MARK: - CustomCodable
+
+/// All persistent custom types should conform to the CustomCodable protocol.
+public protocol CustomCodable: Codable {
```
drochetti, line 13: I think we need a better name for `CustomCodable`. ORMs typically call these *"Embedded"* types, maybe we could follow that?

@jpignata also mentioned the *non-model* thing we keep talking about is not clear, and I agree.

lawmicha:
related regression maybe https://github.com/aws-amplify/amplify-ios/issues/541

Amplify/Categories/DataStore/Model/Embedded.swift
```diff
@@ -0,0 +1,28 @@
...
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+// MARK: - CustomCodable
+
+/// All persistent custom types should conform to the CustomCodable protocol.
```
drochetti, line 12: update description.

also, maybe it would be good to expand on this a bit and explain the different between `Embedded` and `Model`

Amplify/Categories/DataStore/Model/Schema/ModelField+Association.swift
```diff
@@ -183,4 +183,25 @@ extension ModelField {
...
+        }
+    }
+
+    public var isEmbeddedType: Bool {
+        if case .embedded = type {
+            return true
+        }
+        if case .embeddedCollection = type {
+            return true
+        }
```
drochetti, line 204: you could potentially combine both using a `,` in the same `if` clause, or also use a `switch` where the default returns false.

Amplify/Categories/DataStore/Model/Embedded.swift
```diff
@@ -0,0 +1,28 @@
...
+public protocol Embedded: Codable {
```
drochetti, line 13: minor: maybe worth considering calling the protocol `Embeddable`.

It suits well what the industry usually uses (see [Hibernate approach](https://www.baeldung.com/jpa-embedded-embeddable)) and also the iOS naming patterns.

An `Embeddable` is `.embedded(type:)` in a `Model`. I can see that making way more sense than "custom type" or "non-model type". =)

lawmicha:
related Codegen https://github.com/aws-amplify/amplify-cli/pull/4545

Amplify/Categories/DataStore/Model/Embedded.swift
lawmicha: that sounds good, i will make this change!

Amplify/Categories/DataStore/Model/Schema/ModelField+Association.swift
lawmicha: i think `if case .x = type, case .y = type` is checking for both case and will never be true. not completely sure if that is the case when i saw a test failed, so i switched to using the switch statement instead

Issue #540, opened 2020-06-10
awsDynamoDBStoragePlugin is not supported but added to amplifyconfiguration.json


lawmicha:
**Is your feature request related to a problem? Please describe.**
When i add Storage Category and choose to use DynamoDB as storage, there is no corresponding Amplify Plugin for me to use.

**Describe the solution you'd like**
ignore and log unknown configuration keys during Amplify.configure()

**Describe alternatives you've considered**
Possibly remove the option to add dynamoDB as a storage plugin when in an amplify iOS project.

**Additional context**
This was added as part of the `amplifyconfiguration.json` work in CLI 
https://github.com/aws-amplify/amplify-cli/blob/eacaa96038ce109e8491bf9cdabb6ac8de3c02d6/packages/amplify-frontend-android/lib/amplify-config-helper.js#L147

`amplifyconfiguration.json`
```
{
    "UserAgent": "aws-amplify-cli/2.0",
    "Version": "1.0",
    "storage": {
        "plugins": {
            "awsDynamoDbStoragePlugin": {
                "partitionKeyName": "id",
                "region": "us-west-2",
                "arn": "arn:aws:dynamodb:us-123-2:123123123123:table/123123123123-dev",
                "streamArn": "arn:aws:dynamodb:us-123-2:123123123123123:table/123123-dev/stream/2020-06-10T23:21:25.860",
                "partitionKeyType": "S",
                "name": "dynamoc1a123123b-dev"
            }
        }
    }
}
```
`awsconfiguration.json`
```
{
    "UserAgent": "aws-amplify/cli",
    "Version": "0.1.0",
    "IdentityManager": {
        "Default": {}
    },
    "DynamoDBObjectMapper": {
        "Default": {
            "Region": "us-west-2"
        }
    }
}
```
when i run this for JS react, `aws-exports.js`
```
const awsmobile = {
    "aws_project_region": "us-west-2",
    "aws_dynamodb_all_tables_region": "us-west-2",
    "aws_dynamodb_table_schemas": [
        {
            "tableName": "dynamo81236-dev",
            "region": "us-west-2"
        }
    ]
};
```
commands i ran
```
> amplify add storage
? Please select from one of the below mentioned services: NoSQL Database
? Please provide a friendly name for your resource that will be used to label this category in the project: dynamo80123123
? Please provide table name: dynamo8123123
? What would you like to name this column: id
? Please choose the data type: string
? Would you like to add another column? Yes
? What would you like to name this column: name
[...]

PR #543, opened 2020-06-12
fix(Core): Fix plugin configuration validation

palpatim:
Previous validation logic threw an error if it encountered a configuration
stanza with no corresponding plugin registered for that key. However, the
current CLI emits some plugin keys in unexpected situations, breaking this
validation. Remediation would require a customer to edit the config file, which
is possibly dangerous, and definitely inconvenient.

This change:
- Replaces the throwing error with a logged warning when the config encounters
  an unmatched plugin key
- Changes the `Plugin.configure` signature to accept an optional `Any?`
  argument, instead of a non-optional. This will let plugins decide if they
  need to throw or not.
- Fixes a data race in hub test code
- Fix AWSCognitoAuthPlugin and AWSS3StoragePlugin to throw a PluginError
  instead of a category-specific error if they encounter an error during
  configuration

*Issue #, if available:*

#540

### DIFF
--- a/Amplify/Core/Configuration/AmplifyConfiguration.swift
+++ b/Amplify/Core/Configuration/AmplifyConfiguration.swift
@@ -145,4 +145,22 @@ extension Amplify {
 
         try configurable.configure(using: configuration)
     }
+
+    /// Configures a list of plugins with the specified CategoryConfiguration. If any configurations do not match the
+    /// specified plugins, emits a log warning.
+    static func configure(plugins: [Plugin], using configuration: CategoryConfiguration) throws {
+        var pluginConfigurations = configuration.plugins
+
+        for plugin in plugins {
+            let pluginConfiguration = pluginConfigurations[plugin.key]
+            try plugin.configure(using: pluginConfiguration)
+            pluginConfigurations.removeValue(forKey: plugin.key)
+        }
+
+        for unusedPluginKey in pluginConfigurations.keys {
+            log.warn("No plugin found for configuration key `\(unusedPluginKey)`. Add a plugin for that key.")
+        }
+
+    }
+
 }
--- a/Amplify/Core/Configuration/AmplifyConfiguration.swift
+++ b/Amplify/Core/Configuration/AmplifyConfiguration.swift
@@ -160,7 +160,6 @@ extension Amplify {
         for unusedPluginKey in pluginConfigurations.keys {
             log.warn("No plugin found for configuration key `\(unusedPluginKey)`. Add a plugin for that key.")
         }
-
     }
 
 }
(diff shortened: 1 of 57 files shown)


Amplify/Core/Configuration/AmplifyConfiguration.swift
```diff
@@ -145,4 +145,22 @@ extension Amplify {
...
+        for plugin in plugins {
+            let pluginConfiguration = pluginConfigurations[plugin.key]
+            try plugin.configure(using: pluginConfiguration)
+            pluginConfigurations.removeValue(forKey: plugin.key)
+        }
+
+        for unusedPluginKey in pluginConfigurations.keys {
+            log.warn("No plugin found for configuration key `\(unusedPluginKey)`. Add a plugin for that key.")
+        }
+
```
lawmicha, line 163:
```suggestion
```

AmplifyPlugins/API/AWSAPICategoryPluginTests/Configuration/AWSAPICategoryPluginConfigurationTests.swift
```diff
@@ -7,14 +7,27 @@
...
+        let amplifyConfig = AmplifyConfiguration(api: categoryConfig)
+        do {
+            try Amplify.configure(amplifyConfig)
+            XCTFail("Should have thrown a pluginConfigurationError if not supplied with a plugin-specific config.")
+        } catch {
+            if case PluginError.pluginConfigurationError = error {
+                // Pass
+            } else {
+                XCTFail("Should have thrown a pluginConfigurationError if not supplied with a plugin-specific config.")
+            }
```
lawmicha, line 29: better to use guard to avoid comment?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/AWSDataStorePluginConfigurationTests.swift
```diff
@@ -0,0 +1,28 @@
...
+
+import XCTest
+import Amplify
+import AWSDataStoreCategoryPlugin
+
+class AWSDataStorePluginConfigurationTests: XCTestCase {
+
+    // Note this test requires the ability to write a new database in the Documents directcory, so it must be embedded
+    // in a host app
+    func testDoesNotThrowOnMissingConfig() throws {
```
lawmicha, line 16: unrelated to this PR, leaving a note for @DongQuanRui this looks like the right place to put tests for the configuring DataStore when schema version changes

AmplifyPlugins/API/AWSAPICategoryPluginTests/Configuration/AWSAPICategoryPluginConfigurationTests.swift
palpatim: Derp, yes, should have done that

PR #551, opened 2020-06-16
feat(DataStore): database recreation based on new schema

DongQuanRui:
*Description of changes:*
- When user updates schema, the database should be recreated
- Store schema version in the `UserDefaults` database
- Use `ModelRegistration.version` to verify the right one is being used when starting the `StorageAdapter`
- Have log warning that the database will be re-created
- Added 4 unit tests
- Added 2 integration tests
- Need to update DataStore documentation describing the behavior


AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+class DataStoreConfigurationTests: XCTestCase {
+
+    func testConfigWithSameSchema() throws {
+
+        let databaseCreationDone = expectation(description: "Old database deleted and new database recreated")
```
lawmicha, line 30: across different test runs, the database is not created on the first save after configuration. let's just keep it what you are using it for, like `saveSuccess = expectation(description: "Save was successful")` then if you want to explain the overall test, can just add context in a test comment

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+
+        do {
+            let dataStorePlugin = AWSDataStorePlugin(modelRegistration: AmplifyModels())
```
lawmicha, line 33: nit: be explicit then `let previousVersion = 123` and use previousVersion across both configurations in this test method

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+            try Amplify.add(plugin: dataStorePlugin)
+            try Amplify.configure()
+            print("Amplify configured with DataStore plugin")
+        } catch {
+            print("Failed to initialize Amplify with \(error)")
+        }
+
+        let time: Temporal.DateTime = .now()
+        let post = Post(title: "title", content: "content", createdAt: time)
+        let id = post.id
```
lawmicha, line 43: can be removed, use `post.id` at line 68

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+
+        Amplify.DataStore.save(post, completion: { result in
+            switch result {
+            case .success:
+                print("Post saved successfully!")
```
lawmicha, line 48: remove `print` here and else where, you can replace some with `XCTFail(message:)`. If the expecation fails, the expecation also has a message

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+                databaseCreationDone.fulfill()
+            case .failure(let error):
+                print("Error saving post \(error)")
+            }
+        })
+        wait(for: [databaseCreationDone], timeout: 10)
```
lawmicha, line 54: nit: can probably still use `timeout: TestCommonConstants.networkTimeout` even if we know these tests are not making network calls, or store 10 seconds as a constant in class and reuse it, or maybe TestCommonConstants can have a regular one like `timeout = TimeInterval(10)` then you can use it across as `TestCommonConstants.timeout`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+
+        Amplify.reset()
+
+        let databaseRecreationDone = expectation(description: "Database remains")
```
lawmicha, line 58: `querySuccess`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        Amplify.DataStore.query(Post.self, byId: id) { result in
+            switch result {
+            case .success(let postRes):
+                XCTAssertNotNil(postRes)
+                XCTAssertEqual(postRes!.id, id)
+                XCTAssertEqual(postRes!.title, "title")
+                XCTAssertEqual(postRes!.content, "content")
+                XCTAssertEqual(postRes!.createdAt, time)
```
lawmicha, line 75: nit: yes the "postRes" is guarenteed to be not nil from line 71, but i'm still not convinced we should use force unwrapping here and should avoid it if we can. instead of 4 unwraps, you can do it once and provide a more meaningful message when it is `nil`
```
guard let queriedPost = postOptional else {
  XCTFail("could not retrieve post across Amplify re-configure")
  return
}
XCTAssertEqual(queriedPost.id, post.id)
```
you probably don't need to verify the post contents if we are already querying by Id

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+                print("Post saved successfully!")
+                saveToDatabaseDone.fulfill()
+            case .failure(let error):
+                print("Error saving post \(error)")
+            }
+        })
+        wait(for: [saveToDatabaseDone], timeout: 10)
+
+        ModelRegistry.reset()
+        Amplify.reset()
```
lawmicha, line 120: does `ModelRegistry` not get reset after Amplify reset? Amplify reset calls each plugins reset, DataStore should try to reset the Registry as well then, if it's not

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+                XCTFail("Database is not the same")
+            }
+        }
+
+        wait(for: [databaseRecreationDone], timeout: 10)
+
+        Amplify.DataStore.clear(completion: { _ in })
+    }
+
+    func testConfigWithDifferentSchema() throws {
```
lawmicha, line 88: same comments from above test to apply to this test

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+
+        let databaseCreationDone = expectation(description: "Database created")
+
+        do {
+            let dataStorePlugin = AWSDataStorePlugin(modelRegistration: AmplifyModels())
```
lawmicha, line 93: explicit usage of `version` parameter with a version, different from the next

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
```diff
@@ -0,0 +1,127 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+class DataStoreModelsVersionTests: XCTestCase {
```
lawmicha, line 14: usually we have the class name of the test correspond to the class name, so if this is another method on `SQLiteStorageEngineAdapter` then move the tests over to SQLLiteStorageEngineAdapterTests. you can also use `// MARK: <section name>` if the file is getting too big

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
```diff
@@ -0,0 +1,127 @@
...
+
+    func testVersionIsEmpty() {
```
lawmicha, line 16: maybe something like `func testClearIfNewVersionShouldStoreInitialVersion`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
```diff
@@ -0,0 +1,127 @@
...
+        let schemaUpdateInvoked = expectation(description: "Version has updated and old database is deleted")
+
+        guard let userDefaults = UserDefaults.init(suiteName: "testVersionIsEmpty") else {
+            XCTFail("Could not create a UserDafult with this suite name")
+            return
+        }
+
+        let newVersion = "newVersion"
+
+        let mockFileManager = MockFileManager()
```
lawmicha, line 26: you can have an expectation with an inverted fulfill, so it fails if it fulfills
`removeItemInvoked.isInverted = true`

which means you are sure that the file is not deleted

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
```diff
@@ -0,0 +1,127 @@
...
+
+        _ = SQLiteStorageEngineAdapter.clearIfNewVersion(version: newVersion,
+                                                                  databaseName: "database",
+                                                                  userDefaults: userDefaults,
+                                                                  fileManager: mockFileManager)
```
lawmicha, line 31: indentation

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
```diff
@@ -0,0 +1,127 @@
...
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+class DataStoreModelsVersionTests: XCTestCase {
+
+    func testVersionIsEmpty() {
+        let schemaUpdateInvoked = expectation(description: "Version has updated and old database is deleted")
+
+        guard let userDefaults = UserDefaults.init(suiteName: "testVersionIsEmpty") else {
```
lawmicha, line 19: can you make sure `userDefaults` does not contain a version by either
- optimiscally deleting it, ie. `userDefaults.remove(key: SQLiteStorageEngineAdapter.dbVersionKey)`
- or, checking that userDefaults.string(forKey: SQLiteStorageEngineAdapter.dbVersionKey) is nil

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
```diff
@@ -0,0 +1,127 @@
...
+
+        _ = SQLiteStorageEngineAdapter.clearIfNewVersion(version: newVersion,
+                                                                  databaseName: "database",
+                                                                  userDefaults: userDefaults,
+                                                                  fileManager: mockFileManager)
+
+        XCTAssertEqual(userDefaults.string(forKey: SQLiteStorageEngineAdapter.dbVersionKey), newVersion)
+        schemaUpdateInvoked.fulfill()
+
+        wait(for: [schemaUpdateInvoked], timeout: 10)
```
lawmicha, line 36: looks like can be removed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
```diff
@@ -0,0 +1,127 @@
...
+
+        _ = UserDefaults.removeObject(userDefaults)
+    }
+
+    func testVersionRemainsSame() {
```
lawmicha, line 41: similar comments from above test for some clean up in this test

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/TestSupport/Mocks/MockFileManager.swift
```diff
@@ -0,0 +1,28 @@
...
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+enum MockFileManagerError: Error {
+    case failToDeleteDatabase
```
lawmicha, line 11: a file manager doesn't know about the database, so `removeItemError`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -119,7 +121,8 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         storageEngine = try StorageEngine(isSyncEnabled: isSyncEnabled,
                                           dataStoreConfiguration: dataStoreConfiguration,
                                           validAPIPluginKey: validAPIPluginKey,
-                                          validAuthPluginKey: validAuthPluginKey)
+                                          validAuthPluginKey: validAuthPluginKey,
+                                          modelSchemaVersion: modelRegistration.version)
```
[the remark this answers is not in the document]
DongQuanRui: changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -16,12 +16,11 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
 
     internal var connection: Connection!
     private var dbFilePath: URL?
+    static let dbVersionKey = "com.awsamazon.dataStore.dbVersion"
```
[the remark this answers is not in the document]
DongQuanRui: Changed, but in PictureThis the bundle identifier is "com.awsamazon.PictureThis" instead of "com.amazonaws.PictureThis". I guess no matter "awsamazon" or "amazonaws" doesn't matter?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -295,6 +302,34 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
...
+        if previousVersion == version {
+            return .emptyResult
+        }
+
+        log.verbose("\(#function) Warning: Schema change detected, removing your previous database")
+        do {
+            try fileManager.removeItem(at: dbFilePath)
+        } catch {
+            log.error("\(#function) Failed to delete database file located at: \(dbFilePath), error: \(error)")
+            return .failure(causedBy: error)
```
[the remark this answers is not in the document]
DongQuanRui: makes sense. changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -78,10 +78,21 @@ final class StorageEngine: StorageEngineBehavior {
...
+
         let key = kCFBundleNameKey as String
-        let databaseName = Bundle.main.object(forInfoDictionaryKey: key) as? String
-        let storageAdapter = try SQLiteStorageEngineAdapter(databaseName: databaseName ?? "app")
+        let databaseName = Bundle.main.object(forInfoDictionaryKey: key) as? String ?? "app"
+
+        let result = SQLiteStorageEngineAdapter.clearIfNewVersion(version: modelSchemaVersion,
+                                                                  databaseName: databaseName)
+        if case let .failure(error) = result {
+            StorageEngine.log.error("Failed to delete database file called: \(databaseName), error: \(error)")
```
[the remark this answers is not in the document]
DongQuanRui: changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -78,10 +78,21 @@ final class StorageEngine: StorageEngineBehavior {
     convenience init(isSyncEnabled: Bool,
                      dataStoreConfiguration: DataStoreConfiguration,
                      validAPIPluginKey: String = "awsAPIPlugin",
-                     validAuthPluginKey: String = "awsCognitoAuthPlugin") throws {
+                     validAuthPluginKey: String = "awsCognitoAuthPlugin",
+                     modelSchemaVersion: String,
+                     userDefault: UserDefaults = UserDefaults.standard) throws {
+
```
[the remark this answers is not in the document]
DongQuanRui: Added

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+//
+//  DataStoreConfigurationTests.swift
+//  AWSDataStoreCategoryPluginIntegrationTests
+//
+//  Created by Guo, Rui on 6/15/20.
+//  Copyright © 2020 Amazon Web Services. All rights reserved.
+//
```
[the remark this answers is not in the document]
DongQuanRui: removed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -78,10 +78,21 @@ final class StorageEngine: StorageEngineBehavior {
...
         let key = kCFBundleNameKey as String
-        let databaseName = Bundle.main.object(forInfoDictionaryKey: key) as? String
-        let storageAdapter = try SQLiteStorageEngineAdapter(databaseName: databaseName ?? "app")
+        let databaseName = Bundle.main.object(forInfoDictionaryKey: key) as? String ?? "app"
+
+        let result = SQLiteStorageEngineAdapter.clearIfNewVersion(version: modelSchemaVersion,
+                                                                  databaseName: databaseName)
+        if case let .failure(error) = result {
```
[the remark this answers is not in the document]
DongQuanRui: Changed, but not sure if it is the right way. Please take a look

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+            let dataStorePlugin = AWSDataStorePlugin(modelRegistration: AmplifyModels())
+            try Amplify.add(plugin: dataStorePlugin)
+            try Amplify.configure()
+            print("Amplify configured with DataStore plugin")
+        } catch {
+            print("Failed to initialize Amplify with \(error)")
+        }
+
+        let time: Temporal.DateTime = .now()
+        let post = Post(title: "title", content: "content", createdAt: time)
```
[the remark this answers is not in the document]
DongQuanRui: I need to make an assertion at line 68 so that I put .now() in a variable.

AmplifyPlugins/DataStore/DataStoreCategoryPlugin.xcodeproj/project.pbxproj
```diff
@@ -1549,6 +1563,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				2149E60C23886C7B00873955 /* ViewController.swift in Sources */,
+				D8D4E09A2495DE57005594A3 /* SQLiteStorageEngineAdapterTests.swift in Sources */,
```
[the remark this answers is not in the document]
DongQuanRui: unchecked

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: changed, plz take a look

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: removed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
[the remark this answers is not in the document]
DongQuanRui: I see what you meant here. I use post.time at line 68

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Right, removed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Then let me just use networkTimeout

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: true

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
DongQuanRui: Moved to `SQLiteStorageEngineAdapterTests.swift`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
DongQuanRui: Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
DongQuanRui: I may prefer the first option, changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -62,6 +61,14 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
         try connection.execute(databaseInitializationStatement)
     }
 
+    static func getDbFilePath(databaseName: String) -> URL {
+        guard let documentsPath = getDocumentPath() else {
+            preconditionFailure("Could not create the database. The `.documentDirectory` is invalid")
+        }
+        let dbFilePath = documentsPath.appendingPathComponent("\(databaseName).db")
+        return dbFilePath
```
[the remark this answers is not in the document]
DongQuanRui: changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,150 @@
...
+import AWSMobileClient
+import AWSPluginsCore
+
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+class DataStoreConfigurationTests: XCTestCase {
+
+    func testConfigWithSameSchema() throws {
```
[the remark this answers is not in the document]
DongQuanRui: Added a comment before query

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
DongQuanRui: Changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Ya, the ModelRegistry.reset() got called inside DataStore.reset(), I'll remove ModelRegistry.reset()

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
DongQuanRui: Sorry, I don't understand here, one comment below says to remove the wait(for: [schemaUpdateInvoked], timeout: 10), so that expectation is not needed here?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/TestSupport/Mocks/MockFileManager.swift
DongQuanRui: Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,137 @@
...
+import AWSPluginsCore
+
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+class DataStoreConfigurationTests: XCTestCase {
+
+    func testConfigureWithSameSchemaDoesNotDeleteDatabase() throws {
+        Amplify.reset()
```
lawmicha, line 21: can move this to common setUp method

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,137 @@
...
+
+        let prevoisVersion = "123"
+
+        let saveSuccess = expectation(description: "Save was successful")
+
```
lawmicha, line 26: nit: you can reduce the line number where it makes sense, like within the "arrange' part of the test, you probably don't need a new line between these variables, and then add a new line between the "arrange" part and "act" part, and a newline between the "act" and "assert" sections

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,137 @@
...
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+class DataStoreConfigurationTests: XCTestCase {
+
+    func testConfigureWithSameSchemaDoesNotDeleteDatabase() throws {
+        Amplify.reset()
+
+        let prevoisVersion = "123"
```
lawmicha, line 23: `let previousVersion = "previousVersion"`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,137 @@
...
+            case .failure(let error):
+                XCTFail("Error saving post \(error)")
+            }
+        })
+
+        wait(for: [saveSuccess], timeout: TestCommonConstants.networkTimeout)
+
+        Amplify.reset()
+
+        let querySuccess = expectation(description: "Database remains")
```
lawmicha, line 50: "query was successful"

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,137 @@
...
+
+        do {
+            let dataStorePlugin = AWSDataStorePlugin(modelRegistration: AmplifyModels(version: prevoisVersion))
+            try Amplify.add(plugin: dataStorePlugin)
+            try Amplify.configure(AmplifyConfiguration(dataStore: nil))
+        } catch {
+            XCTFail("Failed to initialize Amplify with \(error)")
+        }
+
+        // Here trying to query the saved data from above, because if the schema hasn't been changed that the database and saved data should persist.
```
lawmicha, line 60: suggestion
```
// Query for the previously saved post. Data should be retrieved successfully which indicates that 
// the database file was not deleted after re-configuring Amplify when using the same model registry version
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,137 @@
...
+        Amplify.DataStore.query(Post.self, byId: post.id) { result in
+            switch result {
+            case .success(let postOptional):
+                querySuccess.fulfill()
```
lawmicha, line 64: fulfill should come at the end or else the wait till be released and finish executing at the same time as the XCTAsserts u have below the fulfill

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,137 @@
...
+                guard let queriedPost = postOptional else {
+                    XCTFail("could not retrieve post across Amplify re-configure")
+                    return
+                }
+                XCTAssertEqual(queriedPost.title, "title")
+                XCTAssertEqual(queriedPost.content, "content")
+                XCTAssertEqual(queriedPost.createdAt, post.createdAt)
+            case .failure(let error):
+                XCTFail("Database is not the same, this shouldn't happen, error: \(error)")
```
lawmicha, line 73: Database is not the same? i think just `XCTFail("Failed to query post, error: \(error)")` is enough

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
lawmicha: - testConfigureWithDifferentSchemaClearsDatabase
- previousVersion typo
- nit: reduce new lines

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,137 @@
...
+            let dataStorePlugin = AWSDataStorePlugin(modelRegistration: AmplifyModels(version: "1234"))
+            try Amplify.add(plugin: dataStorePlugin)
+            try Amplify.configure(AmplifyConfiguration(dataStore: nil))
+        } catch {
+            XCTFail("Failed to initialize Amplify with \(error)")
+        }
+
+        Amplify.DataStore.query(Post.self) { result in
+            switch result {
+            case .success(let postRes):
```
lawmicha, line 124: nit: `postRes` -> `postResult`, `postOptional` ?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,137 @@
...
+                XCTAssertTrue(postRes.isEmpty)
+                querySuccess.fulfill()
+            case .failure(let error):
+                XCTFail("Database Recreation Failed due to \(error)")
```
lawmicha, line 128: remove "Database Recreation Failed" message from a failed to query callback

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -375,4 +375,120 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
             XCTFail(String(describing: error))
         }
     }
+
+    func testClearIfNewVersionShouldStoreInitialVersion() {
+
```
lawmicha, line 380: nit: remove newline

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -375,4 +375,120 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
             XCTFail(String(describing: error))
         }
     }
+
+    func testClearIfNewVersionShouldStoreInitialVersion() {
```
lawmicha, line 379: - no longer storing initial version, the scenario is "empty user defaults" like `testClearIfNewVersionWithEmptyUserDefaults`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -375,4 +375,120 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
...
+
+        guard let userDefaults = UserDefaults.init(suiteName: "testVersionIsEmpty") else {
+            XCTFail("Could not create a UserDafult with this suite name")
+            return
+        }
+        userDefaults.removeObject(forKey: SQLiteStorageEngineAdapter.dbVersionKey)
+
+        let newVersion = "newVersion"
+
+        let mockFileManager = MockFileManager()
```
lawmicha, line 389: ensure filemanager removeItem is not called,
one way to achieve this is 
```
fileManager.removeItem = {
 XCTFail("Should not have called removeItem")
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -375,4 +375,120 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
...
+                                                             userDefaults: userDefaults,
+                                                             fileManager: mockFileManager)
+        } catch {
+            XCTFail("Test has failed due to \(error)")
+        }
+
+        _ = UserDefaults.removeObject(userDefaults)
+    }
+
+    func testVersionInUserDefaultsRemainsSameAndFileExists() {
```
lawmicha, line 403: test name like `testClearIfNewVersionWithVersionSameAsPrevious`

```
let previousVersion = "previousVersion"
userDefaults.set(previousVersion ..

try SQLiteStorageEngineAdaapter.clearIfNewVersion(previousVersion,

```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -375,4 +375,120 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
...
+            XCTFail("Could not create a UserDafult with this suite name")
+            return
+        }
+
+        userDefaults.set("previousVersion", forKey: SQLiteStorageEngineAdapter.dbVersionKey)
+
+        let newVersion = "previousVersion"
+
+        let mockFileManager = MockFileManager()
+        mockFileManager.fileExists = true
```
lawmicha, line 416: if fileExists then make sure removeItem is called

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -375,4 +375,120 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
...
+                                                             dbFilePath: URL(string: "dbFilePath")!,
+                                                             userDefaults: userDefaults,
+                                                             fileManager: mockFileManager)
+        } catch {
+            XCTFail("Test has failed due to \(error)")
+        }
+
+        fileExistsInvoked.fulfill()
+
+        wait(for: [fileExistsInvoked], timeout: 10)
```
lawmicha, line 429: what is this fulfill for?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -375,4 +375,120 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
...
+
+        _ = UserDefaults.removeObject(userDefaults)
+    }
+
+    func testVersionInUserDefaultsRemainsSameButFileDoesNotExist() {
```
lawmicha, line 434: `testClearIfNewVersionWithMissingFile`
- same set up as `testClearIfNewVersionWithVersionSameAsPrevious` except your fileExists = false
- make sure fileManager.removeItem is not called

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -375,4 +375,120 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
...
+        }
+
+        fileNotExistsInvoked.fulfill()
+
+        wait(for: [fileNotExistsInvoked], timeout: 10)
+
+        _ = UserDefaults.removeObject(userDefaults)
+    }
+
+    func testVersionInUserDefaultsUpdateFail() {
```
lawmicha, line 464: `testClearIfNewVersionFailure`
- previous != new so you can get pass the first check
- mock fileExist = true
- mock hasError = true
- catch and assert the error that you expected to throw, what type of DataStore error? fulfill not needed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreModelsVersionTests.swift
lawmicha: hmm this comment is outdated now since code has been changed, let me try to explain again in the new code

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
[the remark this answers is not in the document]
lawmicha: oh good point, not sure which is the right one, more important for the codebase to be consistent regardless of PictureThis's bundle since it's external sample app

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Sorry, do you mean only putting this Amplify.reset() in another function maybe called setup() and use it in these two tests?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: True. I toke a look at other test files and it looks better. Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Updated

DongQuanRui: I also put a comment on the second test in this file

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: I see. I thought it is going to be fulfill first and then check the retrieved data

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Ya, it sounds a bit weird. Changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: I changed it to postResult for Both tests

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
DongQuanRui: Removed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
DongQuanRui: En, this is good

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
DongQuanRui: Sorry, I always has trouble thinking about name of test function

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
DongQuanRui: I changed to using this `XCTAssertEqual(error as? MockFileManagerError, MockFileManagerError.removeItemError)` inside catch block

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
DongQuanRui: Actually, here is having an additional var necessary?
`userDefaults.set("previousVersion", forKey: SQLiteStorageEngineAdapter.dbVersionKey)` is OK? Because "previousVersion" is used only once.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
DongQuanRui: I kind of think it's bit weird to have 
`
userDefaults.set("previousVersion", forKey: SQLiteStorageEngineAdapter.dbVersionKey)
let newVersion = "newVersion"
`
because `try SQLiteStorageEngineAdapter.clearIfNewVersion(version: newVersion,
                                                             dbFilePath: URL(string: "dbFilePath")!,
                                                             userDefaults: userDefaults,
                                                             fileManager: mockFileManager)` is not setting newVersion to userDefaults

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Hmmm, but is this necessary?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
lawmicha: can u elaborate on what you mean? the method isn't setting `newVersion` to userDefaults,  it's just clearing if required, or clearing if version is different from what is stored. are u thinking the method should be named differently? 

in terms of a test, it just means that we expect database to be deleted when version differs, and if for some reason it was terminated and restarted then, version still differs (because newVersion is never set) then it will do no-op since file does not exist.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
DongQuanRui: Hmm, I just thought that for every test should have a fulfill but found out that it is not necessary here. Removed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -375,4 +375,108 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
...
+        mockFileManager.hasError = true
+        mockFileManager.fileExists = true
+
+        do {
+            try SQLiteStorageEngineAdapter.clearIfNewVersion(version: newVersion,
+                                                             dbFilePath: URL(string: "dbFilePath")!,
+                                                             userDefaults: userDefaults,
+                                                             fileManager: mockFileManager)
+        } catch {
+            XCTAssertEqual(error as? MockFileManagerError, MockFileManagerError.removeItemError)
```
lawmicha, line 477: assert is DataStoreError.invalidDatabase

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -295,6 +314,33 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
...
+        guard fileManager.fileExists(atPath: dbFilePath.path) else {
+            return
+        }
+
+        log.verbose("\(#function) Warning: Schema change detected, removing your previous database")
+        do {
+            try fileManager.removeItem(at: dbFilePath)
+        } catch {
+            log.error("\(#function) Failed to delete database file located at: \(dbFilePath), error: \(error)")
+            throw error
```
lawmicha, line 340: wrap in `DataStoreError.invalidDatabase(path: dbFilePath.absoluteString, error)`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -375,4 +375,108 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
...
+                                                             fileManager: mockFileManager)
+        } catch {
+            XCTFail("Test failed due to \(error)")
+        }
+
+        _ = UserDefaults.removeObject(userDefaults)
+    }
+
+    func testClearIfNewVersionFailure() {
+        guard let userDefaults = UserDefaults.init(suiteName: "testVersionHasChanged") else {
```
lawmicha, line 459: suiteName is inconsistent with test method name, here and other tests as well

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -295,6 +314,33 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
...
+        }
+
```
lawmicha, line 342: nit: remove newline

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
DongQuanRui: Oh, sorry. I didn't think into this test very clearly when I replied. There is no problem now. 

But in terms of the first bullet in this conversation `previous != new so you can get pass the first check`
what do you mean by this? Could you explain a bit on it?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
lawmicha: the setUp() and tearDown() methods are part of `XCTestCase` and runs for each test method. you don't have to call `setUp()` in your test method and is done automatically for you as part of the test execution. for each test method, it runs setUp(), then your test method, then tearDown(). it is the ideal place to reset the state of the system for your tests

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
```diff
@@ -0,0 +1,138 @@
...
+                XCTAssertEqual(queriedPost.createdAt, post.createdAt)
+                querySuccess.fulfill()
+            case .failure(let error):
+                XCTFail("Failed to query post, error: \(error)")
+            }
+        }
+
+        wait(for: [querySuccess], timeout: TestCommonConstants.networkTimeout)
+
+        Amplify.DataStore.clear(completion: { _ in })
```
lawmicha, line 79: can be moved to tearDown

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
lawmicha: i just meant to arrange the test so that the version in user defaults is different from the version passed in, you are already doing this

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
DongQuanRui: Changed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: What about 
`
func reset() {
    Amplify.reset()
}
`
and
`
func tearDown() {
    Amplify.DataStore.clear(completion: { _ in })
}
`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
DongQuanRui: Removed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
DongQuanRui: Done

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConfigurationTests.swift
DongQuanRui: Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
DongQuanRui: Done

PR #553, opened 2020-06-16
Update CHANGELOG.md for 1.0.2

lawmicha:
*Issue #, if available:*

*Description of changes:*
- Updates for 1.0.2 release changelog
- announcement about Amplify CLI breaking change


CHANGELOG.md
```diff
@@ -5,12 +5,17 @@ All notable changes to this project will be documented in this file. See [standa
...
 
 ### Bug Fixes
 
 * **DataStore:** Fixed a DataStore issue where lazy `List<M>` initialization would fail for relationships 3+ levels deep ([#534](https://github.com/aws-amplify/amplify-ios/pull/534))
-* **Tools:** Update Amplify tools script to resolve node correctly when NVM is installed ([#524](https://github.com/aws-amplify/amplify-ios/pull/524))
+* **DataStore/API:** Add Emeddable type to store schema info for custom types ([#539](https://github.com/aws-amplify/amplify-ios/pull/539))
+* **Tools:** Update Amplify tools script to resolve node correctly when NVM is installed ([#535](https://github.com/aws-amplify/amplify-ios/pull/535))
+* **Auth** Signout will completely delete the session in webui ([#542](https://github.com/aws-amplify/amplify-ios/pull/542))
+* **Core** Fix plugin configuration validation ([#543](https://github.com/aws-amplify/amplify-ios/pull/543))
+* **Build** Add common dependency configuration, standardize environment across all modules, podspec source version tag and url ([#538](https://github.com/aws-amplify/amplify-ios/pull/538))
```
palpatim, line 18: This should be under a miscellaneous section rather than "Bug Fixes"

CHANGELOG.md
```diff
@@ -5,12 +5,17 @@ All notable changes to this project will be documented in this file. See [standa
 ## Unreleased
 
 ### ⚠ BREAKING CHANGES
+* **DataStore/API** category using generated Model classes from Amplify CLI will require the minimum Amplify CLI version of 4.21.4. Current implementations do not expect the field types `embedded(type:)` and `embeddedCollection(of:)` that will break with compile time error. When upgrading Amplify Library, please upgrade Amplify CLI before running `amplify codegen models` to codegen the new `Embeddable` types.
```
palpatim, line 8: If this is a breaking change, we need a major version bump.

1. Is it really a breaking change?
1. If so, is it really critical to release a breaking change now?

lawmicha: i think this is a breaking change because:
the `customType` is not longer generated and is `embedded(type)` and `embeddedCollection(of)` which means if someone upgrades, and they have a schema which had the embedded type in a model, then the app will not build since 'customType` no longer exists in the ModelFieldType's.

without releasing this change means API with customTypes will not generate the correct selection set and deserialization will fail, currently impacting DataStore as well. The impacted is someone using an embedded type in their model.

CHANGELOG.md
lawmicha: moved

CHANGELOG.md
lawmicha: re-worked on the section regarding our decision to do a non-major version bump

CHANGELOG.md
```diff
@@ -2,15 +2,24 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-## Unreleased
+## 1.0.2 (2020-06-18)
 
 ### ⚠ BREAKING CHANGES
```
palpatim, line 7: This isn't a breaking change, move this as instructions under the bug fix below

CHANGELOG.md
```diff
@@ -2,15 +2,24 @@
...
-* **Tools:** Update Amplify tools script to resolve min CLI version for codegen changes ([#554](https://github.com/aws-amplify/amplify-ios/pull/554))
+* **DataStore/API** category using generated Model classes containing embedded types will require Amplify CLI 4.21.4 to generate working Model classes. This is to fix the bug in generating the incorrect selection set when using DataStore with sync enabled and API. The ModelFieldType `.customType` has been replaced with `embedded(type:)` and `embeddedCollection(of:)`. When upgrading to 1.0.2, please also upgrade Amplify CLI before running `amplify codegen models` to generate the new `Embeddable` types. We have considered this breaking change a non-major version bump as existing host applications will continue to experience the existing behavior after upgrading the library if they do not generate the new Model classes.
```
palpatim, line 8: Remove:
> We have considered this breaking change a non-major version bump as existing host applications will continue to experience the existing behavior after upgrading the library if they do not generate the new Model classes.

As discussed, this isn't a breaking change. To fix the bug update the CLI and the library. If you don't update the CLI and update the library, it doesn't break. There's a known incompatibility if someone updates the CLI and *not* the library, but that doesn't make this library release a breaking change.

Feel free to call out that updating the CLI must be accompanied by a library update, but again, not a breaking change in the lib

lawmicha: my bad, gotcha. removing "We have considered..." note and improving wording

> Add Emeddable type to store schema info for custom types ([#539](https://github.com/aws-amplify/amplify-ios/pull/539)). This bug impacts developers with schemas containing embedded types when using DataStore with sync to cloud enabled or using API with Model classes. To fix this bug, upgrade both Amplify CLI to 4.21.4 and Amplify Libraries to 1.0.2. There is a known incompatibility if only upgrading CLI but not the Library. Then run `amplify codegen models` to regenerate the Model classes. The ModelFieldType `.customType` has been replaced with `embedded(type:)` and `embeddedCollection(of:)`.

CHANGELOG.md
lawmicha: removing section

lawmicha:
pre-emptively adding https://github.com/aws-amplify/amplify-ios/pull/551 and https://github.com/aws-amplify/amplify-ios/pull/563 will merge after those are in

PR #562, opened 2020-06-18
fix: Adding back customType with deprecated message

lawmicha:
*Issue #, if available:*

*Description of changes:*
- Add a customType to allow compiling previous code generated code by Amplify CLI versions < 4.21.4

### DIFF
--- a/Amplify/Categories/DataStore/Model/Schema/ModelSchema+Definition.swift
+++ b/Amplify/Categories/DataStore/Model/Schema/ModelSchema+Definition.swift
@@ -34,10 +34,10 @@ public enum ModelFieldType {
     }
 
     @available(*, deprecated, message: """
-        This has been replaced with `.embedded(type)` and `.embeddedCollection(of)`
+        This has been replaced with `.embedded(type)` and `.embeddedCollection(of)` \
         Please use Amplify CLI 4.21.4 or newer to re-generate your Models to conform to Embeddable type.
     """)
-    public static func customType(type: Codable.Type) -> ModelFieldType {
+    public static func customType(_ type: Codable.Type) -> ModelFieldType {
         return .embedded(type: type)
     }
 
--- a/Amplify/Categories/DataStore/Model/Schema/ModelSchema+Definition.swift
+++ b/Amplify/Categories/DataStore/Model/Schema/ModelSchema+Definition.swift
@@ -33,6 +33,14 @@ public enum ModelFieldType {
         }
     }
 
+    @available(*, deprecated, message: """
+        This has been replaced with `.embedded(type)` and `.embeddedCollection(of)`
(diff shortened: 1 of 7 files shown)


Amplify/Categories/DataStore/Model/Schema/ModelSchema+Definition.swift
```diff
@@ -33,6 +33,14 @@ public enum ModelFieldType {
         }
     }
 
+    @available(*, deprecated, message: """
+        This has been replaced with `.embedded(type)` and `.embeddedCollection(of)`
```
palpatim, line 37: nit: add a trailing backslash to avoid line breaks: 
```swift
...and `.embeddedCollection(of)` \
Please use...
```

Podfile.lock
```diff
@@ -23,7 +23,7 @@ DEPENDENCIES:
   - SwiftLint
 
 SPEC REPOS:
-  https://cdn.cocoapods.org/:
+  trunk:
```
drochetti, line 26: I think the CDN is the right one. Something tells me you have an outdated setting in your Cocoapod installation.

AmplifyTestCommon/Models/NonModel/Todo+Schema.swift
```diff
@@ -33,7 +33,7 @@ extension Todo {
       .field(todo.name, is: .required, ofType: .string),
       .field(todo.description, is: .optional, ofType: .string),
       .field(todo.categories, is: .optional, ofType: .embeddedCollection(of: Category.self)),
-      .field(todo.section, is: .optional, ofType: .embedded(type: Section.self)),
+      .field(todo.section, is: .optional, ofType: .customType(type: Section.self)),
```
palpatim, line 36: Is this right? Shouldn't it be embedded?

lawmicha: I was using this to test that using `.customType` compiles. actually let me add another class entirely and put a comment on Amplify CLI version, and deprecated note

Podfile.lock
lawmicha: that's odd, tried with `pod repo remove trunk` and pod install on fresh project. trunk gets added to the repo list every time

AmplifyTestCommon/Models/NonModel/Todo+Schema.swift
lawmicha: updated

Podfile.lock
lawmicha: will remove this change for now to get this change in first, will take a look at this later when doing the release commit

palpatim: I believe this changes back & forth based on the local workstation's cocoapods setup, but it should be innocuous as long as we're all on sufficiently recent versions of CocoaPods (> 1.8.4 IIRC).

PR #571, opened 2020-06-21
feat(Hub): Added Combine support

palpatim:
Extended a few core types to support Combine subscriptions to Hub channels:

- `HubChannel` now conforms to `Hashable`
- Added `AtomicValue.with` method to provide read/write access to a block
- Made the `AWSHubPlugin.dispatchers` method internal, rather than private, to allow tests to validate that only one Hub listener per channel is created per Combine subscriber


AmplifyCombineSupport/AmplifyCombineSupport/Hub.swift
```diff
@@ -0,0 +1,47 @@
...
+    static var `default` = HubListenerMap()
+    var publishers = AtomicValue<[HubChannel: HubPublisher]>(initialValue: [:])
+}
+
+public extension HubCategoryBehavior {
+
+    /// Returns a publisher for all Hub messages on a particular channel
+    ///
+    /// - Parameter channel: The channel to listen for messages on
+    func publisher(for channel: HubChannel) -> HubPublisher {
```
lawmicha, line 26: is there a reason not to support `eventName` and `filter`?

palpatim: Rather than build the filter into the API, we'll let callers build up their desired streams using built-in Combine operators.

PR #575, opened 2020-06-23
feat: Add Combine support for Amplify APIs

palpatim:
Add APIs that return Combine publishers instead of accepting callbacks


jamesonwilliams:
The README.md is top-notch!! Maybe we can use it as a template, for the Android Rx Bindings.

Only design difference I noticed is that the Rx Bindings do return a `Cancellable` (née, `Disposable, in Rx). It wraps the various operations that the framework returns for some APIs.

Rx Bindings return `Observable` for streams, `Single` for single-return emitters, and `Completable` in some cases, where there is no value on success. I guess [the Combine equivalents of these](https://github.com/CombineCommunity/rxswift-to-combine-cheatsheet) would be `Publisher`, `Future`, and ... `Future` with a void value?

palpatim:
> The README.md is top-notch!! Maybe we can use it as a template, for the Android Rx Bindings.

Yeah, I'm really hoping we can keep our in-module documentation up to the high bar set by the main doc site.

> Only design difference I noticed is that the Rx Bindings do return a `Cancellable` (née, `Disposable`, in Rx parlance). It wraps the various operations that the framework returns for some APIs.

Does the Rx `Disposable` propagate a signal to the producer to allow resource cleanup as well, or does it simply dispose of the subscription? One of the things I was concerned about was leaving the customer with no way to cancel an in-process operation. At the same time, I was very reluctant to tie cancellation of the *subscription* to cancellation of the *operation*. It would be very surprising for a customer who, say, starts a download operation, wires up a progress listener to a UI element, and then navigates away, to have the underlying download operation cancelled out from under them. But I couldn't see a good way without creating more custom types to give discoverable and intuitive operation cancellation from a stream publish.

> Rx Bindings return `Observable` for streams, `Single` for single-return emitters, and `Completable` in some cases, where there is no value on success. I guess [the Combine equivalents of these](https://github.com/CombineCommunity/rxswift-to-combine-cheatsheet) would be `Publisher`, `Future`, and ... `Future` with a void value?

That's it exactly. The table of sync/async vs. one/many values is:

image

That is: A Future is one asynchronous value; a Publisher is more than one asynchronous values. 


> The amount of code, ~7700 lines, is sadly about the same as the Rx Bindings. In our case, a lot of it was repetitive in its pattern. I wonder if there would be an opportunity to code-gen this kind of stuff, some day.
> 
> Investing in that code-gen could be a wasted effort if we only did it once. But if we have to do it for every single SDK someday, too....

I looked very briefly at this ([Sourcery](https://github.com/krzysztofzablocki/Sourcery), at @drochetti's suggestion), but couldn't quite work up the ambition to learn a new system to do this, although I acknowledge that it would have made generating the APIs, and tests thereof, easier. It turned out that a great deal of the work was in the test fixtures, which is where Sourcery would really shine, so it's definitely not off the table for a future enhancement.

jamesonwilliams:
> I was very reluctant to tie cancellation of the _subscription_ to cancellation of the _operation_. It would be very surprising for a customer who, say, starts a download operation, wires up a progress listener to a UI element, and then navigates away, to have the underlying download operation cancelled out from under them.

Ah yes, this is a good distinction, and one that I don't believe I make in the Rx Bindings for Android. Currently, the behavior of the Rx Bindings is: `.dispose()` == stop the subscription == stop the underlying operation. (I have [this utility code to convert between an Amplify `Cancelable` and an Rx `Disposable`](https://github.com/aws-amplify/amplify-android/blob/12ec8e9819228f63bd677a0431a5ab191f79d9e1/rxbindings/src/main/java/com/amplifyframework/rx/AmplifyDisposables.java#L40).)

To your earlier question -- `Observable`, `Single`, and `Completable` all do have an [`doOnDispose(() -> { /* do stuff */ })`](http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#doOnDispose-io.reactivex.functions.Action-) hook.

palpatim:
Closing and removing this from the 1.1.0 milestone in favor of https://github.com/aws-amplify/amplify-ios/pull/667 (which is still in process as of this comment, but will be where we track the bulk of the Combine work)

PR #577, opened 2020-06-23
refactor(Auth): Changed the Auth plugins user agent to use Amplify

royjit:
Adds AmplifyServiceConfiguration to Auth category. Now the user agent shows the amplify-io string when making Cognito service calls.

Rev 2
Updated all aws sdk version to 2.13.5


AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
```diff
@@ -36,4 +36,8 @@ public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
                          credentialsProvider: AWSCredentialsProvider) {
         super.init(region: regionType, credentialsProvider: credentialsProvider)
     }
+
+    public init(region regionType: AWSRegionType) {
+        super.init(region: regionType, credentialsProvider: nil)
+    }
```
palpatim, line 42: Do we need this? We already have an initializer that takes a `nil`, as used on L32 above.

AmplifyPlugins/Auth/Podfile.lock
```diff
@@ -26,7 +26,7 @@ PODS:
   - CwlCatchException (1.0.2)
   - CwlPreconditionTesting (1.1.1):
     - CwlCatchException
-  - SwiftFormat/CLI (0.44.13)
+  - SwiftFormat/CLI (0.44.14)
```
palpatim, line 29: Don't we also need to declare a dependency on AWSMobileClient `~> 2.13.5`?

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
```diff
@@ -55,6 +62,35 @@ extension AWSCognitoAuthPlugin {
         }
     }
 
+    func identityPoolServiceConfiguration(from authConfiguration: JSONValue) throws -> AmplifyAWSServiceConfiguration {
+        let regionKeyPath = "CredentialsProvider.CognitoIdentity.Default.Region"
+        guard case .string(let regionString) = authConfiguration.value(at: regionKeyPath) else {
+            let amplifyError = AuthError.configuration(
+                "Error configuring \(String(describing: self))",
+                """
+                Couldnot read Cognito identity pool information from the configuration
```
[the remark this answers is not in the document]
royjit: 1. Fixed
2. Updated the message

AmplifyPlugins/Auth/Podfile.lock
royjit: Updated

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
royjit: That initializer is overriden by the initilizer in line L35 that take a nonnull `credentialsProvider`. So I cannot init this class outside like this: `AmplifyAWSServiceConfiguration(region: region, credentialsProvider: nil)`

PR #588, opened 2020-06-26
ci: add conventional commits support & unstable releases

iartemiev:
* Add support for Conventional Commits by utilizing Fastlane actions from the [release_actions](https://github.com/aws-amplify/amplify-ci-support/pull/44) fastlane plugin @jpignata and I worked on.
  * This will automatically bump the next release version based on the Conventional Commits [spec](https://www.conventionalcommits.org/en/v1.0.0/#summary)
  * Generate a changelog and push it back to **main**

* Add an "unstable" pipeline job that publishes each merge into **main** under a prerelease version available to customers, e.g., `1.0.3-unstable.1`


.circleci/config.yml
```diff
@@ -7,11 +7,10 @@ defaults: &defaults
     xcode: "11.4.0"
   working_directory: ~/amplify-ios
   environment:
-      BUNDLE_PATH: vendor/bundle
+    BUNDLE_PATH: vendor/bundle
 
 references:
-  repo_cache_key: &repo_cache_key
-    v1-repo-{{ .Branch }}-{{ .Revision }}
+  repo_cache_key: &repo_cache_key v1-repo-{{ .Branch }}-{{ .Revision }}
```
jpignata, line 13: Will this work? My vague recollection of yaml is that the alias thing needs to be on its own line.

.circleci/config.yml
```diff
@@ -20,11 +19,9 @@ references:
...
         - v1-repo
 
-  pods_cache_key: &pods_cache_key
-    v3-dependency-pods-{{ checksum "~/amplify-ios/Podfile" }}
-  
-  pods_backup_cache_key: &pods_backup_cache_key
-    v3-dependency-pods
+  pods_cache_key: &pods_cache_key v3-dependency-pods-{{ checksum "~/amplify-ios/Podfile" }}
+
+  pods_backup_cache_key: &pods_backup_cache_key v3-dependency-pods
```
jpignata, line 24: ⬆️

.circleci/config.yml
```diff
@@ -149,19 +146,34 @@ jobs:
...
       - store_test_results:
           path: build/reports
-  
+
   deploy:
     <<: *defaults
+    parameters:
+      lane:
+        type: string
+        default: "unstable"
```
jpignata, line 155: I guess we're all comfortable with the "unstable" token across the three repos? I would have preferred edge or next or something, but I didn't get a vote. I'm furious.

Gemfile
```diff
@@ -3,4 +3,6 @@
 source 'https://rubygems.org'
 gem 'xcpretty'
 gem 'cocoapods'
-gem 'fastlane'
\ No newline at end of file
+gem 'fastlane'
+plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
```
jpignata, line 7: What a weird way to go. I guess Ruby folks gonna Ruby.

Gemfile.lock
```diff
@@ -1,7 +1,15 @@
+GIT
+  remote: https://github.com/aws-amplify/amplify-ci-support
+  revision: 25edc8a763327adda016ffa84ea607ca982f47e0
+  branch: master
+  glob: src/fastlane/release_actions/*.gemspec
+  specs:
+    fastlane-plugin-release_actions (0.1.0)
```
jpignata, line 7: Probably should bump this to 1.0.0 once we put it in service. I neglected to do this.

fastlane/Fastfile
```diff
@@ -1,49 +1,107 @@
...
-  {spec: "AWSPredictionsPlugin.podspec", sleep: false},
-  {spec: "AmplifyPlugins.podspec", sleep: false}
-]
+COCOAPODS_PUSH_WAIT_TIME = 600
 
 platform :ios do
+
+  desc "Create a pre-release version by pushing a tag to GitHub, and pushing pods to CocoaPods"
+  lane :unstable do 
+
```
jpignata, line 14: Nittiest of nits: You may want to lose this whitespace about the block open line as that makes it harder to see a new block.

fastlane/Fastfile
```diff
@@ -1,49 +1,107 @@
...
+    next_version = calculate_next_canary_version
+
+    UI.message("Release unstable version: #{next_version}")
```
jpignata, line 17: Should this be "Releasing.." -- trying to think of what I'd expect to see as a reader of the logs.

fastlane/Fastfile
```diff
@@ -1,49 +1,107 @@
...
+  end
+  
+  desc "Commit and push"
+  private_lane :release_commit do |options|
+    next_version = options[:version]
+
+    sh("git config --global user.email $GITHUB_EMAIL")
+    sh("git config --global user.name $GITHUB_USER")
+
+    commit_message = "chore(release): #{next_version} [skip ci]"
```
jpignata, line 79: Just Bikeshedding on a Friday: Why not just `release: `? Not sure the chore type buys us much, and release is a weird scope.

fastlane/Fastfile
```diff
@@ -1,49 +1,107 @@
...
+    sh("git commit -am '#{commit_message}'")
+    sh("git push origin")
+  end
+
+  desc "Tag in git and push to GitHub"
+  private_lane :add_tag do |options|
+    next_version = options[:version]
+
+    add_git_tag(tag: next_version)
```
jpignata, line 88: TODO: We'll make this better by using the GitHub API to post our changelog in a future iteration.

fastlane/Pluginfile
```diff
@@ -0,0 +1,5 @@
+# Autogenerated by fastlane
+#
+# Ensure this file is checked in to source control!
+
+gem 'fastlane-plugin-release_actions', git: 'https://github.com/aws-amplify/amplify-ci-support', branch: 'master', glob: 'src/fastlane/release_actions/*.gemspec'
```
jpignata, line 5: Hmm, should we lock to a SHA?

.circleci/config.yml
iartemiev: VS Code kindly formatted that file for me, hence the large number of superficial changes (:

But it will work. I validated the config with `circleci config validate`, which checks ensures that the yaml in the config is sound.

Gemfile
iartemiev: Haha yep, autogenerated courtesy of Fastlane

fastlane/Fastfile
iartemiev: `release:` is technically not a valid type (to the best of my knowledge). I went with what we have on JS but can change it to e.g., `chore: release  #{next_version} [skip ci]`

jpignata: There's actually nothing in the specification that prevents you from using your own nouns for types. The only defined types are `fix` and `feat`. The [specification](https://www.conventionalcommits.org/en/v1.0.0/#specification):

> 4. types other than fix: and feat: are allowed, for example @commitlint/config-conventional (based on the the Angular convention) recommends build:, chore:, ci:, docs:, style:, refactor:, perf:, test:, and others.

Commitlint apparently has a default list, but we should feel like we can expand upon this. The specification only concerns itself with `feat` and `fix`.

iartemiev: That's fair, I should've phrased that differently. It's not invalid per se, just not part of the convention

jpignata: I'm being a pedant here, but bear with me. The scope is defined as:

> 4. A scope MAY be provided after a type. A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis, e.g., fix(parser):

Release isn't technically a scope by the above definition. It seems like more of a type. I'm surprised this isn't more commonly used.

Gemfile.lock
jpignata: Done.

fastlane/Fastfile
iartemiev: I agree with your point that `release` is nonsensical as a scope. So I'm suggesting dropping the scope and putting "release" in the description, following a **chore** type.

E.g., `chore: release #{next_version} [skip ci]`

fastlane/Pluginfile
iartemiev: It is in the Gemfile.lock: https://github.com/aws-amplify/amplify-ios/blob/2edfd7bcbe44a68bf5b5f1d94415bb947131fe53/Gemfile.lock#L3

I didn't want to set a `ref` in the Pluginfile in order to allow updating easily with `bundle update` as we release new revisions of the plugin

Gemfile.lock
iartemiev: Gem has been updated

fastlane/Pluginfile
jpignata: My concern is if we update master with a breaking change, that would break the pipelines. Since it's a manual action to bundle, and this isn't a ruby project, it should be fine.

jpignata: Ignore, discussed offline, this only happens on `bundle update`.

fastlane/Fastfile
```diff
@@ -1,49 +1,105 @@
...
-pods = [
-  {spec: "Amplify.podspec", sleep: true},
-  # AmplifyPlugins depends on AWSPluginsCore, but skipping the sleep here
-  # since there's one after CoreMLPredictionsPlugin
-  {spec: "AWSPluginsCore.podspec", sleep: false},
-  {spec: "CoreMLPredictionsPlugin.podspec", sleep: true},
-  {spec: "AWSPredictionsPlugin.podspec", sleep: false},
-  {spec: "AmplifyPlugins.podspec", sleep: false}
-]
+COCOAPODS_PUSH_WAIT_TIME = 600
```
palpatim, line 8: Is this still necessary? I recall we were going to try the new `--synchronous` flag on our pod pushes?

fastlane/Fastfile
```diff
@@ -1,49 +1,105 @@
...
 
 platform :ios do
+  desc "Create a pre-release version by pushing a tag to GitHub, and pushing pods to CocoaPods"
+  lane :unstable do
```
palpatim, line 12: I don't see an entry in `.circleci/config.yml` to trigger this lane. Is that coming in a future PR, or is it added in a way that doesn't reference the token, `"unstable"`?

fastlane/Fastfile
```diff
@@ -1,49 +1,105 @@
...
+        set_info_plist_value(path: plist_path, key: PLIST_KEY, value: version)
+      end
+    end
+  end
+  
+  desc "Commit and push"
+  private_lane :release_commit do |options|
+    next_version = options[:version]
+
+    sh("git config --global user.email $GITHUB_EMAIL")
```
palpatim, line 74: Here and throughout: use the array version of `sh` to avoid shell execution of potentially unsafe user input

fastlane/Fastfile
```diff
@@ -1,49 +1,105 @@
...
+
       if pod[:sleep]
         UI.message("Sleeping for #{COCOAPODS_PUSH_WAIT_TIME}s")
         sleep COCOAPODS_PUSH_WAIT_TIME
       end
-      
-      Actions.sh('pod repo update')
-    }
 
+      Actions.sh('pod repo update')
```
palpatim, line 103: out of curiosity, is it necessary to specify the `Actions` namespace here? I note that we don't do it above

fastlane/Fastfile
iartemiev: I just opened a PR for that fastlane action, adding the synchronous option: https://github.com/fastlane/fastlane/pull/16698

Of course, we can just use a bash command until that gets merged

fastlane/Fastfile
iartemiev: It's been added: https://github.com/aws-amplify/amplify-ios/pull/588/files#diff-1d37e48f9ceff6d8030570cd36286a61R163

I parameterized the deploy job. It passes unstable by default. For the full release, we're passing `release`, i.e., https://github.com/aws-amplify/amplify-ios/pull/588/files#diff-1d37e48f9ceff6d8030570cd36286a61R245

I can take out the default if that'll make it more straightforward

fastlane/Fastfile
iartemiev: Good call, `Actions` isn't necessary

fastlane/Fastfile
iartemiev: update: I've removed the constant and `sleep` logic. Calling `pod trunk push` with `sh` will update to action once someone on the fastlane team merges my PR

fastlane/Fastfile
palpatim: No, parameters are a fine way to do this, I just didn't catch that

PR #596, opened 2020-06-29
Fix: DataStore E2E Integration Tests

DongQuanRui:
*Description of changes:*
Added a filter inside `testCreateMutateDelete()`, `testCreateThenMutateWithCondition()` and `testCreateThenMutateWithConditionFailOnSync()` to make sure these three tests only care about related payload receiving from hub listener.


AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
```diff
@@ -45,8 +45,10 @@ class DataStoreEndToEndTests: SyncEngineIntegrationTestBase {
                         return
                 }
 
+                if post.id != newPost.id { return }
```
lawmicha, line 48: `let post = try? mutationEvent.decodeModel() as? Post` above at line 42 will fail the test if there is an existing mutation in the queue and is not a post. if we are enforcing that the hub listener only reacts to data for this test then it should include decoding the right model as well, ie.
````
guard if let post = try? mutationEvent.decodeModel() as? Post, post.id != newPost.id else {
   return 
}
```

DongQuanRui: Something like this?
`
guard let mutationEvent = payload.data as? MutationEvent
                    else {
                        XCTFail("Can't cast payload as mutation event")
                        return
                }

                guard let post = try? mutationEvent.decodeModel() as? Post, post.id != newPost.id else { return }
`

palpatim: I agree with the intent of this proposal, but a couple of things:

1. You need to fix the `id` equality check:
    ```swift
    guard let post = try? mutationEvent.decodeModel() as? Post, post.id == newPost.id else {
        return
    }
    ```
    Otherwise you'll always be guaranteed to check the wrong post.

2. Please add a comment explaining that this check is to protect against stray events being processed after the test has completed, and that it shouldn't be construed as a pattern necessary for production applications.

PR #603, opened 2020-07-01
ci: fastlane release lane fixes

iartemiev:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Changelog
 
-All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
-
 ## 1.0.3 (2020-06-26)
 
 ### Bug Fixes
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,5 @@
 # Changelog
 
-All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
-
```
iartemiev, line 4: @palpatim since we're auto-generating the changelog by prepending updates to the top of this file, we thought it would make sense to move this notice to the contributing guidelines and/or readme.

palpatim: I'm expecting that we'll have the opportunity to edit the changelog without the release process overwriting our changes, in cases where the commits messages don't properly reflect the intent, and also to provide friendlier "release summaries" so we can call out important pieces of a release without making customers wade through the entire commit history of a release.

As long as that assumption is correct, LGTM.

iartemiev: We can definitely edit the Changelog at will after a release and then commit/push the changes back to main. 

The release pipeline doesn't mess with the existing content of the Changelog, just appends new auto-generated changes to the top of the file.

PR #610, opened 2020-07-01
ci(fastlane): add retry logic for pod push and missing versioning logic

iartemiev:
* Add retry logic for pod push:
  * Retry 5 times if `pod push` trunk encountered a server error
 
* Add auto-versioning for:
  * AmplifyTestCommon.podspec
  * AmplifyAWSServiceConfiguration.swift
  * build-support/dependencies.rb

* Run `pod update` for each plugin to update their respective Podfile.lock files


fastlane/Fastfile
```diff
@@ -111,4 +117,28 @@ platform :ios do
...
+        break
+      rescue => ex
+        unless ex.to_s.include? server_error
+          UI.error "Unrecoverable error in pod push trunk"
+          raise Exception.new ex
+        end
+        i = i + 1
+      end
+    end
+  end
```
jpignata: I'd suggest generalizing this.

```ruby
def with_retry(retries=5, wait=0.2)
  begin
    yield
  rescue Exception => e
    retries -= 1
    raise e if retries.zero?
    UI.error("Error occurred: #{exception}; retrying...")
    sleep(wait)
    retry
  end
```

We can use this as:

```ruby
with_retry do
  sh('bundle', 'exec', 'pod', 'trunk', 'push', pod_path, '--allow-warnings', '--synchronous')
end
```

I'd be tempted to drop the matching of error messages. These are liable to change. Might as well just retry in the face of errors.

jpignata: If we need to maintain the error message check, I'd suggest:

```ruby
with_retry do
  begin:
    sh('bundle', 'exec', 'pod', 'trunk', 'push', pod_path, '--allow-warnings', '--synchronous')
  rescue Exception => exception
    raise unless exception.message =~ /duplicate entry/i
    UI.warn("Version already exists (or something idk")
  end
end
```

.circleci/config.yml
```diff
@@ -44,8 +44,8 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - v1-gems-{{ checksum "~/amplify-ios/Gemfile.lock" }}
-            - v1-gems-
+            - v2-gems-{{ checksum "~/amplify-ios/Gemfile.lock" }}
+            - v2-gems-
```
palpatim, line 48: is the `v2-gems-` key provided by CircleCI? If not, where is the content for that cache saved?

fastlane/Fastfile
```diff
@@ -89,17 +95,20 @@ platform :ios do
   desc "Release pods"
   private_lane :release_pods do
     AmplifyPods.pods.each do |pod|
-      pod_path = "/Users/distiller/amplify-ios/#{pod[:spec]}"
+      if pod[:no_push]
+        next
```
palpatim, line 99: Nit: add a UI.message about skipping this pod, for transparency

fastlane/Fastfile
```diff
@@ -89,17 +95,20 @@ platform :ios do
...
-        sh('bundle', 'exec', 'pod', 'trunk', 'push', pod_path, '--allow-warnings', '--synchronous')
-      rescue => ex
-        unless ex.to_s.include? dupe_error
-          raise Exception.new "Unrecoverable error in pod push trunk"
+      with_retry do
+        begin
+          sh('bundle', 'exec', 'pod', 'trunk', 'push', pod_path, '--allow-warnings', '--synchronous')
+        rescue Exception => exception
+          raise unless exception.message =~ /Unable to accept duplicate entry/i
+          UI.warn("Version already exists. Ignoring")
```
palpatim, line 111: nit: include the exception in the message to help potential troubleshooting efforts

fastlane/Fastfile
```diff
@@ -111,4 +120,37 @@ platform :ios do
...
     end
   end
+
+  desc "Post-release"
+  private_lane :post_release do |options|
+    version = options[:version].to_s
+    # increment plugin lock files by calling pod update on each
+    plugin_root = '/Users/distiller/amplify-ios/AmplifyPlugins'
+  
+    sh('bundle', 'exec', 'pod', 'repo', 'update')
```
palpatim, line 130: I'm confused by the sequence:

- We begin with the repo version strings at 1.0.0, with Podfile.lock at 1.0.0
- We bump repo versions to 1.0.1
- We release the pods at tag `v1.0.1`; but with Podfile.lock still pointing to 1.0.0
- We bump the Podfile.lock to 1.0.1

Doesn't that cause an inconsistency in the versioning, or am I misunderstanding the process? If I'm misunderstanding, a README or documentation in this Fastfile itself would help clarify for future developers

fastlane/Fastfile
```diff
@@ -89,17 +95,20 @@ platform :ios do
...
+      UI.message "Pushing pod #{pod[:spec]}"
 
-      begin
-        sh('bundle', 'exec', 'pod', 'trunk', 'push', pod_path, '--allow-warnings', '--synchronous')
-      rescue => ex
-        unless ex.to_s.include? dupe_error
-          raise Exception.new "Unrecoverable error in pod push trunk"
+      with_retry do
+        begin
+          sh('bundle', 'exec', 'pod', 'trunk', 'push', pod_path, '--allow-warnings', '--synchronous')
```
palpatim, line 108: As I read the code, the default retry period in `with_retry` is 5*60=300s=5 minutes; is that sufficient to catch CDN propagation? I recall that we had to wait up to 10 minutes in the past. Or is the `--synchronous` flag helping reduce that wait requirement?

fastlane/amplify_pods.rb
```diff
@@ -40,11 +40,22 @@ module AmplifyPods
       plist_paths: [
         "AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Resources/Info.plist",
         "AmplifyPlugins/API/AWSAPICategoryPlugin/Info.plist",
-        "AmplifyPlugins/Storage/AWSS3StoragePlugin/Resources/Info.plist"
+        "AmplifyPlugins/Storage/AWSS3StoragePlugin/Resources/Info.plist",
+        "AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Resources/Info.plist"
```
palpatim, line 44: 1. What about `AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Info.plist`?
1. nit: Alphabetize to make it easier to see what's missing

.circleci/config.yml
iartemiev: I invalidated the cache by bumping it to v2. It's saved here: https://github.com/aws-amplify/amplify-ios/pull/610/files#diff-1d37e48f9ceff6d8030570cd36286a61R99

fastlane/Fastfile
iartemiev: The podfile.locks are only used when running the source code locally or running the unit tests, so they don't have to be incremented before the release. I.e. when installing the pods via `pod install`, the lock files aren't downloaded. They can only be retrieved from the git repo directly. 

We can't call `pod update` on these until the pods in question have been pushed to trunk. We could iterate over them and set the values to the latest versions manually, but then the spec checksums would be incorrect, e.g. https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/API/Podfile.lock#L92

fastlane/Fastfile
iartemiev: The purpose of this retry logic is solely to retry during intermittent failures, like the server error we experienced last week.

The `--synchronous` flag handles everything else, e.g., pushing to the trunk repo directly, instead of the CDN

palpatim: Thanks for the clarifications

.circleci/config.yml
palpatim: My read of that file reference is that it sets a cache with a key comprising both the static `v2-gems-` string and the interpolated checksum. I'm wondering about L48, which specifies just the static string `v2-gems-`. It may be innocuous; if so, please disregard

iartemiev: Ah, I see. That's a cache key strategy suggested by CircleCI where it will try to match on the full key first but then falls back to the partial match

iartemiev: So in our case, if it's a partial match, it'll pull the previous gems and then perform a bundle install after. The logic is that it'll be a tad quicker to do that than a complete fresh install

fastlane/Fastfile
palpatim: In the case you describe, the developer would be checking out from `main` or `release`, but it's a common task to troubleshoot errors on a particular release by checking out a tag. E.g.:

- Customer reports a bug on 1.0.1
- Developer checks out source at tag `v1.0.1`
- Developer does a `pod install`
- CocoaPods downloads pods according to the Podfile.lock checked in with tag `v1.0.1`, which point to `v1.0.0` versions
- Hilarity and/or chaos ensues

Am I still misunderstanding? Shouldn't we also move the `v1.0.1` tag from the original commit to the commit that contains the newly updated Podfile.lock?

iartemiev: Got it. I wasn't aware of that development flow.
We can move the remote tag in the post_release step then 👍🏻

palpatim: Actually, thinking about this more, the issue is that we're a monorepo with dependencies to pods contained in our own repo.

The right way to fix this might be to do a `pod update` before each tag, push, and cocoapods push.

fastlane/Fastfile
```diff
@@ -89,26 +93,67 @@ platform :ios do
...
   private_lane :release_pods do
     AmplifyPods.pods.each do |pod|
-      pod_path = "/Users/distiller/amplify-ios/#{pod[:spec]}"
+      if pod[:no_push]
+        UI.message "Skipping pushing pod #{pod[:spec]}"
+        next
+      end
 
-      UI.message("Pushing pod #{pod[:spec]}")
+      pod_path = "/Users/distiller/amplify-ios/#{pod[:spec]}"
```
palpatim, line 101: Extract "/Users/distiller" to a constant, or use the `HOME` env var?

Also, looking ahead to re-using this in the aws-sdk-ios project, consider extracting the project name to a constant.

PR #632, opened 2020-07-10
fix(auth): User pool token, user sub should be returned for signedIn user with no identityPool config

royjit:
*Description of changes:*

Amplify can be used with only Cognito User Pool configured. With the configuration structure shown below. In this case while calling the api `fetchAuthSession` the returned result doesnot provide the right values for usersub and cognito tokens. This PR fixes this behavior.

```

{
    "UserAgent": "aws-amplify-cli/2.0",
    "Version": "1.0",
    "auth": {
        "plugins": {
            "awsCognitoAuthPlugin": {
                "UserAgent": "aws-amplify/cli",
                "Version": "0.1.0",
                "IdentityManager": {
                    "Default": {}
                },
                
                "CognitoUserPool": {
                    "Default": {
                        "PoolId": "xx",
                        "AppClientId": "xx",
                        "AppClientSecret": "xx",
                        "Region": "xx"
                    }
                },
                "Auth": {
                    "Default": {
                        "OAuth": {
                            ...
                            
                        },
                        "authenticationFlowType": "USER_SRP_AUTH"
                    }
                }
            }
        }
    }
}

```

### DIFF
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthorizationProviderAdapter+SignedInSession.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthorizationProviderAdapter+SignedInSession.swift
@@ -112,9 +112,7 @@ extension AuthorizationProviderAdapter {
                 if let urlError = error as NSError?, urlError.domain == NSURLErrorDomain {
                     self.fetchSignedInSessionWithOfflineError(completionHandler)
 
-                } else if let cognitoIdentityPoolError = error as NSError?,
-                    cognitoIdentityPoolError.domain == AWSCognitoIdentityErrorDomain,
-                    cognitoIdentityPoolError.code == AWSCognitoIdentityErrorType.notAuthorized.rawValue {
+                } else if !self.identityPoolConfigured(error: error!) {
                     self.fetchSignedInSessionWithNoIdentityPool(withTokensResult: tokenResult,
                                                                 userSubResult: userSubResult,
                                                                 completionHandler)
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthorizationProviderAdapter+SignedInSession.swift
```diff
@@ -222,4 +220,18 @@ extension AuthorizationProviderAdapter {
                                                 cognitoTokensResult: tokenResult)
         completionHandler(.success(authSession))
     }
+
+    /// Check if the error is related to improper configuration of Cognito Identity Pool
+    private func identityPoolConfigured(error: Error) -> Bool {
```
palpatim, line 225: 1. See note above re method name
1. I don't see this using any instance members, so make this `static`

More importantly, it's hard to wade through the double negatives in this method logic: the method comment asserts that we're looking to see if the error is related to improper configuration of an identity pool, but we return `false` if we find that the error is in fact a `cognitoIdentityPoolNotConfigured` error. How is that error not related to improper configuration? Why are just these two error conditions sufficient to assert that the error isn't related to improper configuration?

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthorizationProviderAdapter+SignedInSession.swift
```diff
@@ -112,9 +112,7 @@ extension AuthorizationProviderAdapter {
                 if let urlError = error as NSError?, urlError.domain == NSURLErrorDomain {
                     self.fetchSignedInSessionWithOfflineError(completionHandler)
 
-                } else if let cognitoIdentityPoolError = error as NSError?,
-                    cognitoIdentityPoolError.domain == AWSCognitoIdentityErrorDomain,
-                    cognitoIdentityPoolError.code == AWSCognitoIdentityErrorType.notAuthorized.rawValue {
+                } else if !self.identityPoolConfigured(error: error!) {
```
palpatim, line 115: More context would be handy here. Why is it safe to ignore what appears just be reading code to be a misconfigured identity pool? Changing the method name to self-document the intent of the check would be helpful, but honestly I can't tell what the method below is actually trying to do.

royjit: Changed the method name

PR #642, opened 2020-07-15
fix: Plugins that are not configured correctly will be error at initial step

royjit:
*Issue https://github.com/aws-amplify/amplify-ios/issues/480*

*Description of changes:*

This change allows to call configure method of each plugin with a nil value. ie even if the corresponding configuration is not present in the configuration file, plugin's configure will be called.

*Issue we are solving:*

Suppose that I added two plugins to my project `AWSAPIPlugin` and `AWSDataStorePlugin`, but forgot to add any configuration for API in my `amplifyconfiguration.json`. Since there is no configuration the current logic does not invoke the `configure(:)` function of `AWSAPIPlugin`. So we end up using a not-configured plugin and end up in unintended error or crash like this - https://github.com/aws-amplify/amplify-ios/issues/480

*Thought process behind this change:*

All category in Amplify conforms to the protocol - [CategoryConfigurable](https://github.com/aws-amplify/amplify-ios/blob/ff8e86c0b8c00a0a1032219a2ee825291761a099/Amplify/Core/Configuration/Internal/CategoryConfigurable.swift#L8), which could infer that the plugins added to every category is [configurable](https://github.com/aws-amplify/amplify-ios/blob/ff8e86c0b8c00a0a1032219a2ee825291761a099/Amplify/Core/Plugin/Plugin.swift#L23). Since each added plugin is configurable it should be given a chance to configure even if the corresponding configuration is missing. This will give the plugin a chance to decide whether it can work in a non-configuration based situation or throw an error.

*Other changes in this PR*
- Fixed an error in storage unit test that caused it to fail when whole storage unit test is executed.

### DIFF
--- a/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPluginTests/AWSPinpointAnalyticsPluginConfigureTests.swift
+++ b/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPluginTests/AWSPinpointAnalyticsPluginConfigureTests.swift
@@ -65,8 +65,8 @@ class AWSPinpointAnalyticsPluginConfigureTests: AWSPinpointAnalyticsPluginTestBa
             try plugin.configure(using: nil)
             XCTFail("Analytics configuration should not succeed")
         } catch {
-            guard let apiError = error as? PluginError,
-                case .pluginConfigurationError(_, _, _) = apiError else {
+            guard let pluginError = error as? PluginError,
+                case .pluginConfigurationError(_, _, _) = pluginError else {
                     XCTFail("Should throw invalidConfiguration exception. But received \(error) ")
                     return
             }
--- a/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPluginTests/AWSPinpointAnalyticsPluginConfigureTests.swift
+++ b/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPluginTests/AWSPinpointAnalyticsPluginConfigureTests.swift
@@ -59,4 +59,17 @@ class AWSPinpointAnalyticsPluginConfigureTests: AWSPinpointAnalyticsPluginTestBa
         }
     }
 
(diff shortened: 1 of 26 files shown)


AmplifyPlugins/Analytics/AWSPinpointAnalyticsPluginTests/AWSPinpointAnalyticsPluginConfigureTests.swift
```diff
@@ -59,4 +59,17 @@ class AWSPinpointAnalyticsPluginConfigureTests: AWSPinpointAnalyticsPluginTestBa
         }
     }
 
+    func testConfigureFailureForNilConfiguration() throws {
+        let plugin = AWSPinpointAnalyticsPlugin()
+        do {
+            try plugin.configure(using: nil)
+            XCTFail("Analytics configuration should not succeed")
+        } catch {
+            guard let apiError = error as? PluginError,
```
palpatim, line 68: I assume we're not really throwing an API error here, it's just a misnamed variable?

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AWSCognitoAuthPluginConfigTests.swift
```diff
@@ -166,4 +166,26 @@ class AWSCognitoAuthPluginConfigTests: XCTestCase {
...
+    /// - Then:
+    ///    - I should get an exception.
+    ///
+    func testConfigureFailureForNilConfiguration() throws {
+        let plugin = AWSCognitoAuthPlugin()
+        do {
+            try plugin.configure(using: nil)
+            XCTFail("Auth configuration should not succeed")
+        } catch {
+            guard let apiError = error as? PluginError,
```
palpatim, line 183: Here and throughout: fix name

PR #644, opened 2020-07-15
[Datastore] sort feature

DongQuanRui:
*Issue:*
Because of some Github tech issue. Created a new PR

*Description of changes:*
- Added a new enum type `QuerySortBy` and a new struct type `QuerySortInput`.
- Added `SortInput` SQLStatement generation code.
- Added 5 unit tests, corresponding to 5 use cases in Quip docs.
- Added 6 integration tests, the first 5 corresponds to 5 use cases in Quip docs, the 6th one is more to verify the correctness of pagination.

Link to old PR: https://github.com/aws-amplify/amplify-ios/pull/580


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -79,11 +79,13 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
 
     public func query<M: Model>(_ modelType: M.Type,
                                 where predicate: QueryPredicate? = nil,
+                                sort sortInput: QuerySortInput? = nil,
```
drochetti, line 82: since we're introducing a new attribute, we should "append" it to the existing ones, meaning that it should come after `paginationInput`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -165,6 +167,7 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
             let metadata = MutationSyncMetadata.keys
             storageEngine.query(MutationSyncMetadata.self,
                                 predicate: metadata.id == model.id,
+                                sort: nil,
```
drochetti, line 170: sorting doesn't make sense in a method that just queries for 1 model by its id.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/QuerySortBy+SQLite.swift
```diff
@@ -0,0 +1,21 @@
...
+
+import Amplify
+import Foundation
+import SQLite
+
+extension QuerySortBy {
+    var sqlOrderBy: String {
+        switch self {
+        case .ascending(let key):
+            return "\(key.stringValue) ASC"
```
drochetti, line 16: 1. it should be `key.columnName` I believe, otherwise you won't get the escaped name with the namespace. Refer to existing SQL generation for a reference.
2. the code is using lowercase for SQL keywords, so prefer `asc` over `ASC`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
```diff
@@ -22,6 +22,7 @@ struct SelectStatementMetadata {
     // TODO remove additionalStatements once sorting support is added to DataStore
```
drochetti, line 22: since sorting is now supported, we should address this `TODO`: `additionalStatements` can be safely replaced with sort

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
DongQuanRui: actually, I think in the review meeting I spoke the reason I put `sort` between `where` and `paginate`. Because in sql statement, this is the required order, first: `where`, second: `order by`, third: `offset` and `limit`. But your point also makes sense. What do you think on the two options? which is better?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/QuerySortBy+SQLite.swift
DongQuanRui: Sorry, but I did google and uppercase should be right. 
Reference: https://www.w3schools.com/sql/sql_orderby.asp
https://www.geeksforgeeks.org/sql-order-by/
http://www.sql-tutorial.com/sql-order-by-sql-tutorial/

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Select.swift
DongQuanRui: removed this line

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
DongQuanRui: Oh, here just need to make compiler happy, like this one: https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/AWSMutationDatabaseAdapter/AWSMutationDatabaseAdapter%2BMutationEventSource.swift#L32
Or maybe I don't get your meaning here

palpatim: You should update the storage engine method signature to make `sort` default to `nil`, rather than requiring it to be specified as `nil` at the call site

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/QuerySortBy+SQLite.swift
palpatim: SQL is case-insensitive, @drochetti's point was a style note. We prefer to use lowercase SQL keywords

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
palpatim: I recall this discussion during the review and agree with the current design: that `sort` should come before `paginate`, since that correctly communicates the order of operations internally: the results will be filtered by `predicate`, sorted by `sortInput`, and then paginated by `paginationInput`.

Because we are adding the parameter with a default value, it should still be a non-breaking change at call sites. Customers who want to adopt sorting will have to specify `sort` in the middle of their arguments, so that is a little bit odd, but I think that's a reasonable tradeoff for the overall clarity of the method signature.

drochetti:
Closing this PR in favor of https://github.com/aws-amplify/amplify-ios/pull/580

PR #647, opened 2020-07-15
fix(Datastore): paginationInput not passed down in query

DongQuanRui:
*Description of changes:*
- Added missing parameter `paginationInput` so that can be called
- Added an integration test to verify if the `paginationInput ` is actually being called.

Issue: https://github.com/aws-amplify/amplify-ios/issues/646

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreLocalStoreTests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreLocalStoreTests.swift
@@ -1,0 +1,85 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+
+import AmplifyPlugins
+import AWSPluginsCore
+
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+class DataStoreLocalStoreTests: LocalStoreIntegrationTestBase {
+
+    /// - Given: 15 posts that has been saved
+    /// - When:
+    ///    - attempt to query existing posts with offset and limit
+    /// - Then:
+    ///    - the existing data that matches offset and limit will be returned
+    func testQueryWithPaginationInput() throws {
+        _ = setUpLocalStore(numberOfPosts: 15)
+        var posts = [Post]()
+        let queryFirstTimeSuccess = expectation(description: "Query post completed")
+
+        Amplify.DataStore.query(Post.self,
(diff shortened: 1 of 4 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreLocalStoreTests.swift
```diff
@@ -0,0 +1,85 @@
...
+            case .success(let returnPosts):
+                posts.append(contentsOf: returnPosts)
+                queryFirstTimeSuccess.fulfill()
+            case .failure(let error):
+                XCTFail("Error querying posts: \(error)")
+            }
+        }
+        wait(for: [queryFirstTimeSuccess], timeout: 10)
+
+        XCTAssertTrue(posts.count == 10)
```
lawmicha, line 41: nit: use `XCTAssertEqual` when doing equality checks for two different fields and leave `XCTAssertTrue` for  boolean fields

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreLocalStoreTests.swift
```diff
@@ -0,0 +1,85 @@
...
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+class DataStoreLocalStoreTests: LocalStoreIntegrationTestBase {
+
+    /// - Given: 15 posts that has been saved
+    /// - When:
+    ///    - attempt to query existing posts with offset and limit
+    /// - Then:
+    ///    - the existing data that matches offset and limit will be returned
```
lawmicha, line 23: i didn't quite understand what it means by "existing data that matches offset", perhaps you can simplify it and say what is happening, like:
```
when:
 - query with pagination input given a page number and limit 10
then:
 - first page returns the 10 (the defined limit) of 15 posts
 - second page returns the remaining 5 posts
 - the 15 retrieved posts have unique identifiers
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreLocalStoreTests.swift
```diff
@@ -0,0 +1,85 @@
...
+            case .failure(let error):
+                XCTFail("Error querying posts: \(error)")
+            }
+        }
+        wait(for: [querySecondTimeSuccess], timeout: 10)
+
+        XCTAssertTrue(posts.count == 15)
+
+        let idArray = posts.map { $0.id }
+        let idSet = Set(idArray)
```
lawmicha, line 59: nit: variables used once can be reduced to one line like `let uniquePosts = Set(posts.map { $0.id })`

PR #660, opened 2020-07-22
fix(predictions): TABLE, CELL & KEY_VALUE_SET blocks are not properly processed

DongQuanRui:
*Description of changes:*
Bug fix according to: https://github.com/aws-amplify/amplify-ios/issues/641

- Refactored `processText(_ textractTextBlocks: [AWSTextractBlock])`
- Fixed the bug where `TABLE`, `CELL` blocks are not properly processed
- Fixed the bug where `KEY_VALUE_SET` blocks are not properly processed

- Updated integration tests
- Update `README.md`
- Replaced `AWSMobileClient` with `Amplify.Auth`


AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -52,6 +52,7 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
                                   identifiedLines: identifiedLines)
     }
 
+    // swiftlint:disable cyclomatic_complexity
```
wooj2, line 55: Instead of disabling swiftlint, I think we should be using it as a hint to simplify our code.  I looked into this, and seems that adding the guard statements is forcing our cyclomatic complexity for this function to go up.  I was able to simplify the code so that we do not need this warning (if you can find out a better/different way to do it, that's fine too.)

I made the following changes:
1.  To avoid `guard let identifier = block.identifier else {` in this function, we just filter out all textractTextBlocks that have an invalid identifier.
2.  To avoid multiple guard/let/else, create sub functions to pass back fully constructed objects, or nil in the event that the block doesn't have enough information
3.  Create additional function to processTables
4.  Create additional function to processKeyValues.

As a result, I ended up with this for `processText(_ textractTextBlocks: [AWSTextractBlock])`:
```
    static func processText(_ textractTextBlocks: [AWSTextractBlock]) -> IdentifyDocumentTextResult {

        var words = [IdentifiedWord]()
        var lines = [String]()
        var linesDetailed = [IdentifiedLine]()
        var selections = [Selection]()
        var fullText = ""
        var blockMap = [String: AWSTextractBlock]()
        var tableBlocks = [AWSTextractBlock]()
        var keyValueBlocks = [AWSTextractBlock]()

        let validTextractTextBlocks = textractTextBlocks.filter { block in
            !(block.identifier ?? "").isEmpty
        }

        for block in validTextractTextBlocks {
            let identifier = block.identifier!
            switch block.blockType {
            case .line:
                if let line = parseLineBlock(block: block) {
                    lines.append(line.text)
                    linesDetailed.append(line)
                }
            case .word:
                if let word = parseWordBlock(block: block) {
                    fullText += word.text + " "
                    words.append(word)
                    blockMap[identifier] = block
                }
            case .selectionElement:
                if let selection = parseSelectionElementBlock(block: block) {
                    selections.append(selection)
                    blockMap[identifier] = block
                }
            case .table:
                tableBlocks.append(block)
            case .keyValueSet:
                keyValueBlocks.append(block)
                blockMap[identifier] = block
            default:
                blockMap[identifier] = block
            }
        }

        let tables = processTables(tableBlocks: tableBlocks, blockMap: blockMap)
        let keyValues = processKeyValues(keyValueBlocks: keyValueBlocks, blockMap: blockMap)

        return IdentifyDocumentTextResult(
            fullText: fullText,
            words: words,
            rawLineText: lines,
            identifiedLines: linesDetailed,
            selections: selections,
            tables: tables,
            keyValues: keyValues)
    }
```

Constructing functions:
```
    static func parseLineBlock(block: AWSTextractBlock) -> IdentifiedLine? {
        guard let text = block.text,
            let boundingBox = processBoundingBox(block.geometry?.boundingBox),
            let polygon = processPolygon(block.geometry?.polygon) else {
                return nil
        }

        return IdentifiedLine(text: text,
                              boundingBox: boundingBox,
                              polygon: polygon,
                              page: Int(truncating: block.page ?? 0))
    }

    static func parseWordBlock(block: AWSTextractBlock) -> IdentifiedWord? {
        guard let text = block.text,
            let boundingBox = processBoundingBox(block.geometry?.boundingBox),
            let polygon = processPolygon(block.geometry?.polygon) else {
                return nil
        }

         return IdentifiedWord(text: text,
                               boundingBox: boundingBox,
                               polygon: polygon,
                               page: Int(truncating: block.page ?? 0))
    }

    static func parseSelectionElementBlock(block: AWSTextractBlock) -> Selection? {
        guard let boundingBox = processBoundingBox(block.geometry?.boundingBox),
            let polygon = processPolygon(block.geometry?.polygon) else {
                return nil
        }
        let selectionStatus = block.selectionStatus == .selected ? true : false
        return Selection(boundingBox: boundingBox, polygon: polygon, isSelected: selectionStatus)
    }
```

Process functions which return `[Table]` and `[AWSTextractBlock]`:
```
    static func processTables(tableBlocks: [AWSTextractBlock],
                              blockMap: [String: AWSTextractBlock]) -> [Table] {
        var tables = [Table]()
        if !tableBlocks.isEmpty {
            for tableBlock in tableBlocks {
                if let table = processTable(tableBlock, blockMap: blockMap) {
                    tables.append(table)
                }
            }
        }
        return tables
    }

    static func processKeyValues(keyValueBlocks: [AWSTextractBlock],
                                 blockMap: [String: AWSTextractBlock]) -> [BoundedKeyValue] {
        var keyValues =  [BoundedKeyValue]()
        if !keyValueBlocks.isEmpty {
            for keyValueBlock in keyValueBlocks where keyValueBlock.entityTypes?.contains("KEY") ?? false {
                if let keyValue = processKeyValue(keyValueBlock, blockMap: blockMap) {
                    keyValues.append(keyValue)
                }
            }
        }
        return keyValues
    }
```

Not ideal, but I think it at least breaks things up a bit more so that `processText` isn't so hairy.  Thoughts?

DongQuanRui: This is a good idea and I will take it. 
But I should actually make this one back to a Draft PR, next week I need more time working on this PR, I believe the code is still missing some behaviors. I am creating an Android App to see how that looks like and also need to sync with Android.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -66,36 +67,55 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
+                guard let polygon = processPolygon(block.geometry?.polygon) else {
+                    continue
+                }
+                let line = IdentifiedLine(text: text,
+                                          boundingBox: boundingBox,
+                                          polygon: polygon,
+                                          page: Int(truncating: block.page ?? 0))
                 lines.append(text)
                 linesDetailed.append(line)
+                blockMap[identifier] = block
```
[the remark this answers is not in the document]
DongQuanRui: I noticed that Android does this for every block. And I also thought about it
As you can see in the latest code, `word`, `keyValue`, `table`, `cell` blocks are being accessed using blockMap. Why not just pushing every block into blockMap. But if we are concerning about performance issue (to reduce the time used to push unnecessary blocks into blockMap), I am OK to put this line separately into each necessary `switch cases`

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -218,6 +238,14 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
     }
 
+    // https://docs.aws.amazon.com/textract/latest/dg/how-it-works-kvp.html
+    /**
+    * Converts a given Amazon Textract block into Amplify-compatible
+    * key-value pair feature. Returns null if not a valid table.
+    * @param block Textract text block
+    * @param blockMap map of Textract blocks by their IDs
+    * @return Amplify KeyValue instance
+    */
```
wooj2, line 248: Lets remove this comment.  We typically have these types of comments for public facing functions that don't change very often.  This seems like a utility function, that is only used local to this file, right?

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -229,32 +257,20 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
-                        }
-                    case .selectionElement:
-                        valueSelected = keyBlock.selectionStatus == .selected ? true : false
-                    default: break
-                    }
-                }
-
+                let valueResult = processValue(ids: ids, blockMap: blockMap)
+                valueText = valueResult.0
+                valueSelected = valueResult.1
```
wooj2, line 273: Admittedly, i have not looked at this carefully enough, but it seems we are looping through a for loop and overwriting a variable keyText -- prior, we were appending text to a variable.  Is this correct?

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/AWSPredictionsPluginTestBase.swift
```diff
@@ -6,97 +6,45 @@
...
-                            "language": "en",
-                            "region": region,
-                            "defaultNetworkPolicy": "auto"
-                        ]
-                    ]
-                ]
-            ]
-        )
+        Amplify.reset()
+        Amplify.Logging.logLevel = .verbose
```
wooj2, line 24: is it needed to turn on verbosity?  If not, lets remove it.

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/AWSPredictionsPluginTestBase.swift
```diff
@@ -6,97 +6,45 @@
...
-        setupAmplify()
-    }
+        super.setUp()
 
-    override func tearDown() {
-        sleep(1)
-        print("Amplify reset")
-        Amplify.reset()
-    }
+        continueAfterFailure = false
```
wooj2, line 21: Why did we turn on continueAfterFailure ?

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/AWSPredictionsPluginTestBase.swift
```diff
@@ -6,97 +6,45 @@
...
-                        "transcription": [
-                            "language": "en",
-                            "region": region,
-                            "defaultNetworkPolicy": "auto"
-                        ]
-                    ]
-                ]
-            ]
-        )
+        Amplify.reset()
```
wooj2, line 23: Why are we calling `Amplify.reset()` in setup and `tearDown`?

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/IdentifyBasicIntegrationTests.swift
```diff
@@ -243,108 +143,143 @@ class IdentifyBasicIntegrationTests: XCTestCase {
...
+            XCTFail("Unable to find image")
+            return
         }
+
         let completeInvoked = expectation(description: "Completed is invoked")
 
-        let operation = Amplify.Predictions.identify(type: .detectText(.table),
+                let operation = Amplify.Predictions.identify(type: .detectText(.table),
                                                      image: url,
                                                      options: PredictionsIdentifyRequest.Options()) { event in
```
wooj2, line 272: Spacing seems off here?

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/IdentifyBasicIntegrationTests.swift
```diff
@@ -219,20 +120,19 @@ class IdentifyBasicIntegrationTests: XCTestCase {
...
-                                                            completeInvoked.fulfill()
-                                                        case .failure(let error):
-                                                            XCTFail("Failed with \(error)")
-                                                        }
+            switch event {
+            case .success:
+                completeInvoked.fulfill()
+            case .failure(let error):
+                XCTFail("Failed with \(error)")
+            }
```
wooj2, line 128: Are you using an app to update the spacing?  What was wrong with the spacing before?  Typically, we try to keep our changesets as small as possible.

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/IdentifyBasicIntegrationTests.swift
```diff
@@ -243,108 +143,143 @@ class IdentifyBasicIntegrationTests: XCTestCase {
...
-                                                        case .failure(let error):
-                                                            XCTFail("Failed with \(error)")
-                                                        }
+            switch event {
+            case .success(let result):
+                let data = result as? IdentifyDocumentTextResult
+                XCTAssertNotNil(data)
+                XCTAssertFalse(data!.keyValues.isEmpty)
+                XCTAssertFalse(data!.fullText.isEmpty)
+                XCTAssertFalse(data!.tables.isEmpty)
```
wooj2, line 207: I agree this is better than what we were doing before, is there anything about the data that is coming back that you'd like to verify?

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/IdentifyBasicIntegrationTests.swift
```diff
@@ -243,108 +143,143 @@ class IdentifyBasicIntegrationTests: XCTestCase {
...
-                                                        case .failure(let error):
-                                                            XCTFail("Failed with \(error)")
-                                                        }
+            switch event {
+            case .success(let result):
+                let data = result as? IdentifyDocumentTextResult
+                XCTAssertNotNil(data)
+                XCTAssertFalse(data!.keyValues.isEmpty)
+                XCTAssertFalse(data!.fullText.isEmpty)
+                completeInvoked.fulfill()
```
wooj2, line 243: Since we've gone through the work to properly parse the TextResults returned from the service seems like we should do some basic validation that we are parsing it correctly?

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/IdentifyBasicIntegrationTests.swift
```diff
@@ -243,108 +143,143 @@ class IdentifyBasicIntegrationTests: XCTestCase {
...
-                                                        case .failure(let error):
-                                                            XCTFail("Failed with \(error)")
-                                                        }
+            switch event {
+            case .success(let result):
+                let data = result as? IdentifyDocumentTextResult
+                XCTAssertNotNil(data)
+                XCTAssertFalse(data!.fullText.isEmpty)
+                XCTAssertFalse(data!.tables.isEmpty)
+                completeInvoked.fulfill()
```
wooj2, line 279: Same comment here, seems like we should add some basic checks to make sure we properly parsed the tables.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
```diff
@@ -67,6 +67,96 @@ class IdentifyResultTransformers {
...
+            let keyBlock = blockMap[keyId]
+            guard let keyBlockType = keyBlock?.blockType else {
+                continue
+            }
+            switch keyBlockType {
+            case .word:
+                if let text = keyBlock?.text {
+                    keyText += text + " "
+                }
+            default: break
```
wooj2, line 83: What other cases are there?  why is breaking on a non-word the correct behavior?

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
```diff
@@ -67,6 +67,96 @@ class IdentifyResultTransformers {
...
+
+    static func processValue(ids: [String],
+                             blockMap: [String: AWSTextractBlock]) -> (String, Bool) {
+        var valueText = ""
+        var valueSelected = false
+
+        // VALUE block has a CHILD list of IDs for the WORD block
+        for valueId in ids {
+            let valueBlock = blockMap[valueId]
+            guard let valueRelatioins = valueBlock?.relationships else {
```
wooj2, line 97: s/relatioins/relations ?

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
```diff
@@ -67,6 +67,96 @@ class IdentifyResultTransformers {
...
+
+    static func parseSelectionElementBlock(block: AWSTextractBlock) -> Selection? {
+        guard let boundingBox = processBoundingBox(block.geometry?.boundingBox),
+            let polygon = processPolygon(block.geometry?.polygon) else {
+                return nil
+        }
+        let selectionStatus = block.selectionStatus == .selected ? true : false
+        return Selection(boundingBox: boundingBox, polygon: polygon, isSelected: selectionStatus)
+    }
+
```
wooj2, line 159: seems like we've added  the following functions to this class:
```
processChild
processValue
parseLineBlock
parseWordBlock
parseSelectionElementBlock
```

Why were they added to the parent level class, and not to IdentifyTextResultTransformers directly?

I can kinda understand why `parseLineBlock`, `parseWordBlock`, and `parseSelectionElementBlock` were added, but we'd probably want to rename these for consistency to `processLineBlock`, `processWordBlock`, and `processSelectionElementBlock` respectively... 

I'm not entirely convinced `processChild`, `processValue` belong in the parent class.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -53,79 +53,52 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
-            let word = IdentifiedWord(text: text,
-                                      boundingBox: boundingBox,
-                                      polygon: polygon,
-                                      page: Int(truncating: block.page ?? 0))
-
-            let line = IdentifiedLine(text: text,
-                                      boundingBox: boundingBox,
-                                      polygon: polygon,
-                                      page: Int(truncating: block.page ?? 0))
+            blockMap[identifier] = block
```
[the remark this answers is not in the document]
DongQuanRui: My mistake here.
I think line 72 should be kept, and line 84, 89 should be removed. 
Reference for how Android does: https://github.com/aws-amplify/amplify-android/blob/71aa76d145812170d2f9d8e3110ef907427e08ef/aws-predictions/src/main/java/com/amplifyframework/predictions/aws/service/AWSTextractService.java#L148
And it makes sense, we don't know what document our customer want to identify, so to put every block into `blockMap` in case need to fetch out in later processing

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -137,15 +110,47 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
+                tables.append(table)
+            }
+        }
+        return tables
+    }
+
+    static func processKeyValues(keyValueBlocks: [AWSTextractBlock],
+                                 blockMap: [String: AWSTextractBlock]) -> [BoundedKeyValue] {
+        var keyValues =  [BoundedKeyValue]()
+        for keyValueBlock in keyValueBlocks where keyValueBlock.entityTypes?.contains("KEY") ?? false {
```
wooj2, line 127: I think it makes sense to do something like:
move:
```
where keyValueBlock.entityTypes?.contains("KEY") ?? false
```
into  the function`processKeyValue()`:
```
 static func processKeyValue(_ keyBlock: AWSTextractBlock, .......... {
        guard keyBlock.entityTypes?.contains("KEY") ?? false,
            let relationships = keyBlock.relationships else {
            return nil
        }
   ...
}
```
((Note that the `keyBlock` parameter does not need to be Optional))

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -229,32 +257,20 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
+
+            guard let ids = keyBlockRelationship.ids else {
+                continue
             }
+
             switch keyBlockRelationship.types {
-            case .child where keyBlock.blockType == .word:
-                     keyText += text + " "
+            case .child:
+                keyText = processChild(ids: ids, blockMap: blockMap)
```
wooj2, line 269: I think a better name for `processChild` would be `processChildofKeyValueSet`

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
wooj2: Looking at the previous implementation, i think the following code would be significantly cleaner.
```
static func process...... {
        var keyText = ""
        for keyId in ids {
            guard let keyBlock = blockMap[keyId],
                let text = keyBlock.text,
                case .word = keyBlock.blockType else {
                continue
            }
            keyText += text + " "
        }
        return keyText.trimmingCharacters(in: .whitespacesAndNewlines)
}
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
wooj2: guard let valueBlockRelationships = valueBlock?.relationships else {

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
```diff
@@ -67,6 +67,96 @@ class IdentifyResultTransformers {
...
+        var valueSelected = false
+
+        // VALUE block has a CHILD list of IDs for the WORD block
+        for valueId in ids {
+            let valueBlock = blockMap[valueId]
+            guard let valueRelatioins = valueBlock?.relationships else {
+                continue
+            }
+            for valueRelation in valueRelatioins {
+                guard let ids = valueRelation.ids else {
```
wooj2, line 101: for valueBlockRelation in valueBlockRelationships {
                guard let ids = valueBlockRelation.ids else {

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
```diff
@@ -67,6 +67,96 @@ class IdentifyResultTransformers {
...
+                    guard let wordValueBlockType = wordBlock?.blockType else {
+                        continue
+                    }
+                    switch wordValueBlockType {
+                    case .word:
+                        if let text = wordBlock?.text {
+                            valueText += text + " "
+                        }
+                    case .selectionElement:
+                        valueSelected = wordBlock?.selectionStatus == .selected ? true : false
```
wooj2, line 115: Seems like we should also do something similar here if we find multiple selectionElements for a specific key/value pair.

Also,

this can be:
```
valueSelected = wordBlock?.selectionStatus == .selected
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
```diff
@@ -67,6 +67,96 @@ class IdentifyResultTransformers {
...
+                    default: break
+                    }
+                }
+            }
+
+        }
+        return (valueText, valueSelected)
```
wooj2, line 122: we should trim trailing whitespace:
```
return (valueText.trimmingCharacters(in: .whitespacesAndNewlines), valueSelected)
```

wooj2:
1.   The employmentapp.png test case comes from here:
https://github.com/aws-samples/amazon-textract-code-samples/tree/master/src-csharp/test-files
a.  The file is still a png file, so, it should not have a `.jpg` extension.  
b.  I think we should just leave it as employmentapp.png

2. testImageTextWithTables.jpg does not seem to be a good test, because it has the same values in each of the tables.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
DongQuanRui: Actually, I haven't really think about the situation of selectionElement. But I agree with your point here

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
DongQuanRui: Great, I was thinking about how to remove the trailing whitespace and you showed me how

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -53,79 +53,52 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
     }
 
     static func processText(_ textractTextBlocks: [AWSTextractBlock]) -> IdentifyDocumentTextResult {
-
+        var fullText = ""
```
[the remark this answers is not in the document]
DongQuanRui: Updated like this

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -137,15 +110,47 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
+        guard let relationships = tableBlock.relationships else {
             return nil
         }
         var table = Table()
         var rows = Set<Int>()
         var cols = Set<Int>()
+
+        // Each TABLE block contains CELL blocks
         for tableRelation in relationships {
             guard let ids = tableRelation.ids else {
```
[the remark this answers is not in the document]
DongQuanRui: true, updated

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -137,15 +110,47 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
+    * table object.
+    * @param block Textract text block
+    * @param blockMap map of Textract blocks by their IDs
+    * @return Amplify Table instance
+    */
+    static func processTable(_ tableBlock: AWSTextractBlock,
                              blockMap: [String: AWSTextractBlock]) -> Table? {
 
-        guard let relationships = block.relationships else {
+        guard let relationships = tableBlock.relationships else {
```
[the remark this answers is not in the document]
DongQuanRui: Actually, here I think might be redundant?, cause in `processText()`, already does a case check only put block of type `TABLE` into `tableBlock`. Is `case .table = tableBlock.blockType` still needed here?

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -162,13 +167,13 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
+                    rows.insert(row)
                 }
                 if !cols.contains(col) {
-                cols.insert(col)
+                    cols.insert(col)
                 }
-                if let cell = constructTableCell(cellBlock) {
-                table.cells.append(cell)
+                if let cell = constructTableCell(cellBlock, blockMap) {
+                    table.cells.append(cell)
```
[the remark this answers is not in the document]
DongQuanRui: I don't know here. For `row` and `col`, I just kept it as before. But I think: `row` and `col` should have value at the same time? How can a table only has row but no col or the other way?

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -178,27 +183,42 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
-        case .word:
-             words += text + " "
-        case .selectionElement:
-            isSelected = selectionStatus == .selected ? true : false
-        default: break
+        // Each CELL block consists of WORD and/or SELECTION_ELEMENT blocks
+        for cellRelation in relationships {
+            guard let ids = cellRelation.ids else {
+                continue
+            }
```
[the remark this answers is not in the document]
DongQuanRui: Oh my, the naming is so long, but to make reading easier.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -178,27 +183,42 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
+                let wordBlock = blockMap[wordId]
+
+                switch wordBlock?.blockType {
+                case .word:
+                    guard let text = wordBlock?.text else {
+                        return nil
+                    }
+                    words += text + " "
+                case .selectionElement:
+                    isSelected = selectionStatus == .selected ? true : false
```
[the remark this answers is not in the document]
DongQuanRui: Oh my, When I started to debug the code, I already thought it is more difficult than expected. but I didn't expect even more😂. No problem, maybe bring this up in ops review? But let me try to create a testable table first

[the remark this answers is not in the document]
DongQuanRui: But I thought about multiple selection elements. I don't really think there will be a cell contains multiple selections, what's the point of that kind of usage? Just wondering, do you have an example of real-life usage?

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
DongQuanRui: The consistency of naming is a good point. But `Child` and `Value` are both `TextractBlock`, the same as `LineBlock`, `WordBlock` and `SelectionElementBlock`, so I put `processChildOfKeyValueSet()` and `processValueOfKeyValueSet()` in parent level

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/IdentifyBasicIntegrationTests.swift
```diff
@@ -243,108 +143,171 @@ class IdentifyBasicIntegrationTests: XCTestCase {
...
+                }
+                XCTAssertFalse(data.fullText.isEmpty)
+                XCTAssertFalse(data.words.isEmpty)
+                XCTAssertEqual(data.words.count, 5)
+                XCTAssertFalse(data.rawLineText.isEmpty)
+                XCTAssertEqual(data.rawLineText.count, 3)
+                XCTAssertFalse(data.identifiedLines.isEmpty)
+                XCTAssertEqual(data.identifiedLines.count, 3)
+                XCTAssertFalse(data.tables.isEmpty)
+                XCTAssertEqual(data.tables.count, 1)
```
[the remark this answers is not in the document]
DongQuanRui: I have the same thought, but the response returned doesn't really have the coordinate of the cell, in a `CELL` you can find `text`, `boundingBox` which is of type `CGRect`, `polygon` of type `Amplify.polygon(four CGPoint)`, `isSelected`, `rowSpan` and `columnSpan`. I mean, I can write a function to process the relative location of each cells in terms of `rowIndex` and `columnIndex`, but that makes the test too complicated, right?

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyResultTransformers.swift
```diff
@@ -67,6 +67,89 @@ class IdentifyResultTransformers {
...
+
+    static func processSelectionElementBlock(block: AWSTextractBlock) -> Selection? {
+        guard let boundingBox = processBoundingBox(block.geometry?.boundingBox),
+            let polygon = processPolygon(block.geometry?.polygon) else {
+                return nil
+        }
+        let selectionStatus = block.selectionStatus == .selected ? true : false
+        return Selection(boundingBox: boundingBox, polygon: polygon, isSelected: selectionStatus)
+    }
+
```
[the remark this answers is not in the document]
DongQuanRui: updated

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/AWSPredictionsPluginTestBase.swift
```diff
@@ -6,97 +6,43 @@
...
 
-        let amplifyConfig = AmplifyConfiguration(predictions: predictionsConfig)
```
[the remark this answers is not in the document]
DongQuanRui: According to Jithin, there is no specific reason why not using one over the other. Since the other categories use file way of configuring, then I would also adopt that way and it's more convenient.

AmplifyPlugins/Predictions/AWSPredictionsPluginIntegrationTests/IdentifyBasicIntegrationTests.swift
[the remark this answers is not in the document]
DongQuanRui: Our model doesn't support the check again this. Should be a bug in iOS code. Will open up an issue to document it and fix it later in another PR.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -137,141 +111,269 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
+        var isSelected = false
+        if !selectionItemFound {
+            selectionItemFound = true
+            //TODO: Support multiple selection items found in a single cell
+            isSelected = selectionStatus == .selected
+        } else {
+            Amplify.log.error("Multiple selection items found in single cell")
         }
+        return (isSelected, selectionItemFound)
+    }
```
wooj2, line 343: I understand trying to re-use code here, but in this case, I think it actually adds complexity.  Lets remove this function.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -137,141 +111,269 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
+                    switch wordValueBlockType {
+                    case .word:
+                        if let text = wordBlock.text {
+                            valueText += text + " "
                         }
                     case .selectionElement:
-                        valueSelected = keyBlock.selectionStatus == .selected ? true : false
+                        let result = processSelectionElement(selectionItemFound, selectionStatus: selectionStatus)
+                        isSelected = result.0
+                        selectionItemFound = result.1
```
wooj2, line 322: Replace with:
```
                        if (!selectionItemFound) {
                            selectionItemFound = true
                            //TODO: http://www.github.com/aws-amplify/amplify-ios/issues/<fill in issue number here>
                            // Support multiple selection items found in a KeyValueSet
                            isSelected = selectionStatus == .selected
                        } else {
                            Amplify.log.error("Multiple selection items found in KeyValueSet")
                        }
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -137,141 +111,269 @@ class IdentifyTextResultTransformers: IdentifyResultTransformers {
...
+                switch wordOrSelectionBlock?.blockType {
+                case .word:
+                    guard let text = wordOrSelectionBlock?.text else {
+                        return nil
+                    }
+                    words += text + " "
+                case .selectionElement:
+                    let result = processSelectionElement(selectionItemFound, selectionStatus: selectionStatus)
+                    isSelected = result.0
+                    selectionItemFound = result.1
```
wooj2, line 212: replace with:
```
                    if (!selectionItemFound) {
                        selectionItemFound = true
                        //TODO: <link>
                        //Support multiple selection items found in a cell
                        isSelected = selectionStatus == .selected
                    } else {
                        Amplify.log.error("Multiple selection items found in a Cell")
                    }
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/IdentifyTextResultTransformers.swift
```diff
@@ -10,6 +10,7 @@ import AWSRekognition
 import Amplify
 import AWSTextract
 
+// swiftlint:disable type_body_length
```
wooj2, line 13: If you want to avoid this, you can create a file like:

`IdentifyTextResultTransformers+Tables.swift` (and similarly with key value)

and then put:
```
extension IdentifyTextResultTransformers {
    static func processTables(...
    static func processTable(_ tableBlock: AWSText...
    static func constructTableCell(_....
}
```

PR #689, opened 2020-08-04
ci(fastlane): correct ruby block in Fastfile (missing end)

iartemiev:
* Fastfile correction found by: @kaichengyan 
* Update Fastlane and other gems


iartemiev:
@palpatim appears to just be dumb luck that it worked. The number of `end`s has not changed in this PR, just their positions, so some of the lanes were nested inside of the `increment_versions` lane but apparently that still works in fastlane. Both the prior and current versions were technically valid Ruby, just with nonsensical block nesting in the former case.

I tested this locally, ensuring that calling `fastlane unstable` and `fastlane release` still executes all of the required private lanes, etc.

PR #690, opened 2020-08-05
fix(api): cognito user pool intercept with accessToken

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/618

*Description of changes:*
This updates the Cognito User Pool interceptor to use the accessToken. Android and JS is using accessToken, while iOS was using idToken before. According to the [AWS docs](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-with-identity-providers.html), both can be passed to the service. 

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/REST/RESTWithUserPoolIntegrationTests/RESTWithUserPoolIntegrationTests.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/REST/RESTWithUserPoolIntegrationTests/RESTWithUserPoolIntegrationTests.swift
@@ -6,64 +6,49 @@
 //
 
 import XCTest
-@testable import Amplify
 import AWSAPICategoryPlugin
-import AWSMobileClient
+import AmplifyPlugins
+
+@testable import Amplify
 @testable import AmplifyTestCommon
 
 class RESTWithUserPoolIntegrationTests: XCTestCase {
+    struct User {
+        let username: String
+        let password: String
+    }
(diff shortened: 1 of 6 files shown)


AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/REST/RESTWithUserPoolIntegrationTests/RESTWithUserPoolIntegrationTests.swift
```diff
@@ -80,8 +65,8 @@ class RESTWithUserPoolIntegrationTests: XCTestCase {
         wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
     }
 
-    func testGetAPIFailedWithNotAuthenticated() {
-        AuthHelper.signOut()
+    func testGetAPIFailedWithSignedOutError() {
+        signOut()
```
royjit, line 69: User is already signout at this stage, do we need to call it again?

lawmicha: ah no need to call it again, i had this when i was debugging the tests, in a normal test run, this is covered already in the `tearDown`. removing this line

lawmicha, line 69:
```suggestion
```

PR #705, opened 2020-08-11
refactor(datastore): Remove the dependency of model type save and query in StorageEngine

royjit:
*Description of changes:*

Removes the dependency on model type in the StorageEngine apis `save` and `query`. Added a new protocol called `JSONValueHolder`, the concrete implementation of this protocol will holds the internal data in a json format. `JSONValueHolder.jsonValue(for:)` can be used to get the value for a particular key inside the json map. 

For example developer can create the following struct using the `JSONValueHolder`.

```swift

struct DynamicModel: JSONValueHolder {

    let values: [String: JSONValue]

    func jsonValue(for key: String) -> Any?? {
        switch values[key] {
        case .some(.array(let deserializedValue)):
            return deserializedValue
        case .some(.boolean(let deserializedValue)):
            return deserializedValue
        case .some(.number(let deserializedValue)):
            return deserializedValue
        case .some(.object(let deserializedValue)):
            return deserializedValue
        case .some(.string(let deserializedValue)):
            return deserializedValue
        case .some(.null):
            return .some(nil)
        case .none:
            return nil
        }
    }
}


```

### DIFF
--- a/Amplify/Categories/DataStore/Model/Internal/ModelRegistry.swift
+++ b/Amplify/Categories/DataStore/Model/Internal/ModelRegistry.swift
@@ -35,17 +35,11 @@ public struct ModelRegistry {
     }
 
     public static func register(modelType: Model.Type) {
-        concurrencyQueue.sync {
-            let modelDecoder: ModelDecoder = { jsonString, jsonDecoder in
-                let model = try modelType.from(json: jsonString, decoder: jsonDecoder)
-                return model
-            }
-
-            modelDecoders[modelType.modelName] = modelDecoder
-
-            modelTypes[modelType.modelName] = modelType
-
-            modelSchemaMapping[modelType.modelName] = modelType.schema
+        register(modelName: modelType.modelName,
+                 modelSchema: modelType.schema,
+                 modelType: modelType) { (jsonString, jsonDecoder) -> Model in
+            let model = try modelType.from(json: jsonString, decoder: jsonDecoder)
+            return model
         }
     }
 
--- a/Amplify/Categories/DataStore/Model/Internal/ModelRegistry.swift
+++ b/Amplify/Categories/DataStore/Model/Internal/ModelRegistry.swift
@@ -49,6 +49,21 @@ public struct ModelRegistry {
         }
     }
 
+    public static func register(modelName: ModelName,
+                                modelSchema: ModelSchema,
+                                modelType: Model.Type,
+                                jsonDecoder: @escaping (String, JSONDecoder?) throws -> Model) {
+        concurrencyQueue.sync {
+            let modelDecoder: ModelDecoder = { jsonString, decoder in
+                return try jsonDecoder(jsonString, decoder)
+            }
+
+            modelSchemaMapping[modelName] = modelSchema
+            modelTypes[modelName] = modelType
+            modelDecoders[modelName] = modelDecoder
+        }
+    }
+
     public static func modelType(from name: ModelName) -> Model.Type? {
         concurrencyQueue.sync {
             modelTypes[name]
--- a/Amplify/Categories/DataStore/Model/Model.swift
+++ b/Amplify/Categories/DataStore/Model/Model.swift
@@ -30,3 +30,16 @@ public protocol Model: Codable {
     var id: Identifier { get }
 
 }
+
+public protocol JsonModel: Model {
+
+    /// Return the value for key
+    func internal_value(for key: String) -> Any?
+}
+
+extension JsonModel {
+
+    public func internal_value(for key: String) -> Any? {
+        return nil
+    }
+}
--- a/Amplify/Categories/DataStore/Model/Model.swift
+++ b/Amplify/Categories/DataStore/Model/Model.swift
@@ -31,15 +31,15 @@ public protocol Model: Codable {
 
 }
 
-public protocol JsonModel: Model {
+public protocol JSONModel: Model {
 
     /// Return the value for key
-    func internal_value(for key: String) -> Any?
+    func internalValue(for key: String) -> Any?
 }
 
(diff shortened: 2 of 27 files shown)


Amplify/Categories/DataStore/Model/Internal/ModelRegistry.swift
```diff
@@ -49,6 +49,21 @@ public struct ModelRegistry {
         }
     }
 
+    public static func register(modelName: ModelName,
```
palpatim, line 52: Let's refactor `register(modelType:)` to invoke this method to make sure all our "initialization" code goes through the same path.

Amplify/Categories/DataStore/Model/Model.swift
```diff
@@ -30,3 +30,16 @@ public protocol Model: Codable {
     var id: Identifier { get }
 
 }
+
+public protocol JsonModel: Model {
+
+    /// Return the value for key
+    func internal_value(for key: String) -> Any?
```
palpatim, line 37: 1. Whence the underscore? This shouldn't even pass linting
1. I'm not really sure what an "internal value" means. What is this actually doing?
1. Follow Swift naming conventions for the protocol: `JSONModel`, but what about this protocol is specific to the JSON format? What is it really doing?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Model+SQLite.swift
```diff
@@ -101,7 +100,7 @@ extension Model {
                 return binding
             } catch {
                 logger.warn("""
-                Error converting \(modelType.modelName).\(field.name) to the proper SQLite Binding.
+                    Error converting \(modelSchema.name).\(field.name) to the proper SQLite Binding.
```
palpatim, line 103: nit: Fix indentation

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+UntypedModel.swift
```diff
@@ -25,7 +25,7 @@ extension SQLiteStorageEngineAdapter {
                 let statement = UpdateStatement(model: untypedModel, modelSchema: modelType.schema)
```
palpatim, line 25: Should we be using `untypedModel.schema` here as well?

palpatim (changes requested):

Amplify/Categories/DataStore/Model/Model.swift
```diff
@@ -30,3 +30,16 @@ public protocol Model: Codable {
     var id: Identifier { get }
 
 }
+
+public protocol JSONModel: Model {
```
palpatim, line 34: 1. As discussed, let's rename this protocol to `JSONValueAccessible` or similar (JSONValueAddressable? JSONValueHolder? JSONValueProperties?). Basically, we want to promise that properties of any type conforming to this protocol can be retrieved with the `jsonValue(forKey:)` method, and that the return value is a `JSONValue?`.
1. This shouldn't derive from Model; it's a separate contract, and you could imagine any type conforming to it, regardless of whether it's also a model. Conforming types that are also Models can conform to both.
1. Probably belongs in a separate file, since it's not a Model.

Amplify/Categories/DataStore/Model/Model.swift
```diff
@@ -30,3 +30,16 @@ public protocol Model: Codable {
...
+
+    /// Return the value for key
+    func internalValue(for key: String) -> Any?
```
palpatim, line 37: 1. As discussed, let's rename this method to `jsonValue(forKey:)`
1. As discussed, this should be returning a `JSONValue?` not an `Any?`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Models/DynamicModel.swift
```diff
@@ -0,0 +1,65 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+
+struct DynamicModel: JSONModel {
```
palpatim, line 10: 1. As noted above, let's make sure this conforms to both `Model` and `JSONAccessible` (or whatever)
1. As discussed, please document this as an example, although I should have caught this by the fact that it was in the Tests package.

palpatim (commented):
As we discussed, let's make the JSON value accessor stuff less closely tied to model, and more universally applicable

Amplify/Categories/DataStore/Model/Model.swift
royjit: 1. Renamed to JSONValueHolder, and it returns Any?? similar to subscript in Model
2. Removed
3. Moved

Amplify/Categories/DataStore/Model/Model.swift
royjit: 1. Renamed
2. Returns an Any??,  should we enforce the concrete class to use `JSONValue` as its internal implementation? They can use their own implementation of JSONValue right? Any gives them the flexibility?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Models/DynamicModel.swift
royjit: 1. Changed
2. Will add

palpatim (approved):

PR #708, opened 2020-08-12
ci(fastlane): fix retry logic and perform a git fetch before lane

iartemiev:
Get latest tags before inferring next version (to prevent race condition in simultaneously running pipelines)

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -6,6 +6,12 @@ default_platform(:ios)
 PLIST_KEY = "CFBundleShortVersionString"
 
 platform :ios do
+  before_all do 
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -6,6 +6,12 @@ default_platform(:ios)
 PLIST_KEY = "CFBundleShortVersionString"
 
 platform :ios do
+  before_all do 
+    # Perform a fetch before inferring the next version 
+    # to prevent race conditions with simultaneous pipelines attempting to create the same tag
```
palpatim, line 11: nit: s/prevent/reduce/

PR #711, opened 2020-08-12
feat(api): Adding custom 'items' deserialization for List

lawmicha:
*Issue #, if available:*

*Description of changes:*
Models with 1-to-many connections contain a List, for example
```
public struct Post: Model {
  public var comments: List<Comment>?
```
The post contains a list of comments. When constructing a Get Query for a Post, the Model-to-GraphQLRequest builders will only generate the first depth of a selection set since comments is not a required association. If the developer wants to retrieve the comments for that particular Post, The developer can create a custom selection set for the GetQuery containing nested selection set with the comments. 
```
query getPost($id: ID!) {
          getPost(id: $id){
            id
            title
            content
            createdAt
            updatedAt
            draft
            rating
            status
            comments {
              items {
                id
                content
                createdAt
                updatedAt
                post {
                  id
                  title
                  content
                  createdAt
                  updatedAt
                  draft
                  rating
                  status
                }
              }
              nextToken
            }
          }
        }
```
Then construct a GraphQLRequest like
```
let graphQLRequest = GraphQLRequest(document: document,
                                            variables: ["id": uuid],
                                            responseType: Post?.self,
                                            decodePath: "getPost")
```

### DIFF
--- a/Amplify/Categories/DataStore/Model/Collection/List+Model.swift
+++ b/Amplify/Categories/DataStore/Model/Collection/List+Model.swift
@@ -112,6 +112,17 @@ public class List<ModelType: Model>: Collection, Codable, ExpressibleByArrayLite
                 let field = Element.schema.field(withName: associatedField)
                 // TODO handle eager loaded associations with elements
                 self.init([], associatedId: associatedId, associatedField: field)
+            } else if case let .array(jsonArray) = list["items"] {
+                let encoder = JSONEncoder()
+                encoder.dateEncodingStrategy = ModelDateFormatting.encodingStrategy
+                let decoder = JSONDecoder()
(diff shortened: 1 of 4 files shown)


richardmcclellan:
This looks great!  This is very similar to the Android implementation.    

One minor thing to note is that this ignores the `nextToken` associated with lists of Models that are not the root Model being returned, so customers won't be able to paginate a list of child Models.  For example, if a customer requests a `Post`, and it has a list of `Comment`s, the customer won't be able to paginate to get more `Comment`s.  They can however make a new request for the Comments and filter for the id of the Post, and then paginate from there.   Android has this issue too.

palpatim:
Released in https://github.com/aws-amplify/amplify-ios/releases/tag/v1.1.2

PR #714, opened 2020-08-13
feat(api): default decodePath behavior

lawmicha:
*Issue #, if available:*

*Description of changes:*
GraphQL responses are returned in the format of
```
{ "data": { }
  "errors": [ ] }
```
A GraphQL request can specify a `decodePath` to extract a specific part of the object at `data` out from the response. In Android, the default behavior is to take the first element at the `data` object. See [comment](https://github.com/aws-amplify/docs/pull/2141#discussion_r462599129)

Before, we would always have to use the decodePath, for example

```
{
 "data": {
   "getTodo": {
      "id": 111
  }
 }
}
```

will require a decodePath = "getTodo" to get to the object `{ "id": 111}` and deserialize successfully when the Todo Model is passed as `responseType`. This PR first
- defaults to the first value of the object under "data" to automatically retrieve `{ "id": 111 }`.

*Customer impact*
- If the customer is using Model-based GraphQLRequest builders - no change, decodePath is removed from the internal implementation
- If the customer is building their own GraphQLRequest with decodePath - no change required, however if they are deserializing to a single model, which is usually the case when using GraphQLTransform provisioned API, then they can also remove `decodePath` which is in alignment with Android
- If customer is building their own GraphQLRequest without decodePath and deserialize to a single object - Change is required since now the nested object is returned so their responseType should be updated, for example
```
struct TodoResponse: Decodable {
    let getTodo: Todo
    struct Todo: Decodable {
      let id: String
    }
}
```
the response type `let responseType = TodoResponse.self` can be updated to `let responseType = Todo.self`

### DIFF
--- a/AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
@@ -29,7 +29,7 @@ protocol ModelGraphQLRequestFactory {
     ///
     /// - seealso: `GraphQLQuery`, `GraphQLQueryType.list`
     static func list<M: Model>(_ modelType: M.Type,
-                               where predicate: QueryPredicate?) -> GraphQLRequest<List<M>>
+                               where predicate: QueryPredicate?) -> GraphQLRequest<[M]>
 
     /// Creates a `GraphQLRequest` that represents a query that expects a single value as a result.
     /// The request will be created with the correct correct document based on the `ModelSchema` and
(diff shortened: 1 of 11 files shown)


AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
```diff
@@ -164,12 +163,11 @@ extension GraphQLRequest: ModelGraphQLRequestFactory {
...
         return GraphQLRequest<M?>(document: document.stringValue,
                                   variables: document.variables,
-                                  responseType: M?.self,
-                                  decodePath: document.name)
+                                  responseType: M?.self)
     }
 
     public static func list<M: Model>(_ modelType: M.Type,
-                                      where predicate: QueryPredicate? = nil) -> GraphQLRequest<[M]> {
+                                      where predicate: QueryPredicate? = nil) -> GraphQLRequest<List<M>> {
```
lawmicha, line 170: reverting this change as this changes the API responseType drastically

lawmicha:
we have decided to hold off on this due to the breaking change on customers (3rd point under customer impact, in the description)

PR #715, opened 2020-08-13
feat(api): PaginatedResult for ModelBased list query

lawmicha:
*Issue #, if available:*

*Description of changes:*
The Model-to-GraphQLRequest builders provides a `.list` query that returns a Collection. This is missing a mechanism for handling pagination from the AppSync API. By providing the `PaginatedResult` return type, the developer can use this to get the next request and perform the next API call to get the next page of results.
- The `PaginatedResult` class is introduced in AWSPluginsCore, with three methods on it: getItems, hasNextResult, and getRequestForNextRequest. Same as [Android docs](https://docs.amplify.aws/lib/graphqlapi/query-data/q/platform/android#list-multiple-pages-of-items)
- The `limit` parameter is added to the static builder for `.list`.

*Customer impact*
So far, customers currently using `.list` to create their GraphQLRequest will receive the first 1000 items (less if they applied a filter, since the filter is applied post-dynamodb-query). This introduces a new static builder `.paginatedList` which returns a `PaginatedResult`. We could also deprecate by removing the old `.list` and to keep the name the same as `list` and will require developers to update how they handle the response 

### DIFF
--- a/Amplify.xcodeproj/project.pbxproj
+++ b/Amplify.xcodeproj/project.pbxproj
@@ -79,6 +79,7 @@
 		214F49CE24898E8500DA616C /* Article+Schema.swift in Sources */ = {isa = PBXBuildFile; fileRef = 214F49CC24898E8500DA616C /* Article+Schema.swift */; };
 		21558E3E237BB4BF0032A5BB /* GraphQLRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21558E3D237BB4BF0032A5BB /* GraphQLRequest.swift */; };
 		21558E40237CB8640032A5BB /* GraphQLError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21558E3F237CB8640032A5BB /* GraphQLError.swift */; };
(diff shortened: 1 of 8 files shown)


richardmcclellan:
On Android, we don't have a separate `paginatedList` method.  Instead, we just have several different overloads of `list`: methods:

```
public static <M extends Model> GraphQLRequest<Iterable<M>> list(Class<M> modelType, QueryPredicate predicate)
public static <M extends Model> GraphQLRequest<Iterable<M>> list(@NonNull Class<M> modelType)
public static <M extends Model> GraphQLRequest<PaginatedResult<M>> list(Class<M> modelType, QueryPredicate predicate, ModelPagination pagination)
public static <M extends Model> GraphQLRequest<PaginatedResult<M>> list(Class<M> modelType, ModelPagination pagination)
```

Essentially, if the `ModelPagination` object parameter is provided, then the response type is a `PaginatedResult`.  If not, the response type is an `Iterable`.  `ModelPagination` is just a wrapper around a `limit`.

Let me know if a similar interface would be possible in Swift, without making any breaking changes.

lawmicha:
closing for https://github.com/aws-amplify/amplify-ios/pull/806

PR #718, opened 2020-08-17
refactor(datastore): Removed the dependency on Model.Type from DataStoreStatement

royjit:
*Description of changes:*

DatastoreStatement that communicate with the underlying storage doesnot need a dependency on Model.Type. This PR is part of the larger work on removing the dependency on the static properties of Model.Type from Datastore.

Instead of Model.Type DatastoreStatement will be using ModelSchema. The `Model.Type` property in `DatastoreStatement` is not used anywhere externally before this change. So make this change should not affect any public apis.

*Current usage of DatastoreStatement:*

`DatastoreStatement` protocol is currently only used as a parent protocol of [`SQLStatement`](https://github.com/aws-amplify/amplify-ios/blob/d88447789d26f042e26f70565a9283dc143a8b22/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement.swift#L15). And `SQLStatement` is adopted by internal structs - `ConditionStatement`, `CreateTableStatement`, `DeleteStatement`, `InsertStatement`, `SelectStatement`and `UpdateStatement`. None of these struct's object uses `Model.Type` where the object is used.

### DIFF
--- a/Amplify/Categories/DataStore/DataStoreStatement.swift
+++ b/Amplify/Categories/DataStore/DataStoreStatement.swift
@@ -15,8 +15,8 @@ public protocol DataStoreStatement {
     /// The type of the variables container related to a concrete statement implementation
     associatedtype Variables
 
-    /// The type of the `Model` associated with a particular statement
-    var modelType: Model.Type { get }
+    /// The model schema of the `Model` associated with a particular statement
+    var modelSchema: ModelSchema { get }
 
     /// The actual string content of the statement (e.g. a SQL query or a GraphQL document)
     var stringValue: String { get }
--- a/Amplify/Categories/DataStore/DataStoreStatement.swift
+++ b/Amplify/Categories/DataStore/DataStoreStatement.swift
@@ -15,6 +15,13 @@ public protocol DataStoreStatement {
     /// The type of the variables container related to a concrete statement implementation
     associatedtype Variables
 
+    /// The type of the `Model` associated with a particular statement
+    @available(*, deprecated, message: """
+        Use of modelType inside the DatastoreStatement is deprecated, instead make use of
+        `ModelSchema` through the `modelSchema` property.
+    """)
+    var modelType: Model.Type { get }
+
     /// The model schema of the `Model` associated with a particular statement
     var modelSchema: ModelSchema { get }
 
--- a/Amplify/Categories/DataStore/DataStoreStatement.swift
+++ b/Amplify/Categories/DataStore/DataStoreStatement.swift
@@ -17,9 +17,9 @@ public protocol DataStoreStatement {
 
     /// The type of the `Model` associated with a particular statement
     @available(*, deprecated, message: """
-        Use of modelType inside the DatastoreStatement is deprecated, instead make use of
-        `ModelSchema` through the `modelSchema` property.
+    Use of modelType inside the DatastoreStatement is deprecated, use modelSchema instead.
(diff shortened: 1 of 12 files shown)


Amplify/Categories/DataStore/DataStoreStatement.swift
```diff
@@ -15,8 +15,8 @@ public protocol DataStoreStatement {
     /// The type of the variables container related to a concrete statement implementation
     associatedtype Variables
 
-    /// The type of the `Model` associated with a particular statement
-    var modelType: Model.Type { get }
+    /// The model schema of the `Model` associated with a particular statement
```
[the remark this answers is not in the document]
royjit: Added back with deprecated warning.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+CreateTable.swift
```diff
@@ -9,22 +9,20 @@ import Amplify
...
-    init(modelType: Model.Type) {
-        self.modelType = modelType
+    init(modelSchema: ModelSchema) {
+        self.modelSchema = modelSchema
     }
 
     var stringValue: String {
-        let schema = modelType.schema
-        let name = schema.name
+        let name = modelSchema.name
```
palpatim, line 21: Did you verify that none of our internal types that conform to `Model` (I'm thinking specifically about `AnyModel`) will have problems with this? I think it's OK, since we don't actually persist any of those types, but a check for any internal uses would be good.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement.swift
```diff
@@ -22,3 +22,11 @@ extension SQLStatement {
         return []
     }
 }
+
+extension SQLStatement {
+
+    // `modelType` is deprecated, instead modelSchema should be used.
+    var modelType: Model.Type {
+        fatalError("Use of modelType is deprecated in SQLStatement")
```
palpatim, line 30: Let's add a bit more context here, along the lines of "`DataStoreStatement.modelType` is deprecated. `SQLStatement` is an internal type and there should be no references to `modelType`. If you encounter this error, please open a GitHub issue <link>."

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement.swift
```diff
@@ -27,6 +27,9 @@ extension SQLStatement {
 
     // `modelType` is deprecated, instead modelSchema should be used.
     var modelType: Model.Type {
-        fatalError("Use of modelType is deprecated in SQLStatement")
+        fatalError("""
+        DataStoreStatement.modelType is deprecated. SQLStatement is an internal type and there should be \
+        no references to modelType. If you encounter this error, please open a GitHub issue.
```
palpatim, line 32: Please include URL to open a GH issue.

royjit: Added

PR #721, opened 2020-08-17
feat(datastore): Dispatch outboxStatus, subscriptionsEstablished, syncQueriesStarted events

DongQuanRui:
*Issue #, if available:*

The rest of 6 events will be implemented in other PRs.
- `Network Status`
- `modelSynced`
- `syncQueriesReady`
- `ready`
- `outboxMutationEnqueued`
- `outboxMutationProcessed`

*Description of changes:*

Implemented three of nine hub events 
- `outboxStatus`
- `syncQueriesStarted`
- `subscriptionEstablished`

### outboxStatus
```
{"isEmpty": true/false}

public struct {
    let isEmpty: Bool
}
```

### subscriptionsEstablished
```
N/A
```
When all subscriptions have been established, 3 per model, for every model, then emit a single subscriptionsEstablished event

### syncQueriesStarted
```
{"models":["Task","Note"]}

struct SyncQueriesStartedEvent {
    let models: [String]
}
```


Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
```diff
@@ -23,4 +23,22 @@ public extension HubPayload.EventName.DataStore {
...
+
+    static let modelSynced = "DataStore.modelSynced"
+
+    static let syncQueriesReady = "DataStore.syncQueriesReady"
+
+    static let ready = "DataStore.ready"
+
+    static let outboxMutationEnqueued = "DataStore.outboxMutationEnqueued"
+
+    static let outboxMutationProcessed = "DataStore.outboxMutationProcessed"
```
palpatim, line 43: 1. Please add doc comments (like above on `conditionalSaveFailed`) describing the conditions under which these events are emitted, so a customer will know how to use them
2. I have concerns about the naming of these events
    1. Can you confirm that these are the same events as are emitted by Android & JS? I'm particularly concerned about `syncQueriesReady`--does that mean "the sync queries are ready to be executed" or "the sync queries have finished executing"?
    2. Can you confirm that the events are named the same (within platform conventions) as Android & JS?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -91,6 +91,13 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         var token: UnsubscribeToken?
         token = Amplify.Hub.listen(to: .dataStore, isIncluded: filter) { _ in
             self.storageEngine.startSync()
+
+            let payloadOfOutboxStatus = HubPayload(eventName: HubPayload.EventName.DataStore.outboxStatusChanged)
+            Amplify.Hub.dispatch(to: .dataStore, payload: payloadOfOutboxStatus)
+
+            let payload = HubPayload(eventName: HubPayload.EventName.DataStore.ready)
+            Amplify.Hub.dispatch(to: .dataStore, payload: payload)
+
```
palpatim, line 100: Once you update the comments explaining when these events are triggered, we may need to discuss why we emit an "outbox status changed" event with no indication as to what the new status is. How is a customer intended to inspect the status of the "outbox" (by which I assume you mean the outgoing mutation queue)?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
```diff
@@ -16,6 +16,7 @@ final class InitialSyncOperation: AsynchronousOperation {
     private weak var reconciliationQueue: IncomingEventReconciliationQueue?
     private weak var storageAdapter: StorageEngineAdapter?
     private let dataStoreConfiguration: DataStoreConfiguration
+    private var modelSyncedPayload: ModelSyncedPayload?
```
palpatim, line 19: 1. `Payload` isn't terribly meaningful here. I think you're using it to report the result of the initial sync operation, so `OperationResult` might be clearer
2. I think you're using this type to report the status of the entire initial sync operation. Are you using it for other contexts? If not, let's call it something like `InitialSyncOperationResult`. If you are using it elsewhere then pick a name that reflects its intent. "Model synced" is ambiguous: are you talking about emitting one of these each time a model instance is synced, or when the initial sync is finished for a single model type, or when the initial sync is finished for all model types? 
3. Why is this optional? You instantiate it in the initializer, and I don't see you setting it to nil anywhere.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
```diff
@@ -157,7 +159,23 @@ final class InitialSyncOperation: AsynchronousOperation {
         let items = syncQueryResult.items
         recordsReceived += UInt(items.count)
```
palpatim, line 160: It seems like we can replace `recordsReceived` with `modelSyncedPayload` or a calculated property thereof.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
```diff
@@ -157,7 +159,23 @@ final class InitialSyncOperation: AsynchronousOperation {
...
+            modelSyncedPayload?.isFullSync = true
+            modelSyncedPayload?.isDeltaSync = false
+        }
+
         for item in items {
+            let itemMetadata = item.syncMetadata
+            if itemMetadata.deleted {
+                modelSyncedPayload?.deleteCount += 1
+            } else if itemMetadata.version == 1 {
+                modelSyncedPayload?.createCount += 1
```
palpatim, line 175: This goes back to a previous discussion: What does "create" mean in this context: that the incoming instance has never been seen *locally*, or that the incoming instance has been created but not updated *on the service*. In other words, in the following case:

- Create Instance1 on console; version == 1
- Update Instance1 on console; version == 2
- Launch app, synchronize
- Receive Instance1; version == 2

Should that be counted as a `create` or an `update`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
```diff
@@ -188,7 +206,7 @@ final class InitialSyncOperation: AsynchronousOperation {
             case .failure(let dataStoreError):
                 self.finish(result: .failure(dataStoreError))
             case .success:
-                self.finish(result: .successfulVoid)
+                self.finish(result: .success(self.modelSyncedPayload))
```
palpatim, line 209: If you make `modelSyncedPayload` non-optional, you may be able to change the signature of the completion block to take a non-optional modelSyncedPayload--I'm not sure where else you use the callback though.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -9,7 +9,7 @@ import Amplify
 import AWSPluginsCore
 
 protocol InitialSyncOrchestrator {
-    func sync(completion: @escaping (Result<Void, DataStoreError>) -> Void)
+    func sync(completion: @escaping (Result<ModelSyncedPayload?, DataStoreError>) -> Void)
```
palpatim, line 12: As noted above, you may be able to make the success case a non-optional `ModelSyncedPayload`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/ModelSyncedPayload.swift
```diff
@@ -0,0 +1,31 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+struct ModelSyncedPayload {
```
palpatim, line 10: Since you're emitting this on the Hub, this needs to be a public type. Each member should also be public. It isn't strictly necessary to make the initializer public, but it's good practice.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/ModelSyncedPayload.swift
```diff
@@ -0,0 +1,31 @@
...
+    var modelName: String
+    var isFullSync: Bool
+    var isDeltaSync: Bool
```
palpatim, line 13: Do the other platforms have two properties to represent this? These are mutually exclusive states; we should either model it as a single bool, or an enum with two cases

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/ModelSyncedPayload.swift
```diff
@@ -0,0 +1,31 @@
...
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+struct ModelSyncedPayload {
+    var modelName: String
```
palpatim, line 11: Favor immutable properties. This may cause you to have to do some additional work in the initial sync operation as you're tallying the counts, but when we emit the payload to the Hub, it should be immutable.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/ModelSyncedPayload.swift
```diff
@@ -0,0 +1,31 @@
...
+         createCount: Int? = nil,
+         updateCount: Int? = nil,
+         deleteCount: Int? = nil) {
+        self.modelName = modelName ?? ""
+        self.isFullSync = isFullSync ?? false
+        self.isDeltaSync = isDeltaSync ?? false
+        self.createCount = createCount ?? 0
+        self.updateCount = updateCount ?? 0
+        self.deleteCount = deleteCount ?? 0
+    }
```
palpatim, line 30: 1. Simplify this by declaring the defaults in the initializer signature and making the argument types non-optional:

```suggestion
    init(modelName: String = "",
         isFullSync: Bool = false,
         isDeltaSync: Bool = false,
         createCount: Int = 0,
         updateCount: Int = 0,
         deleteCount: Int = 0) {
        self.modelName = modelName
        self.isFullSync = isFullSync
        self.isDeltaSync = isDeltaSync
        self.createCount = createCount
        self.updateCount = updateCount
        self.deleteCount = deleteCount
    }
```

2. But even with this, some of these defaults don't really make sense. When would `modelName` be nil or empty? As noted above, what would it mean to have `isFullSync` and `isDeltaSync` both be true or both be false?

My guess is that some of these decisions are so you can build this incrementally inside the operation. In that case, consider a `Builder` pattern, like I've sketched out here:

```swift
public struct ModelSyncedPayload {
    // For internal use only
    enum SyncType {
        case full
        case incremental
        case unspecified
    }
    let syncType: SyncType

    public let modelName: String

    // To conform with other platform specs, if needed
    public var isFullSync: Bool {
        syncType == .full
    }

    public var isDeltaSync: Bool {
        syncType == .incremental
    }

    public let createCount: Int
    public let updateCount: Int
    public let deleteCount: Int

    public init(
        modelName: String,
        syncType: SyncType,
        createCount: Int,
        updateCount: Int,
        deleteCount: Int
    ) {
        self.modelName = modelName
        self.syncType = syncType
        self.createCount = createCount
        self.updateCount = updateCount
        self.deleteCount = deleteCount
    }
}

extension ModelSyncedPayload {
    struct Builder {
        var modelName: String
        var syncType: SyncType
        var createCount: Int
        var updateCount: Int
        var deleteCount: Int
        init () {
            modelName = ""
            syncType = .unspecified
            createCount = 0
            updateCount = 0
            deleteCount = 0
        }

        func build() -> ModelSyncedPayload {
            ModelSyncedPayload(
                modelName: modelName,
                syncType: syncType,
                createCount: createCount,
                updateCount: updateCount,
                deleteCount: deleteCount
            )
        }
    }
}

func atCallSite() {
    var builder = ModelSyncedPayload.Builder()
    builder.modelName = "Foo"
    builder.createCount += 1
    builder.syncType = .full
    // ...
    // callback(.success(builder.build()))
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/TestSupport/Mocks/MockAWSInitialSyncOrchestrator.swift
```diff
@@ -22,7 +22,7 @@ class MockAWSInitialSyncOrchestrator: InitialSyncOrchestrator {
                                        storageAdapter: storageAdapter)
     }
 
-    typealias SyncOperationResult = Result<Void, DataStoreError>
+    typealias SyncOperationResult = Result<ModelSyncedPayload?, DataStoreError>
```
palpatim, line 25: As above: it looks like this can be non-optional

Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
DongQuanRui: 1. Will add comments for them.

2. Rafael J and I had the same concerns on naming. 
Say, `syncQueriesReady ` would better be `syncQueriesCompleted` or `syncQueriesFinished`.
But I have synced up with Manuel 2 weeks ago and Rafael J yesterday. The naming are the same for 3 platform.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/ModelSyncedPayload.swift
DongQuanRui: https://github.com/aws-amplify/amplify-js/blob/eb9719467f7c143a26bfd2d598b3903ff0815731/packages/datastore/src/sync/index.ts#L569
https://github.com/aws-amplify/amplify-android/pull/710/files#diff-bad527d7894aa87e0315e7104f460145R53

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
DongQuanRui: I also synced up with Rafael on this:
From the perspective of local store, the record is new (no matter is the first time to save it or to delete or save it again)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
DongQuanRui: But if I make it non-optional, line here should something back to caller (`RemoteSyncEngine`): https://github.com/aws-amplify/amplify-ios/blob/e6db87a8992926e8fabfcfd0c8953adc65d794b2/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift#L125
I think that's not I want since the caller (`RemoteSyncEngine` class doesn't care about `ModelSyncedPayload`). And that's why I make `ModelSyncedPayload` as an optional value to return.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
DongQuanRui: See: https://github.com/aws-amplify/amplify-ios/pull/721#discussion_r477581573

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
DongQuanRui: 2. when `initialSyncOperaion` for each model is finished, emit the event `ModelSynced`. But in my point of view, if it is called `InitialSyncOperationResult` kind of meaning the result should contains a lot of info since `initialSyncOperaion` does more than simply generating a payload? But maybe call `ModelSyncedResult` which is the result of this specific `ModelSynced` Hub event? What do you think?
3. Ya, this can be non-optional. Let me change it

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
```diff
@@ -157,7 +159,23 @@ final class InitialSyncOperation: AsynchronousOperation {
         let items = syncQueryResult.items
         recordsReceived += UInt(items.count)
 
+        if lastSyncTime != nil {
+            modelSyncedPayload.isFullSync = false
+            modelSyncedPayload.isDeltaSync = true
+        } else {
+            modelSyncedPayload.isFullSync = true
+            modelSyncedPayload.isDeltaSync = false
+        }
```
drochetti, line 168: you can remove the `if/else` entirely:

```swift
modelSyncedPayload.isFullSync = lastSyncTime == nil
modelSyncedPayload.isDeltaSync = lastSyncTime != nil
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/ModelSyncedPayload.swift
palpatim: As discussed, we will leave these as two properties on the public payload, although we'll represent them in the builder with an enum

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
DongQuanRui: Thanks, will update it, but the `ModelSyncedResult` should not be dispatch in this class.

What is made clear in today's (Sep, 1st, 2020) meeting, Android and iOS both should update the implementation of event `modelSynced`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
```diff
@@ -180,31 +180,33 @@ class DataStoreEndToEndTests: SyncEngineIntegrationTestBase {
...
-                    XCTFail("Can't cast payload as mutation event")
-                    return
-            }
-
-            // This check is to protect against stray events being processed after the test has completed,
-            // and it shouldn't be construed as a pattern necessary for production applications.
-            guard let post = try? mutationEvent.decodeModel() as? Post, post.id == newPost.id else {
-                return
-            }
+            if payload.eventName == HubPayload.EventName.DataStore.syncReceived || payload.eventName == HubPayload.EventName.DataStore.conditionalSaveFailed {
```
DongQuanRui, line 183: I put a `if` statement here as a filter filtering `syncReceived` and `conditionalSaveFailed` events since we have added 9 more events.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -71,10 +72,11 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
         }
 
         syncOperationQueue.isSuspended = false
+
+        dispatchSyncQueriesStarted(syncableModels)
```
lawmicha, line 76: thinking about this some more and i think we should actually do the signal to say "sync queries started" and immediately release the suspend on the operation queue. This way, the sync query is not happening concurrent to the dispatch. 

```
dispatchSyncQueriesStarted(syncableModels)
syncOperationQueue.isSuspended = false
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -118,6 +120,15 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
         completion?(.successfulVoid)
     }
 
+    private func dispatchSyncQueriesStarted(_ syncableModels: [Model.Type]) {
+        let modelTask = syncableModels.map { $0.modelName }
```
lawmicha, line 124: let modelNames =

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -118,6 +120,15 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
...
+
+        let syncQueriesStartedEvent = SyncQueriesStartedEvent(models: modelTask)
+        let syncQueriesStartedpayload = HubPayload(eventName: HubPayload.EventName.DataStore.syncQueriesStarted,
```
lawmicha, line 127: `syncQueriesStartedEventPayload`

lawmicha (commented):
review for `syncQueriesReady`

Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
```diff
@@ -23,4 +23,44 @@ public extension HubPayload.EventName.DataStore {
...
+    /// The HubPayload will be a boolean value `isActive` indicating the status of network
+    // TODO: networkStatusChanged to be implemented
+    static let networkStatusChanged = "DataStore.networkStatusChanged"
+
+    /// Dispatched on DataStore start and also every time a local mutation is enqueued and processed in the outbox
+    /// HubPayload `OutboxStatusEvent` contains a boolean value `isEmpty` to notify if there are mutations in the outbox
+    static let outboxStatus = "DataStore.outboxStatus"
+
+    /// Dispatched when all the graphql subscriptions estabilished
+    static let subscriptionEstablished = "DataStore.subscriptionEstablished"
```
lawmicha, line 37: I think it is `subscriptionsEstablished` with a `s`, pluralized subscription

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
DongQuanRui: Makes sense.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
DongQuanRui: Updated, also updated `outboxStatusPayload` to `outboxStatusEventPayload` in `outgoingMutationQueue.swift`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
DongQuanRui: updated

lawmicha (commented):
review `subscriptionsEstablished `

Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
DongQuanRui: Ya, hahaha, code review becomes grammar review

lawmicha (commented):
partial review for `outboxStatus`

Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
```diff
@@ -23,4 +23,44 @@ public extension HubPayload.EventName.DataStore {
...
+    /// Dispatched when network status has changed (active or not)
+    /// The HubPayload will be a boolean value `isActive` indicating the status of network
+    // TODO: networkStatusChanged to be implemented
+    static let networkStatusChanged = "DataStore.networkStatusChanged"
+
+    /// Dispatched on DataStore start and also every time a local mutation is enqueued and processed in the outbox
+    /// HubPayload `OutboxStatusEvent` contains a boolean value `isEmpty` to notify if there are mutations in the outbox
+    static let outboxStatus = "DataStore.outboxStatus"
+
+    /// Dispatched when all the graphql subscriptions estabilished
```
lawmicha, line 36: - `Dispatched when all of the subscriptions to syncable models have been established`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -188,6 +192,12 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
                 self.processSyncMutationToCloudResult(result, mutationEvent: mutationEvent, api: api)
         }
         operationQueue.addOperation(syncMutationToCloudOperation)
+
+        let outboxStatusEvent = OutboxStatusEvent(isEmpty: operationQueue.operationCount == 0 ? true : false)
+        let outboxStatusEventPayload = HubPayload(eventName: HubPayload.EventName.DataStore.outboxStatus,
+                                               data: outboxStatusEvent)
+        Amplify.Hub.dispatch(to: .dataStore, payload: outboxStatusEventPayload)
```
lawmicha, line 199: - can encapsulate into `dispatchOutboxStatusEvent(isEmpty: false)` and called right before adding the operation to operationQueue.
- it should always be true because you are about to add an operation into the operation queue. even if you dispatch after adding it in, most likely the count will be none zero but then you are concurrently executing the operation and performing a check on the operationQueue count.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -241,6 +251,11 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
                 self.log.verbose("mutationEvent deleted successfully")
             }
 
+            let outboxStatusEvent = OutboxStatusEvent(isEmpty: self.operationQueue.operationCount == 0 ? true : false)
+            let outboxStatusEventPayload = HubPayload(eventName: HubPayload.EventName.DataStore.outboxStatus,
+                                                   data: outboxStatusEvent)
+            Amplify.Hub.dispatch(to: .dataStore, payload: outboxStatusEventPayload)
```
lawmicha, line 257: replace with `dispatchOutboxStatusEvent(isEmpty: true)`

lawmicha (commented):
continued review for `outboxStatus`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
DongQuanRui: OK. But is this like hardcoded Although we know we are about to add an operation into queue? 😂

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -79,6 +79,10 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
     func startSyncingToCloud(api: APICategoryGraphQLBehavior,
                              mutationEventPublisher: MutationEventPublisher) {
         log.verbose(#function)
+        let outboxStatusEvent = OutboxStatusEvent(isEmpty: operationQueue.operationCount == 0 ? true : false)
+        let outboxStatusEventPayload = HubPayload(eventName: HubPayload.EventName.DataStore.outboxStatus,
+                                             data: outboxStatusEvent)
+        Amplify.Hub.dispatch(to: .dataStore, payload: outboxStatusEventPayload)
         stateMachine.notify(action: .receivedStart(api, mutationEventPublisher))
```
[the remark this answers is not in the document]
DongQuanRui: So in iOS implementation. when DataStore starts and MutationEvent database is not empty, `requestEvent()` wil be called and then `enqueue()`, so that in `enqueue()` method, `dispatchOutboxStatusEvent(isEmpty: Bool)` already includes two scenarios: 1. when DataStore starts and 2. outboxMutationEnqueued. Is that the correct understanding?

lawmicha:
> So in iOS implementation. when DataStore starts and MutationEvent database is not empty, requestEvent() wil be called and then enqueue(), so that in enqueue() method, dispatchOutboxStatusEvent(isEmpty: Bool) already includes two scenarios: 1. when DataStore starts and 2. outboxMutationEnqueued. Is that the correct understanding?

if `enqueue()` is called when datastore starts, yes the outbox mutation is enqueued, but this is when there mutation events to be processed at datastore start. in OutgoingMutationQueue, there isn't a way to know when DataStore starts, if it is empty. you only know when it is Not empty because your subscriber gets an mutation event from mutationEventPublisher

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/OutboxStatusEvent.swift
```diff
@@ -0,0 +1,16 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
```
lawmicha, line 9: can you mention in comment for each Event payload, which event they are for? `/// Used as payload for the `outboxStatus` event, emitted to the Hub

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/SyncQueriesStartedEvent.swift
```diff
@@ -0,0 +1,16 @@
...
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+public struct SyncQueriesStartedEvent {
+    let models: [String]
```
lawmicha, line 11: comment for fields, like `/// list of model names`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -138,6 +138,7 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
                        mutationEventPublisher: MutationEventPublisher) {
         log.verbose(#function)
         self.api = api
+        dispatchOutboxStatusEvent(isEmpty: operationQueue.operationCount == 0)
```
lawmicha, line 141: `operationQueue.operationCount` is always 0 right here i think, because we haven't `request(.max(1))` yet so no mutation events recieved, and no mutation event added to the operation queue

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
lawmicha: because the operation queue is not suspended so if you add an operation to the queue, you can immediately assume it is in the process of being executed. like the other scenario, first dispatch the hub event then start the operation (by adding it to the operation queue or unsuspend the operation queue).

lawmicha:
can update title to be more explicit on the hub events being added?
 `[DataStore] Triggering hub events` -> `feat(datastore): Dispatch outboxStatus, subscriptionsEstablished, syncQueriesStarted events`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
lawmicha: can you add new line to reduce line length?

lawmicha:
One approach i'm thinking is in OutgoingMutationQueue, there is a state that is `starting` and a responder method that is `start(api:mutationEventPublisher:)`. you can split this into two responder methods

when in `.starting` state, respond to it with `checkStatus()` responder method. In checkStatus, you can have the logic
```
func checkStatus() {
   // use storageAdapter and query for MutationEvents
   // dispatchOutBoxStatusEvent(isEmpty: events.isEmpty)
   stateMachine.notify(action: .dispatchedStatus)
}
```

Then `(.starting, .dispatchedStatus)` can resolve to the state `.subscribe` and triggers `subscribe(api:mutationEvent)` 

``` 
/// note here that the old `start()` was replaced with this new `subscribe()` responder method, but content the same
private func subscribe(api: APICategoryGraphQLBehavior,
                       mutationEventPublisher: MutationEventPublisher) {
        log.verbose(#function)
        self.api = api
        operationQueue.isSuspended = false

        // State machine notification to ".receivedSubscription" will be handled in `receive(subscription:)`
        mutationEventPublisher.publisher.subscribe(self)
    }
```

This approach goes straight to introducing a new state and responder method, however you can try it out by putting `checkStatus()` as a call inside the existing `start()` logic, and you may find that we don't gain that much from introducing a new state in the state machine. the important part here is to get the actual number of mutation events to be processed at DataStore start, is it from storage adapter? is there a way do some sort of peek operation on the mutationEventPublisher?  Once you have the actual number of mutation events to be processed, then you can emit the outbox status event with an accurate `isEmpty` field

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
DongQuanRui: I believe `filter` can only filter on one event instead of multiple events.
Let me first use `guard` and add a new line

lawmicha: looks like there's an `any` filter that you can pass multiple https://github.com/aws-amplify/amplify-ios/blob/main/Amplify/Categories/Hub/HubFilter.swift#L25

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/ModelSyncedPayload.swift
lawmicha: > Since you're emitting this on the Hub, this needs to be a public type. Each member should also be public. It isn't strictly necessary to make the initializer public, but it's good practice.

@DongQuanRui, see this comment, i don't think the members in the other classes are currently public

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/OutboxStatusEvent.swift
```diff
@@ -0,0 +1,16 @@
...
+public struct OutboxStatusEvent {
+    let isEmpty: Bool
```
lawmicha, line 11: this should be `public let` right?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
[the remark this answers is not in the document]
lawmicha: if enqueue() is called when datastore starts, yes the outbox mutation is enqueued, but this is when there are mutation events to be processed at datastore start. in OutgoingMutationQueue, there isn't a way to know when DataStore starts, if there are no mutation events. you only know when it is Not empty because your subscriber gets an mutation event from mutationEventPublisher

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/OutboxStatusEvent.swift
DongQuanRui: updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
DongQuanRui: Thanks. I didn't know this

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -118,6 +119,15 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
         completion?(.successfulVoid)
     }
 
+    private func dispatchSyncQueriesStarted(_ syncableModels: [Model.Type]) {
```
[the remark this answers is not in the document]
DongQuanRui: updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/OutboxStatusEvent.swift
```diff
@@ -0,0 +1,17 @@
...
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// Used as HubPayload for the `OutboxStatus`
+public struct OutboxStatusEvent {
+    public let isEmpty: Bool /// status of outbox: empty or not
```
lawmicha, line 12: place comments on top

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/SyncQueriesStartedEvent.swift
```diff
@@ -0,0 +1,17 @@
...
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// Used as HubPayload for the `SyncQueriesStarted`
+public struct SyncQueriesStartedEvent {
+    public let models: [String] /// list of model names
```
lawmicha, line 12: triple `///` is used for documentation i believe, i don't think putting it on the right side will pick it up

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -138,6 +138,8 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
                        mutationEventPublisher: MutationEventPublisher) {
         log.verbose(#function)
         self.api = api
+
+        dispatchOutboxStatusEvent(isEmpty: checkStatus())
```
lawmicha, line 142: `checkStatus` was suggestion for the new responder method, if you are using the existing start method, then i would expect `checkStatus` to do one thing primary, which is to check the "status" which is "the number of mutation events to be processed" and then secondarily, you can dispatch the event as part of the check... but the function would still return void

if you want it to return a bool, name it like `isMutationEventsEmpty()`, then you can use it to dispatch the outboxStatusEvent

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -241,10 +245,37 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
...
+                             sort: nil,
+                             paginationInput: nil) { result in
+            switch result {
+            case .success(let events):
+                isEmpty = events.isEmpty
+            case .failure(let error):
+                log.error("Error quering mutation events: \(error)")
+            }
+        }
+        return isEmpty
```
lawmicha, line 269: storageAdapter completion closure `{ result in ...` comes back asynchronously, so you will always return isEmpty true at line 269

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
```diff
@@ -0,0 +1,51 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
```
lawmicha, line 9: remove unneeded new lines

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/TestSupport/HubEventsIntegrationTestBase.swift
```diff
@@ -0,0 +1,50 @@
...
+import XCTest
+
+import AmplifyPlugins
+import AWSMobileClient
+
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+class HubEventsIntegrationTestBase: XCTestCase {
```
lawmicha, line 17: why not reuse `SyncEngineIntegrationTestBase`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
```diff
@@ -0,0 +1,51 @@
...
+import AWSPluginsCore
+
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+@available(iOS 13.0, *)
+class DataStoreHubEventTests: HubEventsIntegrationTestBase {
+
+    func testEventsThatDataStoreStartTriggers() throws {
```
lawmicha, line 22: add test comment, and `testDataStoreConfiguredDispatchesHubEvents`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        guard try HubListenerTestUtilities.waitForListener(with: hubListener, timeout: 5.0) else {
+            XCTFail("Listener not registered for hub")
+            return
+        }
+
+        wait(for: [outboxStatusReceived], timeout: networkTimeout)
+        wait(for: [subscriptionsEstablishedReceived], timeout: networkTimeout)
+        wait(for: [syncQueriesStartedReceived], timeout: networkTimeout)
```
lawmicha, line 49: just curious, are these actually received in this order? or can we relax this to `wait(for: [outboxStatusReceived, subscriptionsEstablishedReceived,  syncQueriesStartedReceived], timeout: networkTimeout)` so that it doesn't matter the order as long we we're making sure that these events are emitted ?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
```diff
@@ -0,0 +1,51 @@
...
+                outboxStatusReceived.fulfill()
+            }
+
+            if payload.eventName == HubPayload.EventName.DataStore.subscriptionsEstablished {
+                subscriptionsEstablishedReceived.fulfill()
+            }
+
+            if payload.eventName == HubPayload.EventName.DataStore.syncQueriesStarted {
+                syncQueriesStartedReceived.fulfill()
+            }
```
lawmicha, line 39: can you process the payload just like what the developer would do?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
lawmicha: synced offline on this and it appears storageAdapter calls the completion handler before returning, making it essentially a sycrhonous operation, however we'll still code against it like it is async

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/TestSupport/HubEventsIntegrationTestBase.swift
lawmicha: synced offline, sounds like we need a simplified version of `SyncEngineIntegrationTestBase`. then i suggest additionally, if your Hub events is so closely coupled to the start of DataStore (Amplify.configure()), you may want to hoist the call to `Amplify.configure()` out of the test base's setup method and into your test. Then you'll be able to clearly write a test comment describing the test
```
Given: DataStore with sync enabled
When: `Amplify.configure()`  to start DataStore and the sync engine
Then: The events [X, Y, Z] are emitted as part of the sync engine lifecycle
```

because right now your `Amplify.configure()` is hidden inside the testbase

lawmicha:
Are there any unit tests that you can improve or add?

- For example, `InitialSyncOrchestratorTests.testInvokesCompletionCallback` can probably be improved by adding a hub listener for the event that you are emitted and asserting that the model names match the expected syncable models

- `OutgoingMutationQueueTests`
- `OutgoingMutationQueueMockStateTest`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    /// - Given:
+    ///    - two models: Post, Comment
+    ///    - no pending MutationEvents in MutationEvent database
+    /// - When:
+    ///    - DataStore starts booting up
+    /// - Then:
+    ///    - subscriptionEstablished received, payload should be nil
+    ///    - syncQueriesStarted received, payload should be: {models: ["Post", "Comment"]}
+    ///    - outboxStatus received, payload should be {isEmpty: true}
```
lawmicha, line 28:
```
/// - Given:
    ///    - registered models from `TestModelRegistration` contains `Post` and `Comment`
    ///    - no pending MutationEvents in MutationEvent database // <- how do you ensure this?
    /// - When:
    ///    - DataStore starts booting up  // <- DataStore's remote sync engine is initialized
    /// - Then:
    ///    - `subscriptionEstablished` received, payload should be nil 
    ///    - `syncQueriesStarted` received, payload should be: {models: ["Post", "Comment"]} 
    ///    - outboxStatus received, payload should be {isEmpty: true}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
```diff
@@ -0,0 +1,70 @@
...
+                subscriptionsEstablishedReceived.fulfill()
+            }
+
+            if payload.eventName == HubPayload.EventName.DataStore.syncQueriesStarted {
+                XCTAssertNotNil(payload.data)
+                guard let syncQueriesStartedEvent = payload.data as? SyncQueriesStartedEvent else {
+                    XCTFail("Failed to cast payload data as SyncQueriesStartedEvent")
+                    return
+                }
+                XCTAssertEqual(syncQueriesStartedEvent.models.count, 2)
```
lawmicha, line 47: low: test comment mentions post and comment but this assert does not assert for "Post" and "Comment" in the payload

Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
```diff
@@ -23,4 +23,44 @@ public extension HubPayload.EventName.DataStore {
     /// Dispatched when DataStore receives a sync response from the remote API via the API category. The Hub Payload
     /// will be a `MutationEvent` instance that caused the conditional save failed.
     static let conditionalSaveFailed = "DataStore.conditionalSaveFailed"
+
+    /// Dispatched when network status has changed (active or not)
+    /// The HubPayload will be a boolean value `isActive` indicating the status of network
+    // TODO: networkStatusChanged to be implemented
```
lawmicha, line 29: can you also remove the events not used? reason is this could get released and then would never be emitted (not a huge deal), adding it later allows PR review to comment and help update the description, and there's no real reason now that you've narrowed down the scope of the PR

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -241,10 +244,36 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
                 self.log.verbose("mutationEvent deleted successfully")
             }
 
+            self.dispatchOutboxStatusEvent(isEmpty: true)
             self.stateMachine.notify(action: .processedEvent)
         }
     }
 
+    private func checkMutationEventDatabaseStatus() {
```
lawmicha, line 252: `queryMutationEventsFromStorage` would be closer to what the code is doing

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -138,7 +138,8 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
                        mutationEventPublisher: MutationEventPublisher) {
         log.verbose(#function)
         self.api = api
-        operationQueue.isSuspended = false
+
+        checkMutationEventDatabaseStatus()
 
         // State machine notification to ".receivedSubscription" will be handled in `receive(subscription:)`
         mutationEventPublisher.publisher.subscribe(self)
```
lawmicha, line 145: why isn't this moved as well? you can expose a completion for `checkMutationEventDatabaseStatus()` like
```
checkMutationEventDatabaseStatus(onComplete: {
   operationQueue.isSuspended = false
   // State machine notification to ".receivedSubscription" will be handled in `receive(subscription:)`
    mutationEventPublisher.publisher.subscribe(self)
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
DongQuanRui: - for this one:
```
registered models from `TestModelRegistration` contains `Post` and `Comment`
```
I actually don't guarantee whether the order be `["Post", "Comment"]` or `["Comment", "Post"]`, or is it in this order for sure? So I should probably just say:
```
registered two models from `TestModelRegistration`
```
- and for this one:
```
no pending MutationEvents in MutationEvent database // <- how do you ensure this?
```
By not making any `Amplify.DataStore.save()`? 😂

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
DongQuanRui: Let me just use 
```
waitForExpectations(timeout: networkTimeout, handler: nil)
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
lawmicha: - I don't think the order matters, nor the models that are in there. i'd say just remove the mention of post and comment. `registered two models from `TestModelRegistration` sounds good
- other tests running that does do Save to the same local store and if terminated before the mutation events are enqueued and processed, then this test will run with mutation events in the database. i would say we should just assert for at least one outbox mutation event, and then if there are more that is okay. you can achieve this by having the `expecatation.overfufill = true`

lawmicha (approved):
looks good to me, question about naming for `outboxStatus`. Should we stick with JS/Android or if we should stick to something more common in iOS like `outgoingMutationStatus`?

palpatim:
> looks good to me, question about naming for `outboxStatus`. Should we stick with JS/Android or if we should stick to something more common in iOS like `outgoingMutationStatus`?

@lawmicha We need to keep the names the same for platform consistency. We can revisit naming, event dispatch rules, payloads, etc in the future.

Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
```diff
@@ -23,4 +23,15 @@ public extension HubPayload.EventName.DataStore {
     /// Dispatched when DataStore receives a sync response from the remote API via the API category. The Hub Payload
     /// will be a `MutationEvent` instance that caused the conditional save failed.
     static let conditionalSaveFailed = "DataStore.conditionalSaveFailed"
+
+    /// Dispatched on DataStore start and also every time a local mutation is enqueued and processed in the outbox
```
palpatim, line 27: Clarify wording: "Dispatched when: 1) the DataStore starts; 2) each time a local mutation is enqueued into the outbox; 3) each time a local mutation is finished processing."

Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
```diff
@@ -23,4 +23,15 @@ public extension HubPayload.EventName.DataStore {
...
+    /// HubPayload `OutboxStatusEvent` contains a boolean value `isEmpty` to notify if there are mutations in the outbox
+    static let outboxStatus = "DataStore.outboxStatus"
+
+    /// Dispatched when all of the subscriptions to syncable models have been established
```
palpatim, line 31: "Dispatched when DataStore has finished establishing its subscriptions to all syncable models"

Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
```diff
@@ -23,4 +23,15 @@ public extension HubPayload.EventName.DataStore {
...
+    static let subscriptionsEstablished = "DataStore.subscriptionEstablished"
+
+    /// Dispatched when DataStore is about to start sync queries
```
palpatim, line 34: "Dispatched when DataStore starts establishing its subscriptions to all syncable models"

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/OutboxStatusEvent.swift
```diff
@@ -0,0 +1,18 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
```
palpatim, line 8: Unnecessary import

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/OutboxStatusEvent.swift
```diff
@@ -0,0 +1,18 @@
...
+
+/// Used as HubPayload for the `OutboxStatus`
+public struct OutboxStatusEvent {
+    /// status of outbox: empty or not
```
palpatim, line 12: "`true` if there are no events in the outbox at the time the event was dispatched

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/SyncQueriesStartedEvent.swift
```diff
@@ -0,0 +1,18 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
```
palpatim, line 8: Unnecessary import

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/SyncQueriesStartedEvent.swift
```diff
@@ -0,0 +1,18 @@
...
+
+/// Used as HubPayload for the `SyncQueriesStarted`
+public struct SyncQueriesStartedEvent {
+    /// list of model names
```
palpatim, line 12: "A list of all model names for which DataStore has started establishing subscriptions"

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -241,10 +245,36 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
...
+
+        storageAdapter.query(MutationEvent.self,
+                             predicate: predicate,
+                             sort: nil,
+                             paginationInput: nil) { result in
+            switch result {
+            case .success(let events):
+                self.dispatchOutboxStatusEvent(isEmpty: events.isEmpty)
+            case .failure(let error):
+                log.error("Error quering mutation events: \(error)")
```
palpatim, line 265: s/quering/querying/

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -241,10 +245,36 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
                 self.log.verbose("mutationEvent deleted successfully")
             }
 
+            self.dispatchOutboxStatusEvent(isEmpty: true)
```
palpatim, line 248: How do you know the outbox is empty here? Shouldn't you be querying the mutation events from storage, or checking the `operationCount` of the outbox's operation queue?

Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
DongQuanRui: For `syncQueriesStarted`, it is not `Dispatched when DataStore starts establishing its subscriptions to all syncable models but is `Dispatched when DataStore starts to sync Queries`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/OutboxStatusEvent.swift
DongQuanRui: Removed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
DongQuanRui: Right, it makes sense. 
Incorrectly, I thought because `subscription.request(.max(1))` ensure the queue to have only one `mutationEvent` at a time so I can assert queue is empty when the event is finish processing. I miss the fact that events are stored inside the mutationEventDatabase.
Updated.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/SyncQueriesStartedEvent.swift
DongQuanRui: Removed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Events/SyncQueriesStartedEvent.swift
DongQuanRui: Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
DongQuanRui: updated

Amplify/Categories/DataStore/DataStoreCategory+HubPayloadEventName.swift
DongQuanRui: Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
lawmicha: As i understand it, we were emitting outboxStatus events after the first according to events being processed, not according to what's in storage
- when the remote sync engine starts, check if there are mutation events from storage to determine `isEmpty`
- when a mutationEvent is being enqueued, we get this from the mutationEventPublisher, one mutationEvent at a time. hardcode `isEmpty` to false since we know we are processing one
- when the mutationEvent is done being processed, we emit a following `outboxStatus` event with `isEmpty = true` 

example output
outBoxStatus - isEmpty = false - on start, check mutationEvent from storage, say there 5 mutation events
outboxStatus - isEmpty = false - queue mutationEvent 1
outboxStatus - isEmpty = true - finish processing mutationEvent 1
outboxStatus - isEmpty = false - queue mutationEvent 2
outboxStatus - isEmpty = true - finish processing mutationEvent 2
outboxStatus - isEmpty = false - queue mutationEvent 3
outboxStatus - isEmpty = true - finish processing mutationEvent 3
outboxStatus - isEmpty = false - queue mutationEvent 4
outboxStatus - isEmpty = true - finish processing mutationEvent 4
outboxStatus - isEmpty = false - queue mutationEvent 5
outboxStatus - isEmpty = true - finish processing mutationEvent 5

lawmicha: or is the expected behavior just a ping of the outbox status? for example 
outBoxStatus - isEmpty = false - on start, check mutationEvent from storage, say there 5 mutation events
outboxStatus - isEmpty = false - queue mutationEvent 1
outboxStatus - isEmpty = false - finish processing mutationEvent 1
outboxStatus - isEmpty = false - queue mutationEvent 2
outboxStatus - isEmpty = false - finish processing mutationEvent 2
outboxStatus - isEmpty = false - queue mutationEvent 3
outboxStatus - isEmpty = false - finish processing mutationEvent 3
outboxStatus - isEmpty = false - queue mutationEvent 4
outboxStatus - isEmpty = false - finish processing mutationEvent 4
outboxStatus - isEmpty = false - queue mutationEvent 5
outboxStatus - isEmpty = true - finish processing mutationEvent 5

manueliglesias: @lawmicha the second behavior is the expected one.

For reference, this is the JS implementation:

https://github.com/aws-amplify/amplify-js/blob/a7feacea4ed506340d250249d0b15286fe3ef5fa/packages/datastore/src/sync/index.ts#L358-L365

It checks the queue

palpatim (approved):

PR #722, opened 2020-08-17
refactor(core): Add support for useragent to add platform info

royjit:
*Issue #, if available:*

*Description of changes:*

Added ability to add platform information to the user agent. The user agent can now add in the following format:
```
<Platform-2-Name/Platform-2-Version> <Platform-1-Name/Platform-1-Version> amplify-iOS/xx.yy iOS/13.6 en_US
```

### DIFF
--- a/AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
@@ -9,15 +9,16 @@ import Foundation
 import AWSCore
 
 public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
+    private static let version = "1.1.0"
+
     override public class func baseUserAgent() -> String! {
         //TODO: Retrieve this version from a centralized location:
         //https://github.com/aws-amplify/amplify-ios/issues/276
-        let version = "1.1.0"
-        let sdkName = "amplify-iOS"
+        let platformInfo = AmplifyAWSServiceConfiguration.platformInformation()
         let systemName = UIDevice.current.systemName.replacingOccurrences(of: " ", with: "-")
         let systemVersion = UIDevice.current.systemVersion
         let localeIdentifier = Locale.current.identifier
-        return "\(sdkName)/\(version) \(systemName)/\(systemVersion) \(localeIdentifier)"
+        return "\(platformInfo) \(systemName)/\(systemVersion) \(localeIdentifier)"
     }
 
     override public var userAgent: String {
@@ -39,5 +40,30 @@ public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
 
     public init(region regionType: AWSRegionType) {
         super.init(region: regionType, credentialsProvider: nil)
+    }
+}
+
+extension AmplifyAWSServiceConfiguration {
+
+    static var platformMapping: [Platform: String] = [:]
+
+    public static func addPlatform(_ platform: Platform, version: String) {
+        guard platform != .iOS else {
+            return
+        }
+        platformMapping[platform] = version
+    }
+
+    public enum Platform: String {
+
+        case iOS = "amplify-iOS"
+
+        case flutter = "amplify-flutter"
+    }
+
+    static func platformInformation() -> String {
+        var mapping = platformMapping
+        mapping[.iOS] = version
+        return mapping.map {"\($0.rawValue)/\($1)"}.joined(separator: " ")
     }
 }
--- a/AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
@@ -57,7 +57,7 @@ extension AmplifyAWSServiceConfiguration {
 
     static func platformInformation() -> String {
         let sdkName = "amplify-iOS"
-        let additionalPlatform = platformMapping.map {" \($0.rawValue)/\($1)"}.joined()
-        return "\(sdkName)/\(version)\(additionalPlatform)"
+        let additionalPlatform = platformMapping.map {"\($0.rawValue)/\($1) "}.joined()
+        return "\(additionalPlatform)\(sdkName)/\(version)"
     }
 }
--- a/AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
@@ -48,22 +48,16 @@ extension AmplifyAWSServiceConfiguration {
     static var platformMapping: [Platform: String] = [:]
 
     public static func addUserAgentPlatform(_ platform: Platform, version: String) {
-        guard platform != .ios else {
-            return
-        }
         platformMapping[platform] = version
     }
 
     public enum Platform: String {
-
-        case ios = "amplify-iOS"
-
         case flutter = "amplify-flutter"
     }
 
     static func platformInformation() -> String {
-        var mapping = platformMapping
-        mapping[.ios] = version
-        return mapping.map {"\($0.rawValue)/\($1)"}.joined(separator: " ")
+        let sdkName = "amplify-iOS"
+        let additionalPlatform = platformMapping.map {" \($0.rawValue)/\($1)"}.joined()
+        return "\(sdkName)/\(version)\(additionalPlatform)"
     }
 }
--- a/AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
@@ -47,8 +47,8 @@ extension AmplifyAWSServiceConfiguration {
 
     static var platformMapping: [Platform: String] = [:]
 
-    public static func addPlatform(_ platform: Platform, version: String) {
-        guard platform != .iOS else {
+    public static func addUserAgentPlatform(_ platform: Platform, version: String) {
+        guard platform != .ios else {
             return
         }
         platformMapping[platform] = version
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
```diff
@@ -41,3 +42,28 @@ public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
...
+    public static func addPlatform(_ platform: Platform, version: String) {
+        guard platform != .iOS else {
+            return
+        }
+        platformMapping[platform] = version
+    }
+
+    public enum Platform: String {
+
+        case iOS = "amplify-iOS"
```
palpatim, line 59: Nit: make this all lowercase.

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
```diff
@@ -41,3 +42,28 @@ public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
         super.init(region: regionType, credentialsProvider: nil)
     }
 }
+
+extension AmplifyAWSServiceConfiguration {
+
+    static var platformMapping: [Platform: String] = [:]
+
+    public static func addPlatform(_ platform: Platform, version: String) {
```
palpatim, line 50: This name is ambiguous. It's part of a "service configuration", but we're adding a "platform" but also a version string, with no indication as to its purpose. Can we make its use explicit, as in `addUserAgentPlatform`?

Also I thought we were going to append the additional platform info to the user agent string, so that the UA would include both Amplify iOS and Flutter? Maybe that requirement has changed in other discussions.

AmplifyPlugins/Core/AWSPluginsCoreTests/ServiceConfiguration/AmplifyAWSServiceConfigurationTests.swift
```diff
@@ -26,4 +39,46 @@ class AmplifyAWSServiceConfigurationTests: XCTestCase {
...
+        XCTAssert(configuration.userAgent.contains(AmplifyAWSServiceConfiguration.Platform.flutter.rawValue))
+        XCTAssertEqual(expectedSystem, userAgentParts[2])
+        XCTAssertEqual(expectedLocale, userAgentParts[3])
+    }
+
+    /// Test overriding iOS platform in AmplifyAWSServiceConfiguration
+    ///
+    /// - Given: Amplify library
+    /// - When:
+    ///    - I override iOS platform with a new version
```
palpatim, line 72: A "new" version, or "any" version? If the latter, why do we even expose `.ios` as an enum case?

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
```diff
@@ -9,15 +9,16 @@ import Foundation
 import AWSCore
 
 public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
+    private static let version = "1.1.0"
```
palpatim, line 12: have you verified that this new version declaration works with the fastlane version bump script?

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
royjit: - Changed to `addUserAgentPlatform`. 

- iOS value will always be present when creating the user agent string. Refer line number 66, where we manually add iOS version before creating the platformInformation.

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
royjit: Changed

AmplifyPlugins/Core/AWSPluginsCoreTests/ServiceConfiguration/AmplifyAWSServiceConfigurationTests.swift
royjit: Currently `.ios` case can only be used internally and we ignore the value set by the external developer. This make sure that iOS value is not overriden by developers. There is no need to expose `.ios` but I dont think we can have private enum cases. 

I can make the function that create the platform string like shown below and remove `ios` case:

```swift
    static func platformInformation() -> String {
         let additionalPlatform = platformMapping.map {"\($0.rawValue)/\($1)"}.joined(separator: " ")

         // add iOS here
         let version = "1.1.0"
         let sdkName = "amplify-iOS"
         return "\(sdkName)/\(version) \(additionalPlatform)"
     }
```

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
royjit: Yes tested this with help of @iartemiev . Ran `bundle exec fastlane unstable` and verified that the version is updated.

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
```diff
@@ -41,3 +42,28 @@ public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
...
+    public static func addUserAgentPlatform(_ platform: Platform, version: String) {
+        guard platform != .ios else {
+            return
+        }
+        platformMapping[platform] = version
+    }
+
+    public enum Platform: String {
+
+        case ios = "amplify-iOS"
```
palpatim, line 59: As discussed below, remove

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
```diff
@@ -41,3 +42,28 @@ public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
...
+
+        case flutter = "amplify-flutter"
+    }
+
+    static func platformInformation() -> String {
+        var mapping = platformMapping
+        mapping[.ios] = version
+        return mapping.map {"\($0.rawValue)/\($1)"}.joined(separator: " ")
```
palpatim, line 67: Is the metrics team OK with the order of the UA components being non-deterministic?

AmplifyPlugins/Core/AWSPluginsCoreTests/ServiceConfiguration/AmplifyAWSServiceConfigurationTests.swift
palpatim: Yes, this makes sense

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
royjit: Removed

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
royjit: Will check, hopefully this wont be a problem since we have the same implementation in Android https://github.com/aws-amplify/amplify-android/pull/703/files

royjit: Verified that the value need to be prepended.

AmplifyPlugins/Core/AWSPluginsCoreTests/ServiceConfiguration/AmplifyAWSServiceConfigurationTests.swift
```diff
@@ -26,4 +38,30 @@ class AmplifyAWSServiceConfigurationTests: XCTestCase {
...
+        let currentSystemVersion = UIDevice.current.systemVersion
+        let expectedLocale = Locale.current.identifier
+        let expectedSystem = "\(currentSystemName)/\(currentSystemVersion)"
+
+        let configuration = AmplifyAWSServiceConfiguration()
+
+        XCTAssertNotNil(configuration.userAgent)
+        let userAgentParts = configuration.userAgent.components(separatedBy: " ")
+        XCTAssertEqual(4, userAgentParts.count)
+        XCTAssert(userAgentParts[0].starts(with: AmplifyAWSServiceConfiguration.Platform.flutter.rawValue))
```
palpatim, line 62: You should be able to assert the entire platform & version value right?

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
```diff
@@ -41,3 +42,22 @@ public class AmplifyAWSServiceConfiguration: AWSServiceConfiguration {
...
+        platformMapping[platform] = version
+    }
+
+    public enum Platform: String {
+        case flutter = "amplify-flutter"
+    }
+
+    static func platformInformation() -> String {
+        let sdkName = "amplify-iOS"
+        let additionalPlatform = platformMapping.map {"\($0.rawValue)/\($1) "}.joined()
```
palpatim, line 60: Let's use `joined` all the way down; it'll be less brittle than appending the space at the end of every token:

```swift
var platformTokens = platformMapping.map { "\($0.rawValue)/\($1)" }
platformTokens.append("amplify-iOS/\(version)")
let additionalPlatforms = platformTokens.joined(separator: " ")
```

AmplifyPlugins/Core/AWSPluginsCoreTests/ServiceConfiguration/AmplifyAWSServiceConfigurationTests.swift
royjit: Changed

AmplifyPlugins/Core/AWSPluginsCore/ServiceConfiguration/AmplifyAWSServiceConfiguration.swift
royjit: Refactored

Issue #585, opened 2020-06-26
Search API


TheBenck:
**Is your feature request related to a problem? Please describe.**
There doesn't seem to be any way to query searchable models with the Amplify Libraries API plugin, making Libraries unusable for the search case. A first-class api for this would be very useful.

**Describe the solution you'd like**
Something like list queries but for searching. E.g.:
`Amplify.API.query(request: .search(...`

**Describe alternatives you've considered**
The only alternative is to go back to the mobile SDK. Correct me if I'm missing something.

palpatim:
@TheBenck The [list query](https://docs.amplify.aws/lib/graphqlapi/query-data/q/platform/ios#list-query) API allows you to specify a predicate for searching your models. Does that meet your needs? If not, can you provide some more detail around your use case and explain what features are missing?

TheBenck:
@palpatim Thanks! This is what I needed. Is there more documentation somewhere on the predicate? Also, are sorting and limiting possible like with the functions generated for API.swift?

drochetti:
Hey @TheBenck ,

Search queries are not support out-of-the-box *yet*. There are available workarounds though. The recommended one is to create a custom query, which means create your own function that returns an instance of `GraphQLRequest<[M]>` that contains the search query.

For example:

```swift
import Amplify

let searchTodoQuery = """
query SearchTodos($filter: SearchableTodoFilterInput) {
  searchTodos(filter: $filter) {
    items {
      id
      name
      done
      createdAt
      description
    }
  }
}
"""

extension GraphQLRequest where R == [Todo] {
    static func search(_: Todo.Type, byKeywords keywords: String? = nil) -> Self {
        var variables: [String: Any] = [:]

        if let matches = keywords {
            variables = [
                "filter": [
                    "name": [
                        "match": matches,
                    ],
                ],
            ]
        }

        return GraphQLRequest(document: searchTodoQuery,
                              variables: variables,
                              responseType: [Todo].self,
                              decodePath: "searchTodos.items")
    }
}

```

**Notes:**

- I have a model `Todo` with `@searchable`
- If you check the compiled GraphQL at `amplify/build/schema.graphql`, you will find the `query searchTodos` and the related types
- then I created an extension to `GraphQLRequest where R == [Todo]` (the response type is an array of `Todo`)
- that enables me to call:

```swift
_ = Amplify.API.query(request: .search(Todo.self, byKeywords: "important")) {
    switch $0 {
    case let .success(result):
[...]

PR #742, opened 2020-08-26
feat(API): Model-based GraphQL Search API

lawmicha:
*Related Issues*
https://github.com/aws-amplify/amplify-ios/issues/585

*Previous PR*
https://github.com/aws-amplify/amplify-ios/pull/1009

*Description*
First-class support for models with `@searchable` enabled (https://docs.amplify.aws/cli/graphql-transformer/searchable)

Example schema:
```
type Post @model @searchable {
  id: ID!
  title: String!
  name: String!
  createdAt: String!
  updatedAt: String!
  upvotes: Int
}
```

### Public APIs
1. A public facing GraphQLRequest builder to create a valid search query on a model.
```swift
protocol ModelGraphQLRequestFactory {
    static func search<M: Model>(_ modelType: M.Type,
                                 limit: Int?,
                                 sort: QuerySortBy?) -> GraphQLRequest<List<M>>
}
```
This provides developers an Model-based API to perform search queries on their data. The intention is to get out of the way of developers since they are explicitly enabling `@searchable` on particular models, we do not make the assumption that they always want to perform the search over a list query. The alternative, rather more complex implementation, is for the library to be smart enough to detect that `@searchable` is enabled on the model and to always call search from the `.list` API.

2. QueryPredicate that takes in a custom operator and value
```swift
public struct QueryField {
    public func evaluate(operator: String, value: Persistable) -> QueryPredicateOperation
}
```
This new `QueryPredicationOperation` allows for more advanced use cases of search filters specific to AppSync. This acts as the _escape hatch_, similar to the workaround of using `field(_ name: String) -> QueryField`. This allows developers to fully control their request by providing their own operator and value:
```swift
let predicate = Post.keys.field("title").evaluate(operator: "regexp", value: "k.*y")
```

### Developer experience

1. As a developer, I understand the cost vs performance benefit of performing a search query vs a list query. Based on the app experience I am building, I can decide when perform a search, ie. from map view or search box, or when to perform a list query else where.
```swift
let post = Post.keys
Amplify.API.query(.search(Post.self, where: post.name.contains("myName"), limit: 1000, sort: .ascending(Post.keys.name))
```

2. QueryPredicate parameter `where` allows the developer to search based on a condition. 
```swift
let post = Post.keys
let predicate = post.name.contains("middleName") && post.name.beginsWith("firstName")
```
Internally, the plugin maps these as closely as possible to their expected behavior. Developers should not need to know the following mapping and using the predicates should be intuitive.

| QueryPredicate  | AppSync FilterInput  | ElasticSearch filters (AppSync SearchFilterInput) |
| :------------:|:---------------:|:-----:|
| .notEqual (`!=`)             | "ne" | "ne" |
| .equals (`==`)               |  "eq"  |  "eq" |
| .lessOrEqual (`<=`)      |  "le"  |  "lte" |
| .lessThan (`<`)           |  "lt"  |  "lt" |
| .greaterOrEqual (`>=`)|  "ge"  |  "gte" |
| .greaterThan (`>`)      |  "gt"  |  "gt" |
| .contains             |  "contains"  |  "match" |
| .between            |  "between"  |  "range" |
| .beginsWith         |  "beginsWith"  |  "matchPhrasePrefix" |

*Note*: This mapping is also important for plugin developers to maintain parity across Android and iOS.

Advanced users can use the `evaluate(operator:value)` to pass in any of the Elasticsearch keywords 

```swift
let post = Post.keys
let predicate = post.name.evaluate(operator: "matchPhrase", value: "this is a test")
let predicate = post.name.evaluate(operator: "multiMatch", value: "this is a test")
let predicate = post.name.evaluate(operator: "exists", value: true)
let predicate = post.name.evaluate(operator: "wildcard", value: "S*Elasticsearch!")
let predicate = post.name.evaluate(operator: "regexp", value: "k.*y")
```
Above operators are from [ElasticSearch filters](https://docs.amplify.aws/cli/graphql-transformer/searchable#usage)

4. Pagination. As a developer, I can perform a search query, and retrieve the next page of results.
```swift
let blog = Blog6.keys
let predicate = blog.id.evaluate(operator: "matchPhrasePrefix", value: blogId)
Amplify.API.query(request: .search(Blog6.self, where: predicate, limit: 1)) { result in
    switch result {
    case .success(let result):
        switch result {
        case .success(let blogs):
            for blog in blogs {
               print(blog)
            }
            if blog.hasNextPage() {
                blogs.getNextPage { result in
                        switch result {
                        case .success(let nextPage):
                            print(nextPage)
                        case .failure(let error):
                        }
                    }
           }

```

5. As there are additional parameters and fields that this API does not support. What is not supported:
- `from` parameter in the input
- `total` field in the response
Let us know if this is useful for you and what your use caes is, and as a workaround you create your own custom GraphQLRequest like this [SampleApp](https://github.com/lawmicha/apiSearchable) /  [Code Snippet](https://gist.github.com/lawmicha/d8a9453a3dfdc976f4cfa976a9c4eb30)

### DIFF
--- a/Amplify.xcodeproj/project.pbxproj
+++ b/Amplify.xcodeproj/project.pbxproj
@@ -90,6 +90,7 @@
 		216E460824913D430035E3CE /* Color.swift in Sources */ = {isa = PBXBuildFile; fileRef = 216E460724913D430035E3CE /* Color.swift */; };
 		216E460A249183230035E3CE /* Section.swift in Sources */ = {isa = PBXBuildFile; fileRef = 216E4609249183230035E3CE /* Section.swift */; };
 		216E461A249189050035E3CE /* Embedded.swift in Sources */ = {isa = PBXBuildFile; fileRef = 216E4619249189050035E3CE /* Embedded.swift */; };
+		2177E40625C209C100C4A6FC /* AppSyncGraphQLRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2177E40525C209C100C4A6FC /* AppSyncGraphQLRequest.swift */; };
 		217855C3237F84D700A30D19 /* RESTRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 217855C2237F84D700A30D19 /* RESTRequest.swift */; };
 		217D5EB02577F9DF009F0639 /* Post4.swift in Sources */ = {isa = PBXBuildFile; fileRef = 217D5E9A2577F9DE009F0639 /* Post4.swift */; };
 		217D5EB12577F9DF009F0639 /* Team2+Schema.swift in Sources */ = {isa = PBXBuildFile; fileRef = 217D5E9B2577F9DE009F0639 /* Team2+Schema.swift */; };
@@ -140,10 +141,10 @@
 		21A4ED5E259CD7F400E1047D /* CoreError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21A4ED5D259CD7F400E1047D /* CoreError.swift */; };
 		21A4ED6E259CDA4600E1047D /* List+LazyLoad.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21A4ED6D259CDA4600E1047D /* List+LazyLoad.swift */; };
 		21A4ED77259CDF6800E1047D /* List+Combine.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21A4ED76259CDF6800E1047D /* List+Combine.swift */; };
-		21A7C8AA25ACB6C8004355D6 /* ArrayLiteralListProviderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21A7C8A925ACB6C8004355D6 /* ArrayLiteralListProviderTests.swift */; };
-		21A7C90225ACC4D1004355D6 /* MockDataStoreResponders.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21A7C90125ACC4D1004355D6 /* MockDataStoreResponders.swift */; };
 		21A4F26325A3CD3D00E1047D /* List+Pagination.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21A4F26225A3CD3D00E1047D /* List+Pagination.swift */; };
 		21A4F8F325A77D9100E1047D /* ListPaginationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21A4F8F225A77D9100E1047D /* ListPaginationTests.swift */; };
+		21A7C8AA25ACB6C8004355D6 /* ArrayLiteralListProviderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21A7C8A925ACB6C8004355D6 /* ArrayLiteralListProviderTests.swift */; };
+		21A7C90225ACC4D1004355D6 /* MockDataStoreResponders.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21A7C90125ACC4D1004355D6 /* MockDataStoreResponders.swift */; };
 		21AD424B249BF0DA0016FE95 /* AnyModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = FACBAD522386160100E29E56 /* AnyModel.swift */; };
 		21AD424C249BF0DE0016FE95 /* AnyModel+Codable.swift in Sources */ = {isa = PBXBuildFile; fileRef = FA8EE776238626D60097E4F1 /* AnyModel+Codable.swift */; };
 		21AD424D249BF0E50016FE95 /* AnyModel+Schema.swift in Sources */ = {isa = PBXBuildFile; fileRef = FA8EE78223862DDB0097E4F1 /* AnyModel+Schema.swift */; };
@@ -899,6 +900,7 @@
 		216E460724913D430035E3CE /* Color.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Color.swift; sourceTree = "<group>"; };
 		216E4609249183230035E3CE /* Section.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Section.swift; sourceTree = "<group>"; };
 		216E4619249189050035E3CE /* Embedded.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Embedded.swift; sourceTree = "<group>"; };
+		2177E40525C209C100C4A6FC /* AppSyncGraphQLRequest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppSyncGraphQLRequest.swift; sourceTree = "<group>"; };
 		217855C2237F84D700A30D19 /* RESTRequest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RESTRequest.swift; sourceTree = "<group>"; };
 		217856BA2383320900A30D19 /* GraphQLQueryType.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = GraphQLQueryType.swift; sourceTree = "<group>"; };
 		217856BD2383322700A30D19 /* GraphQLMutationType.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = GraphQLMutationType.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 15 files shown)


randeepbhatia:
Is this feature going to be part of next release of this lib?

lawmicha:
> Is this feature going to be part of next release of this lib?

Hi @randeepbhatia , this feature is work in progress, i'll take a look at this again to see what is the best way forward. If you'd like, there is a recommended workaround for now here: https://github.com/aws-amplify/amplify-ios/issues/585#issuecomment-767875540

lawmicha:
> ### Some alternative ideas
> 1. Instead of creating `API.query(.search(...))`, we could just create an overloaded `API.query(.list(...))`, which takes a `SearchPredicate` instead of a `QueryPredicate`.  This would solve concern (3) mentioned above.  Then, to create a search predicate, a new `search` method should be added to `QueryField`, which can be used like this:
>    ```
>    Amplify.API.query(.list(Post.self, where: post.name.search(operator:"eq", value: "test"))
>    Amplify.API.query(.list(Post.self, where: post.name.search(operator:"matchPhrase", value: "test"))
>    Amplify.API.query(.list(Post.self, where: post.name.search(operator:"matchPhrasePrefix", value: "test"))
>    ```

It looks like by taking in `SearchPredicate`, this allows the the `search` method on `QueryField` to return a `SearchPredicateOperation` which will enforce developers from making the runtime mistake (concern 3): developers can't create a SearchPredicateOperation with a QueryPredicateOperation and vice versa. I also like this as Amplify would expose a "search" method coupling any ElasticSearch terms since the developer is responsible for passing in the "operator".

One call out here is that the overloaded method also allows us to introduce the `SortInput` on the overloaded method without adding that on the original non-overloaded `.list`. (AppSync search API support sort out of the box, but list API does not- except https://github.com/aws-amplify/amplify-ios/issues/1034 but that is for another discussion)

> 2. This option is very similar to (1), with a slightly different way of constructing the `SearchPredicate`.  Create a new method on `QueryField` for each operator, so the DX is like this:
>    ```
>    Amplify.API.query(.list(Post.self, where: post.name.searchEq(value: "test"))
>    Amplify.API.query(.list(Post.self, where: post.name.searchMatchPhrase(value: "test"))
>    Amplify.API.query(.list(Post.self, where: post.name.searchMatchPhrasePrefix(value: "test"))
>    ```

I believe this is a feasible solution if the method names "searchEq", "searchMatchPhrase", etc in AWSPluginsCore package, without adding EleasticSearch terms into Amplify package. This makes sense because developers do not have to learn a mapping (concern 2). 

In terms of the actual Swift implementation (for solution 1 and 2), it might an ambiugious overloaded method unless we rename the 'where' parameter to "search", like `.list(Post.self, search: post.name.search)`

> 3. Or, create the `API.query(.search(...))` as proposed in this PR, but require it to take a `SearchPredicate` rather than a `QueryPredicate`.  Then a `SearchPredicate` could be instantiated by passing the field and value as parameters.
>    ```
>    Amplify.API.query(.search(Post.self, where: .eq(field: post.name, value: "test"))
>    Amplify.API.query(.search(Post.self, where: .matchPhrase(field: post.name, value: "test"))
>    Amplify.API.query(.search(Post.self, where: .matchPhrasePrefix(field: post.name, value: "test"))
>    ```
> 

When looking at this, I am also considering the current experience, where developers build the predicate based on the QueryField and they can group the predicates together, ie.
```
// this doesn't work because QueryField will contain `eq` method that returns both a QueryPredicate and a SearchPredicate and is ambiguous
Amplify.API.query(.search(Post.self, where: post.name.eq == "test" || post.name.matchPhrase("test"))

// this works because the expected type is SearchPredicate
Amplify.API.query(.search(Post.self, where: .eq(field: post.name, value: "test")  || .matchPhase(field: post.name, value: "test"))

```

I imagine the SearchPredicate type could be in Amplify as `.search(operator: String, value: Persistable)` and we build the ElasticSearch specific methods `SearchPredicate.eq`, `SearchPredicate.matchPhase` inside AWSPluginsCore, which makes it easy for developers to use without learning a new mapping.


> All 3 of these approaches would address all 3 of the concerns previously mentioned. My preference would be 3, then 2, then 1. If you go with (3), you could also introduce a similar approach for constructing `QueryPredicate` as well, for congruency. Of course, we'd want to still support the current way as well to avoid any breaking changes.

I'll go ahead and prototype solution 3 to make sure there isn't something we missed in terms of implementaton on the Swift side, and go from there

lawmicha:
@richardmcclellan I ended up trying out the 3rd idea, such that the `GraphQLRequest.search` API takes in a `SearchPredicate` which prevents a incorrect usage at the compile time for either list or search requests.

```swift
protocol ModelGraphQLRequestFactory {
    static func search<M: Model>(_ modelType: M.Type,
                                 where predicate: SearchPredicate?,
                                 limit: Int?,
                                 sort: QuerySortBy?) -> GraphQLRequest<List<M>>
}
```

Method extension on the CodingKey to provide access to the elastic search queries. Similar to idea 1. Here i've added an enum of the cases instead of having the customer pass in operator value. This makes it harder to make a mistake and consolidates all the operators in one place, ie. `.search(.range(value1, value2)` or `.search(.exists(true)`

```swift 
Amplify.API.query(request: .search(Blog.self, where: blog.id.search(.range(blogId1, blog2)) || 
   blog.id.search(.matchPhrasePrefix(blogId1)),  limit: 1))
```

The reason why i could not go with static methods on `SearchPredicate` like so:
```swift 
Amplify.API.query(request: .search(Blog6.self, 
   where: .matchPhrasePrefix(field: blog.id, value: blogId2)) || .match(field: blog.id, value: blogId1)), 
   limit: 1))
```
The error i got was
```
Static member 'match' cannot be used on protocol metatype 'SearchPredicate.Protocol'
```

PR https://github.com/aws-amplify/amplify-ios/pull/1061

richardmcclellan:
> _Proposed API call_
> 
> ```swift
> Amplify.API.query(request: .search(Blog.self, where: blog.id.search(.range(blogId1, blog2)) || 
>    blog.id.search(.matchPhrasePrefix(blogId1)),  limit: 1))
> ```

This looks like a mix of option 1 and option 3.  It seems a little odd that this DX requires typing `search` twice.   Instead of creating a new `.search`, could you just create an overloaded `.list` instead, which takes a `SearchPredicate` instead of `QueryPredicate` (this was my proposed option 1)?

> Static member 'match' cannot be used on protocol metatype 'SearchPredicate.Protocol'

Ahh, right. Because `match` is a value of `SearchPredicateOperation`, not `SearchPredicate`.   I guess this option (3) won't work.

randeepbhatia:
> @richardmcclellan I ended up trying out the 3rd idea, such that the `GraphQLRequest.search` API takes in a `SearchPredicate` which prevents an incorrect usage at the compile time for either list or search requests.
> 
> ```swift
> protocol ModelGraphQLRequestFactory {
>     static func search<M: Model>(_ modelType: M.Type,
>                                  where predicate: SearchPredicate?,
>                                  limit: Int?,
>                                  sort: QuerySortBy?) -> GraphQLRequest<List<M>>
> }
> ```
> 
> A method extension on the CodingKey provides access to the entire set of elastic search queries. As similar to idea 1, but instead I've added an enum of all the cases instead of having the customer pass in operator value. This makes it harder to make a mistake and consolidates all the operators in one place, ie. `.search(.range(value1, value2)`, `.search(.exists(true)`. Making it more generic also requires some thought around when a search operator requires more than one value, ie. `.search(operator:"range", values: [value1, value2])` . Moving to idea 2 is also feasible but increases the API surface area and i found some of the operators a bit confusing, ie. `.searchEq`, `.searchExists`.
> 
> _Proposed API call_
> 
> ```swift
> Amplify.API.query(request: .search(Blog.self, where: blog.id.search(.range(blogId1, blog2)) || 
>    blog.id.search(.matchPhrasePrefix(blogId1)),  limit: 1))
> ```
> 
> This is also different from idea 3, the reason why i could not go with static methods on `SearchPredicate` like
> 
> ```swift
> Amplify.API.query(request: .search(Blog6.self, 
>    where: .matchPhrasePrefix(field: blog.id, value: blogId2)) || .match(field: blog.id, value: blogId1)), 
>    limit: 1))
> ```
> 
> was due to the error I got when using the extensions on the protocol:
> 
> ```
> Static member 'match' cannot be used on protocol metatype 'SearchPredicate.Protocol'
> ```
> 
> PR #1061

Does the search call also allow to filter by GEO_Distance as well?

lawmicha:
> This looks like a mix of option 1 and option 3. It seems a little odd that this DX requires typing `search` twice. Instead of creating a new `.search`, could you just create an overloaded `.list` instead, which takes a `SearchPredicate` instead of `QueryPredicate` (this was my proposed option 1)?

Yes, this looks pretty good, a bit of iOS backstory is that we introduced `.paginatedList` that returns `List` type (this is the API we want to provide that allows pagination) and left `.list` that returns `[]` to be deprecated in a major release. So in doing so we'll have the following API:

```swift
// 1. To deprecated
static func list<M: Model>(_ modelType: M.Type,
                               where predicate: QueryPredicate?) -> GraphQLRequest<[M]>

// 2. Added for pagination functionality, can be deprecated during major release


// 3. To be added during major release and deprecated previous `.list` that returns `[]` (number 1)
static func list<M: Model>(_ modelType: M.Type,

// 4. Proposed Search API with `.list` overload containing SearchPredcate
static func list<M: Model>(_ modelType: M.Type,
                                 where predicate: SearchPredicate?,
                                 limit: Int?,
                                 sort: QuerySortBy?) -> GraphQLRequest<List<M>>

// 5. In the future, we could also add `sort: QuerySortBy` to the `.list` with QueryPredicate (addition to number 3)
static func list<M: Model>(_ modelType: M.Type,
                               limit: Int?,
                               sort: QuerySortBy?) -> GraphQLRequest<List<M>>
```

By having the overload, and different types for the predicate, developers will catch this at compile time:
```
.list(Post.self, where: post.title.search(.match("value")) || post.content.eq("value"))
// Error
Binary operator '||' cannot be applied to operands of type 'SearchPredicateOperation' and 'QueryPredicateOperation'
```

lawmicha:
> Does the search call also allow to filter by GEO_Distance as well?

Hi @randeepbhatia, this PR provides a strongly typed version of the `SearchableFilter<Model>Input` that you can pass to the API call. Depending on your model, as long as you have provisioned an API that allows you to filter on certain fields, you can apply the `range` filter accordingly. Without more context on "GEO_Distance", I can only assume an example of filtering on a set of latitude/longitude would look like this:
```
{
  "filter": {
      "and" : [
        {
          "lat" : {
            "range" : ["40.9", "42"]
          }
        },
        {
          "long" : {
            "range" : ["-111", "-113"]
          }
        }
      ]
    }
}
```
In code (subject to change), you can construct the `[String:Any]` equivalent for the request variables or use the SearchPredicate like `.search(.range(lat1, lat2) && .search(.range(long1, long2)`

randeepbhatia:
> > Does the search call also allow to filter by GEO_Distance as well?
> 
> Hi @randeepbhatia, this PR provides a strongly typed version of the `SearchableFilter<Model>Input` that you can pass to the API call. Depending on your model, as long as you have provisioned an API that allows you to filter on certain fields, you can apply the `range` filter accordingly. Without more context on "GEO_Distance", I can only assume an example of filtering on a set of latitude/longitude would look like this:
> 
> ```
> {
>   "filter": {
>       "and" : [
>         {
>           "lat" : {
>             "range" : ["40.9", "42"]
>           }
>         },
>         {
>           "long" : {
>             "range" : ["-111", "-113"]
>           }
>         }
>       ]
>     }
> }
> ```
> 
> In code , you can construct the `[String:Any]` equivalent for the request variables or use the SearchPredicate (subject to change) like `.search(.range(lat1, lat2) && .search(.range(long1, long2)`

Hi @lawmicha I was referring to a query
where we have a given location and we want to pull datapoints based on distance (like 5 miles away, 10 miles away) and filtered by other fields.

nfarrell36:
Hello @lawmicha, just wanted to see if this addition was dropped. If so is it still recommended to use your solution on issue #585 to extend Amplify to be able to search against elasticsearch?

PR #756, opened 2020-09-04
fix(DataStore): model reconciliation queue publisher

wooj2:
Pushing this out, prior to being ready to push into mainline ( need to update the rest of the unit tests), becuase there is a slight difference from the suggested approach listed here:
https://github.com/aws-amplify/amplify-ios/issues/755

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
@@ -64,7 +64,7 @@ final class AWSModelReconciliationQueue: ModelReconciliationQueue {
     private let modelName: String
 
     private var incomingEventsSink: AnyCancellable?
-    private var reconcileAndLocalSaveOperationSink: AnyCancellable?
+    private var reconcileAndLocalSaveOperationSinks = Set<AnyCancellable?>()
 
     private let modelReconciliationQueueSubject: PassthroughSubject<ModelReconciliationQueueEvent, DataStoreError>
     var publisher: AnyPublisher<ModelReconciliationQueueEvent, DataStoreError> {
@@ -133,11 +133,19 @@ final class AWSModelReconciliationQueue: ModelReconciliationQueue {
     func enqueue(_ remoteModel: MutationSync<AnyModel>) {
         let reconcileOp = ReconcileAndLocalSaveOperation(remoteModel: remoteModel,
                                                          storageAdapter: storageAdapter)
-        reconcileAndLocalSaveOperationSink = reconcileOp.publisher.sink(receiveCompletion: { error in
-            self.modelReconciliationQueueSubject.send(completion: error)
+        var reconcileAndLocalSaveOperationSink: AnyCancellable?
+
+        reconcileAndLocalSaveOperationSink = reconcileOp.publisher.sink(receiveCompletion: { completion in
+            switch completion {
+            case .finished:
+                self.reconcileAndLocalSaveOperationSinks.remove(reconcileAndLocalSaveOperationSink)
+            case .failure:
+                self.modelReconciliationQueueSubject.send(completion: completion)
+            }
         }, receiveValue: { mutationEvent in
             self.modelReconciliationQueueSubject.send(.mutationEvent(mutationEvent))
         })
+        reconcileAndLocalSaveOperationSinks.insert(reconcileAndLocalSaveOperationSink)
         reconcileAndSaveQueue.addOperation(reconcileOp)
     }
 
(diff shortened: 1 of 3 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
```diff
@@ -64,7 +64,7 @@ final class AWSModelReconciliationQueue: ModelReconciliationQueue {
     private let modelName: String
 
     private var incomingEventsSink: AnyCancellable?
-    private var reconcileAndLocalSaveOperationSink: AnyCancellable?
+    private var reconcileAndLocalSaveOperationSinks = Set<AnyCancellable?>()
```
palpatim, line 67: nittiest of nits: I prefer to reserve variable assignments outside of initializers for types that don't have an initializer. Otherwise, a reader has to look in two places to know what the initial state of an object is.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
```diff
@@ -133,11 +133,19 @@ final class AWSModelReconciliationQueue: ModelReconciliationQueue {
...
         let reconcileOp = ReconcileAndLocalSaveOperation(remoteModel: remoteModel,
                                                          storageAdapter: storageAdapter)
-        reconcileAndLocalSaveOperationSink = reconcileOp.publisher.sink(receiveCompletion: { error in
-            self.modelReconciliationQueueSubject.send(completion: error)
+        var reconcileAndLocalSaveOperationSink: AnyCancellable?
+
+        reconcileAndLocalSaveOperationSink = reconcileOp.publisher.sink(receiveCompletion: { completion in
+            switch completion {
+            case .finished:
+                self.reconcileAndLocalSaveOperationSinks.remove(reconcileAndLocalSaveOperationSink)
```
palpatim, line 141: We should remove the sink on any completion, not just a `.finished`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
```diff
@@ -301,6 +303,10 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
         stateMachine.notify(action: .notified)
     }
 
+    private func notifyFinished() {
+        mutationEventPublisher.send(completion: .finished)
```
palpatim, line 307: Why not include this in `finish()` (or include an invocation of this method from `finish()`? Are there cases when we shouldn't notify the publisher of completion when the operation finishes? If so, how are we cleaning up the sink & subscription in those cases?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/SubscriptionSync/ModelReconciliationQueueBehaviorTests.swift
```diff
@@ -89,9 +89,34 @@ class ModelReconciliationQueueBehaviorTests: ReconciliationQueueTestBase {
...
+                }
+            }
+        })
+
         queue.start()
 
-        wait(for: [event1Saved, event2Saved, event3Saved], timeout: 5.0, enforceOrder: true)
+        wait(for: [event1Saved,
+                   event2Saved,
+                   event3Saved], timeout: 5.0, enforceOrder: true)
```
palpatim, line 116: nit: format. Make sure to run swiftlint & swiftformat. If they both pass this, well then, we should revise those rules :)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
wooj2: I tried overriding finish, but it is marked as final.. i guess we could call `finish() `within `notifyFinish()`, but this seems cleaner?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
wooj2: thank you!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
wooj2: Done.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/SubscriptionSync/ModelReconciliationQueueBehaviorTests.swift
wooj2: looks like we'll need to revise the rules.

palpatim (approved):

petermauger:
This PR appears to be causing a crash in one of our environments. We're still trying to pinpoint the details of the crash. Will add a bug when we've figured out what's happening

palpatim:
We've released the fix for the regression in 1.3.1.

PR #763, opened 2020-09-10
AppSyncList POC

lawmicha:
*Issue #, if available:*

*Description of changes:*
Retrieving top level list of models with pagination functionality
```
static func paginatedList<M: Model>(_ modelType: M.Type, where predicate: QueryPredicate?, limit: Int?) 
      -> GraphQLRequest<AppSyncList<M>>
```
This introduces a new Model-to-GraphQLRequest builder that returns `AppSyncList`, and can be used for retrieve subseqential pages of data.

Alternatively, we can introduce a breaking change to replace `.list` which returns `[M]` instead of namespacing this `.paginatedList`.

**High level details**
`AppSyncList` is a subclass of `List` (now made as an open class), and conforms to `Paginatable`.
```
public class AppSyncList<ModelType: Model>: List<ModelType>, Paginatable
```
`List`, made open for `AppSyncList` to subclass and conforms to an empty protocol `ListModel`
```
open class List<ModelType: Model>: Collection, Codable, ExpressibleByArrayLiteral, ListModel
```

`Paginatable`, contains pagination related interface. *To be reviewed*
```
public protocol Paginatable {
    associatedtype Page
    associatedtype PageError: Error
    typealias PageResult = ((Result<Page, PageError>) -> Void)
    func next(onComplete: @escaping PageResult)
    func hasNext() -> Bool
}
```
*note: in lazy load functionality of list, it uses a lock to turn the async operation into syncrhonous, whereas here i've opt-d to keep it a callback*

`ListModel`, empty protocol, used in APIPlugin to indicate a second decode attempt can be done to store additional information in the response.
```
public protocol ListModel { }
```
*TODO: this empty protocol "marker" should probably be conformed to by AppSyncList, not on the List. 

**APIPlugin decoding logic**
APIPlugin does two decode passes. First pass decodes data from AppSync. Second pass is only taken when the data is successfully decoded from the first pass and is of  type `ListModel`. For ListModel types, a second decoding attempt is done to decode with `AppSyncListPayload` as the data.

`AppSyncListPayload`, contains additional request information to be stored in the responseType
```
public struct AppSyncListPayload: Codable {
    let document: String
    let variables: [String: JSONValue]?
    private let graphQLData: JSONValue
}
```

The second decoding attempt is required to store non-response data back in the response, such as the document and variables from the request. Having this request information in the response allows the pagination functionality to work. ie.

```
func hasNext() {
  // retrieve from graphQLData, check if nextToken is nil
}
```

```
func getNextPage() {
   // construct request with document, update variables with nextToken from graphQLData, perform API.query
  // return new results as next page
}
```

### DIFF
--- a/Amplify.xcodeproj/project.pbxproj
+++ b/Amplify.xcodeproj/project.pbxproj
@@ -121,6 +121,11 @@
 		21C395B3245729EC00597EA2 /* AppSyncErrorType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21C395B2245729EC00597EA2 /* AppSyncErrorType.swift */; };
 		21D79FDA237617C60057D00D /* SubscriptionEvent.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21D79FD9237617C60057D00D /* SubscriptionEvent.swift */; };
 		21D79FE32377F4120057D00D /* SubscriptionConnectionState.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21D79FE22377F4120057D00D /* SubscriptionConnectionState.swift */; };
+		21DF40362509CAC000D649A8 /* Paginatable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21DF40352509CAC000D649A8 /* Paginatable.swift */; };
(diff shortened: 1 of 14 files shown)


lawmicha:
closing for https://github.com/aws-amplify/amplify-ios/pull/806

PR #766, opened 2020-09-13
feat(datastore): Dispatch networkStatus event

ruiguoamz:
*Description of changes:*
### networkStatus
```swift
public struct NetworkStatusEvent {
    public let active: Bool

    public init(active: Bool) {
        self.active = active
    }
}
```


AmplifyTestCommon/Mocks/MockAPICategoryPlugin.swift
```diff
@@ -9,7 +9,7 @@ import Amplify
 import Combine
 import Foundation
 
-class MockAPICategoryPlugin: MessageReporter, APICategoryPlugin {
+class MockAPICategoryPlugin: MessageReporter, APICategoryPlugin, APICategoryReachabilityBehavior {
```
lawmicha, line 12: we're added conformance to the protocol but no new methods implemented in the mock, so casting a mock API to `APICategoryReachabilityBehavior` works, but what happens when it tries to use methods from there? 

does `networkReachabilityPublisher = try reachability.reachabilityPublisher()` fall into the catch block? Is there a mock publisher we can return?

DongQuanRui: I found out there is one implemented! (https://github.com/aws-amplify/amplify-ios/blob/1d96b55691de534324f8f1256bc6517cad074982/AmplifyTestCommon/Mocks/MockAPICategoryPlugin.swift#L111)

lawmicha: can we test returning a mocked response? like https://developer.apple.com/documentation/combine/just
```
return Just(ReachabilityUpdate.active)
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
```diff
@@ -28,11 +28,21 @@ class DataStoreHubEventTests: HubEventsIntegrationTestBase {
     ///    - outboxStatus received, payload should be {isEmpty: true}
```
lawmicha, line 28: update testcomment accordingly

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
```diff
@@ -62,6 +72,8 @@ class DataStoreHubEventTests: HubEventsIntegrationTestBase {
             return
         }
 
+        startAmplify()
```
lawmicha, line 75: how does this work? Amplify.Hub.listen is called before `startAmplify()` which calls `Amplify.configure()`

ruiguoamz: According to Tim, listening for events should work prior to `Amplify.configure()` being called. and `Combine` flavor of listening is better to be used because of its better deterministic property. So I will change it to `Hub.publisher.sink` here

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreHubEventsTests.swift
ruiguoamz: updated

lawmicha (approved):

PR #777, opened 2020-09-23
refactor(datastore): Added dependency on ModelSchema in delete api in storage engine

royjit:
*Description of changes:*

Added modelSchema into the storage engine `delete` api. This is part of the larger work to remove dependency on Model.Type used in the Storage engine apis.

Removed model name from `ModelRegistry.register()` since the value is already present inside `ModelSchema`. 

Added public api in the datastore plugin that can be used as escape hatch. The following apis are added: 

```swift

public func save<M: Model>(_ model: M,
                                modelSchema: ModelSchema,
                                where condition: QueryPredicate? = nil,
                                completion: @escaping DataStoreCallback<M>)

public func query<M: Model>(_ modelType: M.Type,
                                 modelSchema: ModelSchema,
                                 sort sortInput: QuerySortInput? = nil,
                                 paginate paginationInput: QueryPaginationInput? = nil,
                                 completion: DataStoreCallback<[M]>)

                                  modelSchema: ModelSchema,
                                  where predicate: QueryPredicate,
                                  completion: @escaping DataStoreCallback<Void>)

```

### DIFF
--- a/Amplify/Categories/DataStore/Model/Internal/ModelRegistry.swift
+++ b/Amplify/Categories/DataStore/Model/Internal/ModelRegistry.swift
@@ -35,23 +35,21 @@ public struct ModelRegistry {
     }
 
     public static func register(modelType: Model.Type) {
-        register(modelName: modelType.modelName,
-                 modelSchema: modelType.schema,
-                 modelType: modelType) { (jsonString, jsonDecoder) -> Model in
+        register(modelType: modelType,
+                 modelSchema: modelType.schema) { (jsonString, jsonDecoder) -> Model in
             let model = try modelType.from(json: jsonString, decoder: jsonDecoder)
             return model
         }
     }
 
-    public static func register(modelName: ModelName,
+    public static func register(modelType: Model.Type,
                                 modelSchema: ModelSchema,
-                                modelType: Model.Type,
                                 jsonDecoder: @escaping (String, JSONDecoder?) throws -> Model) {
         concurrencyQueue.sync {
             let modelDecoder: ModelDecoder = { jsonString, decoder in
                 return try jsonDecoder(jsonString, decoder)
             }
-
+            let modelName = modelSchema.name
             modelSchemaMapping[modelName] = modelSchema
             modelTypes[modelName] = modelType
             modelDecoders[modelName] = modelDecoder
--- a/Amplify/Categories/DataStore/Model/JSONHelper/JSONValueHolder.swift
+++ b/Amplify/Categories/DataStore/Model/JSONHelper/JSONValueHolder.swift
@@ -18,7 +18,7 @@ import Foundation
 ///             let values: [String: Any]
 ///
 ///             public func internalValue(for key: String) -> Any? {
-///                 return values[key] 
+///                 return values[key]
 ///             }
 ///          }
 public protocol JSONValueHolder {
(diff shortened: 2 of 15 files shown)


Amplify/Categories/DataStore/Model/Internal/ModelRegistry.swift
```diff
@@ -35,23 +35,21 @@ public struct ModelRegistry {
...
-                 modelType: modelType) { (jsonString, jsonDecoder) -> Model in
+        register(modelType: modelType,
+                 modelSchema: modelType.schema) { (jsonString, jsonDecoder) -> Model in
             let model = try modelType.from(json: jsonString, decoder: jsonDecoder)
             return model
         }
     }
 
-    public static func register(modelName: ModelName,
+    public static func register(modelType: Model.Type,
```
palpatim, line 45: I've lost track of what we've published--I assume that this change in signature to a `public` method is not a breaking change because we haven't yet released the DataStore changes that start to use `ModelSchema` right?

Amplify/Categories/DataStore/Model/JSONHelper/JSONValueHolder.swift
```diff
@@ -18,7 +18,7 @@ import Foundation
 ///             let values: [String: Any]
 ///
 ///             public func internalValue(for key: String) -> Any? {
-///                 return values[key] 
+///                 return values[key]
```
palpatim, line 21: Thank you, you've warmed my bitter, old, trailing-spaces-hating heart. 🖤

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -163,28 +189,26 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
...
-                                                publisher.send(input: mutationEvent)
-                                            }
-                                        } catch {
-                                            self.log.error(error: error)
-                                        }
-                                    case .failure(let error):
-                                        self.log.error(error: error)
-                                    }
+
+        guard #available(iOS 13.0, *) else {
```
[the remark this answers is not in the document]
palpatim: Nice! You've warmed my bitter, old, nested-indention-hating heart. 🖤

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngineAdapter.swift
```diff
@@ -15,14 +15,17 @@ protocol StorageEngineAdapter: class, ModelStorageBehavior {
     func save(untypedModel: Model, completion: @escaping DataStoreCallback<Model>)
 
     func delete<M: Model>(_ modelType: M.Type,
+                          modelSchema: ModelSchema,
```
palpatim, line 18: I'm a little bit sad that we have to add the `modelSchema` parameters even though it's an internal-only protocol. Is there any benefit to making the old signatures default `modelSchema` to `modelType.schema` for these, to support use cases like we have below where we specify `MutationEvent.self` and `MutationEvent.schema`?

royjit: Cannot give default value in the protocol, so it will be difficult to enforce it. Also if we add default value, there would be a chance that the ModelSchema is not passed in correctly. So I guess keeping it separate will be the safer approach for now.

palpatim: I'm not suggesting to add a default value, I'm suggesting two method signatures with two implementations, one that derives the default value from the incoming `model` and one that is specified. The one that derives would of course call the specific one.

palpatim (approved):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngineAdapter.swift
royjit: Updated as per our conversation, added default methods in the protocol.

PR #779, opened 2020-09-24
chore: Support Xcode 12

palpatim:
Fixes dependencies in Podspec and project to support building with Xcode 12.

**Steps for making this PR live**

- [x] Release AWS SDK 2.17.0
- [x] Update `AmplifyPlugins.podspec` with the SDK-wide `AWSCognitoIdentityProviderASF` podspec version
- [x] Update all dependency versions in Amplify to 2.17.0 (similar to https://github.com/aws-amplify/amplify-ios/pull/779/commits/ccb76cb496bee1a704ec4defcb1c48215acaf5eb)
- [x] Update all Podfile.lock with latest version of dependency pods

Prior to the AWS SDK release, this PR will remain as a Draft, but I am asking @wooj2 to review as if it were live.

*Description of changes:*

Do not merge this until the SDK release is done. We must also update the various Podfile.lock files with the new SDK signature after release.


AmplifyPlugins.podspec
```diff
@@ -46,8 +46,9 @@ Pod::Spec.new do |s|
 
     # This is technically redundant, but adding it here allows Xcode to find it
     # during initial indexing and prevent build errors after a fresh install
-    s.dependency 'AWSAuthCore', $OPTIMISTIC_AWS_SDK_VERSION
-
+    ss.dependency 'AWSAuthCore', $OPTIMISTIC_AWS_SDK_VERSION
+    ss.dependency 'AWSCognitoIdentityProvider', $OPTIMISTIC_AWS_SDK_VERSION
+    ss.dependency 'AWSCognitoIdentityProviderASF', '1.0.2'
```
wooj2, line 51: Seems that `AWSCognitoIdentityProviderASF` is currently `2.16.0` in the develop branch... and it was previously 1.1.0?  Do we want to adopt the 2.x.y scheme, or...?

Note looks like the Podfile.lock files also got updated?

AWSPluginsCore.podspec
```diff
@@ -10,7 +10,7 @@
 $AMPLIFY_VERSION = '1.2.0'
 $AMPLIFY_RELEASE_TAG = "v#{$AMPLIFY_VERSION}"
 
-$AWS_SDK_VERSION = '2.15.0'
+$AWS_SDK_VERSION = '2.16.0'
```
wooj2, line 13: Given that 2.16.0 is already out, and there needs to be another release to support XCode 12, seems like this needs to be 2.17.0 (along with everywhere else that references 2.16.0) ?

.circleci/config.yml
```diff
@@ -4,7 +4,7 @@ version: 2.1
 
 defaults: &defaults
   macos:
-    xcode: '11.4.1'
+    xcode: '12.0.0'
```
wooj2, line 7: Seems fine, but if circleci supports 12.0.1 (since it was released 2d ago, we should probably use that? )

AmplifyPlugins.podspec
palpatim: Locks got updated because I upgraded to 2.16. The ASF version doesn't start conforming with the overall SDK version until 2.17.0.

AWSPluginsCore.podspec
palpatim: Yes, we'll update to 2.17.0, but I want to do it incrementally as noted in the PR discussion above. The idea is to modify the PR incrementally to track progress of the dependencies, rather than a breaking PR that can't access unreleased dependencies.

.circleci/config.yml
palpatim: Circle CI [doesn't support 12.0.1 yet](https://circleci.com/docs/2.0/testing-ios/)

wooj2 (approved):

palpatim:
It's unclear why the `AmplifyTests` unit tests are failing in CircleCI, but they pass locally from both invocation methods (command line `xcodebuild`, running from Xcode). Preliminary research suggests this is a problem with the simulator not starting up in time, although I don't know whether I believe that. In any case, I'm going to investigate it separately and merge this PR in preparation for the release.

PR #781, opened 2020-09-29
refactor(datastore): Remove dependency on ModelType from ModelDecorator

royjit:
*Description of changes:* Removes the modelType from model decorator. This is part of the larger work on removing dependency on Model.Type in Datastore.

### DIFF
--- a/Amplify.xcodeproj/project.pbxproj
+++ b/Amplify.xcodeproj/project.pbxproj
@@ -1638,13 +1638,6 @@
 			path = Decorator;
 			sourceTree = "<group>";
 		};
-		213481D7242AFA58001966DE /* Support */ = {
-			isa = PBXGroup;
-			children = (
-			);
-			path = Support;
-			sourceTree = "<group>";
-		};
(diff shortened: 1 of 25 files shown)


lawmicha (approved):
Since this is related to DataStore, i'm guessing you'll be updating references here as well: https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest%2BAnyModelWithSync.swift to remove dependency on Model.Type

royjit:
> Since this is related to DataStore, i'm guessing you'll be updating references here as well: https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest%2BAnyModelWithSync.swift to remove dependency on Model.Type

Yes, I am doing this change incrementally to reduce the size of PR. I will be changing this in another PR.

PR #788, opened 2020-09-30
fix(DataStore): owner based auth, read operations

wooj2:
In attempts to fix:
https://github.com/aws-amplify/amplify-ios/issues/691

Corresponding android change:
https://github.com/aws-amplify/amplify-android/pull/807/files#diff-5a6612b4d7b74db7c28a89d55c63cd09


wooj2:
There seems to be an issue with using a model like:
```
 type ModelMultipleOwner @model
   @auth(rules: [
     { allow: owner },
     { allow: owner, ownerField: "editors", operations: [update, read] }
   ]) {
   id: ID!
   content: String
   editors: [String]
 }
```
That being said, this is already an issue in our current codebase, and the code proposed in this PR doesn't make this issue worse.  We will fix this issue in a subsequent PR.

richardmcclellan (approved):

lawmicha (approved):
FYI there are also some API integration tests that needs to be updated here https://github.com/aws-amplify/amplify-ios/tree/main/AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/AuthDirective

but i'm guessing these will be vastly improved later anyways on the work you are doing for auth directive support

Issue #485, opened 2020-05-26
Owner-based auth based on IdentityClaim


lawmicha:
Currently, a developer using Auth directive can set their own `identityClaim` which determines which value to use as the owner. DataStore uses `Auth.getCurrentUser` and uses the `username` as owner. This feature is to support different `identityClaims` like cognito sub or something else. 

The impact of not doing this is no support for custom identityClaims.

palpatim:
Code snippet for reference https://github.com/aws-amplify/amplify-ios/blob/4cb4423a6eb3d46dc0caf4bf55392384b0e4acc8/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift#L159-L165

lawmicha:
https://github.com/aws-amplify/amplify-ios/pull/794#discussion_r499046705

wooj2:
This PR should resolve this issue:
https://github.com/aws-amplify/amplify-ios/pull/813

PR #794, opened 2020-10-01
fix(Core): support identityClaim "sub"

wooj2:
In attempts to resolve:
https://github.com/aws-amplify/amplify-ios/issues/485

### DIFF
--- a/AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
@@ -9,8 +9,7 @@ import Foundation
 import Amplify
 
 public enum AuthRuleDecoratorInput {
-    public typealias OwnerId = String
-    case subscription(GraphQLSubscriptionType, OwnerId)
+    case subscription(GraphQLSubscriptionType, AuthUser)
     case mutation
     case query
 }
@@ -60,13 +59,14 @@ public struct AuthRuleDecorator: ModelBasedGraphQLDocumentDecorator {
         let ownerField = authRule.getOwnerFieldOrDefault()
         selectionSet = appendOwnerFieldToSelectionSetIfNeeded(selectionSet: selectionSet, ownerField: ownerField)
 
-        guard case let .subscription(_, ownerId) = input else {
+        guard case let .subscription(_, authUser) = input else {
             return document.copy(selectionSet: selectionSet)
         }
 
         if isOwnerInputRequiredOnSubscription(authRule) {
             var inputs = document.inputs
-            inputs[ownerField] = GraphQLDocumentInput(type: "String!", value: .scalar(ownerId))
+            let identityClaimValue = resolveIdentityClaimValue(authRule, authUser)
+            inputs[ownerField] = GraphQLDocumentInput(type: "String!", value: .scalar(identityClaimValue))
             return document.copy(inputs: inputs, selectionSet: selectionSet)
         }
         return document.copy(selectionSet: selectionSet)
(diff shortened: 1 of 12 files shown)


AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
```diff
@@ -76,6 +76,19 @@ public struct AuthRuleDecorator: ModelBasedGraphQLDocumentDecorator {
...
 
+    private func resolveIdentityClaimValue(_ authRule: AuthRule, _ authUser: AuthUser) -> String {
+        guard let identityClaim = authRule.identityClaim else {
+            return authUser.username
+        }
+        if identityClaim == "username" {
+            return authUser.username
+        } else if identityClaim == "sub" {
+            return authUser.userId
+        }
```
lawmicha, line 87: i think these might be "cognito:username" https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/appsync-swift-visitor.test.ts#L1012

and "sub" looks correct: https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/appsync-swift-visitor.test.ts#L1120

lawmicha (changes requested):

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
richardmcclellan: This differs from the current Android logic, [here](https://github.com/aws-amplify/amplify-android/blob/e1cd01574343c3d0a61e19cdded890307604da0e/aws-api/src/main/java/com/amplifyframework/api/aws/AWSApiPlugin.java#L335-L346).  My understanding is that Android gets the username from the decrypted authToken instead of from `getUsername`, which is necessary to support email/phone sign in.

@TrekSoft can you confirm?

wooj2: Tested w/ schema of:
```
type ScenarioATest6Post @model @auth(rules: [{ allow: owner, identityClaim: "username" }]) {
  id: ID!
  title: String!
}
```
and
```
type ScenarioATest6Post @model @auth(rules: [{ allow: owner}]) {
  id: ID!
  title: String!
}
```

in both cases, in our model generated schema, we see:

```
    
    model.authRules = [
      rule(allow: .owner, ownerField: "owner", identityClaim: "cognito:username", operations: [.create, .update, .delete, .read])
    ]
```

lawmicha: change looks good as incremental improvement as it would support sub and username (iff user uses username for sign in). Considering Android's current implementation, we should probably keep https://github.com/aws-amplify/amplify-ios/issues/485 open to address the username as email and phone sign in to do a similar JWT parse to extract the username out

lawmicha (approved):
approved if u want to merge this in as incremental support, but should consider Android's implementation differences and their requirement for parsing the JWT token

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/AuthRuleDecorator.swift
lawmicha: i wonder, if the auth plugin's get username behaves differently on Android and iOS such that on iOS we can rely on `getUsername`

TrekSoft: We shouldn't rely on getUsername - Cognito is comparing what is sent to what is in the token and it simplifies the code as well to just do it like Android is where we simply pull the right value from the token. With custom identity claims, we can pull whatever other value is specified.

wooj2: @TrekSoft & @richardmcclellan  -- thanks for calling this out.. Looked into implementing the parsing of the JWT token, and it's a bit more work than expecting.  For this reason, we are going to push this code in and view it as an incremental improvement, while keeping #485 open to ultimately fix it in the manner you have suggested.  Thanks again for extra eyes on this.

PR #796, opened 2020-10-02
fix(api): ReachabilityUpdate

ruiguoamz:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
@@ -15,7 +15,7 @@ class NetworkReachabilityNotifier {
     private var reachability: NetworkReachabilityProviding?
     private var allowsCellularAccess = true
 
-    let reachabilityPublisher = CurrentValueSubject<ReachabilityUpdate, Never>(ReachabilityUpdate(isOnline: false))
+    let reachabilityPublisher: CurrentValueSubject<ReachabilityUpdate, Never>
     var publisher: AnyPublisher<ReachabilityUpdate, Never> {
         return reachabilityPublisher.eraseToAnyPublisher()
     }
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
```diff
@@ -40,7 +52,6 @@ class NetworkReachabilityNotifier {
 
     deinit {
         reachability?.stopNotifier()
-        NotificationCenter.default.removeObserver(self)
```
ruiguoamz, line 43: Xcode says it will be automatically removed, so that no need to call `removeObserver()`

AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
```diff
@@ -26,6 +26,18 @@ class NetworkReachabilityNotifier {
...
 
+        let isReachable: Bool
+        switch reachability?.connection {
+        case .wifi:
+            isReachable = true
+        case .cellular:
+            isReachable = allowsCellularAccess
+        case .none, .unavailable, .some(.none):
+            isReachable = false
+        }
```
lawmicha, line 37: this code looks duplicated, how about `getReachabilityUpdate() -> Bool`?

AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
```diff
@@ -60,8 +71,10 @@ class NetworkReachabilityNotifier {
             isReachable = false
         }
 
-        let reachabilityMessageUpdate = ReachabilityUpdate(isOnline: isReachable)
-        reachabilityPublisher.send(reachabilityMessageUpdate)
+        if isReachable != reachabilityPublisher.value.isOnline {
```
lawmicha, line 74: what was the behavior from before? if there was no check for previous value before (`if isReachable != reachabilityPublisher.value.isOnline`) then reachability always provided an update even for connection changes from `wifi` to `cellular`, but now with this check, the reachbility won't send an update

AmplifyPlugins/API/AWSAPICategoryPluginTests/Reachability/NetworkReachabilityNotifierTests.swift
```diff
@@ -81,7 +76,7 @@ class NetworkReachabilityNotifierTests: XCTestCase {
         }, receiveValue: { value in
             values.append(value.isOnline)
             if values.count == 2 {
```
lawmicha, line 78: the setup portion of this test is now related to the assert here, from the reader's prespective, i have no idea why you are checking for count 2 unless i look into the set up of the test. even though all tests are using `MockReachability.iConnection = .wifi` at line 29, i think it's clearer if you put that line in every test and move it out of the setup

lawmicha (changes requested):

PR #804, opened 2020-10-06
refactor(datastore): Removed dependency of Model.Type in Initial sync

royjit:
- Removed dependency of Model.Type in Initial sync. Uses `__typename` to figure out the model name for the untyped model in save
- Mutation event takes a modelSchema now to identity the modelName.

### DIFF
--- a/Amplify/Categories/DataStore/Subscribe/MutationEvent.swift
+++ b/Amplify/Categories/DataStore/Subscribe/MutationEvent.swift
@@ -53,6 +53,21 @@ public struct MutationEvent: Model {
 
     }
 
+    @available(*, deprecated, message: """
+    Init method without ModelSchema is deprecated, use the other init methods.
+    """)
+    public init<M: Model>(model: M,
+                          mutationType: MutationType,
+                          version: Int? = nil,
+                          graphQLFilterJSON: String? = nil) throws {
+        try self.init(model: model,
+                      modelSchema: model.schema,
+                      mutationType: mutationType,
+                      version: version,
+                      graphQLFilterJSON: graphQLFilterJSON)
+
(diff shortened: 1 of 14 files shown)


Amplify/Categories/DataStore/Subscribe/MutationEvent.swift
```diff
@@ -39,13 +39,13 @@ public struct MutationEvent: Model {
     }
 
     public init<M: Model>(model: M,
+                          modelSchema: ModelSchema,
```
palpatim, line 42: Breaking change

palpatim (changes requested):
Change looks OK overall, but you need to create a new initializer for `MutationEvent` and deprecate the existing initializer rather than adding a new required member to it, to avoid a breaking change.

Amplify/Categories/DataStore/Subscribe/MutationEvent.swift
royjit: Added back the old init with deprecated warning.

Amplify/Categories/DataStore/Subscribe/MutationEvent.swift
```diff
@@ -53,6 +53,21 @@ public struct MutationEvent: Model {
 
     }
 
+    @available(*, deprecated, message: """
+    Init method without ModelSchema is deprecated, use the other init methods.
```
palpatim, line 57: Reword to "Initializing from a model without a ModelSchema is deprecated. Use init(model:modelSchema:mutationType:version:graphQLFilterJSON:) instead."

palpatim (approved):
Approved w/changes

PR #806, opened 2020-10-06
Support API Pagination with List refactoring

lawmicha:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Amplify/Categories/DataStore/Model/Collection/List+Model.swift
+++ b/Amplify/Categories/DataStore/Model/Collection/List+Model.swift
@@ -7,125 +7,165 @@
 
 import Foundation
 
-/// `List<ModelType>` is a DataStore-aware custom `Collection` that is capable of loading
-/// records from the `DataStore` on-demand. This is specially useful when dealing with
-/// Model associations that need to be lazy loaded.
-///
-/// When using `DataStore.query(_ modelType:)` some models might contain associations
-/// with other models and those aren't fetched automatically. This collection keeps track
-/// of the associated `id` and `field` and fetches the associated data on demand.
-public class List<ModelType: Model>: Collection, Codable, ExpressibleByArrayLiteral {
+/// Provides page related functionality when conforming to `Paginatable` from an API Category plugin.
+public protocol Paginatable {
 
-    public typealias Index = Int
+    associatedtype Page
+
+    typealias PageResult = ((Result<Page, APIError>) -> Void)
+
+    /// Checks if there is subsequent data.
+    func hasNextPage() -> Bool
+
+    /// Retrieves the next set of data, limited to the `limit` number of results, when completed will be returned as
+    /// a `Page` on `onComplete` closure.
+    func getNextPage(limit: Int?, onComplete: @escaping PageResult)
+}
+
+/// Methods related to lazy loading data
+public protocol LazyLoad {
+    // Persistant operation to set the size for lazy loading of data.
+    func limit(_ limit: Int) -> Self
+}
+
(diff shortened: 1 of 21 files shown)


AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncList+Decoder.swift
```diff
@@ -0,0 +1,42 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+//
+//public class AppSyncListDecoder: ListDecoder {
```
richardmcclellan, line 10: If this class is not used, can it be removed from the PR?

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -7,125 +7,165 @@
...
-        self.limit = limit
-        state = .pending
-        return self
+    // MARK: Paginatable
+
+    open func hasNextPage() -> Bool {
+        return false
+    }
+
+    open func getNextPage(limit: Int? = nil, onComplete: @escaping PageResult) {
```
richardmcclellan, line 139: I played around with something like this on Android, but ultimately we implemented a `getRequestForNextResult` instead, which returns a `GraphQLRequest`, rather than actually executing it.   I think this is a better approach because:
 1.  Better DX - The logic for making subsequent requests is congruent with the first request (Call `Amplify.API.query` for each one)
 2. More extensible - Customers may want the option to modify the `GraphQLRequest` prior to executing it (perhaps apply a decorator to modify some other variable).  By providing the `GraphQLRequest` to the customer, it keeps this possibility available.

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncList.swift
```diff
@@ -0,0 +1,162 @@
...
+import Foundation
+import Amplify
+import AWSPluginsCore
+
+public class AppSyncList<ModelType: Model>: List<ModelType>, ModelListDecoder {
+
+    public typealias Page = AppSyncList<ModelType>
+    public typealias PageError = APIError
+
+    /// The array of `Element` that backs the custom collection implementation.
```
richardmcclellan, line 17: Is this comment meant to be here?  I'm not sure what it is referring to.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList+Decoder.swift
```diff
@@ -0,0 +1,41 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+
+//public class DataStoreListDecoder: ListDecoder {
```
richardmcclellan, line 10: Can this class be removed?

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
lawmicha: thanks for calling this out, i'll review with the iOS folks on these two options and call out that this is what's on Android https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/api/graphql/PaginatedResult.java

PR #810, opened 2020-10-07
fix(auth): parse and surface returned subscription @auth errors

diegocstn:
When a subscription fails due to an authorization issue, parse and surface AppSyncJSONValue errors
in order to let DataStore recover

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
@@ -110,8 +110,7 @@ final public class AWSGraphQLSubscriptionOperation<R: Decodable>: GraphQLSubscri
         case .data(let data):
             onGraphQLResponseData(data)
         case .failed(let error):
-            dispatch(result: .failure(APIError.operationError("subscription item event failed with error", "", error)))
-            finish()
+            onSubscriptionFailure(error)
         }
     }
 
@@ -159,6 +158,21 @@ final public class AWSGraphQLSubscriptionOperation<R: Decodable>: GraphQLSubscri
             dispatch(result: .failure(APIError.operationError("Failed to deserialize", "", error)))
             finish()
         }
+    }
+
+    private func onSubscriptionFailure(_ error: Error) {
+        let errorDescription = "Subscription item event failed with error"
+        if case let ConnectionProviderError.subscription(_, payload) = error,
+           let errors = payload?["errors"] as? AppSyncJSONValue,
+           let graphQLErrors = try? GraphQLResponseDecoder.decodeAppSyncErrors(errors) {
+            let graphQLResponseError = GraphQLResponseError<R>.error(graphQLErrors)
+            dispatch(result: .failure(APIError.operationError(errorDescription, "", graphQLResponseError)))
+            finish()
+            return
+        }
+
+        dispatch(result: .failure(APIError.operationError(errorDescription, "", error)))
+        finish()
     }
 
 }
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/AppSyncJSONValue+toJSONValue.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/AppSyncJSONValue+toJSONValue.swift
@@ -1,0 +1,32 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+import AppSyncRealTimeClient
+
+extension AppSyncJSONValue {
+    static func toJSONValue(_ json: AppSyncJSONValue) -> JSONValue {
+        switch json {
+        case .array(let values):
+            return JSONValue.array(values.map(AppSyncJSONValue.toJSONValue))
+        case .boolean(let value):
+            return JSONValue.boolean(value)
+        case .null:
+            return JSONValue.null
+        case .number(let value):
+            return JSONValue.number(value)
+        case .object(let content):
+            return JSONValue.object(content.reduce(into: [:]) { acc, partial in
+                let (key, value) = partial
+                acc[key] = AppSyncJSONValue.toJSONValue(value)
+            })
+        case .string(let value):
+            return JSONValue.string(value)
+        }
+    }
+}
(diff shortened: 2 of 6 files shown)


wooj2 (commented):
Looks like we are parsing appsync errors and converting them to graphql errors so that we can expose them in the API category?  Seems fine to me, but I'll defer to @lawmicha take a look.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
```diff
@@ -88,6 +88,10 @@ class ProcessMutationErrorFromCloudOperation: AsynchronousOperation {
                 finish(result: .success(nil))
             case .conflictUnhandled:
                 processConflictUnhandled(extensions)
+            case .unauthorized:
+                // TODO: dispatch Hub event
+                log.debug("Unauthorized mutation \(errorType)")
+                finish(result: .success(nil))
```
wooj2, line 94: I'm guessing we'll handle the subscriptions at a later time?

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
```diff
@@ -161,4 +160,27 @@ final public class AWSGraphQLSubscriptionOperation<R: Decodable>: GraphQLSubscri
...
+
+        if let graphQLErrors = try? GraphQLResponseDecoder.decodeAppSyncErrors(errors),
+           let extensions = graphQLErrors.first?.extensions,
+           let recoveryMessage = graphQLErrors.first?.message,
+           case let .string(errorTypeValue) = extensions["errorType"] {
+                let appSyncError = AppSyncErrorType(errorTypeValue)
+                dispatch(result: .failure(APIError.operationError(errorDescription, recoveryMessage, appSyncError)))
+                finish()
+                return
+        }
```
lawmicha, line 180: The field "errorType" is coming back from AppSync services is very AppSync specific and isn't really part of the graphQL spec. I recall [DataStore code](https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift#L81-L94) checks inside the "extensions" for "errorType" instead of just "errorType" at the top level. And this was because the [merging](https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/GraphQLResponseDecoder+DecodeError.swift#L43) happens in the API plugin when decoding a service response with errors like this:
```
{
 "errors": [
   {
     "message":"",
     "extensions: {}
     "errorType:"someError",
   }
 ]
}
```

will transform it into
```
// [GraphQLError] type
{
 "errors": [
   {
     "message":"", // merging logic skips this
     "extensions: { 
        "errorType": "someError"
     }
   }
 ]
}
```

I think the logic you have here to special case it and return an AppSyncErrorType would work for the purpose for DataStore as the client if all it needs is the errorType, but I think it may be more future proof to return `[GraphQLError]` so for all clients dealing with failed subscription connections can handle it in their own way, giving back as much data as possible. How about `let graphQLResponseError = GraphQLResponseError<R>.error(graphQLErrors)` ([GraphQLResponseError](https://github.com/aws-amplify/amplify-ios/blob/main/Amplify/Categories/API/Response/GraphQLResponse.swift))?
```
private func onSubscriptionFailure<R: Decodable>(_ error: Error, responseType: R.Type) {
        let errorDescription = "Subscription item event failed with error"
        if case let ConnectionProviderError.subscription(_, payload) = error,
              let errors = payload?["errors"] as? AppSyncJSONValue,
              let graphQLErrors = try? GraphQLResponseDecoder.decodeAppSyncErrors(errors) {
            let graphQLResponseError = GraphQLResponseError<R>.error(graphQLErrors)
            dispatch(result: .failure(APIError.operationError(errorDescription, "", graphQLResponseError)))
            finish()
            return
        }
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/AppSyncJSONValue+toJSONValue.swift
```diff
@@ -0,0 +1,32 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+import AppSyncRealTimeClient
+
+extension AppSyncJSONValue {
+    static func toJSONValue(_ json: AppSyncJSONValue) -> JSONValue {
```
lawmicha, line 13: this looks good to me, i'm not sure if there's any other way around this? maybe there's a way to just decode the `payload?["errors"]` directly to `[JSONValue.self]` instead of this

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
lawmicha: then on the DataStore client code, you would get the operationError, and `underlyingError as? GraphQLResponseError<MutationSync<AnyModel>>`
https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisher.swift#L22

diegocstn: thanks! makes perfectly sense, I didn't think about returning a `GraphQLResponseError`

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/GraphQLResponseDecoder+DecodeError.swift
```diff
@@ -29,6 +30,20 @@ extension GraphQLResponseDecoder {
         return responseErrors
     }
 
+    static func decodeAppSyncErrors(_ appSyncJSON: AppSyncJSONValue?) throws -> [GraphQLError] {
```
lawmicha, line 33: can the input be non-optional? 
`static func decodeAppSyncErrors(_ appSyncJSON: AppSyncJSONValue) throws -> [GraphQLError]`

If i understand correctly, `let errors = payload?["errors"] as? AppSyncJSONValue`, `errors` should be non-optional

diegocstn: yes, you're right! At that point, `errors` has always a non-optional value, thanks

lawmicha (approved):

PR #812, opened 2020-10-07
feat(datastore): dispatch ready event

ruiguoamz:
*Description of changes:*

In this PR, I will implement:

- ready

The ready events should wait for two conditions:
1. Received a `syncQueriesReady` event
2. The state of `RemoteSyncEngine` is about to become `syncEngineActive` 
(states -> `notifyingSyncStarted` -> `syncEngineActive`, since `syncEngineActive` is not doing anything, so I decided one of two conditions become `syncStarted`)
### ready
```swift
N/A
```


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ReadyEventEmitter.swift
```diff
@@ -0,0 +1,58 @@
...
+import Amplify
+import AWSPluginsCore
+import Combine
+
+@available(iOS 13.0, *)
+final class ReadyEventEmitter {
+    /// Two conditions should be met to emit one `ready` event:
+    /// - receive one `syncQueriesReady` event
+    /// - the state of remoteSyncEngine move to `syncStarted`
+    let conditionCount = AtomicValue(initialValue: 0)
```
ruiguoamz, line 17: the use of `dispatchGroup` causes the unit tests to fail. I agree that using `Semaphore` and `DispatchGroup` may not be a good idea. So I changed to using `conditionCount`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ReadyEventEmitter.swift
```diff
@@ -0,0 +1,58 @@
...
+    let remoteSyncEnginePublisher: AnyPublisher<RemoteSyncEngineEvent, DataStoreError>
+    var remoteEngineSink: AnyCancellable?
+    var syncQueriesReadyEventSink: AnyCancellable?
+
+    init(remoteSyncEnginePublisher: AnyPublisher<RemoteSyncEngineEvent, DataStoreError>) {
+        self.remoteSyncEnginePublisher = remoteSyncEnginePublisher
+    }
+
+    func configure() {
```
ruiguoamz, line 26: Regarding this comment: in this case, I don't need a `configure()`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ReadyEventEmitter.swift
palpatim: You can take better advantage of Combine's composability. Counting conditions isn't meaningful by itself, and you can express your intent precisely using Combine operators:

```swift
final class ReadyEventEmitter {
    var readySink: AnyCancellable?
    init(remoteSyncEnginePublisher: AnyPublisher<RemoteSyncEngineEvent, DataStoreError>) {
        let queriesReadyPublisher = ReadyEventEmitter.makeSyncQueriesReadyPublisher()
        let syncEngineStartedPublisher = ReadyEventEmitter.makeRemoteSyncEngineStartedPublisher(
            remoteSyncEnginePublisher: remoteSyncEnginePublisher
        )
        readySink = Publishers
            .Merge(queriesReadyPublisher, syncEngineStartedPublisher)
            .sink(receiveCompletion: { completion in
                switch completion {
                case .finished:
                    self.dispatchReady()
                    print("Need to do something about the error? \(dataStoreError)")
                }
            }, receiveValue: { _ in })
    }

    private static func makeSyncQueriesReadyPublisher() -> AnyPublisher<Void, DataStoreError> {
        Amplify.Hub
            .publisher(for: .dataStore)
            .filter { $0.eventName == HubPayload.EventName.DataStore.syncQueriesReady }
            .first()
            .map { _ in () }
            .setFailureType(to: DataStoreError.self)
            .eraseToAnyPublisher()
    }

    private static func makeRemoteSyncEngineStartedPublisher(
        remoteSyncEnginePublisher: AnyPublisher<RemoteSyncEngineEvent, DataStoreError>
    ) -> AnyPublisher<Void, DataStoreError> {
        remoteSyncEnginePublisher
            .filter { if case .syncStarted = $0 { return true } else { return false } }
            .first()
            .map { _ in () }
            .eraseToAnyPublisher()
    }

    private func dispatchReady() {
        let readyEventPayload = HubPayload(eventName: HubPayload.EventName.DataStore.ready)
        Amplify.Hub.dispatch(to: .dataStore, payload: readyEventPayload)
    }

}
```

This sample passes unit tests, and expresses your intent more clearly.

palpatim (changes requested):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ReadyEventEmitter.swift
```diff
@@ -0,0 +1,60 @@
...
+            remoteSyncEnginePublisher: remoteSyncEnginePublisher
+        )
+        readySink = Publishers
+            .Merge(queriesReadyPublisher, syncEngineStartedPublisher)
+            .sink(receiveCompletion: { completion in
+                switch completion {
+                case .finished:
+                    self.dispatchReady()
+                case .failure(let dataStoreError):
+                    self.log.error("Failed to emit ready event, error: \(dataStoreError)")
```
ruiguoamz, line 27: Since this is just an event emitter, the root cause of error is not in this class. The real functioning parts should take care of the error, say if `RemoteSyncEngine` cannot move to state `syncEngineActive`, `RemoteSyncEngine` itself should handle the error.

So, I simply put a `log.error` here.

palpatim: Sounds good.

palpatim (approved):

PR #815, opened 2020-10-09
fix(datastore): keep DataStore sync engine running even if model subscriptions fail

diegocstn:
Keep DataStore sync engine running even if model subscriptions fail

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
@@ -43,7 +43,7 @@ final class AWSIncomingEventReconciliationQueue: IncomingEventReconciliationQueu
     private var modelReconciliationQueueFactory: ModelReconciliationQueueFactory
     
     private var isInitialized: Bool {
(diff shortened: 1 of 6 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
```diff
@@ -107,10 +111,18 @@ final class AWSIncomingEventReconciliationQueue: IncomingEventReconciliationQueu
...
             connectionStatusSerialQueue.async {
                 self.reconciliationQueueConnectionStatus[modelName] = true
-                if self.reconciliationQueueConnectionStatus.count == self.reconciliationQueues.count {
+                if self.isInitialized {
+                    self.eventReconciliationQueueTopic.send(.initialized)
+                }
+            }
+        case .disconnected(modelName: let modelName, reason: .unauthorized):
+            connectionStatusSerialQueue.async {
+                self.reconciliationQueues[modelName]?.cancel()
```
wooj2, line 123: in addition to canceling reach of the `reconciliationQueues`, seems like we also need to cancel each one of the `modelReconciliationQueueSinks`?

diegocstn: Good call, I'll update

wooj2 (approved):
lgtm!

PR #830, opened 2020-10-16
fix: auth category should throw AuthError instead of PredictionsError

jtbergman:
*Issue #, if available:*
Fixes https://github.com/aws-amplify/amplify-ios/issues/829

*Description of changes:*
When adding an Auth plugin using `Amplify.add(plugin:)`, if the plugin does not have a key a `PredictionsError` will be thrown. Since this is the `AuthCategory` I would expect an `AuthError` to be thrown. This code changes the error to an `AuthError` and adds a new unit test to verify (1) that an error is thrown and (2) that the error is an `AuthError`.


royjit (approved):

lawmicha:
Thanks @jtbergman! merging this in

PR #843, opened 2020-10-21
Main oidc factory enum

wooj2:
Customer facing call site would look like (owned by customer):
```
            let dataStorePlugin = AWSDataStorePlugin(modelRegistration: AmplifyModels())
            try Amplify.add(plugin: dataStorePlugin)
            try Amplify.add(plugin: AWSAPIPlugin(apiAuthProviderFactory: .oidc({ return MyOIDCAuthProvider() })))
            try Amplify.configure()
```

Impl of`MyOIDCAuthProvider` (owned by customer)
```
class MyOIDCAuthProvider : AmplifyOIDCAuthProvider {
///... whatever implementation they want
}
```

Impl of `APIAuthProviderFactory` (owned by AWS)
```
public enum APIAuthProviderFactory {
    case oidc(() -> AmplifyOIDCAuthProvider)
    case none
}

public protocol AmplifyOIDCAuthProvider {
    func getLatestAuthToken() -> Result<String, Error>
}
```

In the event that we add a new authProvider type:
* (negative) Customers can only choose one auth provider type at a time
* does not break calling site for customers only using single auth provider.
* Required change to Amplify owned APIAuthProviderFactory enum
```
import Foundation

public enum APIAuthProviderFactory {
    case oidc(() -> AmplifyOIDCAuthProvider)
+    case apiKey(() -> AmplifyAPIKeyAuthProvider)
    case none
}

public protocol AmplifyOIDCAuthProvider {
    func getLatestAuthToken() -> Result<String, Error>
}
+public protocol AmplifyAPIKeyAuthProvider {
+    func getLatestApiKey() -> Result<String, Error>
+}

```


wooj2:
Discarding this PR -- decided that this is not a favorable approach.

PR #845, opened 2020-10-21
feat: Support pagination for API and DataStore queries

lawmicha:
*Issue #, if available:*

*Description of changes:*
This PR mainly focuses on introducing a new API for AWSAPIPlugin users without breaking existing APIs. In doing so, List operations related to DataStore have been moved over to DataStore plugin. A "ModelListRegistry" has been added to perform decoding on the fly to instantiate subclasses of List that can perform their specific list operations (AppSyncList or DataStoreList). 

**Public APIs added**
1. `.paginatedList`
```swift
extension GraphQLRequest {
}
```

_note regarding `.paginatedList` vs existing `.list`._ 
```swift
static func list<M: Model>(_ modelType: M.Type,
                               where predicate: QueryPredicate?) -> GraphQLRequest<[M]>
```
Why not just replace the return type from the existing `.list` method from `[]` to `List`? 

The Amplify `List` that conforms to [Collection](https://developer.apple.com/documentation/swift/collection) is not the same as a [Array](https://developer.apple.com/documentation/swift/array). Replacing with`.list` instead of a new method `.paginatedList` will break existing customers that rely on the Array methods such as `contains()` or `append()`. Method overloading will cause a lot of DX pains for having ambiguous return type at the call site and would require the developer to explicitly specify that the return that is either `GraphQLResponse<List<M>>` or `GraphQLResponse<[M]>`. Because of this, we have a phased approach to first introduce `.paginatedList` as a separate API, and provide a breaking change for customers in the future where `.list` returns `List` and can retrieve the underlying array from the list object as `list.elements`.

2. `Paginatable` protocol for `List`
```swift
/// Provides paging related functionality
public protocol Paginatable {

    typealias PageResultCallback = (Result<Page, CoreError>) -> Void
    associatedtype Page

    /// Checks if there is subsequent data to retrieve. If True, retrieve the next page using `fetch()`
    func hasNextPage() -> Bool

    /// Retrieves the next page as a new in-memory object asynchronously.
    func fetch(completion: @escaping PageResultCallback)
}
```
This provides methods on the resulting list object to perform pagination. Currently [Android](https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/api/graphql/PaginatedResult.java) has `getItems()`, `hasNextResult()`, and `getRequestForNextResult()` which provide the same functionality with different calling pattern. 

3. The underlying array has been made public so you can access the elements via
```swift
let array = list.elements
```

**Developer Experience**

**Call site; How to create a List**
```swift
Amplify.API.query(.paginatedList(Comments.self, limit: 10)) { result in
   switch result {
   case .success(let graphQLResponse):
       switch graphqlResponse {
          case .success(let comments):
          case .failure(let graphQLErrorResponse):
       }
   case .failure(let apiError):
   }
}
```

**How to load into a new list: Pagination**

```swift
if comments.hasNextPage() {
    comments.fetch { result in
        switch result {
        case .success(let list):
            print(list)
        case .failure(let coreError):
            print("Failed to fetch next page \(coreError)")
            switch coreError {
               case .validation: break
               case .listOperation: break
               case .unknown: break
            }
        }
    }
}
```

**How to aggregate pages of data**

If you are using the results as a backing store to hydrate some UI components, you should create your own in-memory array to append new pages into it. The reason for this is because there is metadata stored in the object is used to fetch the next page, so any current page only knows how to fetch the next page from the current page. ie. If you append it to the existing page, your `fetch` call will always retrieve the subsequent page of the original page.
```swift
var resultsArray: [Post] = [] // binding to UI components
var currentPage: List<Post>?

_ = Amplify.API.query(request: .paginatedList(Post.self, limit: 1)) { event in
    switch event {
    case .success(let response):
        guard case let .success(graphQLResponse) = response else {
            XCTFail("Missing successful response")
            return
        }

        resultsArray.append(contentsOf: graphQLResponse)
        currentPage = graphQLResponse

    case .failure(let error):
        XCTFail("Unexpected .failure event: \(error)")
    }
}

// respond to UI actions to retrieve next page, ie.
func fetch() {
    if let current = currentPage, current.hasNextPage() {
        current.fetch { result in
            switch result {
            case .success(let list):
                resultsArray.append(contentsOf: list)
                currentPage = list
                fetch()
            case .failure(let coreError):
                print("Failed to fetch next page \(coreError)")
            }
        }
    }
}
```


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList+LazyLoad.swift
```diff
@@ -6,28 +6,19 @@
...
 
-    // MARK: - Asynchronous API
-
-    /// Trigger `DataStore` query to initialize the collection. This function always
-    /// fetches data from the `DataStore.query`.
-    ///
-    /// - seealso: `load()`
-    public func load(_ completion: DataStoreCallback<Elements>) {
-        lazyLoad(completion)
-    }
```
lawmicha, line 29: This is moved over to DataStoreList.swift, that subclasses List, to prevent API breaking change

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList+LazyLoad.swift
```diff
@@ -58,21 +49,6 @@ extension List {
...
-    /// the internal behavior which relies on `DispatchSemaphore` and will block the
-    /// current `DispatchQueue` until data is ready. When operating on large result
-    /// sets, prefer using the asynchronous `load(completion:)` instead.
-    ///
-    /// - Returns: the current instance after data was loaded.
-    /// - seealso: `load(completion:)`
-    public func load() -> Self {
-        lazyLoad()
-        return self
-    }
```
lawmicha, line 74: This is moved over to DataStoreList.swift, that subclasses List, to prevent API breaking change

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -6,126 +6,136 @@
...
-    /// The associatedField represents the field to which the owner of the `List` is linked to.
-    /// For example, if `Post.comments` is associated with `Comment.post` the `List<Comment>`
-    /// of `Post` will have a reference to the `post` field in `Comment`.
-    internal var associatedField: ModelField?
-
-    internal var limit: Int = 100
-
-    /// The current state of lazily loaded list
-    internal var state: LoadState = .pending
+    public var elements: Elements
```
lawmicha, line 25: this has been made public as part of providing developers access to the backing `[]`

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -6,126 +6,136 @@
...
-    public typealias Index = Int
+import Combine
+
+/// `List<ModelType>` provides simple conformance to `Collection` with a backing array of `Model` type elements.
+/// This class acts as an abstract class for plugins to build subclasses that implement their own specific
+/// implementations of a `ModelList`. The decoding logic leverages the `ModelListRegistry` to check for decodability
+/// and decodes to subclasses of this type.
+open class List<ModelType: Model>: ModelList {
+    @available(iOS 13.0, *)
+    public typealias LazyListPublisher = AnyPublisher<Elements, DataStoreError>
```
lawmicha, line 17: no change; this has been migrated over from List+Combine.swift

Amplify/Categories/DataStore/Model/Collection/List+Combine.swift
```diff
@@ -1,32 +0,0 @@
...
-extension List {
-
-    public typealias LazyListPublisher = AnyPublisher<Elements, DataStoreError>
-
-    /// Lazy load the collection and expose the loaded `Elements` as a Combine `Publisher`.
-    /// This is useful for integrating the `List<ModelType>` with existing Combine code
-    /// and/or SwiftUI.
-    ///
-    /// - Returns: a type-erased Combine publisher
-    public func loadAsPublisher() -> LazyListPublisher {
```
lawmicha, line 20: this has been moved to List as an open func for DataStoreList to continue supporting this public API

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/DataStoreListTests.swift
```diff
@@ -27,13 +28,13 @@ class ListTests: BaseDataStoreTests {
         Amplify.DataStore.query(Post.self, byId: postId) {
             switch $0 {
             case .success(let result):
-                if let post = result {
-                    if let comments = post.comments {
+                if let post = result, let postComments = post.comments {
+                    if let comments = postComments as? DataStoreList<Comment> {
                         XCTAssert(comments.state == .pending)
```
lawmicha, line 33: post.comments is a List and `state` is/continues to be internal that was moved from List to DataStoreList, hence the casting before ensuring `state` is accurate

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -6,126 +6,136 @@
...
+        elements.makeIterator()
     }
 
-    // MARK: - Persistent Operations
+    // MARK: - Persistant operations
 
     public var totalCount: Int {
         // TODO handle total count
         return 0
     }
```
lawmicha, line 66: self-review; looks like we should mark this as deprecated as well

lawmicha: actually this should maintain the same implementation for now, until a sub list implementation supports returning totalCount

PR #856, opened 2020-10-26
refactor(datastore): Add modelSchema while getting jsonvalue in JSONValueHolder

royjit:
*Description of changes:*

While fetching json value the model doesnot know about the schema, it just holds a [String: JSONValue] dictionary. So the returned value will not be cast to the right type. Passing `ModelSchema` to `jsonValue()` will allow the implementation to decide the right type before returning.

For example:

```swift
struct DynamicModel: Model, JSONValueHolder {
    public let id: String
    public var values: [String: JSONValue]

    public func jsonValue(for key: String, modelSchema: ModelSchema?) -> Any?? {
        let field = modelSchema?.field(withName: "key")
        
        switch values[key] {
        
        case .some(.number(let deserializedValue)):
            // Uses the field.type from model schema to cast the result to integer.
            if case .int = field?.type {
                return Int(deserializedValue)
            }
            return deserializedValue
            
       ...
        case .none:
            return nil
        }
    }
}
```

- Added public init for `QuerySortDescriptor`

### DIFF
--- a/Amplify/Categories/DataStore/Model/JSONHelper/JSONValueHolder.swift
+++ b/Amplify/Categories/DataStore/Model/JSONHelper/JSONValueHolder.swift
@@ -27,7 +27,7 @@ public protocol JSONValueHolder {
     ///
     /// If a particular key has nil as it value, this method should return .some(nil) as the value.
     func jsonValue(for key: String) -> Any??
-    
+
     /// Return the value for the given key.
     ///
     /// If a particular key has nil as it value, this method should return .some(nil) as the value.
@@ -39,7 +39,7 @@ extension JSONValueHolder {
     func jsonValue(for key: String) -> Any?? {
         return nil
     }
-    
+
     public func jsonValue(for key: String, modelSchema: ModelSchema) -> Any?? {
         return nil
     }
--- a/Amplify/Categories/DataStore/Model/JSONHelper/JSONValueHolder.swift
+++ b/Amplify/Categories/DataStore/Model/JSONHelper/JSONValueHolder.swift
@@ -26,12 +26,21 @@ public protocol JSONValueHolder {
     /// Return the value for the given key.
     ///
     /// If a particular key has nil as it value, this method should return .some(nil) as the value.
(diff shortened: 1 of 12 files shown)


Amplify/Categories/DataStore/Model/JSONHelper/JSONValueHolder.swift
```diff
@@ -27,11 +27,20 @@ public protocol JSONValueHolder {
...
     /// If a particular key has nil as it value, this method should return .some(nil) as the value.
     func jsonValue(for key: String) -> Any??
+
+    /// Return the value for the given key.
+    ///
+    /// If a particular key has nil as it value, this method should return .some(nil) as the value.
+    func jsonValue(for key: String, modelSchema: ModelSchema) -> Any??
 }
 
 extension JSONValueHolder {
```
palpatim, line 37: I'm having second thoughts about this extension. Since JSONValueHolder has a very particular use case (supporting Flutter data store), I'm inclined to make the conforming type implement both methods rather than provide conveniences which have unexpectedly `nil` behavior. Thoughts?

palpatim (approved):
This change looks good. Please see my comment about the extension on JSONValueHolder.

Amplify/Categories/DataStore/Model/JSONHelper/JSONValueHolder.swift
royjit: Agreed and updated PR.

Amplify/Categories/DataStore/Model/JSONHelper/JSONValueHolder.swift
```diff
@@ -26,12 +26,12 @@ public protocol JSONValueHolder {
     /// Return the value for the given key.
     ///
     /// If a particular key has nil as it value, this method should return .some(nil) as the value.
-    func jsonValue(for key: String) -> Any??
+    func jsonValue(for key: String, modelSchema: ModelSchema?) -> Any??
```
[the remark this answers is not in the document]
royjit: Not released

palpatim (approved):

PR #858, opened 2020-10-27
fix(datastore): Issue with graphqlInput not being generated correctly

royjit:
*Description of changes:*
- The value returned by subscript or jsonValue is Any?? and we need to check it properly to get the right value out.
- Enabled AWSPluginsCore test in the CircleCI pipeline


palpatim (approved):
Thanks for catching this!

PR #866, opened 2020-11-03
fix(amplify): change reset method

ruiguoamz:
*Description of changes:*

The existing implementation of `Amplify.reset()` is to call `reset()` on every Categories and put them in global `DispatchGroup` which might cause a race condition. For example, during the `DataStore` reset process, it depends on `Hub` to dispatch some event, but at that point, `Hub` might already been reset and cause a precondition failure on https://github.com/aws-amplify/amplify-ios/blob/main/Amplify/Categories/Hub/HubCategory.swift#L44

The change I made is:
`reset()` on `logging` and `hub` categories should be called when other categories finish resetting.

### DIFF
--- a/Amplify/Core/Configuration/Internal/Amplify+Reset.swift
+++ b/Amplify/Core/Configuration/Internal/Amplify+Reset.swift
@@ -37,7 +37,7 @@ extension Amplify {
             case .predictions:
                 reset(Predictions, in: group) { group.leave() }
             case .hub, .logging:
-                // should be waited until other finish resetting
+                // Hub and Logging should be reset after all other categories
                 break
             }
         }
--- a/Amplify/Core/Configuration/Internal/Amplify+Reset.swift
+++ b/Amplify/Core/Configuration/Internal/Amplify+Reset.swift
@@ -22,7 +22,10 @@ extension Amplify {
 
         let group = DispatchGroup()
 
-        for categoryType in CategoryType.allCases {
+        let manuallyResetCategories = [CategoryType.logging, .hub]
+        let remainingCategories = CategoryType.allCases.filter { !manuallyResetCategories.contains($0) }
+
+        for categoryType in remainingCategories {
             switch categoryType {
             case .analytics:
                 reset(Analytics, in: group) { group.leave() }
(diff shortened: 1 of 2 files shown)


Amplify/Core/Configuration/Internal/Amplify+Reset.swift
```diff
@@ -22,7 +22,12 @@ extension Amplify {
 
         let group = DispatchGroup()
 
-        for categoryType in CategoryType.allCases {
+        let manuallyResetCategories = [CategoryType.logging, .hub]
+        let remainingCategories = CategoryType.allCases.filter {
+            !manuallyResetCategories.contains($0)
+        }
+
+        for categoryType in remainingCategories {
```
diegocstn, line 30: can we maybe use the `partition(by:)` method to get a partitioned array with `.logging` and `.hub` cases at the end of it?

Amplify/Core/Configuration/Internal/Amplify+Reset.swift
```diff
@@ -22,7 +22,12 @@ extension Amplify {
 
         let group = DispatchGroup()
 
-        for categoryType in CategoryType.allCases {
+        let manuallyResetCategories = [CategoryType.logging, .hub]
```
diegocstn, line 25: nit: `let manuallyResetCategories: [CategoryType] = [.logging, .hub]`

diegocstn (commented):
can we maybe use the `partition(by:)` method to get a partitioned array with `.logging` and `.hub` cases at the end?

Amplify/Core/Configuration/Internal/Amplify+Reset.swift
```diff
@@ -32,14 +32,26 @@ extension Amplify {
...
-                reset(Hub, in: group) { group.leave() }
-            case .logging:
-                reset(Logging, in: group) { group.leave() }
             case .storage:
                 reset(Storage, in: group) { group.leave() }
             case .predictions:
                 reset(Predictions, in: group) { group.leave() }
+            case .hub, .logging:
+                // should be waited until other finish resetting
+                break
```
lawmicha, line 41: this class has a lot of hard coding logging and hub categories, maybe you can isolate these two categories that you can reuse in the rest of the code, like
```
let resetLastCategories = [CategoryType.hub, .logging]
```
then you can filter before line 25 by `let resetFirstCategories = CategoryType.allCases.filter { !resetLastCategories.contains($0) }`

With this, would you be able to avoid empty switch cases, similar to code like this? 
```
for categoryType in resetFirstCategories {
    reset(categoryType.category, in: group) { group.leave() }
}

group.wait()

for categoryType in resetLastCategories {
    reset(categoryType.category, in: group) { group.leave() }
}
group.wait()


```

lawmicha (approved):

Amplify/Core/Configuration/Internal/Amplify+Reset.swift
ruiguoamz: Synced offline. You can take a look again.

diegocstn (approved):

palpatim (approved):

PR #868, opened 2020-11-04
chore: Use NSLock instead of DispatchQueue to protect Atomics

palpatim:
*Description of changes:*

My initial implementation of AtomicValue was naively based on a `DispatchQueue` to protect access to the internal value and protect against data races. However, that could result in thread explosions (since not specifying a `target` for the private queue defaults to the global "overcommit" queue), and will require needless context switches. This change refactors `AtomicValue` and `AtomicDictionary` to use an `NSLock` instead of a `DispatchQueue` to protect against data races. As evidenced by the new performance test, this increases performance by about 50% over the previous implementation. It's not likely to make much of a change in our customer apps, since we don't typically have super-high volume use cases for our Atomics, but reducing thread usage is a win regardless.

### DIFF
--- a/.gitallowed
+++ b/.gitallowed
@@ -2,3 +2,4 @@ case slave = "den"
 amazon-textract-code-samples/master/src-csharp
 aws-amplify/amplify-ci-support', branch: 'master'
 branch: master
+build-support/codecov.sh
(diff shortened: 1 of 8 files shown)


royjit (approved):
LGTM

diegocstn (approved):

diegocstn (approved):

PR #885, opened 2020-11-11
fix(amplify): DataStore query fix column missing issue for @connection hasMany schema

ruiguoamz:
*Description of changes:*
1. The PR is to fix this issue: https://github.com/aws-amplify/amplify-ios/issues/512
2. This PR better comes together with a CLI PR by Diego@: https://github.com/aws-amplify/amplify-cli/pull/5857 which is pending review.

**Existing issue**
Take this schema as example:
```graphql
type User @model {
  id: ID!
  fullName: String!
  posts: [Post] @connection(name: "UserPosts")
  createdTimestamp: Int!
}

type Post @model {
  id: ID!
  user: User! @connection(name: "UserPosts")
  createdTimestamp: Int!
}
```
Failure occurs when trying to call
```swift
Amplify.DataStore.query(Post.self, where: Post.keys.user = "1234") { _ in }
```
`Post.keys.user` is picking up the `stringValue` of `case user`, so that the generated SQL statement is 
```
* \"root\".\"user\" = ?" *
```
But Post.keys.user has a `belongs to` association with `targetName: "postUserId"`, the correct SQL statement should be
```
* \"root\".\"postUserId\" = ?" *
```

**Change**
CodingKey should check if there is type `.belongsTo` association, if yes, return targetName, otherwise return `stringValue`
```swift
var columnName: String {
    guard let modelSchema: ModelSchema = ModelRegistry.modelSchema(from: modelName) else {
        log.warn("Please upgrade to the latest version of Amplify CLI and rerun `amplify codegen models`")
        return stringValue
    }

    switch modelType.schema.field(withName: stringValue)?.association {
    case .belongsTo(_, let targetName):
        return targetName ?? stringValue
    default:
        return stringValue
    }
}
```

And also the CodeGen Model should contains the computed property that returns `modelName` so that `modelSchema` can be retrieved from `ModelRegistry`. Here is the PR by Diego@: https://github.com/aws-amplify/amplify-cli/pull/5857


Amplify/Categories/DataStore/Query/ModelKey.swift
```diff
@@ -32,24 +32,35 @@ import Foundation
 ///     post.content != nil
 /// })
 /// ```
-public protocol ModelKey: CodingKey, CaseIterable, QueryFieldOperation {}
+public protocol ModelKey: CodingKey, CaseIterable, QueryFieldOperation {
```
diegocstn, line 35: discussed offline about returning the model name (and use the `ModelRegistry` to resolve the actual model type and its schema) and provide a default implementation to avoid breaking existing models and decouple this PR from the needed CLI changes

AmplifyPlugins/DataStore/Podfile.lock
```diff
@@ -122,4 +122,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 04860e414d616b67d24ed3346a60700f427764b9
 
-COCOAPODS: 1.9.3
+COCOAPODS: 1.10.0
```
diegocstn, line 125: nit: does this change belong to this PR?

Podfile.lock
```diff
@@ -63,4 +63,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: a2acde35ab1bc6b80b362106cfda70905a6a840f
 
-COCOAPODS: 1.9.3
+COCOAPODS: 1.10.0
```
diegocstn, line 66: ditto :)

diegocstn (approved):

AmplifyPlugins/DataStore/Podfile.lock
ruiguoamz: Actually, I am curious what impact does this change have?

Amplify/Categories/DataStore/Query/ModelKey.swift
```diff
@@ -32,24 +32,42 @@ import Foundation
...
+    var modelName: String { get }
+}
+
+extension ModelKey {
+    public var modelName: String { "" }
```
drochetti, line 40: document the behavior. Why is it empty by default? (i.e. explain the backward compatibility)

Amplify/Categories/DataStore/Query/ModelKey.swift
```diff
@@ -32,24 +32,42 @@ import Foundation
...
+}
 
 extension CodingKey where Self: ModelKey {
 
+    var columnName: String {
+        guard let modelSchema: ModelSchema = ModelRegistry.modelSchema(from: modelName) else {
+            return stringValue
```
drochetti, line 47: warn level log to hint customers that they should upgrade their CLI and regenerate their code

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/Decorator/AuthRuleDecorator/ModelWithOwnerAuthAndGroupWithGroupClaim.swift
```diff
@@ -13,63 +13,63 @@ import XCTest
...
+ @auth(
+ rules: [
+ { allow: owner, provider: oidc, identityClaim: "sub"},
+ { allow: groups, provider: oidc, groups: ["Admins"],
+ groupClaim: "https://myapp.com/claims/groups"}
+ ]
+ ) {
+ id: ID!
+ title: String!
+ owner: String
```
drochetti, line 26: nit-picking: leave it formatted (I usually use Prettier playground for GraphQL formatting: https://prettier.io/playground/).

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLDocument/GraphQLListQueryTests.swift
```diff
@@ -115,6 +115,91 @@ class GraphQLListQueryTests: XCTestCase {
...
+    ///   - the query is of type `.list`
+    /// - Then:
+    ///   - check if the generated GraphQL document is valid query:
+    ///     - it contains an `filter` argument of type `ModelCommentFilterInput`
+    ///     - it is named `listComments`
+    ///     - it has a list of fields, one of which is post that also has a list of fields with no nested models
+    ///     - fields are wrapped with `items`
+    ///   - check if generated GraphQL variables is valid:
+    ///     - it contains the correct columnName `commentPostId`
+    func testListGraphQLQueryFromSimpleModelGivenQueryPredicateOnConnectedField() {
```
ruiguoamz, line 131: A new test is added here to test GraphQLList on connected field

Amplify/Categories/DataStore/Query/ModelKey.swift
ruiguoamz: I don't know if there is better way of putting a log than printing here since class `CodingKey` does not support extension on `DefaultLogger`.

diegocstn: would conforming `ModelKey` to `DefaultLogger` help?

ruiguoamz: Also tried, since `ModelKey` is a protocol, it is not working.

Error message: 
```
Extension of protocol 'ModelKey' cannot have an inheritance clause
```

drochetti (approved):
thanks for fixing this! remember not to merge into main until we have more clarify on the CL release with the codegen changes.

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLDocument/GraphQLListQueryTests.swift
```diff
@@ -115,6 +115,91 @@ class GraphQLListQueryTests: XCTestCase {
...
+        let comment = Comment.keys
+        let predicate = comment.id.eq("id")
+            && comment.post.eq("commentPostId")
```
lawmicha, line 134: && comment.post.eq("postId123")

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLDocument/GraphQLListQueryTests.swift
```diff
@@ -115,6 +115,91 @@ class GraphQLListQueryTests: XCTestCase {
...
+          "and" : [
+            {
+              "id" : {
+                "eq" : "id"
+              }
+            },
+            {
+              "commentPostId" : {
+                "eq" : "commentPostId"
+              }
```
lawmicha, line 195: so as i'm looking into these connection directives more, it appears this test does not generate a valid GraphQL input. If you check the provisioned API for the schema that generated these Models, `commentPostId` is not a valid field on the API. So what we know
- We can perform a List query on Comment
- We can pass in a predicate which translate to the "filter" variables
- The filter input available based on the Comment fields's are generated as: `content`, `createdAt`, `id`
- The remaining filter input variables are `and`, `not` and `or`. These are to create groups of filters
- There is no `commentPostId` filter variable

So I would say that it is impossible from the API to list the comments and use a predicate that filters on the connected field. The alternative is to explicitly defining a field on Comment that references the Post's identifier like
```
type Post4 @model {
  id: ID!
  title: String!
  comments: [Comment4] @connection(keyName: "byPost4", fields: ["id"])
}

type Comment4 @model
  @key(name: "byPost4", fields: ["postID", "content"]) {
  id: ID!
  postID: ID! // explicit "postID" on Comment
  content: String!
  post: Post4 @connection(fields: ["postID"])
}
```

With this alternative, developers using API can do the same with the predicate `comment.postID == post.id`. 

This is good intention to add validatin on the API side, but because I stumbled across that we generating an invalid GraphQL input here, then we should remove this unit test because it's validating an expected JSON that is incorrect. Overall, this column name fix is still needed to fix the interaction with DataStore's SQL layer. So i would update the PR title/description to be clear that this is a DataStore query fix

Amplify/Categories/DataStore/Query/ModelKey.swift
```diff
@@ -32,24 +32,43 @@ import Foundation
...
+    var columnName: String {
+        guard let modelSchema: ModelSchema = ModelRegistry.modelSchema(from: modelName) else {
+            log.warn("Please upgrade to the latest version of Amplify CLI and rerun `amplify codegen models`")
+            return stringValue
+        }
+        switch modelSchema.field(withName: stringValue)?.association {
+        case .belongsTo(_, let targetName):
+            return targetName ?? stringValue
+        default:
+            return stringValue
```
lawmicha, line 54: should this also have case for hasOne relationships which have a targetName?

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLDocument/GraphQLListQueryTests.swift
lawmicha: With `Post4` and `Comment4`, these would be good models for API testing since
1. the Model that is generated removes the `postID` from the Comment4 Model.
2. List comment with predicate list.keys.post.eq("postId123") can use the targetName to translate it to
```
{
  "postID": "postId123"
}
```
and this creates a successful GraphQL request

Amplify/Categories/DataStore/Query/ModelKey.swift
lawmicha: we can take the hasOne scenario on as a separate [issue](https://github.com/aws-amplify/amplify-ios/issues/920)

Amplify/Categories/DataStore/Query/ModelKey.swift
```diff
@@ -32,24 +32,43 @@ import Foundation
...
+}
+
+extension ModelKey {
+    public var modelName: String { "" }
+}
 
 extension CodingKey where Self: ModelKey {
 
+    var columnName: String {
+        guard let modelSchema: ModelSchema = ModelRegistry.modelSchema(from: modelName) else {
```
lawmicha, line 46: for customers that don't upgrade and re-codegen, how does modelName get resolved?

lawmicha: oh i see above `public var modelName: String { "" }`

AmplifyTests/CategoryTests/DataStore/CodingKeysTests.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+
+import XCTest
+import Amplify
+@testable import AmplifyTestCommon
+
+class CodingKeysTests: XCTestCase {
+
+    func testSchemaHasCorrectColumnName() throws {
+        ModelRegistry.register(modelType: Comment.self)
```
wooj2, line 15: I believe that these tests will have side effects on other tests running in the suite.

After we register these models, how do we reset `ModelRegistry`? -- seems like we need to add a tearDown() function with calls `ModelRegistry.reset()`?

wooj2: To fix, we should add something like:
```
override func tearDown() {
        guard let api = Amplify.API as? Resettable else {
            XCTFail("API Needs to reset between tests")
            return
        }
        api.reset(onComplete: { })
        XCTAssert(ModelRegistry.models.isEmpty)
    }
```
It's not ideal that we have to use Amplify.API, but ModelRegistry is internal (and rightfully so).  In any case, checking to make sure models.isEmpty before proceeding is a good way to double check that a side effect of calling Amplify.API is actually reseting the ModelRegistry.

AmplifyTests/CategoryTests/DataStore/CodingKeysTests.swift
```diff
@@ -0,0 +1,26 @@
...
+        let commentQPO: QueryPredicateOperation = Comment.keys.id == "1234"
+        XCTAssertEqual(commentQPO.field, "id")
+    }
+
+    func testSchemaWithBelongsToHasCorrectColumnName() throws {
+        ModelRegistry.register(modelType: Comment.self)
+        let commentQPO: QueryPredicateOperation = Comment.keys.post == "5678"
+        XCTAssertEqual(commentQPO.field, "commentPostId")
+    }
+
```
wooj2, line 25: To add test coverage we, should probably add a test for the `MutationEvent`
```
    func testInternalModelType() throws {
        //arrange not required
        
        let mutationEvent = MutationEvent.keys.inProcess == true

        XCTAssertEqual(mutationEvent.field, "inProcess")
    }
```

Amplify/Categories/DataStore/Query/ModelKey.swift
```diff
@@ -32,24 +32,43 @@ import Foundation
...
+            log.warn("Please upgrade to the latest version of Amplify CLI and rerun `amplify codegen models`")
+            return stringValue
+        }
+        switch modelSchema.field(withName: stringValue)?.association {
+        case .belongsTo(_, let targetName):
+            return targetName ?? stringValue
```
wooj2, line 52: We seem to have a unit tests which tests when targetName is not nil, but we should add a test where targetName is nil, so that it just returns the regular `stringValue`.

lawmicha:
I believe this is reverted, see https://github.com/aws-amplify/amplify-ios/pull/965 for more details

PR #889, opened 2020-11-13
fix(DataStore): failed subscriptions lead to instability in sync engine

wooj2:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
@@ -15,6 +15,7 @@ class NetworkReachabilityNotifier {
     private var reachability: NetworkReachabilityProviding?
     private var allowsCellularAccess = true
 
+    // This is a CurrentValueSubject.  Please do not change this unless you talk to wooj2@ or palpatim@ before changing this.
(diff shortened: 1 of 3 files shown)


palpatim (approved):

ruiguoamz (approved):

AmplifyPlugins/API/AWSAPICategoryPlugin/Reachability/NetworkReachabilityNotifier.swift
```diff
@@ -15,6 +15,7 @@ class NetworkReachabilityNotifier {
     private var reachability: NetworkReachabilityProviding?
     private var allowsCellularAccess = true
 
+    // This is a CurrentValueSubject.  Please do not change this unless you talk to wooj2@ or palpatim@ before changing this.
```
royjit, line 18: Can we add a Readme doc for this and its usage as a TODO?

palpatim: This is intended to be a short-term fix to solve the immediate pain. We'll follow up longer term with a more general fix--we just wanted to add this as a warning in the interim period :)

wooj2: This is a workaround to address unblock customers experiencing this bug.  I don't think it requires an update on the readme.

royjit (approved):

PR #896, opened 2020-11-17
fix(Predictions): Callback is not triggered with URLError

ruiguoamz:
*Description of changes:*
PR is a fix for https://github.com/aws-amplify/amplify-ios/issues/678

1. When using schema with unsupported region, which means the host generated cannot be reached and causes a url failure.
2. When the device is not connected to Internet, URLError is also going to be returned

In the above cases, the Callback with error info is not triggered because it is skipping the error check.

The fix is:
1.  Check what type of error return back and avoid calling the next `listen()` if is of type "NSURLErrorDomain".
2. Trigger callback in failure case.


AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/Predictions/AWSPredictionsService+Transcribe.swift
```diff
@@ -31,6 +31,17 @@ extension AWSPredictionsService: AWSTranscribeStreamingServiceBehavior {
         request.mediaSampleRateHertz = 8_000
 
         transcribeClientDelegate.connectionStatusCallback = { status, error in
+            guard status != .closed && error == nil else {
```
ruiguoamz, line 34: Before, the error returned on line 33 is never used, we should not skip error checking.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Internal/NativeWebSocketProvider.swift
```diff
@@ -72,10 +72,15 @@ class NativeWebSocketProvider: NSObject, AWSTranscribeStreamingWebSocketProvider
             switch result {
             case .failure(let error):
                 let status = AWSTranscribeStreamingClientConnectionStatus.closed
+                let nsError = error as NSError?
 
                 self.callbackQueue.async {
                     self.clientDelegate.connectionStatusDidChange(status, withError: error)
                 }
+
+                guard nsError?.domain != NSURLErrorDomain else {
```
ruiguoamz, line 81: If the error is of type NSURLErrorDomain, we do not want to keep calling `listen()`, we should just return

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/Predictions/AWSPredictionsService+Transcribe.swift
lawmicha:
```
if status == .closed && error != nil {
  
} else is status == .connected {

}
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/Predictions/AWSPredictionsService+Transcribe.swift
```diff
@@ -31,6 +31,17 @@ extension AWSPredictionsService: AWSTranscribeStreamingServiceBehavior {
...
+                let error = error as NSError?
+                guard error?.domain != "NSURLErrorDomain" else {
+                    let predictionsErrorString = PredictionsErrorHelper.mapPredictionsServiceError(error!)
```
lawmicha, line 37:
```

if urlError = error as? URLError, let urlErrorTuple = PredictionsErrorHelper.parseURLError(error) {
   onEvent(.failed(...
}


func parseURLError(_ error: URLError) -> (String, String)
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Internal/NativeWebSocketProvider.swift
lawmicha, line 81:
```
self.callbackQueue.async {
                    self.clientDelegate.connectionStatusDidChange(status, withError: error)
                }

if urlError = error as? URLError { 
 // maybe disconnect and clean up?
}
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/Predictions/AWSPredictionsService+Transcribe.swift
lawmicha: `PredictionsErrorHelper.mapPredictionsServiceError` can be something more general like `PredictionsErrorHelper.mapError` and does the chain of error handling

palpatim: Even better, add a condition to `PredictionsError.init` that checks whether the incoming error is a `URLError` and take action appropriately. If the use case is such that supplying a description and recovery suggestion isn't appropriate, then consider a type-specific initializer rather than a "Helper" class with a specialized factory method.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Internal/NativeWebSocketProvider.swift
```diff
@@ -72,10 +72,15 @@ class NativeWebSocketProvider: NSObject, AWSTranscribeStreamingWebSocketProvider
             switch result {
             case .failure(let error):
                 let status = AWSTranscribeStreamingClientConnectionStatus.closed
+                let nsError = error as NSError?
```
palpatim, line 75: Error isn't `nil`, so you should be able to do `let nsError = error as NSError`, and do away with the optional chaining. My recollection is that `Swift.Error` always successfully casts to `NSError`

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Internal/NativeWebSocketProvider.swift
palpatim: Let's not use -1003 as a literal, though. Use the appropriate error code

palpatim (commented):
Comments below. Do you have unit tests to cover this case?

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Internal/NativeWebSocketProvider.swift
ruiguoamz: Removing `?` produces 
```
'Error?' is not convertible to 'NSError'; did you mean to use 'as!' to force downcast?
```
and force cast is also not allowed.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/Predictions/AWSPredictionsService+Transcribe.swift
```diff
@@ -31,7 +31,16 @@ extension AWSPredictionsService: AWSTranscribeStreamingServiceBehavior {
         request.mediaSampleRateHertz = 8_000
 
         transcribeClientDelegate.connectionStatusCallback = { status, error in
-            if status == .connected {
+            if status == .closed && error != nil {
+                let nsError = error as NSError?
+                if nsError?.domain == NSURLErrorDomain {
```
lawmicha, line 36: by working off of URLError (https://developer.apple.com/documentation/foundation/urlerror) you get an object with additional information

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Internal/NativeWebSocketProvider.swift
lawmicha: if you can't, i would go with
```
if let error = error as NSError? {
   
}
```

AmplifyPlugins/Predictions/AWSPredictionsPluginTests/Service/PredictionsTest/PredictionsServiceTranscribeTests.swift
```diff
@@ -112,6 +112,31 @@ class PredictionsServiceTranscribeTests: XCTestCase {
...
+    /// - Given: Predictions service with transcribe behavior
+    /// - When:
+    ///    - I invoke an invalid request wich Unreachable host
+    /// - Then:
+    ///    - I should get back a connection error
+    ///
+    func testTranscribeServiceWithConnectionError() {
+        let mockError = NSError(domain: NSURLErrorDomain,
+                                code: URLError.cannotFindHost.rawValue,
+                                userInfo: [:])
```
lawmicha, line 126: can't you just create a URLError?

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/Predictions/AWSPredictionsService+Transcribe.swift
lawmicha:
```swift
if status == .closed  {
   guard error != nil else {
     return
   }
  guard errorIsNotNormalClosedError(error) else {
     return
  }
   let predictionsError = PredictionsHelper.mapError(nsError)
   onEvent(.failed(predictionsError)
  
} else if status == .connected {

}

```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/PredictionsErrorHelper.swift
```diff
@@ -97,6 +98,28 @@ class PredictionsErrorHelper {
...
+        let defaultError = PredictionsErrorHelper.getDefaultError(error)
+
+        let error = error as? URLError
+        guard let urlError = error else {
+            return defaultError
+        }
+
+        switch urlError.code {
+        case .cannotFindHost:
+            let errorDescription = "The host you are trying to reach cannot be found"
```
lawmicha, line 111:
```suggestion
            let errorDescription = "The host name for a URL couldn’t be resolved."
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/PredictionsErrorHelper.swift
```diff
@@ -97,6 +98,28 @@ class PredictionsErrorHelper {
...
+            let recoverySuggestion = "Please check if you are reaching the correct host"
+            return PredictionsError.network(errorDescription, recoverySuggestion, error)
+        case .notConnectedToInternet:
+            let errorDescription = "Your device is not connected to network"
```
lawmicha, line 115:
```suggestion
            let errorDescription = "A network resource was requested, but an internet connection hasn’t been established and can’t be established automatically."
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/PredictionsErrorHelper.swift
```diff
@@ -97,6 +98,28 @@ class PredictionsErrorHelper {
...
+            let recoverySuggestion = "Please check your network connectivity status"
```
lawmicha, line 116:
```suggestion
            let recoverySuggestion = "Please check your network connectivity status."
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/PredictionsErrorHelper.swift
```diff
@@ -97,6 +98,28 @@ class PredictionsErrorHelper {
...
+
+        let error = error as? URLError
+        guard let urlError = error else {
+            return defaultError
+        }
+
+        switch urlError.code {
+        case .cannotFindHost:
+            let errorDescription = "The host you are trying to reach cannot be found"
+            let recoverySuggestion = "Please check if you are reaching the correct host"
```
lawmicha, line 112:
```suggestion
            let recoverySuggestion = "Please check if you are reaching the correct host."
```

AmplifyPlugins/Predictions/AWSPredictionsPluginTests/Service/PredictionsTest/PredictionsServiceTranscribeTests.swift
```diff
@@ -126,6 +127,60 @@ class PredictionsServiceTranscribeTests: XCTestCase {
...
+    ///
+    /// - Given: Predictions service with transcribe behavior
+    /// - When:
+    ///    - I invoke an invalid request with not connected to Internet
+    /// - Then:
+    ///    - I should get back a connection error
+    ///
+    func testTranscribeServiceWithNotConnectedToInternetError() {
+        let urlError = URLError(.notConnectedToInternet)
+        mockTranscribe.setConnectionResult(result: AWSTranscribeStreamingClientConnectionStatus.closed, error: urlError)
```
lawmicha, line 167: test for .connected without any errors

AmplifyPlugins/Predictions/AWSPredictionsPluginTests/Service/PredictionsTest/PredictionsServiceTranscribeTests.swift
```diff
@@ -126,6 +127,89 @@ class PredictionsServiceTranscribeTests: XCTestCase {
...
 
+    /// Test whether error is correctly propogated
+    ///
+    /// - Given: Predictions service with transcribe behavior
+    /// - When:
+    ///    - I invoke an invalid request
+    /// - Then:
+    ///    - I should get back a service error
+    ///
+    func testTranscribeServiceWith() {
```
lawmicha, line 138: withError?

ruiguoamz: sorry, this one was a mistake.

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/Predictions/AWSPredictionsService+Transcribe.swift
```diff
@@ -31,7 +31,17 @@ extension AWSPredictionsService: AWSTranscribeStreamingServiceBehavior {
...
 
         transcribeClientDelegate.connectionStatusCallback = { status, error in
-            if status == .connected {
+            if status == .closed && error != nil {
+                guard error != nil else {
+                    return
+                }
+                let nsError = error as NSError?
+                guard nsError?.domain != NSURLErrorDomain else {
+                    let predictionsError = PredictionsErrorHelper.mapError(nsError!)
```
lawmicha, line 40: sorry for the iterations on this, but this still doesn't make sense to me because it's an 
1. explicit check on URLError
2. If it is any type of URLError, always generate a PredictionsError and send that back to the caller. from traversing the logic, it will send Predictions.network when URLError is related to connectivity, otherwise send Predictions.unknown

Now there are two places to generate a PredicationsError, in `mapError(_ error: NSError)` and `mapPredictionsServiceError(_ error: NSError)`, how do we as plugin developers know which one to call use?

From the tests, I see that the intent is to only see an PredictionsError on the failed event when the URLError is `notConnectedToInternet` and `cannotFindHost` which means you are missing a test for "Failed event when error is URLError but is not one of the following: notConnectedToInternet or cannotFindHost"

If you want to return an unknown error anytime it's an URLError that is not related to connectivity then you should add a test in there as well.

But i _think_ you are only care to notify the caller on network related URLErrors, then the code should be

```swift
if let urlError = error as? URLError?, (case .notConnected = urlError.code || case .cannotfindHost  == urlErorr.code) {
     // generate PredictionsError generically
    let predictionsError = PredictionsErrorHelper.mapServiceError(error)
   onEvent(.failed(predictionsError)
  }

}
```

lawmicha:
```swift
let predictionsError = PredictionsErorrHelper.mapServiceError(error)
if case .network = predictionsError {
   onEvent(.failed(predictionsError)
}
}
```

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/PredictionsErrorHelper.swift
```diff
@@ -97,6 +98,29 @@ class PredictionsErrorHelper {
         }
     }
 
+    static func mapError(_ error: NSError) -> PredictionsError {
```
lawmicha, line 101:
```swift
case NSURLErorrDomain:
   guard let urlError = error as? URLError else {
       return defaultError 
   }
    return .network(description, recovery, urlError)

```

AmplifyPlugins/Predictions/AWSPredictionsPluginTests/Service/PredictionsTest/PredictionsServiceTranscribeTests.swift
```diff
@@ -126,6 +130,60 @@ class PredictionsServiceTranscribeTests: XCTestCase {
...
+        mockTranscribe.setConnectionResult(result: AWSTranscribeStreamingClientConnectionStatus.closed, error: urlError)
+
+        let errorReceived = expectation(description: "Error should be returned")
+
+        predictionsService.transcribe(speechToText: audioFile, language: .usEnglish) { event in
+            switch event {
+            case .completed(let result):
+                XCTFail("Should not produce result: \(result)")
+            case .failed(let error):
+                XCTAssertNotNil(error, "Should produce an error")
```
lawmicha, line 152: assert it is a network error

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/PredictionsErrorHelper.swift
```diff
@@ -92,11 +93,38 @@ class PredictionsErrorHelper {
                 return defaultError
             }
             return AWSTranscribeStreamingErrorMessage.map(errorType) ?? defaultError
+        case NSURLErrorDomain:
+            return mapError(error)
```
lawmicha, line 97: can cast to URLError now that you know it is NSURLErrorDomain, then call a specific method that takes in `URLError` as a parameter

AmplifyPlugins/Predictions/AWSPredictionsPlugin/Support/Utils/PredictionsErrorHelper.swift
```diff
@@ -92,11 +93,38 @@ class PredictionsErrorHelper {
...
         default:
             return defaultError
         }
     }
 
+    static func mapError(_ nsError: NSError) -> PredictionsError {
```
lawmicha, line 103: this is more like `mapURLError(_ urlError: URLError)`

lawmicha (approved):

Issue #850, opened 2020-10-23
iOS 14 : dev menu shows "Information not available"


sebsto:
**Describe the bug**
I added a DevMenu as per Amplify iOS 1.4.0

On iOS Simulator :
When triggering the DevMenu, the 'Amplify Plugins Information' an 'Developer Environment Information' says 
"version not available" or "information not available"
Also, Log viewers reports "Log Unavailable"

**To Reproduce**
Implement Dev Menu as this 
https://github.com/sebsto/amplify-ios-getting-started/commit/70a42197ebc2975cfb7d6761d9261d759e7f570c

Trigger Dev Menu with a long touch gesture 

Navigate to "environment information"
Navigate to "log viewer"

**Expected behavior**
DevMenu should report valuable data for debugging

**Screenshots**
![Simulator Screen Shot - iPhone 11 - 2020-10-23 at 07 41 03](https://user-images.githubusercontent.com/401798/96960624-208b0900-1503-11eb-817f-0daafcd15b27.png)
![Simulator Screen Shot - iPhone 11 - 2020-10-23 at 07 41 21](https://user-images.githubusercontent.com/401798/96960638-284aad80-1503-11eb-92d6-8c802de67f64.png)

**Environment(please complete the following information):**
 - Amplify Framework Version: 1.4.0
 - Dependency Manager: cocoapods
 - Swift Version : Xcode 12.1 / Swift 5.?

**Device Information (please complete the following information):**
 - Device: simulator
 - iOS Version: 14.1

**Additional context**

sebsto:
Just tested on device - same behaviour

drochetti:
hey @sebsto, thanks for testing this as soon as it came out 🙂 

We [published the docs this morning](https://docs.amplify.aws/lib/debugging/dev-menu/q/platform/ios). I'm working on an update to reflect the new SwiftUI App life cycle, since this was created before that came out.

---

The lack of version is something that I'm fixing right now. As for the logs, you should be able to see it. Can you clarify if you enabled the menu before calling `Amplify.configure()`?

sebsto:
Thank you for the follow up. I saw the doc PR being merged yesterday.
Here are the steps I took (aligned with what I read in the doc) 

1. Add `local-env-info.json` to the project 
2. Add ènableDevMenu`as this : 
```swift 
           try Amplify.add(plugin: AWSCognitoAuthPlugin())
[...]

PR #897, opened 2020-11-18
fix: Use correct category plugins to populate dev menu

palpatim:
*Issue #, if available:*
Refs: #850

*Description of changes:*

Several changes to correctly populate the dev menu:
- Fix & refactor the PluginInfoHelper to simplify the call pattern & remove redundant calls that were easy to accidentally copy/paste incorrectly
- Add `AmplifyVersionable` conformances to all AWS plugins
- Make `AmplifyVersionable` public so it could be consumed by plugins
- Made all members of `EnvironmentInfo` optional so missing fields wouldn't cause the entire thing to not display
- Updated all pod dependencies

See screenshot below for populated menu. Note that I had to manually update the `amplify/.config/local-env-info.json` file, since the CLI is not yet populating those items (at least on my system).

image


Amplify/DevMenu/Data/PluginInfoHelper.swift
```diff
@@ -14,86 +14,70 @@ struct PluginInfoHelper {
...
+                information: versionable.version
+            )
+        } else {
+            pluginInfoItem = PluginInfoItem(
+                displayName: pluginKey,
+                information: DevMenuStringConstants.versionNotAvailable
+            )
+        }
+        return pluginInfoItem
+    }
```
royjit, line 81: guard statement?
```suggestion
   
      guard let versionable = versionable else {
           return PluginInfoItem(
                displayName: pluginKey,
                information: DevMenuStringConstants.versionNotAvailable
            )
        } 
        return PluginInfoItem(displayName: pluginKey, information: versionable.version)
    }
```

Amplify/DefaultPlugins/AWSUnifiedLoggingPlugin/AWSUnifiedLoggingPlugin.swift
```diff
@@ -99,3 +99,11 @@ extension AWSUnifiedLoggingPlugin {
         return wrapper
     }
 }
+
+extension AWSUnifiedLoggingPlugin: AmplifyVersionable {
+    public var version: String {
+        let bundle = Bundle(for: type(of: self))
+        let version = bundle.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String
+        return version ?? "Not Available"
```
royjit, line 107: Can we give default behavior for this protocol, instead of repeating the same code?

Amplify/DevMenu/Data/PluginInfoHelper.swift
```diff
@@ -14,86 +14,70 @@ struct PluginInfoHelper {
     static func getPluginInformation() -> [PluginInfoItem] {
         var pluginList = [PluginInfoItem]()
 
-        // Add Analytics plugins information
```
royjit, line 17: Does this work?
```swift
let plugins = Amplify.Analytics.plugins 
                     + Amplify.API.plugins 
                     + Amplify.Auth.plugins
                     + Amplify.DataStore.plugins ...
pluginList.append(contentsOf: plugins.map {
                 makePluginInfoItem(for: $0.key, versionable: $0.value as? AmplifyVersionable)
             })
```

Amplify/DevMenu/DevEnvironmentInfo.swift
```diff
@@ -9,12 +9,12 @@ import Foundation
...
-    let amplifyCLIVersion: String
-    let podVersion: String
-    let xcodeVersion: String
-    let osVersion: String
+    let nodejsVersion: String?
+    let npmVersion: String?
+    let amplifyCLIVersion: String?
+    let podVersion: String?
+    let xcodeVersion: String?
+    let osVersion: String?
```
royjit, line 17: This will show `Not Available` if the value is nil?

Amplify/DevMenu/Data/PluginInfoHelper.swift
palpatim: Hmm. In this case, I'm less instinctively inclined to apply a `guard` because it's not actually an early exit. But thinking about it more, I like how it makes the "I need this info to do the happy path" intent easier to follow, so yeah! Guard! :)

Amplify/DevMenu/DevEnvironmentInfo.swift
palpatim: Yep, that is the way the Dev Menu view is currently written.

Amplify/DefaultPlugins/AWSUnifiedLoggingPlugin/AWSUnifiedLoggingPlugin.swift
palpatim: When I originally thought about that I was stymied by `type(of:)` expecting a class, and `AmplifyVersionable` not being constrained to classes. But now that I'm thinking about it, I could write a conditional extension. Derp. Will fix.

Amplify/DevMenu/Data/PluginInfoHelper.swift
palpatim: No, because `plugins` is an `internal` implementation detail of the category. We're diving into it here because we happen to be in the same package, but you'll note special case handling for `API` and `Logging` categories. If we wanted to abstract it out, we could write a `allPlugins` property that would normalize that, but it's not something we have a public facing use case for just now, so the logic would be manually attached to each category. Since we're only doing it for this dev menu, I'm disinclined to add API surface over on each category separated from the call site. (The very reasonable counterargument of course is that if a future internal use case needs something like an `allPlugins` property, it will be blind luck for someone to remember "Oh yeah, the Dev Menu is doing special case handling for this, I should change it." But I'm going to solve for the current problem rather than adding complexity for a hypothetical future case.)

Amplify/DevMenu/Data/PluginInfoHelper.swift
palpatim: Thinking about it even *more* though, it's even cleaner to just worry about the version:

```swift
let version = versionable?.version ?? DevMenuStringConstants.versionNotAvailable
return PluginInfoItem(displayName: pluginKey, information: version)
```

drochetti (approved):
lgtm, I mostly +1 some Jithin's comments and your own comments, so I'm leaving my preemptive "ship it!"

royjit (approved):
LGTM

PR #904, opened 2020-11-20
fix(predictions): use string's unicode scalars view to compute indexes for Comprehend results

### LINKED ISSUE #873: AWS Predictions (AWSPredictionsService) Index out of Range due to Emojis
**Describe the bug**

AWS Predictions (text comprehend) crashes when passing in a string of text that contains specific emojis. I am pretty sure this is because some of these emojis count for 1 character but when getting the endIndex swift will return 1 more than the count.

The particular emoji I am referring to is one:👇🏽. This increases the string character count `theString.count` by 1 but when taking `theString.endIndex`, it will return 2. I am sure there are many other emojis with the same issue as well. It may have to do with the different colors that are selectable with the emojis, but I am uncertain.

The crash happens around line 140 in the file "AWSPredictionsService+Comprehend.swift" when we try to define the endIndex `let endIndex = text.index(text.startIndex, offsetBy: endOffset)`.

I have implemented a temporary guard bandaid to skip the syntax if the endIndex is out of range, but whenever there is time to fix this, the issue is there. I have a feeling that putting the emoji in the middle of a longer string may also mess up the indexes of the words that are interpreted and return inaccurate strings of the key phrases of the string, but I have not tested this.

**To Reproduce**
1. Create a string with 👇🏽 at the end of the string.
2. Pass into Amplify.Predictions.interpret()
3. Crash occurs.

**Expected behavior**
No crash and accurate interpretation of string returned.

**Screenshots**
Unnecessary.

Thank you for the great platform otherwise!

diegocstn:
Amazon Comprehend returns an array of Entity https://docs.aws.amazon.com/comprehend/latest/dg/API_Entity.html .
Each entity has a _beginOffset_ and _endOffset_ properties, which are positions of each UTF-8 "code point" in the string.

A [code point](https://www.unicode.org/glossary/#code_point) is a value in the Unicode codespace, the range of integers from 0 to 10FFFF, not all code points are assigned to encoded chars.
We need to use Swift strings underlying scalar view to compute indexes correctly and avoid crashes at runtime.

fix #873 

### DIFF
--- a/AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/Predictions/AWSPredictionsService+Comprehend.swift
+++ b/AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/Predictions/AWSPredictionsService+Comprehend.swift
@@ -137,8 +137,8 @@ extension AWSPredictionsService: AWSComprehendServiceBehavior {
                 }
                 let beginOffSet = syntax.beginOffset?.intValue ?? 0
                 let endOffset = syntax.endOffset?.intValue ?? 0
-                let startIndex = text.index(text.startIndex, offsetBy: beginOffSet)
-                let endIndex = text.index(text.startIndex, offsetBy: endOffset)
+                let startIndex = text.unicodeScalars.index(text.startIndex, offsetBy: beginOffSet)
+                let endIndex = text.unicodeScalars.index(text.startIndex, offsetBy: endOffset)
                 let range = startIndex ..< endIndex
 
                 let score = comprehendPartOfSpeech.score?.floatValue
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/Predictions/AWSPredictionsService+Comprehend.swift
```diff
@@ -174,8 +174,8 @@ extension AWSPredictionsService: AWSComprehendServiceBehavior {
 
                 let beginOffSet = keyPhrase.beginOffset?.intValue ?? 0
                 let endOffset = keyPhrase.endOffset?.intValue ?? 0
-                let startIndex = text.index(text.startIndex, offsetBy: beginOffSet)
-                let endIndex = text.index(text.startIndex, offsetBy: endOffset)
+                let startIndex = text.unicodeScalars.index(text.startIndex, offsetBy: beginOffSet)
+                let endIndex = text.unicodeScalars.index(text.startIndex, offsetBy: endOffset)
```
royjit, line 178: Should we fix this Line 233 below as well?

royjit (approved):
LGTM

PR #905, opened 2020-11-20
fix(DataStore): Nested query predicates are not stored properly

wooj2:
For Reference, android code:
https://github.com/aws-amplify/amplify-android/blame/12ec8e9819228f63bd677a0431a5ab191f79d9e1/core/src/main/java/com/amplifyframework/core/model/query/predicate/QueryPredicateGroup.java#L77


drochetti (approved):
makes more sense now! thanks for catching that

wooj2:
> makes more sense now! thanks for catching that

Shout out to @richardmcclellan for helping me debug this!

PR #917, opened 2020-11-29
feat(DataStore): Start/Stop implementation

wooj2:
* Implementing start/stop was more difficult than expected due to supporting an offline use case.
I am leaving the PR in draft for historical purposes.  This approach attempts to introduce a state into the AWSDataStorePlugin to keep track of whether or not we've issued a start() to the sync engine.  Doing so, in my opinion, overcomplicates the design. 
 In addition, it increases overhead by starting a mostly no-op thread for each data store operation.

A couple of test cases to consider:
1.  No API Configured, customer does not call start(), but publisher needs to be setup
2. API is configured, start(), query(), stop(), start() -- explicit case
3. API is configured, start(), query(), clear(), start() -- explicit case
4. API is configured, query(), query(), stop(), query() -- implicit case, where the first query() will start the sync engine

Known issues:
There is a known issue with this implementation.  If customers attempt to call `stop`, and then immediately call `clear()`, the local database will not be cleared because the instance to the storage engine is set to `nil`.  I don't think this is a use case we really have to worry about, so I am holding off on implementing this.

### DIFF
--- a/Amplify/Categories/DataStore/DataStoreCategory+Behavior.swift
+++ b/Amplify/Categories/DataStore/DataStoreCategory+Behavior.swift
@@ -38,6 +38,14 @@ extension DataStoreCategory: DataStoreBaseBehavior {
         plugin.delete(modelType, withId: id, completion: completion)
     }
 
+    public func start(completion: @escaping DataStoreCallback<Void>) {
+        plugin.start(completion: completion)
+    }
+
+    public func stop(completion: @escaping DataStoreCallback<Void>) {
+        plugin.stop(completion: completion)
+    }
+
     public func clear(completion: @escaping DataStoreCallback<Void>) {
         plugin.clear(completion: completion)
     }
--- a/Amplify/Categories/DataStore/DataStoreCategoryBehavior.swift
+++ b/Amplify/Categories/DataStore/DataStoreCategoryBehavior.swift
@@ -34,6 +34,10 @@ public protocol DataStoreBaseBehavior {
                           withId id: String,
                           completion: @escaping DataStoreCallback<Void>)
 
+    func start(completion: @escaping DataStoreCallback<Void>)
+
+    func stop(completion: @escaping DataStoreCallback<Void>)
+
     func clear(completion: @escaping DataStoreCallback<Void>)
 }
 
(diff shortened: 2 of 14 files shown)


wooj2:
Not particularly happy with this this because:
* If customers do not have API configured, and have logs set to info, they will repeatedly get a info message of `"Unable to find suitable API plugin for syncEngine. syncEngine will not be started"`
* Having to create a serial queue to kick off sync engine is not preferable, plus, this happens each time there is a save/delete/etc

To remedy this, it seems that the only way to get around this is to either expose the state of the sync engine somehow -- either through a published combine event, or a member variable.. or something like this.

In any case, I'm sending this review out a bit early in case anyone can think of a better way of implementing this that is thread safe.

wooj2:
> Not particularly happy with this this because:
> 
> * If customers do not have API configured, and have logs set to info, they will repeatedly get a info message of `"Unable to find suitable API plugin for syncEngine. syncEngine will not be started"`
> * Having to create a serial queue to kick off sync engine is not preferable, plus, this happens each time there is a save/delete/etc
> 
> To remedy this, it seems that the only way to get around this is to either expose the state of the sync engine somehow -- either through a published combine event, or a member variable.. or something like this.
> 
> In any case, I'm sending this review out a bit early in case anyone can think of a better way of implementing this that is thread safe.

I took another look at the code.  Previously, the code storageEngine being instantiated was synonymous with the syncEngine being started (yes, you read that right.. these two actions were tightly coupled).  Since we are changing the behavior to not automatically start the syncEngine, there is not a good way to separate out instantiation with starting the sync engine without introducing some state.

Option 1 (what is currently implemented):
Have the storage engine (but could be the sync engine) keep track of the state of whether or not we have already started the sync engine.
**Thoughts:** Not ideal because we have to have a SerialQueue to on each and every save()/delete() etc.  This creates a constant overhead for each call the customer makes.

Option 2:
Change the behavior of the AWSDataStorePlugin so that the storage engine is not instantiated on configure().  Only a call to query/start/etc will instantiate the storage engine and then start it.
**Thoughts:** - This breaks a number of our unit tests because the AWSDataStorePlugin is setup to accept a StorageEngine in it's init().  Making this change feels invasive (at least on the surface), but I suspect this is what we really ought to do, rather than keep track of the state of the sync engine.

PR #926, opened 2020-12-01
fix: add targetName to hasOne relationships

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/920

*Description of changes:*
The fix in this PR is enabled by the [CLI codegen changes](https://github.com/aws-amplify/amplify-ios/issues/920) which adds the `targetName` to hasOne relationships so that this becomes an additive, non-breaking change for developers.

The fix targets making the following schema usable by API, and thus DataStore when sync is enabled:
```
# 2 - Project with explicit field for team’s id
type Project2 @model {
  id: ID!
  name: String
  teamID: ID!
  team: Team2 @connection(fields: ["teamID"])
}

type Team2 @model {
  id: ID!
  name: String!
}
```
See https://github.com/aws-amplify/amplify-ios/issues/920 for more details


AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario1Tests.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+
+import XCTest
+import AWSMobileClient
+@testable import AWSAPICategoryPlugin
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSAPICategoryPluginTestCommon
+
+class GraphQLConnectionScenario1Tests: XCTestCase {
```
palpatim, line 15: Here & throughout: "scenario1" isn't meaningful without a reference. Rather than making someone look up the README, consider also including the scenario definitions in the class comments of this file.

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -18,76 +18,129 @@ extension Model {
...
-                if let modelValue = value as? Model {
-                    input[fieldName] = modelValue.id
-
-                } else if let value = value as? [String: JSONValue],
-                          case .string(let primaryKeyValue) = value[modelSchema.primaryKey.name] {
-                    input[fieldName] = primaryKeyValue
-                } else {
-                    input[fieldName] = nil
-                }
+                let fieldName = getFieldNameForAssociatedModels(modelField: modelField) ?? modelField.name
```
palpatim, line 55: tiny nit: If the correct behavior is always to use the `modelField.name`, would it make sense to make that the return value from `getFieldNameForAssociatedModels`? Or is `nil` a meaningful value in that case?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -18,76 +18,129 @@ extension Model {
...
         }
+
+        return fixHasOneAssociationsWithExplicitFieldOnModel(input, modelSchema: modelSchema)
+    }
+
+    // This is to account for Models with an explicit field on the Model along with an object representing the hasOne
+    // association to another model. See https://github.com/aws-amplify/amplify-ios/issues/920 for more details.
+    // When the associated object is `nil`, remove the key from the GraphQL input to prevent runtime failures.
+    // When the associated object is found, take this value over the explicit field's value by replacing the correct
+    // entry for the field name of the associated model.
```
palpatim, line 86: I'm not necessarily disagreeing, but why is it right to take the object rather than the specified field? How do we let the customer know which one will be used?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -18,76 +18,129 @@ extension Model {
...
+        let defaultFieldName = modelName.camelCased() + modelField.name + "Id"
+        if case let .belongsTo(_, targetName) = modelField.association {
+            return targetName ?? defaultFieldName
+        } else if case let .hasOne(_, targetName) = modelField.association {
+            return targetName ?? defaultFieldName
+        }
+
+        return nil
+    }
+
```
palpatim, line 145: Thanks for this refactor!

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -18,76 +18,129 @@ extension Model {
...
+            return jsonModel.jsonValue(for: modelFieldName, modelSchema: modelSchema) ?? nil
+        } else {
+            return self[modelFieldName] ?? nil
+        }
+    }
+
+    // Retrieves the GraphQL field name that associates the current model with the target model.
+    // By default, this is the current model + the associated Model + "Id", For example "comment" + "Post" + "Id"
+    // This information is also stored in the schema as `targetName` which is codegenerated to be the same as the
+    // default or an explicit field specified by the developer.
```
palpatim, line 134: Wee nit: making this a doc comment means devs can see it directly in the IDE.
```suggestion
    /// Retrieves the GraphQL field name that associates the current model with the target model.
    /// By default, this is the current model + the associated Model + "Id", For example "comment" + "Post" + "Id"
    /// This information is also stored in the schema as `targetName` which is codegenerated to be the same as the
    /// default or an explicit field specified by the developer.
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
lawmicha: actually the previous code is `var fieldName = modelName.camelCased() + name` (name is resolved earlier as `name + "Id"`) which means it will always resolve to the default "postCommentId" for the associated models. I don't believe there's a case for `getFieldNameForAssociatedModels` to return nil for it to take `modelField.name`, i will update this to remove `modelField.name`

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
lawmicha: updating here and elsewhere

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
lawmicha: If they do any of the below, the correct team identifier will be used
```
let project = Project2(teamID: team.id, team: team)
let project = Project2(teamID: team.id) 
// note `teamID` is always required
```
So the decision to take the object over the specified field only matters when the developer sets their values differently, ie.
```
let project = Project2(teamID: "nonExistentTeamID", team: team)
```

If the developer started off with
```
let project = Project2(teamID: "nonExistentTeamID")
```
This will be persisted but the association points to an non existent team. They'll eventually end up with one of the three scenarios once they get working code, by persisting/retrieving the team object before using it in the project object
```
let project = Project2(teamID: team.id) // 1. they updated their code to use team.id, and it works without passing in an optional `team`
let project = Project2(teamID: team.id, team: team) // 2. they updated it in both places, because it's confusing where to add it to
let project = Project2(teamID: "nonExistentTeamID", team: team) // 3. they only passed it into `team` parameter, but it starts working. I'm guessing the teamID here is gibberish
```
Another thought I had for this was thinking that we will omit the specific field in the future with a breaking codegen change, so that our documentation always promotes the use of passing the object (since belongsTo associations already omit the specific field). So where it's not omitted, we would continue to document it as `let project = Project2(teamID: team.id, team: team)` to communicate that developers should work with the model, while satisfying requirement that `teamID` is a required field

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario3Tests.swift
```diff
@@ -0,0 +1,276 @@
...
+                XCTAssertEqual(queriedComment, comment)
+                queriedCommentCompleted.fulfill()
+            case .failure(let error):
+                XCTFail("failed \(error)")
+            }
+        }
+        wait(for: [queriedCommentCompleted], timeout: networkTimeout)
+    }
+
+    // TODO: Issue https://github.com/aws-amplify/amplify-ios/issues/939
```
lawmicha, line 95: https://github.com/aws-amplify/amplify-ios/issues/939

Amplify/Categories/DataStore/Model/Internal/Schema/ModelField+Association.swift
```diff
@@ -88,7 +88,7 @@ import Foundation
 ///   by host applications. The behavior of this may change without warning.
 public enum ModelAssociation {
     case hasMany(associatedFieldName: String?)
-    case hasOne(associatedFieldName: String?)
+    case hasOne(associatedFieldName: String?, targetName: String?)
```
royjit, line 91: Is this backward compatible, this is a public enum. Might be used by CLI for codegen

palpatim (approved):

Amplify/Categories/DataStore/Model/Internal/Schema/ModelField+Association.swift
lawmicha: thanks for catching this! updating with default `nil` parameter

AmplifyTests/CategoryTests/DataStore/ModelFieldAssociationTests.swift
```diff
@@ -36,12 +36,13 @@ class ModelFieldAssociationTests: XCTestCase {
     }
 
     func testHasOneWithCodingKeys() {
```
royjit, line 38: Can we have some test with the old code, to make sure it is backward compatible. Or keep this test as it is, and write another with the targetName fields.

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -18,76 +18,129 @@ extension Model {
...
+            return self[modelFieldName] ?? nil
+        }
+    }
+
+    /// Retrieves the GraphQL field name that associates the current model with the target model.
+    /// By default, this is the current model + the associated Model + "Id", For example "comment" + "Post" + "Id"
+    /// This information is also stored in the schema as `targetName` which is codegenerated to be the same as the
+    /// default or an explicit field specified by the developer.
+    private func getFieldNameForAssociatedModels(modelField: ModelField) -> String {
+        let defaultFieldName = modelName.camelCased() + modelField.name + "Id"
```
royjit, line 136: "Id" was not added before this change.. is this a bug fix ? Earlier it was `modelName.camelCased() + field.graphQLName`

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -18,76 +18,129 @@ extension Model {
...
-            }
+            let fieldValueOptional = getFieldValue(for: modelField.name, modelSchema: modelSchema)
 
             // Since the returned value is Any?? we need to do the following:
             // - `guard` to make sure the field name exists on the model
             // - `guard` to ensure the returned value isn't nil
-            guard let fieldValue = fieldOptionalValue else {
-                input.updateValue(nil, forKey: name)
+            guard let fieldValue = fieldValueOptional else {
+                input.updateValue(nil, forKey: modelField.name)
```
royjit, line 35: We used `field.graphQLName` before? Just confirming both works?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -18,76 +18,129 @@ extension Model {
...
+
+    private func getModelId(from value: Any, modelSchema: ModelSchema) -> String {
+        if let modelValue = value as? Model {
+            return modelValue.id
+        } else if let value = value as? [String: JSONValue],
+                  case .string(let primaryKeyValue) = value[modelSchema.primaryKey.name] {
+            return primaryKeyValue
+        }
+
+        preconditionFailure("Could not retrieve the identifier from associated model: \(value)")
```
royjit, line 120: Hope this wont break? Previous code returned nil here.

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+AnyModelWithSync.swift
```diff
@@ -64,13 +64,13 @@ extension GraphQLRequest: ModelSyncGraphQLRequestFactory {
                                                    decodePath: document.name)
     }
 
-    public static func createMutation(of model: Model, version: Int?) -> GraphQLRequest<MutationSyncResult> {
```
royjit, line 67: Why do we need these changes?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
lawmicha: yes graphQLName returns name by default, and only in the case it's a belongsTo it will return `name.pascalCase()+"Id"`. i've refactored the logic down into the `case model` scenario

https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelField+GraphQL.swift

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
lawmicha: i wrote this on the assumption that it should always get an id from the value, but i'm more inclined to return nil here to introduce less changes to the code

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
lawmicha: `graphQLName` internally resolves to `name.pascalCased() + "Id"` for associations.  oh shoot. this is incorrect then it should be:
`modelName.camelCased() + modelField.name.pascalCased() + "Id"`

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+AnyModelWithSync.swift
lawmicha: I noticed in the integration tests over in AWSAPIPlugin that tests these paths, for some reason were not compiling. until i added the default parameters here. Not sure how those were written without the nil default parameter on these methods

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
lawmicha: if it is belongsTo and returns targetName because targetName exists, then it may set targetName to nil here. I don't think this is ever the case because you cannot have nil belongsTo assocations, but i will revert this change and only make it if we're sure that there isn't a scenario that i've missed

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
lawmicha: updated to correct association field, we always pascal case the `modeField.name` for the associated field

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
lawmicha: reverted this change

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
royjit: Cool! Wondering whether we have coverage on this logic. It would nice if you can add this or add a backlog to add this logic to unit test.

royjit (approved):

PR #927, opened 2020-12-01
fix(API): subscription cancel unit test failing intermittently

wooj2:
While implementing start/stop for datastore, I noticed these tests were failing because the `SubscriptionConnectionFactory` was able to throw an exception and fail, prior to `cancel()` execution happening.  When this happens, the `operation` ends up in a `finished` state rather than the expected `canceled` state.

So.. to remedy this, I re-worked these tests by using the existing `MockSubscriptionConnectionFactory` to test various cases.

To reproduce the unexpected behavior, revert these changes, and add a `sleep(2)` right before the call to `operation.cancel()`, then run the test.

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPluginTests/Operation/AWSGraphQLSubscriptionOperationCancelTests.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginTests/Operation/AWSGraphQLSubscriptionOperationCancelTests.swift
@@ -78,7 +78,6 @@ class AWSGraphQLSubscriptionOperationCancelTests: XCTestCase {
         })
         setUp(subscriptionConnectionFactory: mockSubscriptionConnectionFactory)
 
-        let connectingSemaphore = DispatchSemaphore(value: 0)
         let request = GraphQLRequest(apiName: apiName,
                                      document: testDocument,
                                      variables: nil,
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/API/AWSAPICategoryPluginTests/Operation/AWSGraphQLSubscriptionOperationCancelTests.swift
```diff
@@ -44,12 +121,101 @@ class AWSGraphQLSubscriptionOperationCancelTests: AWSAPICategoryPluginTestBase {
...
+                receivedCompletion.fulfill()
+            }
+        }
+
+        let operation = apiPlugin.subscribe(
+            request: request,
+            valueListener: valueListener,
+            completionListener: completionListener
+        )
+        creatingConnectionSemaphore.wait()
```
lawmicha, line 215: my previous comments suggested to replace the semaphore with expectations since they were fulfilling expectations right before signaling the semaphore, but for this one seems like this makes sense that it's a semaphore

lawmicha (approved):
suggestion on the use of expectation over semaphore

AmplifyPlugins/API/AWSAPICategoryPluginTests/Operation/AWSGraphQLSubscriptionOperationCancelTests.swift
wooj2: I think this one can use an expectation as well

PR #930, opened 2020-12-02
Make fields in SelectionSetField public to be able to filter out some…

dvasileuski:
… fields in a custom model decorator for  graphql doc generation

*Issue #, if available:*

*Description of changes:*

I need this change to be able write the following code:

```swift
/// Excludes from the GraphQL document "selectionSet" fields from fieldsToExclude set.
public struct ExcludeFieldsFromSelectionSetDecorator: ModelBasedGraphQLDocumentDecorator {

    private let excludeFields: Set<String>

    public init(excludeFields: Set<String>) {
        self.excludeFields = excludeFields
    }

    public func decorate(_ document: SingleDirectiveGraphQLDocument,
                         modelType: Model.Type) -> SingleDirectiveGraphQLDocument {
        guard var selectionSet = document.selectionSet else {
            fatalError("Document should contain selectionSet here.")
        }

        selectionSet = filter(selectionSet: selectionSet)

        return document.copy(selectionSet: selectionSet)
    }

    func filter(selectionSet: SelectionSet) -> SelectionSet {
        let result = SelectionSet(value: selectionSet.value)

        for child in selectionSet.children {
            if let name = child.value.name {
                if !excludeFields.contains(name) {
                    result.addChild(settingParentOf: child)
                }
            }
        }

        return result
    }
}
```

### DIFF
--- a/AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSet.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSet.swift
@@ -18,8 +18,8 @@ public enum SelectionSetFieldType {
 }
 
 public class SelectionSetField {
-    var name: String?
-    var fieldType: SelectionSetFieldType
+    public var name: String?
+    public var fieldType: SelectionSetFieldType
     public init(name: String? = nil, fieldType: SelectionSetFieldType) {
         self.name = name
         self.fieldType = fieldType


lawmicha:
Hi @dvasileuski, looks like you are creating your own custom GraphQL document using the internal builders we are using to create the static model-to-graphQL request helpers. They're public because of how we've structured our code architecture but they are "public but internal and subject to change". I believe shipping this PR may unblock you for now, but in the future- the more your codebase relies on the "public but internal" types, the easier your code may be broken in the future. Before we add this, can you expand on your use case? Do you have an example of the calling code that uses `ExcludeFieldsFromSelectionSetDecorator` ? Currently, our recommendation is to draft the GraphQL document without relying on these decorator types if you need something custom: https://docs.amplify.aws/lib/graphqlapi/advanced-workflows/q/platform/ios/

lawmicha:
Hi @dvasileuski, it's been awhile since hearing back on your use case for creating `ExcludeFieldsFromSelectionSetDecorator`. Will close this PR for now, but feel free to reach out to us to let us know your use case for modifying the types used by the decorators. You can reach us via a Github Issue/Discussion or on Discord. From there we can explore moving this PR forward or even other options such as adding the decorator you have into the Amplify codebase.

PR #941, opened 2020-12-04
fix: Add plugin conformances to AmplifyVersionable

palpatim:
*Issue #, if available:* https://github.com/aws-amplify/amplify-ios/issues/850

*Description of changes:*

I removed plugin conformances to `AmplifyVersionable` when I added the default `version` property in the previous PR, which of course breaks since the Plugin Info helper scans plugins looking for those that conform to `AmplifyVersionable`. 

This reinstates the conformance. Tested locally, plugin versions now show up as expected. I did add conformances to the Predictions Plugins, but didn't add those plugins because my test project didn't support iOS 13:

image


royjit (approved):
LGTM... can we improve the unit test so that we could caught this before?

palpatim:
> LGTM... can we improve the unit test so that we could caught this before?

Yep, good callout! Done.

royjit (approved):

PR #947, opened 2020-12-07
fix(Auth): Dismiss empty UIViewController

ruiguoamz:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/936

*Description of changes:*

The buggy behavior is:
If user taps on `Cancel` button, the WebView is going to disappear without affecting other ui components, because it is cancelling both ViewControllers. But when user taps outside of the WebView, then other UI components will be blocked by the empty transparent `UIViewController`.

As we investigated, we believe it is a regression from Apple, it only happens on iPad running on iOS 14+.

Thanks for Wooj2's solution: disable view from disappearing if users tap outside.

### DIFF
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
@@ -225,7 +225,6 @@ extension AuthenticationProviderAdapter {
 class PresentingNavigationController: UINavigationController {
 
     override func present(_ viewControllerToPresent: UIViewController, animated flag: Bool, completion: (() -> Void)? = nil) {
-        print("present is being called")
         if #available(iOS 13, *) {
             viewControllerToPresent.isModalInPresentation = true
         }
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
@@ -146,8 +146,6 @@ extension AuthenticationProviderAdapter {
 
         parentViewController?.present(navController, animated: false, completion: {
 
-            navController.dismiss(animated: false, completion: {})
-
             self.awsMobileClient.showSignIn(navigationController: navController,
                                             signInUIOptions: SignInUIOptions(),
                                             hostedUIOptions: hostedUIOptions) { [weak self] state, error in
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
```diff
@@ -146,6 +146,8 @@ extension AuthenticationProviderAdapter {
 
         parentViewController?.present(navController, animated: false, completion: {
 
+            navController.dismiss(animated: false, completion: {})
```
royjit, line 149: do we need this now? It should work without dismissing the presentation

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
```diff
@@ -221,3 +221,14 @@ extension AuthenticationProviderAdapter {
     }
 
 }
+
+class PresentingNavigationController: UINavigationController {
+
+    override func present(_ viewControllerToPresent: UIViewController, animated flag: Bool, completion: (() -> Void)? = nil) {
+        print("present is being called")
```
royjit, line 228: Remove this print statement.

wooj2, line 228: remove print statement

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
```diff
@@ -221,3 +221,14 @@ extension AuthenticationProviderAdapter {
     }
 
 }
+
+class PresentingNavigationController: UINavigationController {
```
wooj2, line 225: Is it possible to make this an inner class and... private?

royjit (approved):
LGTM: Added a minor comment.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
```diff
@@ -134,7 +134,7 @@ extension AuthenticationProviderAdapter {
                                               signOutURIQueryParameters: request.options.signOutQueryParameters)
 
         // Create a navigation controller with an empty UIViewController.
```
wooj2, line 136: I don't think this comment is needed

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
wooj2, line 225: Maybe `ModalPresentingNavigationController` might be a better name?

PR #956, opened 2020-12-09
chore: Improve codecov reporting

palpatim:
*Description of changes:*

- Improve codecov labels; limit packages to be reported
- Add codecov reporting for AWSPluginsCore
- Added a config file to set the failure threshold

This is an attempt to reduce spurious code coverage failures and alarming reports of low coverage for things like test targets.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  branch: main
+  status:
+    project:
+      default:
+        threshold: 1%
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,5 @@
 coverage:
+  branch: main
   status:
     project:
       default:
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,7 @@
 coverage:
-  branch: main
   status:
     project:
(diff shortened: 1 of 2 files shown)


palpatim:
If I am interpreting the codecov report correctly, the drop in coverage is entirely due to test files being removed from the project.

.codecov.yml
```diff
@@ -0,0 +1,6 @@
+coverage:
+  branch: main
```
ruiguoamz, line 2: I saw in this link: https://docs.codecov.io/docs/commit-status#branches
that the yaml looks like:
```yml
coverage:
  status:
    project:
      default:
        threshold: 1%
        branches: 
          - main
```
But the current change also works, I don't know if there is difference. I guess it's fine.

ruiguoamz (approved):

.codecov.yml
palpatim: I added that per https://docs.codecov.io/docs/codecov-yaml#master-copy

Since we have a `main`, not a `master`, branch, I want to make sure that codecov picks the correct branch to cache its YAML

PR #957, opened 2020-12-10
fix(Amplify): remove a warn level log from CodingKey

ruiguoamz:
*Issue #, if available:*

*Description of changes:*
the log is printed out too often even `query` is not called

### DIFF
--- a/Amplify/Categories/DataStore/Query/ModelKey.swift
+++ b/Amplify/Categories/DataStore/Query/ModelKey.swift
@@ -44,7 +44,6 @@ extension CodingKey where Self: ModelKey {
 
     var columnName: String {
         guard let modelSchema: ModelSchema = ModelRegistry.modelSchema(from: modelName) else {
-            log.warn("Please upgrade to the latest version of Amplify CLI and rerun `amplify codegen models`")
             return stringValue
         }
         switch modelSchema.field(withName: stringValue)?.association {


lawmicha (approved):

Amplify/Categories/DataStore/Query/ModelKey.swift
```diff
@@ -44,7 +44,6 @@ extension CodingKey where Self: ModelKey {
 
     var columnName: String {
         guard let modelSchema: ModelSchema = ModelRegistry.modelSchema(from: modelName) else {
-            log.warn("Please upgrade to the latest version of Amplify CLI and rerun `amplify codegen models`")
```
palpatim, line 47: Why is removing the message the right thing here? What should a customer do to fix this if not upgrade & regen models?

ruiguoamz: The message is printed out too often even you don't explicitly call `DataStore.query()`.

For example: 
During `Amplify.DataStore.start()`, `clearStateOutgoingMutations` will be called, it creates a predicate which accesses `columnName`, but `MutationEvent` is not in `ModelRegistry` so that it fails the guard comment and print out the warning message.
```swift
func clearStateOutgoingMutations(completion: @escaping BasicClosure) {
        let predicate = fields.inProcess == true
        let sort = QuerySortDescriptor(fieldName: fields.createdAt.stringValue, order: .ascending)
        storageAdapter.query(MutationEvent.self,
                             predicate: predicate,
                             sort: [sort],
                             paginationInput: nil) { result in
                                switch result {
                                    log.error("Failed on clearStateOutgoingMutations: \(dataStoreError)")
                                case .success(let mutationEvents):
                                    if !mutationEvents.isEmpty {
                                        updateMutationsState(mutationEvents: mutationEvents,
                                                             completion: completion)
                                    } else {
                                        completion()
                                    }
                                }
        }
    }
```
But like you said, the removal of message won't tell customer the fix unless they create a Github issue and we reply to tell them.

ruiguoamz:
Closing because of a PR reverting the change: https://github.com/aws-amplify/amplify-ios/pull/963

PR #965, opened 2020-12-11
fix(Model): retrieve correct associated target name

ruiguoamz:
*Issue #, if available:*
Alternative of fixing: https://github.com/aws-amplify/amplify-ios/issues/512

*Description of changes:*
An alternative for this PR which was reverted: https://github.com/aws-amplify/amplify-ios/pull/885

Moved `func modelName(forField: fieldName)` to `AmplifyPluginsCore` so that both API and DataStore can reuse the logic.

- API: While the getGraphQLFilter(_ modelSchema: ModelSchema) is returned with the correct fieldName
- DataStore: At the time sql is translated from `QueryPredicate`, the correct columnName for `belongsTo` and `hasOne` field can be retrieved from `modelSchema.association`. In this case, CLI change is not required.


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Condition.swift
```diff
@@ -32,7 +32,7 @@ private func translateQueryPredicate(from modelSchema: ModelSchema,
         let indent = String(repeating: indentPrefix, count: indentSize)
         if let operation = pred as? QueryPredicateOperation {
             let logicalOperator = groupOpened ? "" : "\(groupType.rawValue) "
-            let column = operation.operator.columnFor(field: operation.field,
```
lawmicha, line 35: what type is `operation.field`? can we build an extension method on `operation.field` in AWSPluginsCore like
```swift
func getColumnName(_ modelSchema: ModelSchema) {
        switch modelSchema.field(withName: operation.field)?.association {
        case .belongsTo(_, let targetName):
            return targetName ?? self
        default:
            break
        }
        return self
}
```
then you can use it from both DataStore and API

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
```diff
@@ -206,7 +206,7 @@ extension GraphQLRequest: ModelGraphQLRequestFactory {
         documentBuilder.add(decorator: DirectiveNameDecorator(type: .list))
 
         if let predicate = predicate {
-            documentBuilder.add(decorator: FilterDecorator(filter: predicate.graphQLFilter))
+            documentBuilder.add(decorator: FilterDecorator(filter: predicate.graphQLFilter(modelType.schema)))
```
lawmicha, line 209: this looks like `graphQLFilter` should be deprecated, and all usages should be replaced

lawmicha: line 172 and 177 and in syncQuery over in `+AnyModel.swift`

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -47,11 +48,22 @@ public struct GraphQLFilterConverter {
...
+            return operation.graphQLFilter(modelSchema)
+        } else if let group = self as? QueryPredicateGroup {
+            return group.graphQLFilter(modelSchema)
+        }
+
+        preconditionFailure(
+            "Could not find QueryPredicateOperation or QueryPredicateGroup for \(String(describing: self))")
+    }
+
     public var graphQLFilter: GraphQLFilter {
```
lawmicha, line 62: deprecate

Amplify/Categories/DataStore/Model/Internal/Schema/ModelSchema.swift
```diff
@@ -100,6 +100,17 @@ public struct ModelSchema {
         return fields[name]
     }
 
+    public func columnName(withName name: String) -> String {
```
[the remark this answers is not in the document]
lawmicha: consider: `columnName(for field: String)`

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -60,24 +72,49 @@ extension QueryPredicate {
...
         return [field: [self.operator.graphQLOperator: self.operator.value]]
     }
 }
 
 extension QueryPredicateGroup: GraphQLFilterConvertible {
+    func graphQLFilter(_ modelSchema: ModelSchema? = nil) -> GraphQLFilter {
+        let logicalOperator = type.rawValue
+        switch type {
+        case .and, .or:
+            var graphQLPredicateOperation = [logicalOperator: [Any]()]
```
diegocstn, line 92: ultra nit: can we narrow down `Any` to a more specific type?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -47,11 +48,22 @@ public struct GraphQLFilterConverter {
...
+        }
+
+        preconditionFailure(
+            "Could not find QueryPredicateOperation or QueryPredicateGroup for \(String(describing: self))")
+    }
+
     public var graphQLFilter: GraphQLFilter {
         if let operation = self as? QueryPredicateOperation {
-            return operation.graphQLFilter
+            return operation.graphQLFilter(nil)
```
diegocstn, line 64: nit: do we need to pass `nil` as argument here and on line 66? Looks like method argument has `nil` as default value

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -60,24 +72,49 @@ extension QueryPredicate {
 }
 
 extension QueryPredicateOperation: GraphQLFilterConvertible {
+    func graphQLFilter(_ modelSchema: ModelSchema? = nil) -> GraphQLFilter {
+        guard let modelSchema = modelSchema else {
+            return [field: [self.operator.graphQLOperator: self.operator.value]]
+        }
+        return [modelSchema.columnName(withName: field): [self.operator.graphQLOperator: self.operator.value]]
```
diegocstn, line 79:
```suggestion
        let filterValue = [self.operator.graphQLOperator: self.operator.value]
        guard let modelSchema = modelSchema else {
            return [field: filterValue]
        }
        return [modelSchema.columnName(withName: field): filterValue]
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
ruiguoamz: Changed to `graphQLFilter()`

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -47,6 +52,20 @@ public struct GraphQLFilterConverter {
...
+        } else if let group = self as? QueryPredicateGroup {
+            return group.graphQLFilter(modelSchema)
+        }
+
+        preconditionFailure(
+            "Could not find QueryPredicateOperation or QueryPredicateGroup for \(String(describing: self))")
+    }
+
+    @available(*, deprecated, message: """
+    Computed property use of GraphQLFilter is deprecated, instead use the function one.
```
lawmicha, line 67: Is there a Swift way to write a function in the message? like
```
Use `graphQLFilter(_)` instead. See https://github.com/aws-amplify/amplify-ios/pull/965 for more details. 
```

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLDocument/GraphQLListQueryTests.swift
```diff
@@ -115,14 +116,127 @@ class GraphQLListQueryTests: XCTestCase {
...
+        {
+          "postID" : {
+            "eq" : "post4Id"
+          }
+        }
+        """
+        XCTAssertEqual(String(data: filterJSON!, encoding: .utf8), expectedFilterJSON)
+    }
+
+    func testComment4BelongsToPost4SuccessScenario2() {
```
lawmicha, line 171: what is "Scenario2" in the second test here? is the only difference the predicate used? If so, i'd say keep `testComment4BelongsToPost4SuccessScenario2` and rename as `testComment4BelongsToPost4Success` so that you don't duplicate the test but keep the more complex coverage. and then write your different scenarios against `graphQLFilter(_)`

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -60,12 +79,44 @@ extension QueryPredicate {
...
+        guard let modelSchema = modelSchema else {
+            return [field: filterValue]
+        }
+        return [modelSchema.columnName(forField: field): filterValue]
+    }
+
+    @available(*, deprecated, message: """
+    Computed property use of GraphQLFilter is deprecated, instead use the function one.
+    """)
     var graphQLFilter: GraphQLFilter {
```
lawmicha, line 93: this doesn't look public so no need to add deprecation

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -20,7 +25,7 @@ public struct GraphQLFilterConverter {
     /// Serialize the translated GraphQL query variable object to JSON string.
     public static func toJSON(_ queryPredicate: QueryPredicate,
                               options: JSONSerialization.WritingOptions = []) throws -> String {
-        let graphQLFilterData = try JSONSerialization.data(withJSONObject: queryPredicate.graphQLFilter,
+        let graphQLFilterData = try JSONSerialization.data(withJSONObject: queryPredicate.graphQLFilter(),
```
lawmicha, line 28: this method is using a deprecated method. At first glance, it looks like the unit tests are using this it serialize to a JSON sting so it's easy to assert what a QueryPredicate looks like in JSON, but it's public, so maybe it's being used elsewhere like in DataStore? If not, we should still treat this the same way and deprecate and pass in `modelSchema` and pass that to `graphQLFilter(modelSchema)`

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/Support/QueryPredicateGraphQLTests.swift
```diff
@@ -10,6 +10,7 @@ import XCTest
 @testable import AmplifyTestCommon
 @testable import AWSPluginsCore
 
+// swiftlint:disable type_body_length
 class QueryPredicateGraphQLTests: XCTestCase {
```
lawmicha, line 14: If you are looking for a good place to test your QueryPredicate to GraphQL filter input, then you can update `GraphQLFilterCoverted.toJSON(predicate, options)` to take in model schema
`GraphQLFilterConverter.toJSON(predicate, modelSchema, options: [.prettyPrinted])`.

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -11,6 +11,11 @@ import Amplify
 public typealias GraphQLFilter = [String: Any]
 
 protocol GraphQLFilterConvertible {
+    func graphQLFilter(_ modelSchema: ModelSchema?) -> GraphQLFilter
```
lawmicha, line 14: this may be my fault as I may have said to pass in optional ModelSchema, but by passing in `nil` means that it will call the deprecated computed property version of `GraphQLFilter` and expose the bug that you are fixing. Deprecrating the computed property is correct because there's a bug if it is used. Introducing a new public method that allows the bug to happen is incorrect then, so the signature should always require the model schema: `graphQLFilter(_ modelSchema: ModelSchema)`

lawmicha (dismissed):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario1Tests.swift
```diff
@@ -219,7 +219,6 @@ class DataStoreConnectionScenario1Tests: SyncEngineIntegrationTestBase {
         wait(for: [getProjectAfterDeleteCompleted], timeout: TestCommonConstants.networkTimeout)
     }
 
-    // TODO: This test will fail until https://github.com/aws-amplify/amplify-ios/pull/885 is merged in
     func testListProjectsByTeamID() throws {
```
lawmicha, line 222: you also need to re-enable the test by right clicking on the test from the Test Explorer and click "Enable testListProjectsByTeamID"

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario4Tests.swift
```diff
@@ -172,7 +172,6 @@ class DataStoreConnectionScenario4Tests: SyncEngineIntegrationTestBase {
         wait(for: [getCommentAfterDeleteCompleted], timeout: TestCommonConstants.networkTimeout)
     }
 
-    // TODO: This test will fail until https://github.com/aws-amplify/amplify-ios/pull/885 is merged in
```
lawmicha, line 175: re-enable test as well

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario1Tests.swift
lawmicha: this test calls `startAmplifyAndWaitForSync()` so you have to rebase from main branch to get the DataStore.start after configuring or add it in yourself
```
Amplify.DataStore.start { result in
                    guard case .success = result else {
                        XCTFail("Failed to start")
                        return
                    }
                }
```

to run the test successfully

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario5Tests.swift
```diff
@@ -46,7 +46,6 @@ import AWSMobileClient
 
 class DataStoreConnectionScenario5Tests: SyncEngineIntegrationTestBase {
 
-    // TODO: This test will fail until https://github.com/aws-amplify/amplify-ios/pull/885 is merged in
     func testListPostEditorByPost() throws {
```
lawmicha, line 49: re-enable

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario5Tests.swift
```diff
@@ -74,7 +73,6 @@ class DataStoreConnectionScenario5Tests: SyncEngineIntegrationTestBase {
         wait(for: [listPostEditorByPostIDCompleted], timeout: TestCommonConstants.networkTimeout)
     }
 
-    // TODO: This test will fail until https://github.com/aws-amplify/amplify-ios/pull/885 is merged in
```
lawmicha, line 77: re-enable

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLStatementTests.swift
```diff
@@ -309,6 +310,45 @@ class SQLStatementTests: XCTestCase {
...
+            && comment.createdAt == nil
+            && comment.post == "PostID"
+
+        let statement = SelectStatement(from: Comment.schema, predicate: predicate)
+        let expectedStatement = """
+        select
+          "root"."id" as "id", "root"."content" as "content", "root"."createdAt" as "createdAt",
+          "root"."commentPostId" as "commentPostId", "post"."id" as "post.id", "post"."content" as "post.content",
+          "post"."createdAt" as "post.createdAt", "post"."draft" as "post.draft", "post"."rating" as "post.rating",
+          "post"."status" as "post.status", "post"."title" as "post.title", "post"."updatedAt" as "post.updatedAt"
```
lawmicha, line 334: "root"."commentPostId" gets generated by using `field.columnName(forNamespace: rootNamespace) + " as " + field.columnAlias()` which resolves the column name using 
```
extension ModelField: SQLColumn {

    var sqlName: String {
        if case let .belongsTo(_, targetName) = association {
            return targetName ?? name + "Id"
        }
        return name
    }

   func columnName(forNamespace namespace: String? = nil) -> String {
        var column = sqlName.quoted()
        if let namespace = namespace {
            column = namespace.quoted() + "." + column
        }
        return column
    }
}
```

ConditionStatement also generates the columnName differently, using the extension you added on the ModelSchema. Although it would be great if a column name is generated from the same place, but more importantly there seems to be a bug because the select statement generates it from one place while condition statement generates it from another place. if targetName does not exist then it will generate different things. select statement will generate "commentPostId" while the condition statement will fall back to the field "post"

lawmicha:
looks like there's an extension on ModelField+Schema that already generates the columnName 
https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/ModelSchema+SQLite.swift#L42-L73

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario1Tests.swift
lawmicha: this test doesn't seem to work at all without client side predicate matching

ruiguoamz: What this mean? the test passes successfully

AmplifyPlugins/DataStore/Podfile.lock
```diff
@@ -118,4 +118,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 04860e414d616b67d24ed3346a60700f427764b9
 
-COCOAPODS: 1.9.3
+COCOAPODS: 1.10.0
```
ruiguoamz, line 121: will remove later

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario1Tests.swift
lawmicha: I thought Project1 and Team1 does not have the appropriate list Project operation to filter by teamId

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -60,12 +99,38 @@ extension QueryPredicate {
...
+        return [modelSchema.columnName(forField: field): filterValue]
+    }
+
     var graphQLFilter: GraphQLFilter {
         return [field: [self.operator.graphQLOperator: self.operator.value]]
     }
 }
 
 extension QueryPredicateGroup: GraphQLFilterConvertible {
+    func graphQLFilter(_ modelSchema: ModelSchema) -> GraphQLFilter {
```
lawmicha, line 113: the public method should not take an optional ModelSchema, but you can replace the computed property with the internal method taking in an operational model schema, right? this would reduce the duplication of code

Amplify/Categories/DataStore/Model/Collection/List+LazyLoad.swift
```diff
@@ -32,20 +32,12 @@ extension List {
 
         // if the collection has no associated field, return the current elements
         guard let associatedId = self.associatedId,
-              let associatedField = self.associatedField else {
+              let associatedField = associatedField else {
```
lawmicha, line 35: looks like u removed the use of `self` for associatedField, do the same for the first?`let associatedId = associatedId` as well?

Amplify/Categories/DataStore/Model/Collection/List+LazyLoad.swift
```diff
@@ -32,20 +32,12 @@ extension List {
...
-        // TODO: this is currently done by specific plugin implementations (API or DataStore)
-        // How to add this name resolution to Amplify?
-        let modelName = Element.modelName
-        var name = modelName.camelCased() + associatedField.name.pascalCased() + "Id"
-        if case let .belongsTo(_, targetName) = associatedField.association {
-            name = targetName ?? name
-        }
-
-        let predicate: QueryPredicate = field(name) == associatedId
+        let predicate: QueryPredicate = field(associatedField.name) == associatedId
```
lawmicha, line 40: great! lazy load will always pass the name of the associated field, and let it be resolved to the column name internally, just like how the customer would

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelSchema+Extension.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+extension ModelSchema {
+
+    public func columnName(forField fieldName: String) -> String {
```
lawmicha, line 13: the scenarios we talked about for DataStore queries apply for API queries as well
1. If the field cannot be resolved to the ModelField, then use it as is, so `let columnName = fieldName`.This scenario is when the developer has explicitly specified the correct column name to use such as "commentPostId" or "postID".
2. If the field can be resolved to the ModelField, then retrieve it's association, if it is a `belongsTo` or `hasOne` then return targetName
3. If for 2, there is no targetName available, then default to the expected column name provisioned by AppSync
```
let defaultFieldName = modelSchema.name.camelCased() + fieldName.pascalCased() + "Id"
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelSchema+Extension.swift
```diff
@@ -0,0 +1,24 @@
...
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+extension ModelSchema {
```
lawmicha, line 11: this no longer needs to be a generic `+Extension` for both DataStore and API since for DataStore, we are using the columnName from `ModelField: SQLColumn` that returns a column name specific for SQL statements, ie. quoting the fields and prepending namespace. So this can be moved over to a more GraphQL specific extension

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/ModelSchema+Extension.swift
lawmicha: take a look at `getFieldNameForAssociatedModels` for more info

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -60,30 +104,50 @@ extension QueryPredicate {
...
-        return [field: [self.operator.graphQLOperator: self.operator.value]]
+    func graphQLFilterInternal(_ modelSchema: ModelSchema? = nil) -> GraphQLFilter {
+        let filterValue = [self.operator.graphQLOperator: self.operator.value]
+        guard let modelSchema = modelSchema else {
+            return [field: filterValue]
+        }
+        return [columnName(modelSchema, forField: field): filterValue]
+    }
+
+    func columnName(_ modelSchema: ModelSchema, forField fieldName: String) -> String {
```
lawmicha, line 115: if `field` is on the QueryPredicateOperation, then you don't need to pass it in to the extension method and just access it in the method itself:
```
guard let modelField = modelSchema.field(withName: field) else {
```

lawmicha: in the latest code
```swift
func columnName(_ modelSchema: ModelSchema, forField fieldName: String) -> String {
    guard let modelField = modelSchema.field(withName: fieldName) else { // no need to pass in fieldName
            return field // you already have the field
```
you can also update the signature of the method to
`func columnName(_ modelSchema: ModelSchema)`

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -60,30 +104,50 @@ extension QueryPredicate {
...
+        guard let modelField = modelSchema.field(withName: fieldName) else {
+            return field
```
lawmicha, line 117: code coverage for when you pass in a `field` that does not resolve to a ModelField

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -46,12 +66,36 @@ public struct GraphQLFilterConverter {
 
 /// Extension to translate a `QueryPredicate` into a GraphQL query variables object
 extension QueryPredicate {
+    public func getGraphQLFilter(_ modelSchema: ModelSchema) -> GraphQLFilter {
```
wooj2, line 69: This function seems to be exactly the same as `graphQLFilterInternal` -- it seems that `graphQLFilterInternal` is now required since it is part of the `GraphQLFilterConvertible` protocol...

We do we need to introduce getGraphQLFilter?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -46,12 +66,36 @@ public struct GraphQLFilterConverter {
...
         if let operation = self as? QueryPredicateOperation {
-            return operation.graphQLFilter
+            return operation.graphQLFilterInternal()
+        } else if let group = self as? QueryPredicateGroup {
+            return group.graphQLFilterInternal()
+        }
+
+        preconditionFailure(
+            "Could not find QueryPredicateOperation or QueryPredicateGroup for \(String(describing: self))")
+    }
```
wooj2, line 92: Is there a reason for keeping the `graphQLFliter()` around as a deprecated method that can work in some cases (and fail in others)?  Seems like we can just change the signature and force a minor version update?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLStatementTests.swift
```diff
@@ -836,4 +877,18 @@ class SQLStatementTests: XCTestCase {
...
+
+        let statement = ConditionStatement(modelSchema: Post.schema, predicate: predicate, namespace: "root")
+
+        XCTAssertEqual("""
+          and "root"."commentPostId" = ?
+        """, statement.stringValue)
+
+        let variables = statement.variables
+        XCTAssertEqual(variables[0] as? String, "postID")
+    }
```
wooj2, line 893: Unit tests should be written in the following way.  There are always exceptions to the rule, but in general, we follow the following pattern:

```
func testDescriptiveTestName() {
    // Arrange
    Instantiate variables, etc.

    // Act
    Call the code to execute the test

    // Assert
    XCTAssert(.....)
    XCTAssert(.....)
    XCTAssert(.....)
}
```
1.  Notice that there here is a single whitespace between each section.
2.  Please update the function name to describe what we are testing.  The test name is very similar to the test above, but we seem to be testing something different in this test.

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
ruiguoamz:
```swift
protocol GraphQLFilterConvertible {
    func graphQLFilterInternal(_ modelSchema: ModelSchema?) -> GraphQLFilter
}
```
the `graphQLFilterInternal(_ modelSchema: ModelSchema? = nil)` takes `ModelSchema` as optional, but we want the public facing one `getGraphQLFilter` to take `ModelSchema` as required, in this way, the `getGraphQLFilter(_ modelSchema: ModelSchema)` can call `graphQLFilterInternal(_ modelSchema: ModelSchema? = nil)`.

And the reason to have a method that takes optional `ModelSchema` is deprecated `graphQLFilter` in `extension QueryPredicate` used to be a public computed property, better not to remove it immediately but to deprecate it. And it should call `graphQLFilterInternal()` with passing nil `ModelSchema`

ruiguoamz: But I agree this is duplicate
I have changed it to
```swift
public func getGraphQLFilter(_ modelSchema: ModelSchema) -> GraphQLFilter {
    return graphQLFilterInternal(modelSchema)
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Condition.swift
```diff
@@ -65,6 +64,21 @@ private func translateQueryPredicate(from modelSchema: ModelSchema,
         }
     }
 
+    func resolveColumn(_ operation: QueryPredicateOperation) -> String {
```
wooj2, line 67: This function is ambiguous, because if you have a namespace, it has the potential to get overwritten if you also have a modelField. (e.g. if line 74 get executed, then line 70 will also get executed -- this seems inefficient)

Is it possible to write some unit tests to demonstrate the intent of this function?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLStatementTests.swift
ruiguoamz: 1. Ya, I agree this pattern is more clear, I was just following the pattern of other tests in this file.
2. I will change the function name

ruiguoamz: Regarding bullet 1. Took another thought
```
func testDescriptiveTestName() {
    // Arrange
    Instantiate variables, etc.

    // Act
    Call the code to execute the test

    // Assert
    XCTAssert(.....)

    // Act
    Call the code to execute the test

    // Assert
    XCTAssert(.....)
}
```
Just my idea, It is clearer to show what action we are asserting for
eg. First, `ConditionStatement` is called and `stringValue` is asserted, Secondly, `let variables = statement.variables` and assert for `variables` alone

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Condition.swift
ruiguoamz: e.g.
- `testTranslateQueryPredicateWithNameSpaceWhenFieldNameSpecified` 

hits 
```swift
if let namespace = namespace {
    column = String(namespace).quoted() + "." + column
}
```
but skips
```swift
if let modelField = modelSchema.field(withName: operation.field) { // modelField doesn't return
    if let namespace = namespace {
        column = modelField.columnName(forNamespace: String(namespace)) // execute this line because name space given
    } else {
        column = modelField.columnName()
    }
}
```
because explicit filename is given like this
```swift
let predicate = field("commentPostId") == "postID"
```

- `testTranslateSingleQueryPredicateWithNamespace`

hits both
```swift
if let namespace = namespace {
    column = String(namespace).quoted() + "." + column
}
```
and
```swift
if let modelField = modelSchema.field(withName: operation.field) { // modelField is return
    if let namespace = namespace {
        column = modelField.columnName(forNamespace: String(namespace)) // execute this line because name space given
    } else {
        column = modelField.columnName()
    }
}
```
because it's using CodingKey
```swift
let post = Post.keys
let predicate = post.id != nil
```

- `testTranslateSingleQueryPredicate`

skips
```swift
if let namespace = namespace {
    column = String(namespace).quoted() + "." + column
}
```
but hits
```swift
if let modelField = modelSchema.field(withName: operation.field) { // modelField is return
    if let namespace = namespace {
        column = modelField.columnName(forNamespace: String(namespace))
    } else {
        column = modelField.columnName() // execute this line because namespace not given
    }
}
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
wooj2: If the implementation of `graphQLFilter` is broken, does it make sense to keep it around?  Seems to me it would be better to remove it immediately.

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
lawmicha: Would removing a public API that has a bug for another API be major version bump? I believe if it is a major version bump, we should maintain the deprecated API for now and do the major version bump later (2.0.0), but if it is just a minor version then we should do it now

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/SQLStatement+Condition.swift
wooj2: I think you may have misunderstood me.  Lets start over.

Is there a cleaner way to write the code so that when we enter the function, we set `column` once and only once?  We shouldn't need to make column a `var`.  If needed, we could break this out into separate functions.

ruiguoamz: I came up with code like this, return a string if it falls into each of the four cases, the force-unwrapping is safe because I make sure that it's not nil
```swift
func resolveColumn(_ operation: QueryPredicateOperation) -> String {
    let namespace = namespace
    let modelField = modelSchema.field(withName: operation.field)

    if namespace != nil && modelField != nil {
        return modelField!.columnName(forNamespace: String(namespace!))
    } else if namespace == nil && modelField != nil {
        return modelField!.columnName()
    } else if namespace != nil && modelField == nil {
        return String(namespace!).quoted() + "." + operation.field.quoted()
    } else {
        return operation.field.quoted()
    }
}
```

wooj2: The code block you provided seems like it would work, and i agree the force-unwrapping is safe, but... it can be written in the following manner:
```
        let modelField = modelSchema.field(withName: operation.field)
        if let namespace = namespace, let modelField = modelField {
            return modelField.columnName(forNamespace: String(namespace))
        } else if let modelField = modelField {
            return modelField.columnName()
        } else if let namespace = namespace {
            return String(namespace).quoted() + "." + operation.field.quoted()
        }
        return operation.field.quoted()
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLStatementTests.swift
```diff
@@ -836,4 +877,17 @@ class SQLStatementTests: XCTestCase {
...
+        let predicate = field("commentPostId") == "postID"
+
+        let statement = ConditionStatement(modelSchema: Post.schema, predicate: predicate, namespace: "root")
+        XCTAssertEqual("""
+          and "root"."commentPostId" = ?
+        """, statement.stringValue)
+
+        let variables = statement.variables
+        XCTAssertEqual(variables[0] as? String, "postID")
+    }
```
wooj2, line 892: With XCTAssertEqual, I believe the second parameter is suppose to be the expected value.  how about we change this code to match the other functions in this file:
```
    func testTranslateQueryPredicateWithNameSpaceWhenFieldNameSpecified() {
        let predicate = field("commentPostId") == "postID"

        let statement = ConditionStatement(modelSchema: Post.schema, predicate: predicate, namespace: "root")

        let expectStatement =
            """
              and "root"."commentPostId" = ?
            """
        let expectedVariable = "postID"
        let variables = statement.variables
        XCTAssertEqual(statement.stringValue, expectStatement)
        XCTAssertEqual(variables[0] as? String, expectedVariable)
    }
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
ruiguoamz: Discussed with team offline, we will deprecate it first and at some point completely remove it

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -47,11 +69,26 @@ public struct GraphQLFilterConverter {
 /// Extension to translate a `QueryPredicate` into a GraphQL query variables object
 extension QueryPredicate {
 
+    public func graphQLFilter(for modelSchema: ModelSchema?) -> GraphQLFilter {
+        Amplify.log.warn("\(#function) is for internal use only")
```
wooj2, line 73: 1.  Remove this warning statement
2. Add a comment to the effect of:
```
/// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly
///   by host applications. The behavior of this may change without warning.
```
(search code to copy and paste)

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -47,11 +69,26 @@ public struct GraphQLFilterConverter {
...
+        } else if let group = self as? QueryPredicateGroup {
+            return group.graphQLFilter(for: modelSchema)
+        }
+
+        preconditionFailure(
+            "Could not find QueryPredicateOperation or QueryPredicateGroup for \(String(describing: self))")
+    }
+
+    @available(*, deprecated, message: """
+    Use `graphQLFilter(_)` instead. See https://github.com/aws-amplify/amplify-ios/pull/965 for more details.
```
wooj2, line 85: change to: 
Use `graphQLFilter(for:)` instead

wooj2 (approved):

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -47,11 +70,27 @@ public struct GraphQLFilterConverter {
...
+    ///   by host applications. The behavior of this may change without warning.
+    public func graphQLFilter(for modelSchema: ModelSchema?) -> GraphQLFilter {
+        if let operation = self as? QueryPredicateOperation {
+            return operation.graphQLFilter(for: modelSchema)
+        } else if let group = self as? QueryPredicateGroup {
+            return group.graphQLFilter(for: modelSchema)
+        }
+
+        preconditionFailure(
+            "Could not find QueryPredicateOperation or QueryPredicateGroup for \(String(describing: self))")
```
lawmicha, line 83: no action needed; just a thought exercise here; If developers use `QueryPredicateConstant.all` then it will fail. Is there a way for developers to get into this state? Could they somehow pass the predicate as `QueryPredicateConstant.all` and then that gets synced to the API and fail here? If so, what should API do when they see a predicate that is ".all"? 

Secondly, precondionFailure affects live apps as well. In one of my latest PR's, I've started using `assert(false, "message")` instead so that it crashes during debug builds

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -47,11 +70,27 @@ public struct GraphQLFilterConverter {
 /// Extension to translate a `QueryPredicate` into a GraphQL query variables object
 extension QueryPredicate {
 
+    /// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly
+    ///   by host applications. The behavior of this may change without warning.
+    public func graphQLFilter(for modelSchema: ModelSchema?) -> GraphQLFilter {
```
lawmicha, line 75: add doc comment for the public API

lawmicha (approved):

PR #970, opened 2020-12-14
fix: add blog post comment schema and tests

lawmicha:
*Issue #, if available:*

*Description of changes:*
- Add the blog, post, comment schema that developers can get when they run `amplify add api`
- Add placeholder for API integration tests (this is in anticipation of lazy loading test for infinitely traversable data)
- Add datastore integration test

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Helpers.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Helpers.swift
@@ -1,0 +1,104 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+import AWSMobileClient
+@testable import AWSAPICategoryPlugin
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSAPICategoryPluginTestCommon
+
+extension GraphQLConnectionScenario3Tests {
+
+    func createPost(id: String = UUID().uuidString, title: String) -> Post3? {
+        let post = Post3(id: id, title: title)
+        var result: Post3?
+        let completeInvoked = expectation(description: "request completed")
+        Amplify.API.mutate(request: .create(post)) { event in
+            switch event {
+            case .success(let data):
+                switch data {
+                case .success(let post):
+                    result = post
+                default:
+                    XCTFail("Could not get data back")
+                }
+                completeInvoked.fulfill()
+            case .failure(let error):
+                XCTFail("Failed \(error)")
+            }
+        }
+        wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result
+    }
+
+    func createComment(id: String = UUID().uuidString, postID: String, content: String) -> Comment3? {
+        let comment = Comment3(id: id, postID: postID, content: content)
+        var result: Comment3?
+        let completeInvoked = expectation(description: "request completed")
+        Amplify.API.mutate(request: .create(comment)) { event in
+            switch event {
+            case .success(let data):
+                switch data {
+                case .success(let comment):
+                    result = comment
+                default:
+                    XCTFail("Could not get data back")
+                }
+                completeInvoked.fulfill()
+            case .failure(let error):
+                XCTFail("Failed \(error)")
+            }
+        }
+        wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result
+    }
+
+    func updatePost(id: String, title: String) -> Post3? {
+        var result: Post3?
+        let completeInvoked = expectation(description: "request completed")
+
+        let post = Post3(id: id, title: title)
+        _ = Amplify.API.mutate(request: .update(post)) { event in
+            switch event {
+            case .success(let data):
+                switch data {
+                case .success(let post):
+                    result = post
+                default:
+                    XCTFail("Could not get data back")
+                }
+                completeInvoked.fulfill()
+            case .failure(let error):
+                print(error)
+            }
+        }
+        wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result
+    }
+
+    func deletePost(post: Post3) -> Post3? {
+        var result: Post3?
+        let completeInvoked = expectation(description: "request completed")
+        _ = Amplify.API.mutate(request: .delete(post)) { event in
+            switch event {
+            case .success(let data):
+                switch data {
+                case .success(let post):
+                    result = post
+                default:
+                    XCTFail("Could not get data back")
+                }
+                completeInvoked.fulfill()
+            case .failure(let error):
+                print(error)
+            }
+        }
+        wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
+        return result
+    }
+}
--- a/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Helpers.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Helpers.swift
@@ -6,7 +6,6 @@
 //
 
 import XCTest
-import AWSMobileClient
 @testable import AWSAPICategoryPlugin
 @testable import Amplify
 @testable import AmplifyTestCommon
@@ -17,7 +16,7 @@ extension GraphQLConnectionScenario3Tests {
     func createPost(id: String = UUID().uuidString, title: String) -> Post3? {
         let post = Post3(id: id, title: title)
         var result: Post3?
-        let completeInvoked = expectation(description: "request completed")
+        let requestInvokedSuccessfully = expectation(description: "request completed")
         Amplify.API.mutate(request: .create(post)) { event in
             switch event {
             case .success(let data):
@@ -27,19 +26,19 @@ extension GraphQLConnectionScenario3Tests {
                 default:
                     XCTFail("Could not get data back")
                 }
-                completeInvoked.fulfill()
+                requestInvokedSuccessfully.fulfill()
             case .failure(let error):
                 XCTFail("Failed \(error)")
             }
         }
-        wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
+        wait(for: [requestInvokedSuccessfully], timeout: TestCommonConstants.networkTimeout)
         return result
     }
 
     func createComment(id: String = UUID().uuidString, postID: String, content: String) -> Comment3? {
         let comment = Comment3(id: id, postID: postID, content: content)
         var result: Comment3?
-        let completeInvoked = expectation(description: "request completed")
+        let requestInvokedSuccessfully = expectation(description: "request completed")
         Amplify.API.mutate(request: .create(comment)) { event in
             switch event {
             case .success(let data):
@@ -49,20 +48,18 @@ extension GraphQLConnectionScenario3Tests {
                 default:
                     XCTFail("Could not get data back")
                 }
-                completeInvoked.fulfill()
+                requestInvokedSuccessfully.fulfill()
             case .failure(let error):
                 XCTFail("Failed \(error)")
             }
         }
-        wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
+        wait(for: [requestInvokedSuccessfully], timeout: TestCommonConstants.networkTimeout)
         return result
     }
 
-    func updatePost(id: String, title: String) -> Post3? {
+    func mutatePost(post: Post3) -> Post3? {
         var result: Post3?
-        let completeInvoked = expectation(description: "request completed")
-
-        let post = Post3(id: id, title: title)
+        let requestInvokedSuccessfully = expectation(description: "request completed")
         _ = Amplify.API.mutate(request: .update(post)) { event in
             switch event {
             case .success(let data):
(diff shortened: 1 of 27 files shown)


AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Helpers.swift
```diff
@@ -0,0 +1,104 @@
...
+import XCTest
+import AWSMobileClient
+@testable import AWSAPICategoryPlugin
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSAPICategoryPluginTestCommon
+
+extension GraphQLConnectionScenario3Tests {
+
+    func createPost(id: String = UUID().uuidString, title: String) -> Post3? {
```
wooj2, line 17: Might be better to decouple the creation of the `Post3` with calling mutate.  For example, something like:
`func createPost(id: title) -> Post3`
`func mutatePost(post: Post) -> Post3`

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Helpers.swift
```diff
@@ -0,0 +1,104 @@
...
+            case .success(let data):
+                switch data {
+                case .success(let post):
+                    result = post
+                default:
+                    XCTFail("Could not get data back")
+                }
+                completeInvoked.fulfill()
+            case .failure(let error):
+                print(error)
```
wooj2, line 77: Maybe would be better to XCTFail on this fail rather than deal with the expectation?, similar to how we are doing it on mutate.

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Helpers.swift
```diff
@@ -0,0 +1,104 @@
...
```
wooj2, line 98: change print to XCTFail?

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Subscribe.swift
```diff
@@ -0,0 +1,232 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+import AWSMobileClient
```
wooj2, line 9: remove awsmobileclient if we can.

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Subscribe.swift
```diff
@@ -0,0 +1,232 @@
...
+                    switch state {
+                    case .connecting:
+                        break
+                    case .connected:
+                        connectedInvoked.fulfill()
+                    case .disconnected:
+                        disconnectedInvoked.fulfill()
+                    }
+                case .data:
+                    progressInvoked.fulfill()
```
wooj2, line 38: Rather than expecting twice -- seems like it would be better to match the contents of the data and have two different expectations to make sure it is the correct data?

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Subscribe.swift
```diff
@@ -0,0 +1,232 @@
...
+        operation.cancel()
+        wait(for: [disconnectedInvoked, completedInvoked], timeout: TestCommonConstants.networkTimeout)
+        XCTAssertTrue(operation.isFinished)
+    }
+
+    func testOnDeletePostSubscriptionWithModel() {
+        let connectedInvoked = expectation(description: "Connection established")
+        let disconnectedInvoked = expectation(description: "Connection disconnected")
+        let completedInvoked = expectation(description: "Completed invoked")
+        let progressInvoked = expectation(description: "progress invoked")
```
wooj2, line 131: i'm guessing progress invoked is only invoked once here because we are doing 1 create and 1 delete?

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
```diff
@@ -53,6 +53,37 @@ class GraphQLConnectionScenario3Tests: XCTestCase {
...
+        let completeInvoked = expectation(description: "request completed")
+        _ = Amplify.API.query(request: .get(Post3.self, byId: post.id)) { event in
+            switch event {
+            case .success(let graphQLResponse):
+                guard case let .success(data) = graphQLResponse else {
+                    XCTFail("Missing successful response")
+                    return
+                }
+                guard let resultPost = data else {
+                    XCTFail("Missing post from querySingle")
```
wooj2, line 71: s/querySingle/query

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
```diff
@@ -53,6 +53,37 @@ class GraphQLConnectionScenario3Tests: XCTestCase {
         Amplify.reset()
     }
 
+    func testQuerySinglePost() {
+        guard let post = createPost(title: "title") else {
+            XCTFail("Failed to set up test")
+            return
+        }
+
+        let completeInvoked = expectation(description: "request completed")
```
wooj2, line 62: s/completeInvoked/queryInvokedSuccessfully/

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
```diff
@@ -152,6 +153,83 @@ class GraphQLConnectionScenario3Tests: XCTestCase {
...
+        post.title = updatedTitle
+        let completeInvoked = expectation(description: "request completed")
+        _ = Amplify.API.mutate(request: .update(post)) { event in
+            switch event {
+            case .success(let data):
+                switch data {
+                case .success(let post):
+                    XCTAssertEqual(post.title, updatedTitle)
+                case .failure(let error):
+                    print(error)
```
wooj2, line 174: s/print/XCTFail

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
```diff
@@ -152,6 +153,83 @@ class GraphQLConnectionScenario3Tests: XCTestCase {
...
+                }
+                completeInvoked.fulfill()
+            case .failure(let error):
+                print(error)
```
wooj2, line 178: s/print/XCTFail

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
```diff
@@ -152,6 +153,83 @@ class GraphQLConnectionScenario3Tests: XCTestCase {
...
+        }
+        wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
+    }
+
+    func testDeleteExistingPost() {
+        let uuid = UUID().uuidString
+        let testMethodName = String("\(#function)".dropLast(2))
+        let title = testMethodName + "Title"
+        guard let post = createPost(id: uuid, title: title) else {
+            XCTFail("Failed to ensure at least one Post to be retrieved on the listQuery")
```
wooj2, line 189: is this the right message? we seem to be creating a post here, not doing a list query?

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
```diff
@@ -152,6 +153,83 @@ class GraphQLConnectionScenario3Tests: XCTestCase {
...
+        let completeInvoked = expectation(description: "request completed")
+
+        _ = Amplify.API.mutate(request: .delete(post)) { event in
+            switch event {
+            case .success(let data):
+                switch data {
+                case .success(let post):
+                    XCTAssertEqual(post.title, title)
+                case .failure(let error):
+                    print(error)
```
wooj2, line 202: s/print/XCTFail

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
```diff
@@ -152,6 +153,83 @@ class GraphQLConnectionScenario3Tests: XCTestCase {
...
+                }
+                completeInvoked.fulfill()
+            case .failure(let error):
+                print(error)
```
wooj2, line 206: s/print/XCTFail

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
```diff
@@ -152,6 +153,83 @@ class GraphQLConnectionScenario3Tests: XCTestCase {
...
+            }
+        }
+        wait(for: [completeInvoked], timeout: TestCommonConstants.networkTimeout)
+
```
wooj2, line 210: If we fail to wait for completeInvoked, i think the test will continue to execute here (right?)... is this what we want, or should we just fail?

ruiguoamz (approved):
Looks good to me, you can address John's comments

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Helpers.swift
lawmicha: that sounds good, `mutatePost` will make more sense since it's passing in some post that was retrieved instead of creating a dummy post (this was also bad because the dummy post created for the update will overwrite some other fields that take default values in the constructor)

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Helpers.swift
lawmicha: applying `XCTFail` this across all other places using `print` as well

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests+Subscribe.swift
lawmicha: i believe it is because the subscription is `onDelete` mutations only so only 1 delete will corespond to the subscription event

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
lawmicha: replacing with requestInvokedSuccessfully as it is easiest since a lot of helper code that uses this

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
lawmicha: updating copypasta error

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario3Tests.swift
lawmicha: if it fails to wait for the expectation, it should fail the test and shouldn't continue to execute. we want it to continue when the expectation is met to then query for the deleted post to check that it's not there

AmplifyPlugins/API/AWSAPICategoryPluginFunctionalTests/GraphQLModelBased/GraphQLConnectionScenario6Tests.swift
```diff
@@ -0,0 +1,129 @@
...
+   content: String!
+ }
+ ```
+ */
+class GraphQLConnectionScenario6Tests: XCTestCase {
+
+    override func setUp() {
+        do {
+            Amplify.Logging.logLevel = .verbose
+            try Amplify.add(plugin: AWSAPIPlugin())
```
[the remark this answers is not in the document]
lawmicha: i don't think so, just easier to set up in the test code without having to create another class that calls the register methods

wooj2 (approved):
Chatted offline, lawmicha@ will work on refactoring tests to improve readability of unit tests

PR #974, opened 2020-12-15
fix: refactor List to DataStoreList and AppSyncList

lawmicha:
*Issue #, if available:*

*Description of changes:*
This moves the implementation of List into their respective subclasses DataStoreList and AppSyncList


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -0,0 +1,172 @@
...
+/// Model associations that need to be lazy loaded.
+///
+/// When using `DataStore.query(_ modelType:)` some models might contain associations
+/// with other models and those aren't fetched automatically. This collection keeps track
+/// of the associated `id` and `field` and fetches the associated data on demand.
+public class DataStoreList<ModelType: Model>: List<ModelType>, ModelListDecoder {
+
+    /// If the list represents an association between two models, the `associatedId` will
+    /// hold the information necessary to query the associated elements (e.g. comments of a post)
+    internal var associatedId: Model.Identifier?
```
palpatim, line 23: No need to specify `internal` if both getter and setter are internal
```suggestion
    var associatedId: Model.Identifier?
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    /// The associatedField represents the field to which the owner of the `List` is linked to.
+    /// For example, if `Post.comments` is associated with `Comment.post` the `List<Comment>`
+    /// of `Post` will have a reference to the `post` field in `Comment`.
+    internal var associatedField: ModelField?
```
palpatim, line 28:
```suggestion
    var associatedField: ModelField?
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    internal var limit: Int = 100
```
palpatim, line 30:
```suggestion
    var limit: Int = 100
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    /// The current state of lazily loaded list
+    internal var state: LoadState = .pending
```
palpatim, line 33:
```suggestion
    var state = LoadState.pending
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -0,0 +1,172 @@
...
+        self.init(elements, associatedId: nil, associatedField: nil)
+        self.state = .loaded
+    }
+
+    init(_ elements: Elements,
+         associatedId: Model.Identifier? = nil,
+         associatedField: ModelField? = nil) {
+        super.init(elements)
+        self.associatedId = associatedId
+        self.associatedField = associatedField
```
palpatim, line 47: Why is the state still `pending` if the initializer specifies `elements`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    // MARK: - ExpressibleByArrayLiteral
+
+    required convenience public init(arrayLiteral elements: Element...) {
+        self.init(elements)
+    }
+
+    // MARK: - Collection conformance
+
+    public override var startIndex: Index {
```
palpatim, line 58: Here and anywhere we do a `loadIfNeeded`: note that this method implicitly lazy loads, so it isn't O(1) unless the collection has already been `load`ed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -0,0 +1,172 @@
...
+    /// - Returns: the current instance after data was loaded.
+    /// - seealso: `load(completion:)`
+    public override func load() -> Self {
+        lazyLoad()
+        return self
+    }
+
+    // MARK: Combine
+
+    /// Lazy load the collection and expose the loaded `Elements` as a Combine `Publisher`.
```
palpatim, line 153: We should be clear about this--are we lazily loading as soon as this is called, or only when it receives the first subscription? Combine supports both kinds of operations, so we need to specify

palpatim (commented):
Comments below. I've noted some missing test coverages, but stopped partway through because I remember you're adding them separately. To be clear, I'm not worried about missing the `fatalError` cases where we have deprecated methods or methods that must be overridden by subclasses, but for the rest of the new code, we need to make sure we're thoroughly covering all supported use cases of this API. We'll see those in a separate PR, I assume.

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -23,35 +23,12 @@ public class List<ModelType: Model>: Collection, Codable, ExpressibleByArrayLite
...
-    /// The associatedField represents the field to which the owner of the `List` is linked to.
-    /// For example, if `Post.comments` is associated with `Comment.post` the `List<Comment>`
-    /// of `Post` will have a reference to the `post` field in `Comment`.
-    internal var associatedField: ModelField?
-
-    internal var limit: Int = 100
-
-    /// The current state of lazily loaded list
-    internal var state: LoadState = .pending
+    public var elements: Elements
```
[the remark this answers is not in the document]
lawmicha: No reason not to use `private(set)`, in the latest iteration, i've moved the storage of `elements` into their respective subclasses. I think this is better so that all logic is in the subclasses and testing is isolated to the plugin, without having to know what the super class does to get the full picture of the implementation details.

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -62,70 +39,58 @@ public class List<ModelType: Model>: Collection, Codable, ExpressibleByArrayLite
 
     // MARK: - Collection conformance
 
-    public var startIndex: Index {
-        loadIfNeeded()
+    open var startIndex: Index {
```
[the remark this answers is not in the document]
lawmicha: i couldn't change this or I may be misunderstanding the comment. Curretly it comes from the conformance to Collection and so I needed to add `open` for subclasses to override. I've updated this to throw fatalError and subclasses are responsible for using its computed property `elements` in the Collection computed properties.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
lawmicha: the decoder was calling `self.init([], associatedId, associatedField)` so it was passing in empty array. updated the initializer to not take in elements, and decoder to call `init(associatedid, assocaiatedField)`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
lawmicha: Updated comment on this, i tested a sample example in playgrounds and obvserved that lazy loading happens as soon as it's called and the publisher is then returned. The `sink` returns the result of the call that occurred already.

Amplify/Categories/DataStore/Model/Internal/ModelListDecoder.swift
```diff
@@ -0,0 +1,30 @@
...
+/// Registry of `ModelListDecoder`'s used to retrieve decoders for checking if decodable `List<ModelType>` subclasses.
+/// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly
+///   by host applications. The behavior of this may change without warning.
+public struct ModelListDecoderRegistry {
+    public static var listDecoders: [ModelListDecoder.Type] = []
+
+    /// Register a decoder when the plugin is configured to be used for custom decoding to plugin subclasses of
+    /// `List<ModelType>`.
+    public static func registerDecoder(_ listDecoder: ModelListDecoder.Type) {
+        listDecoders.append(listDecoder)
```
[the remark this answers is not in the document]
lawmicha: added some tests in here around ModelList using `BasicList` stub

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -85,35 +92,35 @@ open class List<ModelType: Model>: ModelList {
...
+        let elements = [ModelType]()
         try elements.encode(to: encoder)
     }
 
     // MARK: - Asynchronous API
 
-    /// Use to initialize the collection.
+    /// Call this to initialize the collection if you have retrieved the list by traversing from your model objects
+    /// to its associated children objects. For example, a Post model may contain a list of Comments. By retrieving the
+    /// post object and trvasering to the comments, the comments are not retrieved from the data source until this
```
palpatim, line 107:
```suggestion
    /// post object and traversing it to the comments, the comments are not retrieved from the data source until this
```

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -85,35 +92,35 @@ open class List<ModelType: Model>: ModelList {
...
+    /// method is called. Data will be retrieved based on the plugin's data source and may have different failure
+    /// conditions, such as one that requires connectivity may fail with a network error. You can alternatively access
+    /// the Collection methods on the list to trigger the retrieval of data, but also expect that any operation to
+    /// perform as fast as it takes to make a request to the datastore and back.
```
palpatim, line 111: Suggested reword. Please format comment appropriately.

```suggestion
    /// conditions--for example, a data source that requires network connectivity may fail if the network is unavailable.
    /// Alternately, you can trigger an implicit `load` by invoking
    /// the Collection methods (such as using `map`, or iterating in a `for/in` loop)  on the List, which will retrieve data if it hasn't already been retrieved. In such cases, the time to perform that operation will include the time required to request data from the underlying data source.
```

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -85,35 +92,35 @@ open class List<ModelType: Model>: ModelList {
...
     ///
-    /// - seealso: `load()`
+    /// If you have directly created this list object then the collection has already been initialized and calling this
```
palpatim, line 113: Suggested reword. Please fill in with the appropriate initializer spelling and reformat comment as appropriate

```suggestion
    /// If you have directly created this list object (for example, by calling `List(elements:)` then the collection has already been initialized and calling this
```

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -85,35 +92,35 @@ open class List<ModelType: Model>: ModelList {
...
+    /// method will have no effect. For example, when you directly retrieve a list of comments by its post id, upon
```
palpatim, line 114: This is confusing following on the heels of you talking about navigating the object graph from Post to Comments. Suggest you include a method call to illustrate the "directly created".

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -123,14 +130,15 @@ open class List<ModelType: Model>: ModelList {
 
     // MARK: Combine
 
-    /// Lazy load the collection and expose the loaded `Elements` as a Combine `Publisher`.
-    /// This is useful for integrating the `List<ModelType>` with existing Combine code
-    /// and/or SwiftUI.
+    /// Load data into the collection from the data source and expose the data as a Combine `Publisher`. This is useful
+    /// for integrating the `List<ModelType>` with existing Combine code and/or SwiftUI. This method has been deprecated
+    /// and is currently not a supported API.
```
palpatim, line 135: Update docs and deprecation message to note that it will always crash with a fatal error

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncList.swift
```diff
@@ -10,6 +10,48 @@ import Amplify
...
+        super.init()
+    }
+
+    // MARK: - ExpressibleByArrayLiteral
+
+    required convenience init(arrayLiteral elements: List<ModelType>.Element...) {
+        self.init(elements)
+    }
+
+    // MARK: Collection conformance
```
palpatim, line 33: nit
```suggestion
    // MARK: - Collection conformance
```

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
lawmicha, line 107:
```suggestion
    /// post object and traversing to the comments, the comments are not retrieved from the data source until this
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncList.swift
```diff
@@ -10,6 +10,48 @@ import Amplify
...
+
+    public override subscript(position: Int) -> Element {
+        return elements[position]
+    }
+
+    public override __consuming func makeIterator() -> IndexingIterator<Elements> {
+        return elements.makeIterator()
+    }
+
     // MARK: Codable
```
palpatim, line 55: nit: 
```suggestion
    // MARK: - Codable
```

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
lawmicha, line 111:
```suggestion
    /// perform as fast as it takes to make a request to the data source and back.
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncList.swift
```diff
@@ -10,6 +10,48 @@ import Amplify
...
+
+    // MARK: - ExpressibleByArrayLiteral
+
+    required convenience init(arrayLiteral elements: List<ModelType>.Element...) {
+        self.init(elements)
+    }
+
+    // MARK: Collection conformance
+
+    public override var startIndex: Index {
```
palpatim, line 35: I don't see that these are implicitly loading?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -20,31 +20,37 @@ public class DataStoreList<ModelType: Model>: List<ModelType>, ModelListDecoder
...
     /// The associatedField represents the field to which the owner of the `List` is linked to.
     /// For example, if `Post.comments` is associated with `Comment.post` the `List<Comment>`
     /// of `Post` will have a reference to the `post` field in `Comment`.
-    internal var associatedField: ModelField?
+    var associatedField: ModelField?
 
-    internal var limit: Int = 100
+    var storedElements: [Element]
+
+    var limit: Int = 100
```
palpatim, line 32: nit
```suggestion
    var limit = 100
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -55,12 +61,32 @@ public class DataStoreList<ModelType: Model>: List<ModelType>, ModelListDecoder
 
     // MARK: - Collection conformance
 
+    /// This method implicitly lazy loads the underlying collection so the runtime of using Collection methods that
+    /// are normally O(1) are decreased to the time it takes to retrieve the data from the underlying SQL data source
+    /// for the initial method call. After the first access, the data will be retrieved and subsequent calls will be
+    /// O(1) runtime.
```
palpatim, line 67: suggested reword here and throughout

```suggestion
    /// for the initial method call. After the first access the data is already retrieved, so subsequent calls will be
    /// performed in O(1).
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -55,12 +61,32 @@ public class DataStoreList<ModelType: Model>: List<ModelType>, ModelListDecoder
...
     public override var startIndex: Index {
         loadIfNeeded()
         return elements.startIndex
     }
 
-    public __consuming override func makeIterator() -> IndexingIterator<Elements> {
+    public override var endIndex: Index {
+        return elements.endIndex
```
palpatim, line 74: No implicit load?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -55,12 +61,32 @@ public class DataStoreList<ModelType: Model>: List<ModelType>, ModelListDecoder
...
+    }
+
+    public override func index(after index: Index) -> Index {
+        return elements.index(after: index)
```
palpatim, line 78: no implicit load?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -55,12 +61,32 @@ public class DataStoreList<ModelType: Model>: List<ModelType>, ModelListDecoder
...
+    }
+
+    public override subscript(position: Int) -> Element {
+        return elements[position]
```
palpatim, line 82: no implicit load?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreList.swift
```diff
@@ -150,8 +183,9 @@ public class DataStoreList<ModelType: Model>: List<ModelType>, ModelListDecoder
 
     // MARK: Combine
 
-    /// Lazy load the collection and expose the loaded `Elements` as a Combine `Publisher`.
-    /// This is useful for integrating the `List<ModelType>` with existing Combine code
+    /// Lazy load the collection and expose the loaded `Elements` as a Combine `Publisher`. The collection will
+    /// immediately retrieve the data from when the publisher is returned. You can call `sink` to get the reuslts of
+    /// the load in O(1). This is useful for integrating the `List<ModelType>` with existing Combine code
```
palpatim, line 188: Suggested reword

```suggestion
    /// Lazy load the collection and expose the loaded `Elements` as a Combine `Publisher`. The List will
    /// retrieve the data from the underlying data source before the publisher is returned. This is useful for integrating the `List<ModelType>` with existing Combine code
```

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
lawmicha: going over the latest comments to apply to latest PR; the fatal error is no longer relevant since List is no longer an "abstract" class, the deprecation remains

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncList.swift
lawmicha: going over the latest comments to apply to latest PR; from discussion with @drochetti, the observed behavior is that `makeIterator()` is called when startIndex is called, thus the "loadIfNeeded()" is in fact called. In the latest PR, we're performing the load when the computed property `elements` is accessed, so this will also be covered by the computed property logic

PR #979, opened 2020-12-18
fix(DataStore): [DRAFT] cascade delete parent/child

wooj2:
**tldr; -- Don't review this PR**.  This PR is superseded by:
https://github.com/aws-amplify/amplify-ios/pull/986
https://github.com/aws-amplify/amplify-ios/pull/987

~This PR started out as demonstrating synchronous unit tests to demonstrate improved readability.  Notice:~
* ~Absence of parent base class~
* ~XCTAsserts ONLY belong in the actual test~
* ~Ability to linearly read from the top of the function to the bottom~

~However, this PR has grown into the first version of cascade delete, to address [issue #993](https://github.com/aws-amplify/amplify-ios/issues/933). This PR will be here for historical purposes, but will work on breaking this PR up into separate PRs in order to make it easier to review.  I will break this up into two PRs:
**1st PR:** Refactor - move `queryAndDeleteTransaction` and `collpaseResults`, `syncDeletion`, into an extension (with no functional changes) 
**2nd PR:**
a.  Addition of `recurseQueryAssociatedModels` w/ unit tests
b.  Removing `private` from  `submitToSyncEngine` -- i prefer not to do this, but StorageEngine is getting large, and syncDeletion methods need to be able to call this function.~

~Main reason for doing splitting up this PR is to clearly see the diff of the implementation of `queryAndDeleteTransaction` and `collapseResults` in a single file.~

~Note that this PR relies on: https://github.com/aws-amplify/amplify-ios/pull/965~

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
@@ -1,0 +1,155 @@
+//
+// Copyright 2018-2020 Amazon.com,
+// Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+import Combine
+import Foundation
+import AWSPluginsCore
+
+extension StorageEngine {
+    func queryAndDeleteTransaction<M: Model>(_ modelType: M.Type,
+                                             modelSchema: ModelSchema,
+                                             predicate: QueryPredicate) -> DataStoreResult<([M], [Model])> {
+        var queriedResult: DataStoreResult<[M]>?
+        var deletedResult: DataStoreResult<[M]>?
+        var associatedModels: [Model] = []
+
+        let queryCompletionBlock: DataStoreCallback<[M]> = { queryResult in
+            queriedResult = queryResult
+            if case .success(let queriedModels) = queryResult {
+                let modelIds = queriedModels.map {$0.id}
+                associatedModels = self.recurseQueryAssociatedModels(modelSchema: modelSchema, ids: modelIds)
+
+                let deleteCompletionWrapper: DataStoreCallback<[M]> = { deleteResult in
+                    deletedResult = deleteResult
+                }
+                self.storageAdapter.delete(modelType,
+                                           modelSchema: modelSchema,
+                                           predicate: predicate,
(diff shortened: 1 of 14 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Storage/StorageEngineTests.swift
```diff
@@ -0,0 +1,156 @@
...
+        guard case .success(let comments) =
+            queryModelSynchronous(modelType: Comment.self, predicate: Comment.keys.post == savedPost1.id) else {
+            XCTFail("Failed to query comment")
+            return
+        }
+        XCTAssertEqual(comments.count, 2)
+    }
+
+    /**
+     * Below are synchronous conveinence methods.  Please do not add any calls to XCTFail()
```
lawmicha, line 89: removing all calls to XCTFail is good addition! it makes it easier to move out of the test class

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Storage/StorageEngineTests.swift
```diff
@@ -0,0 +1,156 @@
...
+     * in these conveinence methods.  Failures should be handled in the body of the unit test.
+     */
+    func saveModelSynchronous<M: Model>(model: M) -> DataStoreResult<M> {
```
lawmicha, line 92: i wonder how well this translate to API results as that results a success/failure and inside the success is a graphQL response with the data or graphql error that has transformation/partial/error cases. This nested response means that the helper for API requests will result in aggregating the results in the helper method and returning `Result<M>` so that the caller can simply do `guard case .success(let m) = ..`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
```diff
@@ -0,0 +1,155 @@
...
+    func resolveAssociatedModels<M: Model>(_ result: DataStoreResult<([M], [Model])>) -> [Model] {
+        switch result {
+        case .success(let results):
+            return results.1
+        case .failure:
+            return []
+        }
+    }
+
+    func generateMutationsFromModels(models: [Model]) -> [MutationEvent] {
```
wooj2, line 142: Remove, no longer needed.

PR #982, opened 2020-12-19
feat(auth): Fix cancelling hostedUI returning a generic error

royjit:
*Issue #, if available:* https://github.com/aws-amplify/amplify-ios/issues/967

*Description of changes:*

- This is a behavior change and potential break the current customers, please see Before and After effect below
- Cancelling HostedUI signOut now returns `userCancelled` similar to the cancelling signIn
- Added unit test for the hostedUI flow

*Other platform behavior:*

**Android**

Android currently returns AuthException that contains AWSMobileClient error AuthNavigationException. This certainly needs to change because Amplify apis should not return AWSMobileClient specific errors. 

```
AuthException{
  message=Sign in with web UI failed
  cause=AuthNavigationException{
    message=user cancelled
  }
  recoverySuggestion=See attached exception for more details
}
```
Android PR to fix - https://github.com/aws-amplify/amplify-android/pull/1073

**Flutter**

Flutter has not implemented hostedUI yet and currently in PR - https://github.com/aws-amplify/amplify-flutter/pull/154 So, if we can push this change sooner Flutter can support this.

---
Before this PR

```swift
        Amplify.Auth.signInWithWebUI(presentationAnchor: self.view.window!) { (result) in
            
            switch result {
            case .success(let result):
                print(result)
            case .failure(let error):
                if case AuthError.unknown(_, let underlyingError) = error,
                   case .canceledLogin = (underlyingError as? SFAuthenticationError)?.code {
                       print("User cancelled")
                }
            }
        }

         // SignOut
        Amplify.Auth.signOut(listener: { (result) in
            
            switch result {
            case .success:
                print("SignOut")
            case .failure(let error):
                if case AuthError.unknown(_, let underlyingError) = error,
                   case .canceledLogin = (underlyingError as? SFAuthenticationError)?.code {
                       print("User cancelled")
                }
            }
        })
```

---
After this PR

```swift

Amplify.Auth.signInWithWebUI(presentationAnchor: self.view.window!) { (result) in
            
            switch result {
            case .success(let result):
                print(result)
            case .failure(let error):
                if case AuthError.service(_, _, let underlyingError) = error,
                   case .userCancelled = (underlyingError as? AWSCognitoAuthError) {
                    print("User cancelled")
                }
            }
        }

// SignOut

Amplify.Auth.signOut(listener: { (result) in
            
            switch result {
            case .success:
                print("SignOut")
            case .failure(let error):
                if case AuthError.service(_, _, let underlyingError) = error,
                   case .userCancelled = (underlyingError as? AWSCognitoAuthError) {
                    print("User cancelled")
                }
            }
        })

```

### DIFF
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Support/Utils/AuthErrorHelper.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Support/Utils/AuthErrorHelper.swift
@@ -7,6 +7,8 @@
 
 import Amplify
 import AWSMobileClient
+import SafariServices
+import AuthenticationServices
 
 struct AuthErrorHelper {
 
@@ -164,5 +166,22 @@ struct AuthErrorHelper {
         }
         return AuthError.unknown("An unknown error occurred", error)
 
+    }
+
+    static func didUserCancelHostedUI(_ error: Error) -> Bool {
+        if let sfAuthError = error as? SFAuthenticationError,
+           case SFAuthenticationError.Code.canceledLogin = sfAuthError.code {
+            return true
+        }
+
+        if #available(iOS 12.0, *) {
+
+            if let asWebAuthError = error as? ASWebAuthenticationSessionError,
+               case ASWebAuthenticationSessionError.Code.canceledLogin = asWebAuthError.code {
+                return true
+            }
+        }
+
+        return false
     }
 }
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderSigninWithSocialWebUITests.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderSigninWithSocialWebUITests.swift
@@ -95,13 +95,49 @@ class AuthenticationProviderSigninWithSocialWebUITests: BaseAuthenticationProvid
     /// - Given: an auth plugin with mocked service.
     ///
     /// - When:
-    ///    - I invoke signInWithWebUI and mock cancel
+    ///    - I invoke signInWithWebUI and mock cancel using SFAuthenticationError
     /// - Then:
-    ///    - I should get a SFAuthenticationError.canceledLogin error
+    ///    - I should get a AWSCognitoAuthError.userCancelled error
     ///
     func testCancelSignIn() {
-        let mockError = NSError(domain: "com.apple.SafariServices.Authentication",
+        let mockError = NSError(domain: SFAuthenticationErrorDomain,
                                 code: SFAuthenticationError.canceledLogin.rawValue,
+                                userInfo: nil)
+        mockAWSMobileClient?.showSignInMockResult = .failure(mockError)
+        let options = AuthWebUISignInRequest.Options()
+
+        let resultExpectation = expectation(description: "Should receive a result")
+        _ = plugin.signInWithWebUI(for: .amazon, presentationAnchor: window, options: options) { result in
+            defer {
+                resultExpectation.fulfill()
+            }
+            switch result {
+            case .success(let signinResult):
+                XCTFail("Should throw user cancelled error, instead - \(signinResult)")
+            case .failure(let error):
+                guard case .service(_, _, let underlyingError) = error,
+                      case .userCancelled = (underlyingError as? AWSCognitoAuthError) else {
+                    XCTFail("Should produce SFAuthenticationError error but instead produced \(error)")
+                    return
+                }
+            }
+        }
+        wait(for: [resultExpectation], timeout: apiTimeout)
+    }
+
+    /// Test a signInWithWebUI when the user cancel
+    ///
+    /// - Given: an auth plugin with mocked service.
+    ///
+    /// - When:
+    ///    - I invoke signInWithWebUI and mock cancel using ASWebAuthenticationSessionError
+    /// - Then:
+    ///    - I should get a AWSCognitoAuthError.userCancelled error
+    ///
+    @available(iOS 12.0, *)
+    func testASWebAuthenticationSessionError() {
+        let mockError = NSError(domain: ASWebAuthenticationSessionErrorDomain,
+                                code: ASWebAuthenticationSessionError.canceledLogin.rawValue,
                                 userInfo: nil)
         mockAWSMobileClient?.showSignInMockResult = .failure(mockError)
         let options = AuthWebUISignInRequest.Options()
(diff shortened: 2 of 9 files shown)


AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderSigninWithWebUITests.swift
```diff
@@ -6,11 +6,229 @@
...
+
+    /// Test a signIn with valid inputs
+    ///
+    /// - Given: Given an auth plugin with mocked service.
+    ///
+    /// - When:
+    ///    - I invoke signIn with valid values
+    ///    - Mock service returns invalid response
+    /// - Then:
+    ///    - I should get a .done response
```
ruiguoamz, line 67: I should get a .unknown error

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderSigninWithWebUITests.swift
```diff
@@ -6,11 +6,229 @@
...
+    }
+
+    /// Test a signIn with invalid id token error
+    ///
+    /// - Given: Given an auth plugin with mocked service.
+    ///
+    /// - When:
+    ///    - I invoke signIn and mock idTokenAndAcceessTokenNotIssued
+    /// - Then:
+    ///    - I should get a .service error
```
ruiguoamz, line 201: should get a .unknown error

ruiguoamz (approved):
Had it tested locally, looks good to me

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderSigninWithWebUITests.swift
palpatim: +1

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Support/Utils/AuthErrorHelper.swift
```diff
@@ -7,6 +7,10 @@
 
 import Amplify
 import AWSMobileClient
+import SafariServices
+#if canImport(AuthenticationServices)
+import AuthenticationServices
+#endif
```
palpatim, line 13: Please verify that we actually need the `canImport` guard for this. My recollection is that Apple frameworks are constructed with appropriate availability guards to avoid this. Test by importing a module with this code change in a project with an iOS 9 deployment target (using SFSafari style hosted UI), and one iOS 14 (using ASAuthenticationSession). If they both work as expected, then we shouldn't need this `canImport`

palpatim (approved):
Discussed offline: works without the `canImport`. Once that's removed, and @ruiguoamz's comments are addressed, this LGTM

palpatim (changes requested):
Actually, thinking about this, we need to validate the behavior on Android (for platform parity) and Flutter (for platform support) before committing this--as a behavior change, this has the potential to be breaking, even though it's the right thing to do.

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderSigninWithWebUITests.swift
royjit: Updated

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderSigninWithWebUITests.swift
royjit: Updated

royjit:
> Actually, thinking about this, we need to validate the behavior on Android (for platform parity) and Flutter (for platform support) before committing this--as a behavior change, this has the potential to be breaking, even though it's the right thing to do.

I have updated this in the PR description above.

jamesonwilliams:
> > Actually, thinking about this, we need to validate the behavior on Android (for platform parity) and Flutter (for platform support) before committing this--as a behavior change, this has the potential to be breaking, even though it's the right thing to do.
> 
> I have updated this in the PR description above.

@palpatim: @royjit and I discussed this today on Slack. Here's the analogous PR for Android: https://github.com/aws-amplify/amplify-android/pull/1073. We don't have good coverage over auth like iOS does, unfortunately, but I validated the change in a sample app.

Flutter currently doesn't have support for sign-in with web UI, at all. But @aws-amplify/amplify-flutter may like to be aware of this change in their upcoming work around error handling.

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderSigninWithSocialWebUITests.swift
```diff
@@ -115,8 +115,44 @@ class AuthenticationProviderSigninWithSocialWebUITests: BaseAuthenticationProvid
             case .success(let signinResult):
                 XCTFail("Should throw user cancelled error, instead - \(signinResult)")
             case .failure(let error):
-                guard case .unknown(_, let underlyingError) = error,
-                      case .canceledLogin = (underlyingError as? SFAuthenticationError)?.code else {
+                guard case .service(_, _, let underlyingError) = error,
+                      case .userCancelled = (underlyingError as? AWSCognitoAuthError) else {
+                    XCTFail("Should produce SFAuthenticationError error but instead produced \(error)")
```
palpatim, line 120:
```suggestion
                    XCTFail("Should produce AWSCognitoAuthError error but instead produced \(error)")
```

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderSigninWithSocialWebUITests.swift
```diff
@@ -115,8 +115,44 @@ class AuthenticationProviderSigninWithSocialWebUITests: BaseAuthenticationProvid
...
+            defer {
+                resultExpectation.fulfill()
+            }
+            switch result {
+            case .success(let signinResult):
+                XCTFail("Should throw user cancelled error, instead - \(signinResult)")
+            case .failure(let error):
+                guard case .service(_, _, let underlyingError) = error,
+                      case .userCancelled = (underlyingError as? AWSCognitoAuthError) else {
                     XCTFail("Should produce SFAuthenticationError error but instead produced \(error)")
```
palpatim, line 156:
```suggestion
                    XCTFail("Should produce AWSCognitoAuthError error but instead produced \(error)")
```

palpatim (approved):
Fix error messages (either "AWSCognitoAuthError" or "userCancelled error") otherwise LGTM

PR #998, opened 2020-12-30
chore(DataStore): Dispose initial sync resource

ruiguoamz:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/823

*Description of changes:*

- ModelSyncedEventEmitter

This instance won't be deallocated automatically, the reason I guess is: the two sinks/subscribers(syncOrchestratorSink & reconciliationQueueSink) in this instance never receive `.finished` completion. So that I added a `cancel()` method to manually cancel these two sinks/subscribers.

- SyncEventEmitter

This instance will be automatically deallocated once it receives completion (all ModelSyncedEventEmitter send out completion and got merged many in `SyncEventEmitter`)

- ReadyEventEmitter

For this one, I added a `completion`, so that when it sends out the `ready` event to `Hub` and returns using `completion`, I can set it to nil to reduce the ARC

So that I added a new method called:
```swift
func cancelEmitters() {
    syncEventEmitter = nil
    readyEventEmitter = nil
}
```

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
@@ -70,13 +70,6 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
         self.initialSyncOrchestratorTopic = PassthroughSubject<InitialSyncOperationEvent, DataStoreError>()
     }
 
-    deinit {
-        initialSyncOperationSinks.forEach { _, initialSyncOperationSink in
-            initialSyncOperationSink.cancel()
-        }
-        syncOperationQueue.cancelAllOperations()
-    }
-
     /// Performs an initial sync on all models. This should only be called by the
     /// RemoteSyncEngine during startup. Calling this multiple times will result in
     /// undefined behavior.
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
@@ -70,6 +70,13 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
         self.initialSyncOrchestratorTopic = PassthroughSubject<InitialSyncOperationEvent, DataStoreError>()
     }
 
+    deinit {
+        initialSyncOperationSinks.forEach { _, initialSyncOperationSink in
+            initialSyncOperationSink.cancel()
+        }
+        syncOperationQueue.cancelAllOperations()
+    }
+
     /// Performs an initial sync on all models. This should only be called by the
     /// RemoteSyncEngine during startup. Calling this multiple times will result in
     /// undefined behavior.
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ModelSyncedEventEmitter.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ModelSyncedEventEmitter.swift
@@ -64,6 +64,10 @@ final class ModelSyncedEventEmitter {
             })
     }
 
+    deinit {
+        cancel()
+    }
+
     /// Filtering `InitialSyncOperationEvent`s that come from `InitialSyncOperation` of the same ModelType
     private func filterSyncOperationEvent(_ value: InitialSyncOperationEvent) -> Bool {
         switch value {
@@ -136,6 +140,10 @@ final class ModelSyncedEventEmitter {
         modelSyncedEventTopic.send(completion: .finished)
     }
 
+    func cancel() {
+        syncOrchestratorSink?.cancel()
+        reconciliationQueueSink?.cancel()
+    }
 }
 
 @available(iOS 13.0, *)
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ModelSyncedEventEmitter.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ModelSyncedEventEmitter.swift
@@ -64,10 +64,6 @@ final class ModelSyncedEventEmitter {
             })
     }
 
-    deinit {
-        cancel()
-    }
-
     /// Filtering `InitialSyncOperationEvent`s that come from `InitialSyncOperation` of the same ModelType
     private func filterSyncOperationEvent(_ value: InitialSyncOperationEvent) -> Bool {
         switch value {
@@ -138,6 +134,7 @@ final class ModelSyncedEventEmitter {
                                                  data: modelSyncedEventBuilder.build())
         Amplify.Hub.dispatch(to: .dataStore, payload: modelSyncedEventPayload)
         modelSyncedEventTopic.send(completion: .finished)
+        cancel()
     }
 
     func cancel() {
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ReadyEventEmitter.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ReadyEventEmitter.swift
@@ -12,7 +12,10 @@ import Combine
 @available(iOS 13.0, *)
 final class ReadyEventEmitter {
     var readySink: AnyCancellable?
-    init(remoteSyncEnginePublisher: AnyPublisher<RemoteSyncEngineEvent, DataStoreError>) {
+    let callBack: () -> Void
+    init(remoteSyncEnginePublisher: AnyPublisher<RemoteSyncEngineEvent, DataStoreError>,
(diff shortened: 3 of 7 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -70,6 +70,13 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
         self.initialSyncOrchestratorTopic = PassthroughSubject<InitialSyncOperationEvent, DataStoreError>()
     }
 
+    deinit {
+        initialSyncOperationSinks.forEach { _, initialSyncOperationSink in
+            initialSyncOperationSink.cancel()
```
palpatim, line 75: 1. Have you validated that this is actually necessary? My recollection is that Combine `AnyCancellable` instances are automatically released, and the related subscription is `cancel()`ed, when their containing object is released. What does this code do that the default behavior does not?
2. Here and throughout--please make sure you've added tests to cover this new functionality. While you may not be able to assert that the resources are actually disposed of without convoluted mocking, you can at least validate that the system behaves after disposing of the resources. Perhaps you could test the `start()-stop()-start()` flow to make sure everything works as expected? Coverage reports would ensure that you're at least passing through this code block.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -70,6 +70,13 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
...
+        }
+        syncOperationQueue.cancelAllOperations()
```
palpatim, line 77: As above, let's make sure this actually gets called. If any operations on the queue are holding strong references to the orchestrator, it would result in a retain cycle that would mean `deinit` never gets invoked.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ModelSyncedEventEmitter.swift
```diff
@@ -64,6 +64,10 @@ final class ModelSyncedEventEmitter {
             })
     }
 
+    deinit {
```
palpatim, line 67: As above, is this necessary, since Combine sinks are automatically cancelled upon release?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ModelSyncedEventEmitter.swift
```diff
@@ -134,8 +134,13 @@ final class ModelSyncedEventEmitter {
                                                  data: modelSyncedEventBuilder.build())
         Amplify.Hub.dispatch(to: .dataStore, payload: modelSyncedEventPayload)
         modelSyncedEventTopic.send(completion: .finished)
+        cancel()
     }
 
+    func cancel() {
+        syncOrchestratorSink?.cancel()
+        reconciliationQueueSink?.cancel()
+    }
```
ruiguoamz, line 143: These two should be cancelled manually, because they never receives completion from downstream from my finding

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -287,7 +288,8 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
         syncEventEmitter = SyncEventEmitter(initialSyncOrchestrator: initialSyncOrchestrator,
                                             reconciliationQueue: reconciliationQueue)
 
-        readyEventEmitter = ReadyEventEmitter(remoteSyncEnginePublisher: publisher)
+        readyEventEmitter = ReadyEventEmitter(remoteSyncEnginePublisher: publisher,
+                                              callBack: { self.cancelEmitter() })
```
ruiguoamz, line 292: The `callBack` I added in ReadyEventEmitter is returned here so that it calls the new method `cancelEmitter()`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -365,6 +367,15 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
     }
 
+    func cancelEmitter() {
+        if syncEventEmitter != nil {
+            syncEventEmitter = nil
+        }
+        if readyEventEmitter != nil {
+            readyEventEmitter = nil
+        }
+    }
```
ruiguoamz, line 377: Decrement the ARC ([Automatic Reference Count](https://docs.swift.org/swift-book/LanguageGuide/AutomaticReferenceCounting.html)) to 0 so that the instance will be deallocated by the system automatically.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine+Retryable.swift
```diff
@@ -21,6 +21,7 @@ extension RemoteSyncEngine {
             scheduleRestart(advice: advice)
         } else {
             remoteSyncTopicPublisher.send(completion: .failure(DataStoreError.api(error)))
+            cancelEmitter()
```
ruiguoamz, line 24: InitialSync failure also should call `cancelEmitter()`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -218,6 +218,7 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
 
     func terminate() {
         remoteSyncTopicPublisher.send(completion: .finished)
+        cancelEmitter()
```
ruiguoamz, line 221: Terminate also should call `cancelEmitter()`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -365,6 +367,15 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
         Amplify.Hub.dispatch(to: .dataStore, payload: networkStatusEventPayload)
     }
 
+    func cancelEmitter() {
```
palpatim, line 370: nit: you're cancelling multiple emitters right? :) 
```suggestion
    func cancelEmitters() {
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -365,6 +367,15 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
...
+        if syncEventEmitter != nil {
+            syncEventEmitter = nil
+        }
+        if readyEventEmitter != nil {
+            readyEventEmitter = nil
+        }
```
palpatim, line 376: Nit, it's innocuous to assign these emitters to `nil` even if they're already `nil`, and avoids the (very unlikely) possibility of a race condition on the `if` check.
```suggestion
        syncEventEmitter = nil
        readyEventEmitter = nil
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
palpatim: Welcome to [memory management in Swift](https://github.com/palpatim/StrongReferenceDemo/blob/master/20160406%20-%20Swift%20Strong%20Reference%20Cycles.key)! :)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ReadyEventEmitter.swift
```diff
@@ -12,7 +12,10 @@ import Combine
 @available(iOS 13.0, *)
 final class ReadyEventEmitter {
     var readySink: AnyCancellable?
-    init(remoteSyncEnginePublisher: AnyPublisher<RemoteSyncEngineEvent, DataStoreError>) {
+    let callBack: () -> Void
```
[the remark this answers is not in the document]
ruiguoamz: 1. yes, updated
2. renamed it to `completion`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
ruiguoamz: Updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
ruiguoamz: Updated

palpatim (approved):

PR #1000, opened 2020-12-31
feat: Separate DataStore List logic out to list provider

lawmicha:
*Issue #, if available:*

### Public APIs Added
- `CoreError` - An error type that is returns `listOperation` and `clientValidation`, useful for plugin agnostic failure types for `ModelListProvider` APIs

### API deprecated

- `loadAsPublisher()` - This returns a DataStore specific publisher error, developers can extend `load(completion)` on their own to implement similar functionality
- `load() -> Self` - This is a fairly dangers API as it calls fatalError, updated to assert and deprecated, developers can also call `load(completion:)` and turn it synchronous to their own implementation.
- limit()  - This has not been implemented so marking as deprecated until it is implemented and usable for the developers. Current implementation has not been changed (it currently sets the state to "notLoaded", but i don't suspect any meaningful usage)
- `load(completion: DataStoreCallback)` API currently returns a specific DataStore error. Next PR will also deprecate this and introduce one that returns CoreError, like `fetch(completion: Result<Void, CoreError>`.

The reason for these deprecations is mainly due to the fact that the APIs are DataStore specific. In the follow up PRs, we will be introducing a plugin agnostic API for lazy loading the data, `fetch()` which will return a CoreError. The existing `load(completion:)`, `load()-> Self`, and `loadAsPublisher()` all depend on the `DataStoreError`. Since these are helper functions that simply wrap the existing load with completion method, we'll opt to introduce them again for the `fetch` API as a feature

### Implementation notes
- Separated out the behavior `List` into a `ModelListProvider`. Logic from the previous implementation of the List has been moved over to `DataStoreListProvider` in DataStore plugin
- `ModelListDecoderRegistry` has been added to allow creating the plugin specific list providers at runtime
- `ModelListProvider`'s `hasNextPage` and `getNextPage` will call fatalError() when called from any instances. No new APIs added to `List` that exposes the providers.
 - No behavior changes to AppSync usages of List, implemented in `AppSyncListProvider`, which is almost identical to ArrayLiteralListProvider and does not support lazy load or pagination.

- _synchronous ModelListProvider.load()_: The signature i've updated from @palpatim POC `load() -> [Elements]` to `load() -> Result<[Elements], CoreError>` to represent a synchronous version of the async call which may fail with `CoreError`. Plugin implementations can decide to crash the app if it fails to load. Default behavior of using this API will the List state as `.notLoaded` when `listProvider.load()` returns a failure.

In previous code, when the sychronous `List.load()` fails, it would call fatalError. In the current PR, this delegates the call to the list provider's load() and in DataStoreListProvider, it has been updated to perform Amplify logging and call `assert`. I believe this is an improvement to the previous behavior for DataStore as the assert will not crash a live app and only applies to debug mode (https://www.hackingwithswift.com/read/18/3/debugging-with-assert), while maintaining the same behavior as before.
A note about future implementation of `AppSyncListProvider`: As it performs a call to AppSync service on the synchronous load(), i'm reluctant to implement it the same way as DataStore's list provider as fatalError/assert may occur more frequently as a result of network calls. 

The next PR in the chain for AppSync lazy loading and pagination support is https://github.com/aws-amplify/amplify-ios/pull/1009


Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -6,82 +6,97 @@
...
 /// Model associations that need to be lazy loaded.
-///
-/// When using `DataStore.query(_ modelType:)` some models might contain associations
-/// with other models and those aren't fetched automatically. This collection keeps track
-/// of the associated `id` and `field` and fetches the associated data on demand.
 public class List<ModelType: Model>: Collection, Codable, ExpressibleByArrayLiteral {
-
     public typealias Index = Int
     public typealias Element = ModelType
     public typealias Elements = [Element]
```
palpatim, line 17: nit: Can we remove this if it's not required for protocol conformance? I don't think it really adds anything to understanding, unless we plan on changing the underlying storage of elements at some point

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -6,82 +6,97 @@
...
 
-    internal var limit: Int = 100
+    /// The provider for fulfilling list behaviors
+    let listProvider: AnyModelListProvider<Element>
 
-    /// The current state of lazily loaded list
-    internal var state: LoadState = .pending
+    /// The array of `Element` that backs the custom collection implementation.
+    /// Attempting to access the list object will attempt to retrieve the elements in memory or retrieve it from the
+    /// provider's data source. This is not thread safe as it can be performed from multiple threads, however the
```
palpatim, line 33: Let's add the note about thread safety to the doc comment of the main type as well

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -6,82 +6,97 @@
...
+                case .failure(let error):
+                    Amplify.log.error(error: error)
+                    return []
+                }
+            }
+        }
+        set {
+            switch loadedState {
+            case .loaded:
+                return
```
palpatim, line 55: Log an error here also since we explicitly disallow setting list elements after they're already loaded

Amplify/Categories/DataStore/Model/Internal/ModelListDecoder.swift
```diff
@@ -0,0 +1,41 @@
...
+///
+/// - Warning: Although this has `public` access, it is intended for internal & codegen use and should not be used
+/// directly by host applications. The behavior of this may change without warning. Though it is not used by host
+/// application making any change to these `public` types should be backward compatible, otherwise it will be a breaking
+/// change.
+public struct ModelListDecoderRegistry {
+    public static var listDecoders: [ModelListDecoder.Type] = []
+
+    /// Register a decoder when the plugin is configured to be used for custom decoding to plugin subclasses of
+    /// `List<ModelType>`.
```
lawmicha, line 20:
```suggestion
    /// Register a decoder as part of the plugin configuration flow.
```

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
palpatim: And add a note on the main type doc comment

Amplify/Categories/DataStore/Model/Internal/ModelListDecoder.swift
```diff
@@ -0,0 +1,41 @@
...
+        listDecoders.append(listDecoder)
+    }
+}
+
+extension ModelListDecoderRegistry {
+    static func reset() {
+        listDecoders = []
+    }
+}
+/// `ModelListDecoder` provides decodability checking and decoding functionality.
```
lawmicha, line 31:
```suggestion

/// `ModelListDecoder` provides decodability checking and provides providers to perform list behaviors.
```

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -91,48 +106,43 @@ public class List<ModelType: Model>: Collection, Codable, ExpressibleByArrayLite
...
     // MARK: - Codable
 
+    /// The decoding logic uses `ModelListDecoderRegistry` to find available decoders to decode to plugin specific
+    /// implementations of a `ModelListProvider` for `List<Model>`. The decoders should be added to the registry by the
+    /// plugin as part of its configuration steps. By delegating responsibility to the `ModelListDecoder`, it is up to
+    /// the plugin to successfully return an instance of `ModelListProvider`.
     required convenience public init(from decoder: Decoder) throws {
+        for listDecoder in ModelListDecoderRegistry.listDecoders {
+            if listDecoder.shouldDecode(decoder: decoder) {
+                let listProvider = try listDecoder.getListProvider(modelType: ModelType.self, decoder: decoder)
```
palpatim, line 126: 👍

Amplify/Categories/DataStore/Model/Internal/ModelListDecoder.swift
```diff
@@ -0,0 +1,41 @@
...
+    public static var listDecoders: [ModelListDecoder.Type] = []
+
+    /// Register a decoder when the plugin is configured to be used for custom decoding to plugin subclasses of
+    /// `List<ModelType>`.
+    public static func registerDecoder(_ listDecoder: ModelListDecoder.Type) {
+        listDecoders.append(listDecoder)
+    }
+}
+
+extension ModelListDecoderRegistry {
```
palpatim, line 26: make conform to `Resettable` (I think that's the protocol name)

Amplify/Categories/DataStore/Model/Internal/ModelListProvider.swift
```diff
@@ -0,0 +1,76 @@
...
+    associatedtype Element: Model
+
+    /// Retrieve the array of `Element` from the data source.
+    func load() -> Result<[Element], CoreError>
+
+    ///  Retrieve the array of `Element` from the data source asychronously.
+    func load(completion: (Result<[Element], CoreError>) -> Void)
+
+    /// Checks if there is subsequent data to retrieve. If True, the next page can be retrieved using.
+    /// `getNextPage(completion)`
```
palpatim, line 25:
```suggestion
    /// Checks if there is subsequent data to retrieve. If true, the next page can be retrieved using
    /// `getNextPage(completion:)`
```

Amplify/Core/Error/CoreError.swift
```diff
@@ -0,0 +1,72 @@
...
+    case listOperation(ErrorDescription, RecoverySuggestion, Error? = nil)
+
+    /// A client side validation error occured.
+    case clientValidation(ErrorDescription, RecoverySuggestion, Error? = nil)
+
+    /// An error to encapsulate an error received by a dependent plugin
+    case pluginError(AmplifyError)
+
+    /// An unknown error occurred
+    case unknown(ErrorDescription, Error? = nil)
```
palpatim, line 21: Are we using this?

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListDecoder.swift
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        return false
+    }
+
+    public static func getListProvider<ModelType: Model>(modelType: ModelType.Type,
+                                                         decoder: Decoder) throws -> AnyModelListProvider<ModelType> {
+        let json = try JSONValue(from: decoder)
+
+        if case let .object(jsonObject) = json,
```
palpatim, line 32: nit: prefer guard to handle exceptional early exits

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListDecoder.swift
```diff
@@ -0,0 +1,51 @@
...
+                let serializedJSON = try encoder.encode(jsonElement)
+                return try decoder.decode(ModelType.self, from: serializedJSON)
+            }
+
+            return AppSyncListProvider(elements).eraseToAnyModelListProvider()
+        }
+
+        let message = "AppSyncListProvider could not be created from \(String(describing: json))"
+        Amplify.DataStore.log.error(message)
+        assert(true, message)
```
palpatim, line 48: What is the intent of this `assert`? It will never crash even in debug mode

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreListDecoder.swift
```diff
@@ -0,0 +1,65 @@
...
+               let field = ModelType.schema.field(withName: associatedField) {
+                return DataStoreListProvider<ModelType>(associatedId: associatedId,
+                                                        associatedField: field).eraseToAnyModelListProvider()
+            } else {
+                let message = """
+                DataStoreListProvider could not be created. Failed to store association data for \(modelType.modelName)
+                given: \(associationData)
+                """
+                Amplify.DataStore.log.error(message)
+                assert(true, message)
```
palpatim, line 55: As above, this `assert` will never trigger

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreListDecoder.swift
```diff
@@ -0,0 +1,65 @@
...
+                return ArrayLiteralListProvider<ModelType>(elements: []).eraseToAnyModelListProvider()
+            }
+        default:
+            let message = "DataStoreListProvider could not be created from \(String(describing: json))"
+            Amplify.DataStore.log.error(message)
+            assert(true, message)
```
palpatim, line 61: As above, this `assert` will never trigger

Amplify/Categories/DataStore/Model/Internal/ModelListDecoder.swift
lawmicha: that one provides `reset` on instances, so i don't think we have one at the static method level

Amplify/Core/Error/CoreError.swift
lawmicha: we are hardcoding to `unknown` in DataStoreListProvider's synchronous `load()` as the default case before using a DispatchSemaphore on the async load. So it will never occur and will hang on the sempahore if the query doesn't return. I have a few options for this, most likey i can do `.pluginError(DataStore.operationError)` or some sort as the default 

One reason was was because `AmplifyError` requires an init, which we've implemented across other AmplifyError's default to setting it to `.unknown` in the else case:
```
public init(
        errorDescription: ErrorDescription = "An unknown error occurred",
        recoverySuggestion: RecoverySuggestion = "(Ignored)",
        error: Error
    ) {
        if let error = error as? Self {
            self = error
        } else {
            self = .unknown(errorDescription, error)
        }
    }
```

`.unknown` does have a slight difference from the others, for example, we will return recoverySuggestion as `AmplifyErrorMessages.shouldNotHappenReportBugToAWS()` which will guide developers to open a bug with the issue that occured:
```
There is a possibility that there is a bug if this error persists. Please take a look at \
        https://github.com/aws-amplify/amplify-ios/issues to see if there are any existing issues that \
        match your scenario, and file an issue with the details of the bug if there isn't. Issue encountered \
        at:
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListDecoder.swift
lawmicha: will keep this in mind but would like to keep as is, this logic will change drastically with the double decoding logic that's performed to support pagination

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListDecoder.swift
lawmicha: this is a coding error, should be `assert(false, message)`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreListDecoder.swift
lawmicha: updating to pass in false

Amplify/Categories/DataStore/Model/Collection/ArrayLiteralListProvider.swift
```diff
@@ -0,0 +1,34 @@
...
+        completion(.success(elements))
+    }
+
+    public func hasNextPage() -> Bool {
+        false
+    }
+
+    public func getNextPage(completion: @escaping (Result<List<Element>, CoreError>) -> Void) {
+        completion(.failure(CoreError.clientValidation("No pagination on an array literal",
+                                                       "Don't call this method",
```
palpatim, line 31: Very nice actionable recovery suggestion. :D

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -6,82 +6,120 @@
...
-/// `List<ModelType>` is a DataStore-aware custom `Collection` that is capable of loading
-/// records from the `DataStore` on-demand. This is specially useful when dealing with
-/// Model associations that need to be lazy loaded.
-///
-/// When using `DataStore.query(_ modelType:)` some models might contain associations
-/// with other models and those aren't fetched automatically. This collection keeps track
-/// of the associated `id` and `field` and fetches the associated data on demand.
+import Combine
+
+/// `List<ModelType>` is a custom `Collection` that is capable of loading records from a data source. This is specially
```
palpatim, line 11:
```suggestion
/// `List<ModelType>` is a custom `Collection` that is capable of loading records from a data source. This is especially
```

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
```diff
@@ -91,48 +129,43 @@ public class List<ModelType: Model>: Collection, Codable, ExpressibleByArrayLite
         return 0
     }
 
+    /// `limit` is currently not a supported API.
+    @available(*, deprecated, message: "Not supported.")
     public func limit(_ limit: Int) -> Self {
         // TODO handle query with limit
-        self.limit = limit
-        state = .pending
+        loadedState = .notLoaded
```
palpatim, line 136: This method shouldn't have any side effects if it's not supported. This implementation would allow someone to call `limit` and reset the loaded state of the list, thus potentially working around some of our state checks elsewhere.

Amplify/Categories/DataStore/Model/Internal/ModelListDecoder.swift
```diff
@@ -0,0 +1,41 @@
...
+    static func reset() {
+        listDecoders.set([ModelListDecoder.Type]())
+    }
+}
+/// `ModelListDecoder` provides decoding and list functionality.
```
palpatim, line 31: nit: vertical whitespace
```suggestion

/// `ModelListDecoder` provides decoding and list functionality.
```

Amplify/Categories/DataStore/Model/Internal/ModelListProvider.swift
```diff
@@ -0,0 +1,76 @@
...
+public protocol ModelListProvider {
+    associatedtype Element: Model
+
+    /// Retrieve the array of `Element` from the data source.
+    func load() -> Result<[Element], CoreError>
+
+    ///  Retrieve the array of `Element` from the data source asychronously.
+    func load(completion: @escaping (Result<[Element], CoreError>) -> Void)
+
+    /// Checks if there is subsequent data to retrieve. If true, the next page can be retrieved using.
```
palpatim, line 24:
```suggestion
    /// Checks if there is subsequent data to retrieve. If true, the next page can be retrieved using
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreListProvider.swift
```diff
@@ -0,0 +1,97 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+import Combine
+
+/// `DataStoreList<ModelType>` is a DataStore-aware custom `Collection` that is capable of loading
+/// records from the `DataStore` on-demand. This is specially useful when dealing with
```
palpatim, line 13:
```suggestion
/// records from the `DataStore` on-demand. This is especially useful when dealing with
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreListProvider.swift
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+    }
+
+    public func hasNextPage() -> Bool {
+        fatalError("Not implemented.")
+    }
+
+    public func getNextPage(completion: (Result<List<Element>, CoreError>) -> Void) {
+        fatalError("Not implemented.")
```
palpatim, line 95: Why fatalError instead of a `.failure` result like you did for the ArrayLiteral?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreListProvider.swift
```diff
@@ -0,0 +1,97 @@
...
+                    completion(.failure(CoreError.listOperation("Failed to Query DataStore.",
+                                                                "See underlying DataStoreError for more details.",
+                                                                error)))
+                }
+            }
+        }
+    }
+
+    public func hasNextPage() -> Bool {
+        fatalError("Not implemented.")
```
palpatim, line 91: Why a fatalError instead of just `false`?

Amplify/Categories/DataStore/Model/Collection/List+Model.swift
lawmicha: my original thinking was to keep the original implementation since that would be the "backwards compatible" thing to do. But the previous implementation was not complete, and since we are marking this as not supported and the internal state management has changed drastically in this PR, it will be safer to remove the implementation in `limit` all together.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Core/DataStoreListProvider.swift
lawmicha: In this PR, i wanted to keep the logic to a minimal, even the pagination APIs are not added to `List` type. I think i have added the actual implementation and testing in the next PR

palpatim (approved):

PR #1009, opened 2021-01-08
feat: Support lazy load and pagination for API

lawmicha:
*Related Issues*
https://github.com/aws-amplify/amplify-ios/issues/860

*Previous PR*
https://github.com/aws-amplify/amplify-ios/pull/1000

### Public APIs
`.paginatedList` is useful for Developers to create a `List<M>`. 
```swift
extension GraphQLRequest {
}
```
Currently exposed `.list` API returns array literal. We can rename `.paginatedList`  to `.list` as a major version change.

`fetch` is the new proposed API name for lazy loading data.

```swift
extension List {
       public func fetch(_ completion: @escaping (Result<Void, CoreError>) -> Void)
}
```

On success, this returns Void to avoid any confusing on what this API does. After calling fetch, on success, developers continue to use the original loaded list.

`.elements`

```swift
public class List {
  public internal(set) var elements: [Element]
}
```
Access to elements allows the developers to perform on an [Array Literal type](https://developer.apple.com/documentation/swift/array). 

### Deprecated APIs

```swift
extension List {
   @available(*, deprecated, message: "Use fetch(completion:) instead.")
    public func load(_ completion: DataStoreCallback<[Element]>) 
}
```
This has been deprecated due to its tight coupling to a DataStoreError in its failure result type.

### Developer Experience

**How to create a List: Query for a list of items**

```swift
Amplify.API.query(.paginatedList(Comments.self, limit: 10)) { result in
   switch result {
   case .success(let graphQLResponse):
       switch graphqlResponse {
          case .success(let comments):
          case .failure(let graphQLErrorResponse):
       }
   case .failure(let apiError):
   }
}
```

The list will be loaded since this is directly loading the items from AppSync into the List.

**Given a model, traverse to its associated array data**
```swift
Amplify.API.query(request: .get(Post3.self, byId: post.id)) { result in
    switch result {
    case .success(let result):
        switch result {
        case .success(let queriedPostOptional):
            guard let queriedPost = queriedPostOptional else { // Post object
                return
            }
            guard let comments = queriedPost.comments else { // instantiated List object
                return
            }
```
The comments List object is deserialized with the the associated data (post field and postId) to allow lazy loading the comments for that post. The developer can explicitly call fetch to load the comments into the List object
```swift
// Then load the data
comments.fetch { fetchResults in // fetches up to 1000
    switch fetchResults {
    case .success: 
    case .failure(let error):
    }
}

for comment in comments {
}
```
Or comments can be implicitly fetched by performing operations on the object such as
```swift
// accessing the count
comments.count 
// or iterating over the comments
for comment in comments {
}
```

**Pagination support: How to get next page**

```swift
if comments.hasNextPage() {
    comments.getNextPage { result in
        switch result {
        case .success(let nextPageList):
            print(nextPageList)
        case .failure(let coreError):
            print("Failed to fetch next page \(coreError)")
            switch coreError {
               case .validation: break
               case .listOperation: break
            }
        }
    }
}
```

**How to aggregate pages of data**

If you are using the results to hydrate UI components, you should create your own in-memory array to append new pages into it. The reason for this is because there is metadata stored in the object that is used to get the next page, so any current page only knows how to get the next page. 
```swift
var resultsArray: [Post] = [] // binding to UI components
var currentPage: List<Post>?

_ = Amplify.API.query(request: .paginatedList(Post.self, limit: 1)) { event in
    switch event {
    case .success(let response):
        guard case let .success(graphQLResponse) = response else {
            XCTFail("Missing successful response")
            return
        }
        resultsArray.append(contentsOf: graphQLResponse)
        currentPage = graphQLResponse
    case .failure(let error):
        XCTFail("Unexpected .failure event: \(error)")
    }
}

// respond to UI actions to retrieve next page, ie.
func loadMore() {
    if let current = currentPage, current.hasNextPage() {
        current.getNextPage { result in
            switch result {
            case .success(let list):
                resultsArray.append(contentsOf: list)
                currentPage = list
                loadMore()
            case .failure(let coreError):
                print("Failed to fetch next page \(coreError)")
            }
        }
    }
}
```


Amplify/Categories/DataStore/Model/Collection/List+Pagination.swift
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+    }
+
+    /// Retrieve the next page as a new in-memory List object. Calling this will load the underlying elements from the
+    /// data source if not yet loaded before
+    public func getNextPage(completion: @escaping (Result<List<Element>, CoreError>) -> Void) {
+        switch loadedState {
+        case .loaded:
+            listProvider.getNextPage(completion: completion)
```
lawmicha, line 37: In terms of platform parity, for Android in its current state-

- Android does not have lazy loading, and a list will always be created in a "loaded" state.
- getNextRequest returns nil when there is no next page. 

on IOS, when it is loaded, we delegate the call to the list provider to getNextPage. In the case where there is no next page, we return a failure. 

previous conversation: https://github.com/aws-amplify/amplify-ios/pull/990/files#r548594733

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/GraphQLRequest+toListQuery.swift
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Amplify
+import AWSPluginsCore
+
+extension GraphQLRequest {
+
+    /// Retrieve a GraphQL List operation request
+    static func listQuery<ResponseType: Decodable>(responseType: ResponseType.Type,
```
lawmicha, line 14: this method was refactored and moved over here from the Provider/List implementation 

previous conversation: https://github.com/aws-amplify/amplify-ios/pull/990/files#r548597032

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListPayload.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// Extract from the `variables` object the original GraphQL "filter" is required to perform pagination by
+    /// the `AppSyncListProvider` based on the original request filter criteria that was passed in as
+    /// a `QueryPredicate`. This may be the association data used if lazy loading was originally performed, or the
+    /// predicate that was passed in when retrieving a list (`Amplify.API.query(.paginatedList(where: predicate)`)
+    var graphQLFilter: [String: Any]? {
+        guard let storedVariables = variables,
+           let filters = storedVariables["filter"],
+           case let .object(filterValue) = filters else {
+            return nil
+        }
```
lawmicha, line 36: updated to multiple guard statements

previous conversation: https://github.com/aws-amplify/amplify-ios/pull/990/files#r548597651

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListPayload.swift
```diff
@@ -0,0 +1,65 @@
...
+        let encoder = JSONEncoder()
+        encoder.dateEncodingStrategy = ModelDateFormatting.encodingStrategy
+
+        // At this point, there exists a "filter" object that is expected to be successful serialized back to a
+        // `GraphQLFilter`.
+        guard let filterVariablesData = try? encoder.encode(filterValue),
+              let filterVariablesJSON = try? JSONSerialization.jsonObject(with: filterVariablesData)
+                as? GraphQLFilter else {
+
+            assert(false, "Filter variables is not a valid JSON object: \(filterValue)")
```
lawmicha, line 47: updated this to `assert` over `fatal`

previous conversation: https://github.com/aws-amplify/amplify-ios/pull/990/files#r548598473

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListDecoder.swift
```diff
@@ -9,16 +9,37 @@ import Foundation
 import Amplify
 
 public struct AppSyncListDecoder: ModelListDecoder {
+
+    /// Return true when the incoming payload to decode can be detected as one of the following scenarios:
```
lawmicha, line 13: explaining the flow a bit here

previous conversation: https://github.com/aws-amplify/amplify-ios/pull/990/files#r548609876

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Decode/GraphQLResponseDecoder+DecodeData.swift
```diff
@@ -21,17 +21,30 @@ extension GraphQLResponseDecoder {
...
             let anyModel = try AnyModel(modelJSON: graphQLData)
-            let serializedJSON = try encoder.encode(anyModel)
-            return try decoder.decode(request.responseType, from: serializedJSON)
+            serializedJSON = try encoder.encode(anyModel)
+        } else if request.responseType is ModelListMarker.Type {
+            let payload = AppSyncListPayload(graphQLData: graphQLData,
+                                             apiName: request.apiName,
+                                             variables: try getVariablesJSON())
+            serializedJSON = try encoder.encode(payload)
+        } else if AppSyncModelMetadataUtils.shouldAddMetadata(toModel: graphQLData) {
```
lawmicha, line 35: no logical ordering of these statements, just a bit of refactoring to avoid double decoding flows (by checking `responseType` instead of decoding once and checking `response`)

previous conversation: https://github.com/aws-amplify/amplify-ios/pull/990/files#r548616921

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncModelMetadata.swift
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        let encoder = JSONEncoder()
+        encoder.dateEncodingStrategy = ModelDateFormatting.encodingStrategy
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = ModelDateFormatting.decodingStrategy
+        // Iterate over the associations of the model and for each association, store its association data when
+        // the object at the association is empty. For example, if the modelType is a Post and has a field that is an
+        // array association like Comment, store the post's identifier and the ModelField name of the parent, ie.
+        // "post" in the comments object as metadata.
```
lawmicha, line 79: attempting to clarify this a bit more

previous conversation: https://github.com/aws-amplify/amplify-ios/pull/990/files#r548622749

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListProvider.swift
```diff
@@ -7,27 +7,223 @@
...
+                completion(.failure(.listOperation("The AppSync request failed",
+                                                   "See underlying `APIError` for more details.",
+                                                   apiError)))
+            }
+        }
+    }
+
+    /// Check if there is subsequent data to retrieve. This method always returns false if the underlying provider is
+    /// not loaded. Make sure the underlying data is loaded by calling `load(completion)` before calling this method.
+    /// If true, the next page can be retrieved using `getNextPage(completion:)`.
```
palpatim, line 169: nit: Not going to swear to it, but I think Xcode will auto-populate the doc comments from the protocol if you leave the override comments blank. Not a big deal, but it saves you from having to try to keep comments in sync across the different contexts (protocol, different implementations)

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncModelMetadata.swift
```diff
@@ -0,0 +1,100 @@
...
+public struct AppSyncModelMetadata: Codable {
+    let appSyncAssociatedId: String
+    let appSyncAssociatedField: String
+    let apiName: String?
+
+    init(associatedId: String, associatedField: String, apiName: String?) {
+        self.appSyncAssociatedId = associatedId
+        self.appSyncAssociatedField = associatedField
+        self.apiName = apiName
+    }
```
palpatim, line 20: nit: if you're not changing the access level from `internal`, you don't need to explicitly write it: the compiler auto-synthesizes a member-wise initializer for structs.

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Decode/GraphQLResponseDecoder+DecodeData.swift
```diff
@@ -43,4 +56,12 @@ extension GraphQLResponseDecoder {
 
         return model
     }
+
+    private func getVariablesJSON() throws -> [String: JSONValue]? {
+        if let variables = request.variables {
```
palpatim, line 61: nit: prefer guard/let for early exit conditions

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListDecoder.swift
```diff
@@ -27,25 +48,27 @@ public struct AppSyncListDecoder: ModelListDecoder {
...
-            encoder.dateEncodingStrategy = ModelDateFormatting.encodingStrategy
-            let decoder = JSONDecoder()
-            decoder.dateDecodingStrategy = ModelDateFormatting.decodingStrategy
-            let elements = try jsonArray.map { (jsonElement) -> ModelType in
-                let serializedJSON = try encoder.encode(jsonElement)
-                return try decoder.decode(ModelType.self, from: serializedJSON)
-            }
-
-            return AppSyncListProvider(elements).eraseToAnyModelListProvider()
+    static func getAppSyncListProvider<ModelType: Model>(modelType: ModelType.Type,
```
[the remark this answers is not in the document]
lawmicha: Does this comment apply to the protocol for the method `getListProvider`? This appears to be the case, but i see the comment is on the internal `getAppSyncListProvider` method, so just want to clarify

Current
```swift
public protocol ModelListDecoder {
    static func shouldDecode<ModelType: Model>(modelType: ModelType.Type, decoder: Decoder) -> Bool
    static func getListProvider<ModelType: Model>(
        modelType: ModelType.Type, decoder: Decoder) throws -> AnyModelListProvider<ModelType>
}
```

As per guideline"

```swift
public protocol ModelListDecoder {
    static func shouldDecode<ModelType: Model>(modelType: ModelType.Type, decoder: Decoder) -> Bool
    static func makeListProvider<ModelType: Model>(
        modelType: ModelType.Type, decoder: Decoder) throws -> AnyModelListProvider<ModelType>
}
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListProvider.swift
lawmicha: removing the doc comment optimiscally and seeing, can always add it back if it doesn't, but it sounds like the expected behavior of Xcode

lawmicha:
Added iteration on the doc comment into separate readme: https://github.com/aws-amplify/amplify-ios/blob/ecb66ee966ab67a5118e2525aa84558800ecaa73/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
```diff
@@ -0,0 +1,93 @@
...
+- `AppSyncListDecoder` checks that the data can be successfully decoded to `AppSyncListResponse` and instantaites a loaded list of post and list of comments respectively in the chain of decoders. 
+
+This advanced use case is not without caveats on integrity of the subsequent APIs calls performed from the list object. For example, `hasNextPage` relies on the selection set to contain `nextToken`, so if this is excluded from the selection set, then `hasNextPage` will always return false. `getNextPage` does not retrieve the next page according to the associated parent since association data was never added to the list provider. By providing this successful decoding flow `AppSyncListResponse`, the data represents more of a snapshot, and assumes that the developer understand what they are trying to achieve with the customization. Alternatively, developers can go to the full extent of modifying the response type as well to `AppSyncListResponse` to decode exactly what the AppSync service returns in the response.
```
palpatim, line 93:
```suggestion
This is an advanced use case, and has some caveats regarding the subsequent API calls performed from the `List`:
- `hasNextPage()` relies on the selection set to contain `nextToken`, so if this is excluded from the selection set, then `hasNextPage()` will always return `false`.
- `getNextPage(completion:)` does not retrieve the next page according to the associated parent since association data was never added to the `List` provider.

In this flow the returned data represents more of a snapshot, and assumes that the developer understands what they are trying to achieve with the customization.

Alternately, developers can go to the full extent of modifying the response type as well to `AppSyncListResponse` to control exactly what the AppSync service returns in the response.
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListDecoder.swift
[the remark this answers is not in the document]
palpatim: Yeah, I just missed that one :)

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
```diff
@@ -0,0 +1,93 @@
...
+- `GraphQLResponseDecoder` will decode the "post" response post data to a `JSONValue` object as an intermediate step to allow modifications to the data.
+- The object is analyzed according to the model schema, association data (post.id and "post" field name) is stored at "comments" by creating and storing `AppSyncModelMetadata` at the "comments" key.
+- The object is then serialized and then decoded to the post instance. The `Post` decoder will instantiate the fields as normally while the "comments" field will be decoded by the `List<Comment>` that delegates its logic to the `AppSyncModelListDecoder`. 
+- `AppSyncListDecoder` checks that the data can be successfully decoded to an `AppSyncModelMetadata` and stores this information in an `AppSyncListProvider` when instantiating a "not loaded" list with association data.
```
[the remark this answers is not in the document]
lawmicha: woops, yes, `AppSyncModelListDecoder ` is a type and should be `AppSyncListDecoder` at line 54 above this line

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
```diff
@@ -0,0 +1,93 @@
...
+The developer has the post instance and can lazy load the comments upon accessing it.
+```swift
+foreach comment in post.comments {
+   /// comments is loaded before the first element is returned.
+}
+```
+- The comments are implicitly loaded upon access by taking the association data and performing a query to get the list of comments by its post id. Storing the result in the list object, and returning that to the caller that is performing the access.
```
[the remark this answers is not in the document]
lawmicha: it's was decoded from one of the types, in this case `AppSyncModelMetadata`, before making the list provider, so at this point of the flow it is an `AppSyncListProvider` that is holding the association data

AmplifyPlugins/API/AWSAPICategoryPlugin/Core/AppSyncListDecoder.swift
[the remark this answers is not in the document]
lawmicha: updated!

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
```diff
@@ -0,0 +1,93 @@
...
+
+1. As a developer, I want to retrieve a simple model object. 
+```swift
+Amplify.API.query(.get(SimpleModel.self, byId: id)
+
+// A Simple Model
+struct SimpleModel: Model {
+  let id: String
+}
+```
```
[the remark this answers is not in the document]
lawmicha: thanks, i've applied this else where

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
```diff
@@ -0,0 +1,93 @@
...
+- `GraphQLResponseDecoder` will decode the "post" response post data to a `JSONValue` object as an intermediate step to allow modifications to the data.
+- The object is analyzed according to the model schema, association data (post.id and "post" field name) is stored at "comments" by creating and storing `AppSyncModelMetadata` at the "comments" key.
+- The object is then serialized and then decoded to the post instance. The `Post` decoder will instantiate the fields as normally while the "comments" field will be decoded by the `List<Comment>` that delegates its logic to the `AppSyncModelListDecoder`. 
+- `AppSyncListDecoder` checks that the data can be successfully decoded to an `AppSyncModelMetadata` and stores this information in an `AppSyncListProvider` when instantiating a "not loaded" list with association data.
+The developer has the post instance and can lazy load the comments upon accessing it.
```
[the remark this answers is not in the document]
lawmicha: added the `fetch` api call example below this line

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
```diff
@@ -0,0 +1,93 @@
...
+- Upon getting a successful response from the service, the data portion of the response which represents the model instance, is extracted by `GraphQLResponseDecoder`.
+- The model instance data is decoded to the response type. As an end result, the instance of the model is returned.
+
+2. As a developer, I want to retrieve a list of simple model objects.
+```swift
+Amplify.API.query(.paginatedList(SimpleModel.self)
+```
+- `.paginatedList(SimpleModel.self)` will create a `GraphQLRequest` with selection set containing "items" and "nextToken", response type `List<SimpleModel>`, and variables containing the `limit` of 1000.
+- `GraphQLResponseDecoder` will detect that the response type is a `List` by checking against `ModelListMarker`, before encapsulating the original request and response using an `AppSyncListPayload`, and decoding the `AppSyncListPayload` to the list type.
+- `List`'s custom decoder will use the `ModelListDecoderRegistry` to find a registered decoder and attempt to decode the response to retrieve an `ModelListProvider`.
```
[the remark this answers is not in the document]
lawmicha: missed replying to this comment, but i have tried to reword the entire section here, like so:

- `GraphQLResponseDecoder` checks if the response type conforms to `ModelListMarker`. If so, it encapsulates the original request and response using an `AppSyncListPayload`, and decodes the `AppSyncListPayload` to the `List` type.
- `List`'s custom decoder logic will detect the `AppSyncListPayload` by first finding registered decoders in the `ModelListDecoderRegistry`. A registered decoder is used to decode the response into a `ModelListProvider` before instantiating the list instance.
- The `AppSyncListDecoder` is returned from the `ModelListDecoderRegistry`. At config time, the `AWSAPIPlugin` registers `AppSyncListDecoder` with `ModelListDecoderRegistry` to provide runtime decoding functionality. 
- `AppSyncListDecoder` checks that the data can be successfully decoded to an `AppSyncListPayload`, extracts the original request variables and response, vends a `AppSyncListProvider`, and instanatites a `List` with the list provider.

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
```diff
@@ -1,68 +1,181 @@
...
+APIPlugin at its core provides functionality to make requests to a GraphQL service. There are key developer use cases that can be achieved when using Amplify's `Model` types generated from Amplify CLI after provisioning a GraphQL AppSync service through the `amplify add api` command. Use-case centric APIs (`Amplify.API.query`, `Amplify.API.mutate`, and `Amplify.API.subscribe`), coupled with the `GraphQLRequest` builders (`.get`, `.create`, etc), to provide a simple way to perform operations on a model or to retrieve instances of the model. The following goes over the algorithms used to perform successful operations for retrieving a model, a model with associations, and a list of models.
```
palpatim, line 3:
```suggestion
APIPlugin at its core provides functionality to make requests to a GraphQL service. There are key developer use cases that can be achieved when using Amplify's `Model` types generated from Amplify CLI after provisioning a GraphQL AppSync service through the `amplify add api` command. Use-case centric APIs (`Amplify.API.query`, `Amplify.API.mutate`, and `Amplify.API.subscribe`), coupled with the `GraphQLRequest` builders (`.get`, `.create`, etc), provide a simple way to perform operations on a model or to retrieve instances of the model. The following goes over the algorithms used to perform successful operations for retrieving a model, a model with associations, and a list of models.
```

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
```diff
@@ -1,68 +1,181 @@
...
-- The model instance data is decoded to the response type. As an end result, the instance of the model is returned.
+
+**App code**
+
+```swift
+Amplify.API.query(request: .get(SimpleModel.self, byId: id))
+```
+- A `GraphQLRequest` is created from `.get(SimpleModel.self, byId: id)` containing the request payload such as the document, variables, and `SimpleModel` response type. 
+
+**GraphQL document**
```
palpatim, line 30: "GraphQL document generated by ..." to make it clear what's doing the work

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
```diff
@@ -1,68 +1,181 @@
...
+**First level of results**
+```
+query getPost($input: GetPostInput!) {
+  getPost(input: $input) {
+    id
+  }
+}
+```
+
+**Second level of results**
```
palpatim, line 122: Nice!

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
```diff
@@ -1,68 +1,181 @@
...
+- `GraphQLResponseDecoder` serializes object and decodes it to a `Post` instance as described above. The `Post` model-specific decoder instantiates scalar fields normally, while decoding the `comments` field into a `List<Comment>` that delegates its logic to a `AppSyncListDecoder`.
 - `AppSyncListDecoder` checks that the data can be successfully decoded to an `AppSyncModelMetadata` and stores this information in an `AppSyncListProvider` when instantiating a "not loaded" list with association data.
-The developer has the post instance and can lazy load the comments upon accessing it.
+
+The developer now has the `Post` instance and can either explicitly load the comments or lazy load the comments upon accessing iterator methods
+
+**Explicit load** 
+```swift
+if let comments = post.comments {
+  comments.fetch { fetchResults in
```
palpatim, line 160: very much a nit, but `fetch` is a Void return type right, where success just indicates that the results are now available in the `List`? If so, let's rename this `fetchResult`, and change the "print" statement below to something like `"fetch completed, list data is now loaded into memory"` or similar.

If in fact the `fetchResults` include data, do something with them in the example to make that clear.

palpatim (approved):
Approved w/comments

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.md
lawmicha: yup this returns void to avoid the customer confusion, will update to `result` since this is a Result type and update print statement

PR #1013, opened 2021-01-11
fix(DataStore): support for QueryPredicateConstant.all for graphql queries

wooj2:
Customers will experience a crash in the event they define a sync query with the `.all` constant.  For example:
```
  let syncExpression1 = DataStoreSyncExpression.syncExpression(Todo.schema, where: {
      return QueryPredicateConstant.all
  }
````

This code change remedies this by using `[:]` as the filter value -- effectively disabling the selective sync expression.

### DIFF
--- a/AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
@@ -102,7 +102,7 @@ extension QueryPredicate {
 
 extension QueryPredicateConstant: GraphQLFilterConvertible {
     func graphQLFilter(for modelSchema: ModelSchema?) -> GraphQLFilter {
-        if (self == .all) {
+        if self == .all {
             return [:]
         }
         preconditionFailure("Could not find QueryPredicateConstant \(self)")
--- a/AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
@@ -77,6 +77,8 @@ extension QueryPredicate {
             return operation.graphQLFilter(for: modelSchema)
         } else if let group = self as? QueryPredicateGroup {
             return group.graphQLFilter(for: modelSchema)
+        } else if let constant = self as? QueryPredicateConstant {
+            return constant.graphQLFilter(for: modelSchema)
         }
 
         preconditionFailure(
@@ -95,6 +97,15 @@ extension QueryPredicate {
 
         preconditionFailure(
             "Could not find QueryPredicateOperation or QueryPredicateGroup for \(String(describing: self))")
+    }
+}
+
+extension QueryPredicateConstant: GraphQLFilterConvertible {
+    func graphQLFilter(for modelSchema: ModelSchema?) -> GraphQLFilter {
+        if (self == .all) {
+            return [:]
+        }
+        preconditionFailure("Could not find QueryPredicateConstant \(self)")
     }
 }
 
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOperationTests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOperationTests.swift
@@ -739,8 +739,8 @@ class InitialSyncOperationTests: XCTestCase {
         let apiWasQueried = expectation(description: "API was queried with sync expression")
         let responder = QueryRequestListenerResponder<PaginatedList<AnyModel>> { request, listener in
             XCTAssertEqual(request.document, """
-            query SyncMockSynceds($filter: ModelMockSyncedFilterInput, $limit: Int) {
-              syncMockSynceds(filter: $filter, limit: $limit) {
+            query SyncMockSynceds($limit: Int) {
+              syncMockSynceds(limit: $limit) {
                 items {
                   id
                   __typename
@@ -753,11 +753,7 @@ class InitialSyncOperationTests: XCTestCase {
               }
             }
             """)
-            guard let filter = request.variables?["filter"] as? [String: Any?] else {
-                XCTFail("Unable to get filter")
-                return nil
-            }
-            XCTAssert(filter.isEmpty)
+            XCTAssertNil(request.variables?["filter"])
 
             let list = PaginatedList<AnyModel>(items: [], nextToken: nil, startedAt: nil)
             let event: GraphQLOperation<PaginatedList<AnyModel>>.OperationResult = .success(.success(list))
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOperationTests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOperationTests.swift
@@ -731,4 +731,80 @@ class InitialSyncOperationTests: XCTestCase {
         waitForExpectations(timeout: 1)
         sink.cancel()
     }
+
+    func testBaseQueryWithSyncExpressionConstantAll() throws {
+        let storageAdapter = try SQLiteStorageEngineAdapter(connection: Connection(.inMemory))
+        try storageAdapter.setUp(modelSchemas: StorageEngine.systemModelSchemas + [MockSynced.schema])
+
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by lawmicha** [comment]
are there other ways for the developer to get into this state?
1. Calling API directly and passing in the `.all` predicate- what is the expected behavior? `[:]` filter?
2. Calling DataStore save with a condition, does this get synced using API? 
3. and i believe the syncExpression is the scenario that you have mentioned, if they configure DataStore with a selective sync, it should would have crashed without this PR

**Comment by Kilo-Loco** [comment]
> are there other ways for the developer to get into this state?
> 
> 1. Calling API directly and passing in the `.all` predicate- what is the expected behavior? `[:]` filter?
> 2. Calling DataStore save with a condition, does this get synced using API?  is this even a valid use case?
> 3. Calling DataStore delete with a condition that is `.all` ? This is the other usage of `.all` right?
> 4. and i believe the syncExpression is the scenario that you have mentioned, if they configure DataStore with a selective sync, it would have crashed the app without this PR. Is it common to Selective sync with `.all` ? wouldn't they just not pass anything in to sync everything?

The requirement to return `QueryPredicateConstant.all` is to be able to write logic inside of the resolver block that checks against a `nil` value. My use case is as follows:

```swift
let syncExpression = DataStoreSyncExpression.syncExpression(TVSeries.schema) {
    let rating = UserDefaults.standard.integer(forKey: "preferredRating")
    guard rating > 0 else {
        return QueryPredicateConstant.all
    }
    return TVSeries.keys.rating >= 4
}
let config = DataStoreConfiguration.custom(syncExpressions: [syncExpression])
```

There's also an example of this kind of use case found in [Android DataStore: Sync data to cloud](https://docs.amplify.aws/lib/datastore/sync/q/platform/android#reevaluate-expressions-at-runtime) docs

wooj2:
> Calling API directly and passing in the .all predicate- what is the expected behavior? [:] filter?

I believe passing in `[:]` is the correct behavior.  We are effectively saying include "all", which i believe translates to not having a filter.


> * Calling DataStore save with a condition, does this get synced using API?  is this even a valid use case?

In the case of `save`, using the `.all` constant actually crashes because we are calling into the deprecated version of toJSON(_:options) (without the schema).  We should update our code so that we are no longer calling the deprecated methods.  I can do this, but prefer not to do that as part of this PR.

> Calling DataStore delete with a condition that is .all ? This is the other usage of .all right?

Delete with predicate on a model type is currently not exposed to customer.  we expose `delete(_ model: M, modelSchema:where:completion)`, but completely ignore the predicate we pass in.

> and i believe the syncExpression is the scenario that you have mentioned, if they configure DataStore with a selective sync, it would have crashed the app without this PR. Is it common to Selective sync with .all ? wouldn't they just not pass anything in to sync everything?

See @Kilo-Loco 's response

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -98,6 +100,15 @@ extension QueryPredicate {
     }
 }
 
+extension QueryPredicateConstant: GraphQLFilterConvertible {
+    func graphQLFilter(for modelSchema: ModelSchema?) -> GraphQLFilter {
+        if (self == .all) {
+            return [:]
```
lawmicha, line 106: have you made sure this works at the integration level? a request that makes it to AppSync service with the GraphQLRequest, 
document: 
```
query SyncMockSynceds($filter: ModelMockSyncedFilterInput, $limit: Int) {
              syncMockSynceds(filter: $filter, limit: $limit) {
                items {
                  id
                  __typename
                  _version
                  _deleted
                  _lastChangedAt
                }
                nextToken
                startedAt
              }
            }
```

what actually ends up being in the `variables` object?

lawmicha: when i try on the AWS Console on a simplifed example

### docment with filter parameter, and no "filter" variables
document
```
query MyQuery($filter: ModelCommentFilterInput) {
  listComments(filter: $filter, limit: 10) {
    items {
      content
    }
  }
}

```
variables is empty

```
```

This works and response is successful

### document with filter parameter, and "filter" variables
```
query MyQuery($filter: ModelCommentFilterInput) {
  listComments(filter: $filter, limit: 10) {
    items {
      content
    }
  }
}

```
variables is empty object: `{}`

```
{
  "filter": {}
}
```

response:
```
{
  "data": {
    "listComments": null
  },
  "errors": [
    {
      "path": [
        "listComments"
      ],
      "data": null,
      "errorType": "DynamoDB:DynamoDbException",
      "errorInfo": null,
      "locations": [
        {
          "line": 2,
          "column": 3,
          "sourceName": null
        }
      ],
      "message": "ExpressionAttributeNames must not be empty (Service: DynamoDb, Status Code: 400, Request ID: 8P5G95PBK3R7GTA1MLHJ5VKFMNVV4KQNSO5AEMVJF66Q9ASUAAJG, Extended Request ID: null)"
    }
  ]
}
```

lawmicha:
> In the case of `save`, using the `.all` constant actually crashes because we are calling into the deprecated version of toJSON(_:options) (without the schema). We should update our code so that we are no longer calling the deprecated methods. I can do this, but prefer not to do that as part of this PR.

nice catch, opened https://github.com/aws-amplify/amplify-ios/issues/1014

lawmicha:
> Delete with predicate on a model type is currently not exposed to customer. we expose `delete(_ model: M, modelSchema:where:completion)`, but completely ignore the predicate we pass in.

This seems like we should be crashing the app during development time if `predicate` is passed in for delete call. Thanks for checking, opening issue https://github.com/aws-amplify/amplify-ios/issues/1015

lawmicha:
Thanks for clarifying @Kilo-Loco @wooj2

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
lawmicha: So i'm guessing the filter object `[:]` gets added to variables, which gets added to the request payload as an object that looks like this:
```
let queryDocument = 
[ 
 "document": "some graphql document that references filter input",
 "variables": [
    "filter": [:]
  ]
]
```
and [serialized here](https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLOperation.swift#L81) using JSONSerialization, that probably looks like this as the JSON string?
```
{ 
 "document": "some graphql document that references filter input",
 "variables": {
    "filter": {}
  }
}
```
is this what's happening? can we verify this use case is working in an integration test?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
```diff
@@ -98,6 +100,15 @@ extension QueryPredicate {
     }
 }
 
+extension QueryPredicateConstant: GraphQLFilterConvertible {
+    func graphQLFilter(for modelSchema: ModelSchema?) -> GraphQLFilter {
+        if self == .all {
+            return [:]
```
lawmicha, line 106: just thinking whether the return value should just be `nil` and the method return `GraphQLFilter?` type. I think this is keeping it `GraphQLFilter` is fine since FilterDecorator should be checking if the filter is empty anyways, just to be safe.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOperationTests.swift
```diff
@@ -731,4 +731,80 @@ class InitialSyncOperationTests: XCTestCase {
...
+                nextToken
+                startedAt
+              }
+            }
+            """)
+            guard let filter = request.variables?["filter"] as? [String: Any?] else {
+                XCTFail("Unable to get filter")
+                return nil
+            }
+            XCTAssert(filter.isEmpty)
```
lawmicha, line 760: i don't think this is the assert scenario anymore, since the filter doesn't get added, so `request.variables["filter"]` should be nil

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/QueryPredicate+GraphQL.swift
wooj2: Hmm I think it would be useful to change if we had a use case of detecting nil vs an empty GraphQLFilter.  In this case, of QueryPredicateConstant.all, i think that returning [:] is an accurate representation of `.all` -- that is, there is no filter -- so at this point, i'm not seeing a strong case for changing this to an optional.

lawmicha (approved):

PR #1018, opened 2021-01-14
fix(DataStore): Remove AWSDataStoreCategoryPlugin.Cancellable in favor of Amplify.Cancellable

### LINKED ISSUE #328: Remove AWSDataStoreCategoryPlugin.Cancellable in favor of Amplify.Cancellable
This PR:
https://github.com/aws-amplify/amplify-ios/pull/323/files

Introduces a new Cancellable Protocol.  We should dig into why we are unable to use Amplify.Cancellable in the DataStore Category.

dvanwinkle:
*Issue #, if available:*
Fixes #328 

*Description of changes:*
Removed `AWSDataStoreCategoryPlugin.Cancellable` and replaced it with `Amplify.Cancellable`. This isn't possible without explicitly using `import protocol` due to there being a class named `Amplify` inside the module `Amplify` ***and*** `Cancellable` having a name collision with `Amplify.Cancellable` and `Combine.Cancellable`.

Alternative solutions, although likely with worse consequences, would be to rename `Amplify.Amplify`, rename `Amplify.Cancellable`, or rename the Swift module `Amplify`.

Of note, this may have exposed an error with SwiftLint as well. If you do not disable the check for `duplicate_imports`, SwiftLint complains with a `duplicate_imports` warning. If you have `duplicate_imports` disabled, SwiftLint complains that it is not needed. I erred on the side of caution knowing this bug could be fixed in the future and disabled `duplicate_imports`.


dvanwinkle:
@aws-amplify/amplify-native Sorry for not adding you as a reviewer. Please review.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/MutationEvent/MutationEventPublisher.swift
```diff
@@ -8,6 +8,8 @@
 import Amplify
 import Combine
 
+import protocol Amplify.Cancellable // swiftlint:disable:this duplicate_imports
+
 /// Publishes mutation events to downstream subscribers for subsequent sync to the API.
 @available(iOS 13.0, *)
 protocol MutationEventPublisher: class, Cancellable {
```
lawmicha, line 15: does it work if we just use `Amplify.Cancellable` here?
```swift
protocol MutationEventPublisher: class, Amplify.Cancellable {
```

or a public typealias as `AmplifyCancellable`?

Was reading @drochetti's comment https://github.com/aws-amplify/amplify-ios/issues/328#issuecomment-636173785

dvanwinkle: You cannot do `Amplify.Cancellable` due to the Amplify module having a class Amplify. You can do the typealias, but you'd have to do it in a file where you still need to `import protocol Amplify.Cancellable`

lawmicha: Thanks for explaining, for the typealias I meant `AmplifyCancellable` would be over in Amplify target, then over here in the plugin, you would just replace `Cancellable` with AmplifyCancellable` since there wouldn't be any conflicts and the existing `import Amplify` should pull in `AmplifyCancellable`. I'm assuming if the file is over in Amplify, you wouldn't need to `import protocol Amplify.Cancellable`, right?

dvanwinkle: Yes, that would work. It would probably make more sense to change the protocol name if you put it in the Amplify module. This problem really just exists because of the double name collision.

dvanwinkle: @lawmicha @drochetti Do either of you have a preferred direction?

lawmicha: what was the previous and current developer's experience when dealing with `DataStore.Cancellable` and with this change to return `Amplify.Cancellable`, is there any impact to the developers?

For example, if app code contains an instance property of the DataStore.Cancellable, now that we are removing the duplicate type, do they have to change anything such as
- an `import protocol` in their code to avoid the conflict with Combine's Cancellable? 
- If developers had code which was `var cancellable: DataStore.Cancellable?` then this would be breaking, right?

dvanwinkle: @lawmicha Good point. That would be a breaking change as it's already been released. External users would only need to do an `import protocol` if they are importing both Amplify and Combine into their swift file, which would be rare.

The discussion in #328 with @wooj2 and @drochetti seems to point to adding a typealias in the DataStore project, but I'm okay with any direction.

Let me know the direction that you'd like me to take this and I can make all of the necessary changes.

lawmicha: i think we should hold this until we release a major version since developer has to change their code when they upgrade and additionally address developer experience if there is a pain point or if this change introduces the pain point. Using `import protocol` in our plugin code looks perfectly fine, but seems a bit unintuitive for developer apps. Having both Amplify and Combine sounds kind of common scenario to me if they're writing a SwiftUI app so they pull in Combine to do debouncing on their views, while pulling in Amplify to access the category APis. The direction I vote for would be to finalize the decision on one of the options that @wooj2 listed in the issue with the team in regards to some developer experience

Developer experience after removing DataStore's Cancellable (hand wrote quickly- may be incorrect)
```swift
import Amplify
import AWSDataStorePlugin
import Combine

import protocol Amplify.Cancellable

class MyClass {
  let cancellable: Cancellable? // Without the `import protocol`, fails with Error?
  let combineCancellable: AnyCancellable?  
}
```

With typealias
```swift
mport Amplify
import AWSDataStorePlugin
import Combine

class MyClass {
  let cancellable: AmplifyCancellable?
  let combineCancellable: AnyCancellable?  
}
```

dvanwinkle: That makes sense. Would you like me to close this PR and rewrite it using AmplifyCancellable so it's ready for the next release?

lawmicha: Thanks for the help and discussion! I believe @brennanMKE took a crack at this issue and a similar approach was added over in https://github.com/aws-amplify/amplify-ios/pull/1044 Since that PR is merged, I think we're good to close off this PR

PR #1030, opened 2021-01-20
fix(datastore): delete model with predicate

lawmicha:
*Issue #, if available:*

*Description of changes:*

This PR adds logic to use the condition passed into `delete(model:where:)` and `delete(withId:where:)`. 
 
This is an existing API, which did not use the passed in `predicate` value.
```swift
public func delete<M: Model>(_ model: M,
                                 modelSchema: ModelSchema,
                                 completion: @escaping DataStoreCallback<Void>) 
```

Updated API, no changes to developer call site
```swift
                                 withId id: String,
                                 where predicate: QueryPredicate? = nil, // added optional `where`
                                 completion: @escaping DataStoreCallback<Void>)
```

Update API,  no changes to developer call site
```swift
public func delete<M: Model>(
        _ modelType: M.Type,
        withId id: String,
        where predicate: QueryPredicate? = nil // added optional `where`
    ) -> DataStorePublisher<Void> 
```

Because of the existing API which exposes the `where` parameter, i've opt to treat this as a fix rather than a feature. However, we are increasing the surface area for the latter two APIs, so we could also treat this as a feature.

### Delete behaviors

1. Deleting a model that does not exist returns successful
2. Deleting a model that exists returns successful
3. Deleting a model with a condition and the model does not exist, returns successful
4. Deleting a model with a condition and the model exists, and the condition does not match, returns invalid condition failure
5. Deleting a model with a condition and the mdoel exists, and condition matches the model, returns successful.


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
```diff
@@ -11,9 +11,43 @@ import Foundation
 import AWSPluginsCore
 
 extension StorageEngine {
+
+    enum DeleteInput {
+        case withId(id: Model.Identifier, predicate: QueryPredicate?)
+        case withPredicate(predicate: QueryPredicate)
```
diegocstn, line 17: would having three separates cases `withId(id:)`, `withIdAndPredicate(id:predicate:)`, `withPredicate(predicate:)` and a static factory method that returns the proper case simplify the `predicate` property and make the `isDeleteByIdWithPredicate` property unnecessary?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
```diff
@@ -11,9 +11,43 @@ import Foundation
...
+                } else {
+                    return field("id").eq(id)
+                }
+            case .withPredicate(let predicate):
+                return predicate
+            }
+        }
+
+        /// Returns the `id` if the delete input is by `id` and contains `predicate`, return `nil` otherwise
+        var isDeleteByIdWithPredicate: Model.Identifier? {
```
diegocstn, line 34: nit: the property name sounds like an assertion but doesn't return a boolean maybe we can just call it `id` or `modelIdentifier`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
lawmicha: i tried this out and it solves the original problem much better! thanks, i've updated to the 3 cases and removed `isDeleteByIdWithPredicate` since i can just check the case `withIdAndPredicate`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
lawmicha: no longer needed due to having the 3 cases

diegocstn (approved):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
diegocstn: thanks for the change!

PR #1037, opened 2021-01-25
chore(auth): Updated the signInWithWebUI to use ASWebAuthenticationSession

royjit:
*Issue #, if available:* https://github.com/aws-amplify/aws-sdk-ios/issues/3141

*Description of changes:*

- HostedUI signIn defaults to `ASWebAuthenticationSession` in iOS 13.0+
- Added a new parameter in `AWSAuthWebUISignInOptions` called `preferPrivateSession` which will start the hostedUI in `prefersEphemeralWebBrowserSession` of `ASWebAuthenticationSession`.

_*Developer Experience*_

1. There is no change in the signIn or signOut api for existing developers if they do not need to add the new feature of preferPrivateSession.
2. Existing developer (including Flutter) need to update the userCancelled error if they are dependent on it, explained here - https://github.com/aws-amplify/amplify-ios/pull/982
3. New feature is only available for iOS >= 13.0, so if the app runs in a iOS version lesser than 13.0, the `preferPrivateSession` will not have any effect and the signOut will see the popup.
4. PrivateSession is guaranteed to work in Safari, and might not work if the default browser of the user is different, developer should be aware of this. 
5. Developer code after the new properties are added:

```swift
// Sign in With private session
let pluginOption = AWSAuthWebUISignInOptions(preferPrivateSession: true)
let webUISignInoptions = AuthWebUISignInRequest.Options(pluginOptions: pluginOption)
Amplify.Auth.signInWithWebUI(presentationAnchor: self.view.window!, 
                             options: webUISignInoptions) { ... }

// Using the static builder options to signin with private session
Amplify.Auth.signInWithWebUI(presentationAnchor: self.view.window!, 
                             options: .preferPrivateSession()) { ... }


// Sign in Without private session
Amplify.Auth.signInWithWebUI(presentationAnchor: self.view.window!) { ... }

```

### DIFF
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
@@ -49,7 +49,7 @@ extension AuthenticationProviderAdapter {
                 completionHandler(.failure(error))
                 return
             }
-            AWSCognitoAuthPluginUserDefaults.storePreferredBrowserSession(privateSessionPrefered: false)
+            self.userdefaults.storePreferredBrowserSession(privateSessionPrefered: false)
             let authResult = AuthSignInResult(nextStep: signInNextStep)
             completionHandler(.success(authResult))
         }
@@ -205,7 +205,7 @@ extension AuthenticationProviderAdapter {
             completionHandler(.failure(error))
             return
         }
-        AWSCognitoAuthPluginUserDefaults.storePreferredBrowserSession(privateSessionPrefered: preferPrivateSession)
+        userdefaults.storePreferredBrowserSession(privateSessionPrefered: preferPrivateSession)
         let authResult = AuthSignInResult(nextStep: .done)
         completionHandler(.success(authResult))
     }
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
@@ -8,10 +8,12 @@
 import Amplify
 import AWSMobileClient
 
+typealias SigInResultCompletion = (Result<AuthSignInResult, AuthError>) -> Void
+
 extension AuthenticationProviderAdapter {
 
     func signIn(request: AuthSignInRequest,
-                completionHandler: @escaping (Result<AuthSignInResult, AuthError>) -> Void) {
+                completionHandler: @escaping SigInResultCompletion) {
 
         // AuthSignInRequest.validate method should have already validated the username and the below line
         // is just to avoid optional unwrapping.
@@ -47,6 +49,7 @@ extension AuthenticationProviderAdapter {
                 completionHandler(.failure(error))
                 return
             }
+            AWSCognitoAuthPluginUserDefaults.storePreferredBrowserSession(privateSessionPrefered: false)
             let authResult = AuthSignInResult(nextStep: signInNextStep)
             completionHandler(.success(authResult))
         }
@@ -54,7 +57,7 @@ extension AuthenticationProviderAdapter {
     }
 
     func signInWithWebUI(request: AuthWebUISignInRequest,
-                         completionHandler: @escaping (Result<AuthSignInResult, AuthError>) -> Void) {
+                         completionHandler: @escaping SigInResultCompletion) {
 
         let window = request.presentationAnchor
         DispatchQueue.main.async { [weak self] in
@@ -67,7 +70,7 @@ extension AuthenticationProviderAdapter {
     }
 
     func confirmSignIn(request: AuthConfirmSignInRequest,
-                       completionHandler: @escaping (Result<AuthSignInResult, AuthError>) -> Void) {
+                       completionHandler: @escaping SigInResultCompletion) {
 
         let userAttributes = (request.options.pluginOptions as? AWSAuthConfirmSignInOptions)?.userAttributes ?? []
         let mobileClientUserAttributes = userAttributes.reduce(into: [String: String]()) {
@@ -108,16 +111,17 @@ extension AuthenticationProviderAdapter {
     // MARK: - Internal methods
     private func showSignInWebView(window: UIWindow,
                                    request: AuthWebUISignInRequest,
-                                   completionHandler: @escaping (Result<AuthSignInResult, AuthError>) -> Void) {
+                                   completionHandler: @escaping SigInResultCompletion) {
 
         // Stop the execution here if we are not running on the main thread.
         // There is no point on returning an error back to the developer, because
         // they do not control how the UI is presented.
         dispatchPrecondition(condition: .onQueue(DispatchQueue.main))
 
-        let idpIdentifier = (request.options.pluginOptions as? AWSAuthWebUISignInOptions)?.idpIdentifier
-        let federationProviderName = (request.options.pluginOptions as? AWSAuthWebUISignInOptions)?
-            .federationProviderName
+        let pluginOptions = request.options.pluginOptions as? AWSAuthWebUISignInOptions
+        let idpIdentifier = pluginOptions?.idpIdentifier
+        let federationProviderName = pluginOptions?.federationProviderName
+        let preferPrivateSession = pluginOptions?.preferPrivateSession ?? false
         let hostedUIOptions = HostedUIOptions(disableFederation: false,
                                               scopes: request.options.scopes,
                                               identityProvider: request.authProvider?.toCognitoHostedUIString(),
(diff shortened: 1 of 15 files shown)


pedrofspinho (approved):

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignIn.swift
```diff
@@ -146,28 +167,49 @@ extension AuthenticationProviderAdapter {
...
+                                      completionHandler: @escaping SigInResultCompletion) {
+        if let error = error {
+            let authError = convertSignInUIErrorToAuthError(error)
+            completionHandler(.failure(authError))
+            return
+        }
+
+        guard let state = state, state == .signedIn else {
+
+            let error = AuthError.unknown("signInWithWebUI did not produce a valid result.")
```
palpatim, line 204: Would be helpful to include the state in the error message if present but not signed in

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Models/Options/AWSAuthWebUISignInOptions.swift
```diff
@@ -24,8 +25,28 @@ public struct AWSAuthWebUISignInOptions {
     /// are using Auth0, specify the `federationProviderName` as <your_domain>.auth0.com.
     public let federationProviderName: String?
 
-    public init(idpIdentifier: String? = nil, federationProviderName: String? = nil) {
+    /// Start the webUI signin in a private browser session.
```
palpatim, line 28:
```suggestion
    /// Starts the webUI signin in a private browser session, if supported by the current browser.
```

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Models/Options/AWSAuthWebUISignInOptions.swift
```diff
@@ -24,8 +25,28 @@ public struct AWSAuthWebUISignInOptions {
...
+    ///
+    /// Internally this sets the ASWebAuthentication session's prefersEphemeralWebBrowserSession = true.
+    /// Note that this value internally set `prefersEphemeralWebBrowserSession` in ASWebAuthenticationSession.
```
palpatim, line 31:
```suggestion
    /// Note that this value internally sets `prefersEphemeralWebBrowserSession` in ASWebAuthenticationSession.
```

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Models/Options/AWSAuthWebUISignInOptions.swift
```diff
@@ -24,8 +25,28 @@ public struct AWSAuthWebUISignInOptions {
...
+    }
+}
+
+extension AuthWebUISignInRequest {
+
+    public static func optionWithPrivateSession() -> AuthWebUISignInOperation.Request.Options {
+        let pluginOptions = AWSAuthWebUISignInOptions(preferPrivateSession: true)
+        let options = AuthWebUISignInOperation.Request.Options(pluginOptions: pluginOptions)
+        return options
     }
```
palpatim, line 51: This seems more appropriate as an extension on the options object
```suggestion
extension AWSAuthWebUISignInOptions {
    public static func preferPrivateSession() -> AuthWebUISignInOperation.Request.Options {
        let pluginOptions = AWSAuthWebUISignInOptions(preferPrivateSession: true)
        let options = AuthWebUISignInOperation.Request.Options(pluginOptions: pluginOptions)
        return options
    }
```

What does the call site look like using your code?

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Service/AWSMobileClient/AWSMobileClientAdapter.swift
```diff
@@ -100,6 +100,15 @@ class AWSMobileClientAdapter: AWSMobileClientBehavior {
                                    completionHandler)
     }
 
+    @available(iOS 13, *)
+    func showSignIn(uiwindow: UIWindow,
```
palpatim, line 104: Here and below:

Don't we already have a typealias like "AWSMobileClientPresentationContext" or similar for UIWindow? Aren't we requiring presentation context in other calls? If so, seems like we should make that consistent.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Service/AWSMobileClient/AWSMobileClientBehavior.swift
```diff
@@ -44,6 +44,11 @@ protocol AWSMobileClientBehavior {
                     hostedUIOptions: HostedUIOptions?,
                     _ completionHandler: @escaping (UserState?, Error?) -> Void)
 
+    @available(iOS 13, *)
```
palpatim, line 47: Same note here & below regarding presentation context param name & type alias

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Support/Utils/AWSCognitoAuthPluginUserDefaults.swift
```diff
@@ -0,0 +1,21 @@
...
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+struct AWSCognitoAuthPluginUserDefaults {
+
+    private static let preferPrivateSessionKey = "privateSessionKey"
```
palpatim, line 12: Prefer a more granular namespace: "AWSCognitoAuthPluginUserDefaults.privateSessionKey" or similar. While admittedly unlikely, "privateSessionKey" runs a higher risk of colliding with some other library or app logic.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Support/Utils/AWSCognitoAuthPluginUserDefaults.swift
```diff
@@ -0,0 +1,21 @@
...
+
+    static func storePreferredBrowserSession(privateSessionPrefered: Bool) {
+        UserDefaults.standard.setValue(privateSessionPrefered, forKey: preferPrivateSessionKey)
```
palpatim, line 15: Are we using UserDefaults anywhere else? It would be good to abstract this into a behavior so we could inject our own user defaults provider for testing

royjit: We are not using anywhere else in Auth currently. Moved this to a behavior.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Models/Options/AWSAuthWebUISignInOptions.swift
palpatim: The intent would be to provide a DX like:

```swift
Amplify.Auth.signInWithWebUI(presentationAnchor: window, options: .preferPrivateSession())
```

So that means you want the extension to be on `AWSAuthWebUISignInOptions`

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/Mocks/MockUserDefaults.swift
```diff
@@ -0,0 +1,32 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+@testable import AWSCognitoAuthPlugin
+
+struct MockUserDefaults: AWSCognitoAuthPluginUserDefaultsBehavior {
+
+    let userDefaults: UserDefaults
```
palpatim, line 13: Any reason not to just make this something like a dictionary? No particular reason you should be writing to disk for the unit test.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+SignOut.swift
```diff
@@ -12,6 +12,15 @@ extension AuthenticationProviderAdapter {
 
     func signOut(request: AuthSignOutRequest, completionHandler: @escaping (Result<Void, AuthError>) -> Void) {
 
+        // If developer had signed in using private session, we just need to signout the user locally.
+        guard userdefaults.isPrivateSessionPreferred() == false else {
```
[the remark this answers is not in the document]
royjit: Changed

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderSigninTests.swift
```diff
@@ -46,6 +46,8 @@ class AuthenticationProviderSigninTests: BaseAuthenticationProviderTest {
                     return
                 }
                 XCTAssertTrue(signinResult.isSignedIn, "Signin result should be complete")
+                XCTAssertFalse(self.mockUserDefault.isPrivateSessionPreferred(),
+                              "Prefer private session userdefaults should not be set.")
```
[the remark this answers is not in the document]
royjit: Fixed

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/Mocks/MockUserDefaults.swift
royjit: Changed to use NSDictionary inside user defaults mock.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Models/Options/AWSAuthWebUISignInOptions.swift
royjit: Yes, I updated the code.

palpatim (approved):

PR #1044, opened 2021-01-29
chore(DataStore): defines typealias for AmplifyCancellable

brennanMKE:
*Issue #, if available:*

#328

*Description of changes:*

This change is related to #328. It prevents a conflict with the `Combine` type for `Cancellable`. With this change `AmplifyCancellable` is a type alias of `Cancellable` from the `Amplify` module and it is public. It can now be used by any plugin and prevent the name conflict with `Combine`.


brennanMKE:
I originally defined the type alias in the DataStore plugin but realized this will be an issue with any module which also imports Combine, even consumers of Amplify. So I chose to define it in the core module to make it an option for everyone to prevent this name conflict.

Amplify/Core/Support/Cancellable.swift
```diff
@@ -10,3 +10,5 @@
 public protocol Cancellable {
     func cancel()
 }
+
+public typealias AmplifyCancellable = Cancellable
```
palpatim, line 14: Can you add a brief note explaining why we have this, and that it may be removable once we drop support for iOS < 13?

palpatim (commented):
LGTM with one small comment add. As discussed, you can consider editing the title to be `chore`, but it's not really meaningful. Please do add a colon after the tag though "refactor(DataStore): ..." -- conventional commits require a colon and space after the type and scope.

palpatim:
> LGTM with one small comment add. As discussed, you can consider editing the title to be `chore`, but it's not really meaningful. Please do add a colon after the tag though "refactor(DataStore): ..." -- conventional commits require a colon and space after the type and scope.

I should have also noted: it's not strictly necessary to change the PR title, but since it's pretty common to copy/paste the PR title as the commit message of the squash/merge, it's convenient to keep them in sync.

Amplify/Core/Support/Cancellable.swift
brennanMKE: I've added a comment to provide context for this addition.

palpatim (approved):

PR #950, opened 2020-12-07
amplify-xcode: foundation and import-config command

diegocstn:
This PR introduces a basic architecture implementation of a new tool aimed at improving Amplify Xcode integration.

### Structure
The Swift package has been broken down in two separate "components" with clear responsibilities to favor re-usability and keep the doors open to other type of integrations different from a CLI interface.

#### Core
The Core component adopts a command-based architecture, it's responsible of the business logic and completely decoupled from any interface.
An `AmplifyCommand` represents the largest unit of work (a command), constituted of a set of smaller and composable units of work of type `AmplifyCommandTask`.
Each task is a pure function receiving a `AmplifyCommandEnvironment` as argument in order to have access to file system and Xcode specific utilities. 

##### Commands:
- `CommandImportConfig` given an Amplify iOS project, adds a new "AmplifyConfig" group with necessary configuration files to the Xcode project

#### CLI
The CLI component exposes commands provided by `Core` to users through a command line interface.

#### BuildTools
To overcome SwiftPM tooling limitations, an additional Swift package (without source code) lives in this folder and it's being used by the main `AmplifyCLIiOS` package.

#### TODOs
- [x] public functions/classes/structs docs
- [x] unit tests
- [x] home dir `~` path resolution

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1055, opened 2021-02-05
amplify-xcode: Import models command

diegocstn:
#950 follow up PR.

### Description of changes:
- `import-models` command
- improved error messages
- re-enabled SwiftLint + SwiftFormat
- CLI exit code based on success/failure

### DIFF
--- a/AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLI.swift
+++ b/AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLI.swift
@@ -10,48 +10,9 @@ import ArgumentParser
 import AmplifyXcodeCore
 
 /// This module defines a CLI (Command Line Interface) to commands defined in `Core/Commands`.
-/// Each "CLI command" defined below is the actual executor of an `AmplifyCommand`, thus it's responsible
+/// Each "CLI command"  is the actual executor of an `AmplifyCommand`, thus it's responsible
 /// for providing an environment, instantiate and execute a command.
 /// The `CommandExecutable` protocol glues an `AmplifyCommand` and the environment provided by the executor.
-
-/// CLI command invoking `CommandImportConfig`.
-struct CLICommandImportConfig: ParsableCommand, CommandExecutable, CLICommandReportable {
-    static let configuration = CommandConfiguration(
-        commandName: "import-config",
-        abstract: CommandImportConfig.description
-    )
-
-    @Option(name: .shortAndLong, help: "Project base path")
-    private var path: String = Process().currentDirectoryPath
-
-    var environment: AmplifyCommandEnvironment {
-        CommandEnvironment(basePath: path, fileManager: FileManager.default)
-    }
-
-    func run() throws {
-        let output = exec(command: CommandImportConfig())
-        report(result: output)
-    }
-}
-
-struct CLICommandImportModels: ParsableCommand, CommandExecutable, CLICommandReportable {
-    public static let configuration = CommandConfiguration(
-        commandName: "import-models",
-        abstract: CommandImportModels.description
-    )
-
-    @Option(name: .shortAndLong, help: "Project base path")
-    private var path: String = Process().currentDirectoryPath
-
-    var environment: AmplifyCommandEnvironment {
-        CommandEnvironment(basePath: path, fileManager: FileManager.default)
-    }
-
-    func run() throws {
-        let output = exec(command: CommandImportModels())
-        report(result: output)
-    }
-}
 
 /// CLI interface entry point `amplify-xcode`
 struct AmplifyXcode: ParsableCommand {
--- a/AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLI.swift
+++ b/AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLI.swift
@@ -34,10 +34,28 @@ struct CLICommandImportConfig: ParsableCommand, CommandExecutable, CLICommandRep
     }
 }
 
+struct CLICommandImportModels: ParsableCommand, CommandExecutable, CLICommandReportable {
+    public static let configuration = CommandConfiguration(
+        commandName: "import-models",
+        abstract: CommandImportModels.description
+    )
+
+    @Option(name: .shortAndLong, help: "Project base path")
+    private var path: String = Process().currentDirectoryPath
+
+    var environment: AmplifyCommandEnvironment {
+        CommandEnvironment(basePath: path, fileManager: FileManager.default)
+    }
+
+    func run() throws {
+        let output = exec(command: CommandImportModels())
+        report(result: output)
+    }
+}
+
 /// CLI interface entry point `amplify-xcode`
 struct AmplifyXcode: ParsableCommand {
     static let configuration = CommandConfiguration(
         abstract: "Amplify Xcode CLI",
-        subcommands: [CLICommandImportConfig.self]
-    )
+        subcommands: [CLICommandImportConfig.self, CLICommandImportModels.self])
 }
--- a/AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLICommandReportable.swift
+++ b/AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLICommandReportable.swift
@@ -25,7 +25,7 @@ extension CLICommandReportable {
     }
 
     private func reportCommandFailure(_ error: AmplifyCommandError) {
-        print(error.debugDescription)
+        print("🚫 \(error.debugDescription)")
     }
 
     func report(result: AmplifyCommandResult) {
--- a/AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLICommandReportable.swift
+++ b/AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLICommandReportable.swift
@@ -32,9 +32,11 @@ extension CLICommandReportable {
         switch result {
         case .success(let intermediateRes):
             reportCommandSuccess(for: intermediateRes)
+            exit(EXIT_SUCCESS)
 
         case .failure(let error):
             reportCommandFailure(error)
+            exit(EXIT_FAILURE)
         }
     }
 }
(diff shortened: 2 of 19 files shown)


AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLI.swift
```diff
@@ -34,10 +34,28 @@ struct CLICommandImportConfig: ParsableCommand, CommandExecutable, CLICommandRep
     }
 }
 
+struct CLICommandImportModels: ParsableCommand, CommandExecutable, CLICommandReportable {
```
palpatim, line 37: Consider refactoring the CLICommands into separate files for modularity

AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLICommandReportable.swift
```diff
@@ -25,16 +25,18 @@ extension CLICommandReportable {
...
     private func reportCommandFailure(_ error: AmplifyCommandError) {
-        print(error.debugDescription)
+        print("🚫 \(error.debugDescription)")
     }
 
     func report(result: AmplifyCommandResult) {
         switch result {
         case .success(let intermediateRes):
             reportCommandSuccess(for: intermediateRes)
+            exit(EXIT_SUCCESS)
```
palpatim, line 35: Let's not put such a drastic side effect (exiting the program) inside a `report` method. Call `exit` explicitly at the call site where you're reporting, or less preferably, refactor into a `reportAndExit` method that does both.

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/AmplifyCommandError.swift
```diff
@@ -54,13 +56,18 @@ public struct AmplifyCommandError: Error {
 
 public extension AmplifyCommandError {
     var debugDescription: String {
-        var components = ["\(type): "]
+        let underlyingErrors = self.underlyingErrors ?? []
```
palpatim, line 59: Since you're going to the trouble of an early exit condition for 1 underlying error, you could probably early exit for 0 underlying errors too, maybe as a `guard let` here?
```suggestion
        guard let underlyingErrors = self.underlyingErrors else {
            return "whatever a 0 underlying error message should be"
        }
```

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportConfig.swift
```diff
@@ -17,7 +17,7 @@ enum ImportConfigTasks {
 
         return .failure(AmplifyCommandError(
                             .folderNotFound,
-                            error: nil,
+                            errorDescription: "We couldn't find an Amplify project at \(environment.basePath).",
                             recoverySuggestion: "Please run `amplify init` to initialize an Amplify project."))
```
palpatim, line 21: Avoid "we" in our error messages, prefer more terse constructions. Similarly, we avoid "please" in messages & constructions, and prefer being direct and prescriptive.
```suggestion
                            errorDescription: "Amplify project not found at \(environment.basePath).",
                            recoverySuggestion: "Run `amplify init` to initialize an Amplify project."))
```

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportModels.swift
```diff
@@ -0,0 +1,67 @@
...
+//
+
+import Foundation
+import PathKit
+
+enum CommandImportModelsTasks {
+    static func projectHasGeneratedModels(environment: AmplifyCommandEnvironment,
+                                          args: CommandImportModels.TaskArgs) -> AmplifyCommandTaskResult {
+        let modelsPath = environment.path(for: args.generatedModelsPath)
+        if environment.directoryExists(atPath: modelsPath) {
```
palpatim, line 15: nit: It's a bit of a stylistic thing, but I prefer to use `guard` conditions to indicate early exit conditions or non-happy-path flows. In this case, the happy path is that the directory exists, so the failure condition could be modeled as a `guard` with a `return .failure(...)`. As mentioned, it's a stylistic preference, but see what you think.

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportModels.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    public var tasks: [AmplifyCommandTask<CommandImportModelsArgs>] = [
+        .run(CommandImportModelsTasks.projectHasGeneratedModels),
+        .run(CommandImportModelsTasks.addGeneratedModelsToProject)
+    ]
+
+    public init() {}
+
+    public func onFailure() {
+    }
```
palpatim, line 66: I assume `onFailure` is a callback used to clean up resources if a command fails? If so, and it's relatively common to have a no-op failure like this, consider adding a default implementation on the protocol:

```swift
public extension AmplifyCommand {
    public func onFailure() {
        // Do nothing by default
    }
}
```

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportModels.swift
```diff
@@ -0,0 +1,67 @@
...
+    public static var description = "Import Amplify models"
+
+    public var taskArgs = CommandImportModelsArgs()
+
+    public var tasks: [AmplifyCommandTask<CommandImportModelsArgs>] = [
+        .run(CommandImportModelsTasks.projectHasGeneratedModels),
+        .run(CommandImportModelsTasks.addGeneratedModelsToProject)
+    ]
+
+    public init() {}
```
palpatim, line 63: Is the no-arg constructor a requirement of the protocol? Not sure why we have it here since structs get member-wise initializers for free.

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportModels.swift
```diff
@@ -0,0 +1,67 @@
...
+    public struct CommandImportModelsArgs {
+        let modelsGroup = "AmplifyModels"
+        let generatedModelsPath = "amplify/generated/models"
+    }
+
+    public typealias TaskArgs = CommandImportModelsArgs
+
+    public static var description = "Import Amplify models"
+
+    public var taskArgs = CommandImportModelsArgs()
```
palpatim, line 56: Is this truly variable?

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportModels.swift
```diff
@@ -0,0 +1,67 @@
...
+
+public struct CommandImportModels: AmplifyCommand {
+    public struct CommandImportModelsArgs {
+        let modelsGroup = "AmplifyModels"
+        let generatedModelsPath = "amplify/generated/models"
+    }
+
+    public typealias TaskArgs = CommandImportModelsArgs
+
+    public static var description = "Import Amplify models"
```
palpatim, line 54:
```suggestion
    public static let description = "Import Amplify models"
```

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportConfig.swift
diegocstn: Will do, thanks

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportModels.swift
diegocstn: Good call, thanks

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportModels.swift
diegocstn: Unfortunately member-wise initializer in a struct are being marked `internal` by design, so when used in a different module it's necessary to provide your own public initializer. There's an old Swift evolution proposal that also addresses this issue but got deferred.

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportModels.swift
diegocstn: no they're not (both `description` and `taskArgs`), thanks for catching that.

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/CommandImportModels.swift
diegocstn: I usually rely on `guard` mostly to indicate early exit conditions, but your style recommendation is a good call so I'll apply it.

AmplifyTools/AmplifyXcode/Sources/AmplifyXcode/CLICommandReportable.swift
diegocstn: Sure, I'll move it to the call site. I've also just found that `ArgumentParser` library exposes an `Exit.failure` error that can be thrown from the `run` method.

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/AmplifyCommandError.swift
```diff
@@ -12,7 +12,7 @@ public struct AmplifyCommandError: Error {
         case unknown
         case fileNotFound
         case folderNotFound
-        case xcodeProject
+        case xcodeProject /// @see `XcodeProjectError`
```
palpatim, line 15: nit: style: we avoid trailing comments. Also, does a trailing doc-comment show up in the IDE associated with the correct member?

palpatim (approved):
formatting nit otherwise LGTM

AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Commands/AmplifyCommandError.swift
diegocstn: No, it doesn't properly show up. Removed :)

Issue #977, opened 2020-12-17
API request signature is generated prior to URL encoding


edwinveger:
**Describe the bug**
It appears request signing takes place prior to URL encoding. This fails the request if there are query parameters contains characters that need to be encoded, e.g. `@`.

**To Reproduce**
We have a GET endpoint that require a query parameter (`/my/endpoint/path/:id?email=my@email.address.com`).
We create a RESTRequest like so:
```swift
let request = RESTRequest(
    path: "/my/endpoint/path/\(id)",
    queryParameters: ["email": "my@email.address.com"]
)

Amplify.API.get(request: request) { 
    // etc.
}
```

This request fails with a signing error.
Status code: 403
Body:
```json
{
  "error": "The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.\n\nThe Canonical String for this request should have been\n'GET\n/my/endpoint/path/:id\nemail=my%40email.address.com\ncontent-type:application/json\nhost:REDACTED\nuser-agent:amplify-iOS/(Function) iOS/14.3 en\nx-amz-date:20201217T105627Z\nx-amz-security-token:REDACTED\n\ncontent-type;host;user-agent;x-amz-date;x-amz-security-token\REDACTED'\n\nThe String-to-Sign should have been\n'AWS4-HMAC-SHA256\REDACTED\n20201217/us-east-1/execute-api/aws4_request\REDACTED'\n"
}
```

**Expected behavior**
A successful request (i.e. not receiving a signing error).

**Environment:**
 - Amplify Framework Version: 1.4.3 (also reproducible on 1.5.1)
 - Dependency Manager: Cocoapods
 - Swift Version : 5.3.2

**Device Information:**
 - Device: Simulator
 - iOS Version: iOS 14.3

edwinveger:
Although we thought up a workaround for above particular case, this issue has surfaced in another URL request. In this case we pass a timestamp (e.g. `2021-06-18T09:00:00Z`) in which the `:` characters are encoded (as `%3A0`).

This is actively blocking our development. It seems like a very common use case, and I must say I am quite confused that this issue has not been reported/addressed earlier.

If there is a simple workaround that I have missed so far, please provide directions.

diegocstn:
PR in review: https://github.com/aws-amplify/amplify-ios/pull/1068

palpatim:
Removed "closing soon" label pending PR approval & release.

PR #1068, opened 2021-02-14
feat(api): fix querystring encoding according to AWS SigV4

diegocstn:
*Issue #977*

*Description of changes:*
This PR fixes an issue where querystring  parameters aren't correctly encoded according per [AWS reference guide ](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html)
`URLComponents` encodes queryItems values by strictly following [`RFC 3986`](https://tools.ietf.org/html/rfc3986), thus characters in the set ` !*';:+$,/?()` aren't percent-encoded.

This PR adds necessary changes to make the querystring compliant to AWS guidelines without double encoding those characters already encoded by `URLComponents`.

**AWSAPIGatewayClient** adopts a similar strategy by encoding the request URL before generating the signature:
https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSAPIGateway/AWSAPIGatewayClient.m#L97

**Reference**
- http://www.openradar.me/24076063

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Internal/URLComponents+sigv4Encoding.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Internal/URLComponents+sigv4Encoding.swift
@@ -1,0 +1,36 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// Per AWS reference guide https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
+/// URL querystring should be encoded according to the following rules::
+/// - percent-encode with %XY (X and Y are hexadecimal characters) all characters
+///   but any of the __unreserved characters__ defined by `RFC3986` (A-Za-z0-9-_.~)
+/// - encode spaces with %20
+/// - double encode any equals in params values
+///
+/// `URLComponents` encodes queryItems values by strictly following `RFC 3986`.
+/// This function encode missing characters needed to make the querystring compliant to AWS guidelines
+/// without double encoding those characters already encoded by `URLComponents`
+extension URLComponents {
+    static var sigV4UnreservedCharacters: CharacterSet = {
+        var sigV4UnreservedCharacters = CharacterSet(charactersIn: "A" ... "Z")
+        sigV4UnreservedCharacters = sigV4UnreservedCharacters.union(CharacterSet(charactersIn: "a" ... "z"))
+        sigV4UnreservedCharacters = sigV4UnreservedCharacters.union(CharacterSet(charactersIn: "0" ... "9"))
+        sigV4UnreservedCharacters = sigV4UnreservedCharacters.union(CharacterSet(charactersIn: "-_~.%=&"))
+        return sigV4UnreservedCharacters
+    }()
+
+    mutating func percentEncodeQueryBySigV4Rules() {
+        guard let percentEncodedQuery = self.percentEncodedQuery else {
+            return
+        }
+        self.percentEncodedQuery = percentEncodedQuery.addingPercentEncoding(
+            withAllowedCharacters: Self.sigV4UnreservedCharacters)
+    }
+}
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Internal/URLComponents+sigv4Encoding.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Internal/URLComponents+sigv4Encoding.swift
@@ -15,8 +15,6 @@ import Foundation
 /// - double encode any equals in params values
 ///
 /// `URLComponents` encodes queryItems values by strictly following `RFC 3986`.
-/// This function encode missing characters needed to make the querystring compliant to AWS guidelines
-/// without double encoding those characters already encoded by `URLComponents`
 extension URLComponents {
     static var sigV4UnreservedCharacters: CharacterSet = {
         var sigV4UnreservedCharacters = CharacterSet(charactersIn: "A" ... "Z")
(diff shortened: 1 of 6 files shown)


AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Utils/RESTOperationRequestUtils.swift
```diff
@@ -29,10 +29,11 @@ final class RESTOperationRequestUtils {
             components.path.append(path)
         }
 
-        if let queryParameters = queryParameters {
+        if let queryParameters = queryParameters?.sorted(by: { $0.key < $1.key }) {
```
palpatim, line 32: Reordering the query parameters is not necessarily an idempotent operation. Depending on the service implementation, `http://www.foo.com?foo[]=a&foo[]=b` may result in a different interpretation than `http://www.foo.com?foo[]=b&foo[]=a`. What is the reason for reordering if the intent is to simply encode the keys & values?

diegocstn: The only reason for reordering is for making testing reliable as dictionaries don't guarantee any order on keys.
I also considered changing the method signature to accept [`KeyValuePairs`](https://developer.apple.com/documentation/swift/keyvaluepairs) instead of a `Dictionary` but it would have generated more changes

palpatim: As discussed, remove the reordering and change your test assertion to compare one parameter key/value pair at a time

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Internal/URLComponents+sigv4Encoding.swift
```diff
@@ -22,15 +22,29 @@ extension URLComponents {
...
+        /// removingPercentEncoding returns `nil` if called on a value
+        /// that hasn't been prior encoded
+        let unencoded = value.removingPercentEncoding ?? value
+
+        /// apply sigv4 encoding
+        return unencoded.addingPercentEncoding(
+            withAllowedCharacters: Self.sigV4UnreservedCharacters) ?? value
+    }
+
+    mutating func queryItemsEncodedPerSigV4Rules(_ queryItems: [String: String]?) {
```
palpatim, line 39: nit, since this method doesn't return a value, but rather mutates in place, consider renaming to something that is imperative rather than declarative, such as `encodeQueryItemsPerSigV4Rules` or similar. (Similar constructions in the Swift stdlib with things like `sort()` and `sorted()`)

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Internal/URLComponents+sigv4Encoding.swift
```diff
@@ -22,15 +22,29 @@ extension URLComponents {
...
+        guard let queryItems = queryItems else {
             return
         }
-        self.percentEncodedQuery = percentEncodedQuery.addingPercentEncoding(
-            withAllowedCharacters: Self.sigV4UnreservedCharacters)
+        percentEncodedQuery = queryItems.map { (name, value) -> String in
```
palpatim, line 43: Style nit: is the full type signature required here? (It might be--Swift's type inference is great all the way up until it isn't. :) )

```suggestion
        percentEncodedQuery = queryItems.map { name, value in
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Internal/URLComponents+sigv4Encoding.swift
```diff
@@ -22,15 +22,29 @@ extension URLComponents {
...
+            let encodedName = encodeQueryParamValueBySigV4Rules(name)
+            let encodedValue = encodeQueryParamValueBySigV4Rules(value)
```
palpatim, line 45: Such nit, very picky: if it wouldn't be a breaking change...you're encoding an arbitrary string value, not just a query parameter value. More specifically, you're encoding both a query "field" and "value". Consider name change to something like `encodeQueryParamItemBySignV4Rules(_:)`. Or even `encodeBySigV4Rules(_:)`?

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Internal/URLComponents+sigv4Encoding.swift
```diff
@@ -22,15 +22,29 @@ extension URLComponents {
...
 
-    mutating func percentEncodeQueryBySigV4Rules() {
-        guard let percentEncodedQuery = self.percentEncodedQuery else {
+    private func encodeQueryParamValueBySigV4Rules(_ value: String) -> String {
+        /// removingPercentEncoding returns `nil` if called on a value
+        /// that hasn't been prior encoded
+        let unencoded = value.removingPercentEncoding ?? value
+
+        /// apply sigv4 encoding
+        return unencoded.addingPercentEncoding(
```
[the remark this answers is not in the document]
diegocstn: 1. 🙃
2. I'm not sure about under what conditions `addingPercentEncoding` would return return nil, Apple docs only say _"if the transformation is not possible"_ -- possibly if the UTF-8 encoding fails or similar edge case?  I'll update to throw an error that `constructURL` can catch and eventually re-throws an `APIError`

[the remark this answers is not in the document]
palpatim: As discussed, since this is an internal extension, I don't object to throwing an APIError directly from the extension, rather than adding a custom type of limited value. Let's just make sure the APIError has enough detail that the customer can diagnose the problem if they encounter the rare case where a URL can't be encoded.

AmplifyPlugins/API/AWSAPICategoryPlugin/Support/Internal/URLComponents+sigv4Encoding.swift
```diff
@@ -41,14 +42,15 @@ extension URLComponents {
         percentEncodedQuery = try queryItems.map { name, value in
             guard let encodedName = encodeQueryParamItemBySigV4Rules(name),
                   let encodedValue = encodeQueryParamItemBySigV4Rules(value) else {
-                throw SigV4Error.invalidQueryItem(name, value)
+                throw APIError.invalidURL(
+                    "Invalid query parameter.",
+                    """
+                    Review your API plugin configuration and make sure to pass valid query parameters in your request.
```
palpatim, line 48: Nit: Would the plugin configuration be the issue, or is it more likely to be something in app code or even user-provided data? Maybe something like:
```suggestion
                    Review your Amplify.API call to make sure you are passing valid UTF-8 query parameters in your request.
```

(Appropriately line-breaked, of course :) )

palpatim (approved):
Minor rewording comment, otherwise LGTM

palpatim (approved):

PR #1076, opened 2021-02-24
fix(API): Return response body for non-2xx failure cases

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/655

*Description of changes:*
This adds a non-breaking API change to propagate the response body as an `HTTPURLResponse` for non 2xx responses (`APIError.httpStatusError`)

**The developer experience**

When using `AWSAPIPlugin`, you can optionally type cast `HTTPURLResponse` to an instance of `AWSHTTPURLResponse`. The response body can be accessed from the `body: Data?` property. For example, when the `APIError` is an `.httpStatusError(StatusCode, HTTPURLResponse)`, then access the response body by type casting the response to an `AWSHTTPURLResponse` and retrieve the `body` field.
```swift
if case let .httpStatusError(statusCode, response) = error, let awsResponse = response as? AWSHTTPURLResponse {
    if let responseBody = awsResponse.body {
        print("Response contains a response body")
    }
}
```

*Check points:*

- [x] All unit tests pass

*Integration testing done*
API Integration testing in RESTWithIAMIntegrationTests.swift


AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/APIOperationResponse.swift
```diff
@@ -32,7 +34,11 @@ extension APIOperationResponse {
 
             let successStatusCodes = 200 ..< 300
             if !successStatusCodes.contains(statusCode) {
-                throw APIError.httpStatusError(statusCode, response)
+                if let restResponse = AWSHTTPURLResponse.init(response: response, body: responseData) {
```
palpatim, line 37: Prefer the type constructor over calling `.init` as a method

```suggestion
                if let restResponse = AWSHTTPURLResponse(response: response, body: responseData) {
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSHTTPURLResponse.swift
```diff
@@ -0,0 +1,80 @@
...
+    public let body: Data?
+
+    private let response: HTTPURLResponse
+
+    init?(response: HTTPURLResponse, body: Data?) {
+        self.body = body
+        self.response = response
+
+        // Call the super class initializer with dummy values to satisfy the requirement of the inheritance.
+        super.init(url: URL(string: "dummyURL")!,
```
palpatim, line 38: Why can't we call this with `response.url` and only default to the dummy if that isn't present?

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSHTTPURLResponse.swift
```diff
@@ -0,0 +1,80 @@
...
+///         print("Response contains a response body")
+///     }
+/// }
+/// ```
+/// **Note**: The class inheritance to `HTTPURLResponse` is to provide above mechanism, and actual
+/// implementation acts as a facade that stores an instance of `HTTPURLResponse` that delegates overidden methods to
+/// this stored property.
+public class AWSHTTPURLResponse: HTTPURLResponse {
+
+    /// The response body of the response
```
palpatim, line 28:
```suggestion
    /// The body of the response, if available
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSHTTPURLResponse.swift
```diff
@@ -0,0 +1,80 @@
...
+/// `AWSHTTPURLResponse` contains the response body and metadata associated with the response to an HTTP request.
+///
+/// When using `AWSAPIPlugin`, you can optionally type cast `HTTPURLResponse` to an instances of `AWSHTTPURLResponse`.
+/// The response body can be accessed from the `body: Data?` property. For example, when the `APIError` is an
+/// `.httpStatusError(StatusCode, HTTPURLResponse)`, then access the response body by type casting the response to
+/// an `AWSHTTPURLResponse` and retrieve the `body` field.
+/// ```swift
+/// if case let .httpStatusError(statusCode, response) = error, let awsResponse = response as? AWSHTTPURLResponse {
+///     if let responseBody = awsResponse.body {
+///         print("Response contains a response body")
```
palpatim, line 19: Nit to clarify that the response body is a Data object, and also to avoid compiler warnings in copy/pasted code :)

```suggestion
///         print("Response contains a \(responseBody.count) byte long response body")
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSHTTPURLResponse.swift
```diff
@@ -0,0 +1,80 @@
...
+
+        // Call the super class initializer with dummy values to satisfy the requirement of the inheritance.
+        super.init(url: URL(string: "dummyURL")!,
+                   statusCode: 0,
+                   httpVersion: nil,
+                   headerFields: nil)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented and is not expected to be called.")
```
palpatim, line 45: Why can't we delegate to super instead of crashing?

```suggestion
        super.init(coder: coder)
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSHTTPURLResponse.swift
```diff
@@ -0,0 +1,80 @@
...
+/// implementation acts as a facade that stores an instance of `HTTPURLResponse` that delegates overidden methods to
+/// this stored property.
+public class AWSHTTPURLResponse: HTTPURLResponse {
+
+    /// The response body of the response
+    public let body: Data?
+
+    private let response: HTTPURLResponse
+
+    init?(response: HTTPURLResponse, body: Data?) {
```
palpatim, line 33: If we're subclassing, why assign to a private instance rather than invoking [`super.init(url:statusCode:httpVersion:headerFields:)`](https://developer.apple.com/documentation/foundation/httpurlresponse/1415870-init). That way we don't have to proxy all the calls below either.

lawmicha: I felt that this initializer was a bit limiting (lost some information about the parent `URLResponse` class) which led me to think its purpose is meant for mocking. If we take a look at the super.init's parameters, I could avoid the proxying by calling super.init with the following values
- `url` - can directly use `response.url`
- `statusCode` - can directly use `response.statusCode`
- `httpVersion` - could not find this on the response object, could hard code to "HTTP/1.1" as a dummy value
- `headerFields` - translate from `response.allHeaderValues` of type `[AnyHashable: Any]` to [String: String]`

By doing so, the values returned are correct (header fields, status code). My concern with this was developers could also access the `URLResponse` methods from the `HTTPURLResponse` instance, which would then be missing information such as `mimeType`, `expectedContentLength`, `textEncoding`. So because I could not fully re-create an instance that had the same information as the original instance, I had decided to hold onto the original instance and proxy the calls, incase this missing information causes a problem for existing developers.

palpatim: That makes a lot of sense to me, thanks for explaining.

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSHTTPURLResponse.swift
lawmicha: we can, but given the implementation is to proxy the calls to the stored property, I did not want to introduce logic that could confuse the reader such as "why are we storing passing the `response.url` to the super.init` but still store the httpURLResponse?

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSHTTPURLResponse.swift
lawmicha: yes we can, this change appears to be the safer option for sure

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSHTTPURLResponse.swift
```diff
@@ -42,7 +42,9 @@ public class AWSHTTPURLResponse: HTTPURLResponse {
     }
 
     required init?(coder: NSCoder) {
-        fatalError("init(coder:) has not been implemented and is not expected to be called.")
+        self.body = nil
+        self.response = HTTPURLResponse()
```
palpatim, line 46: This should be decoding the incoming `response` on the coder, or else the decoded instance won't have any values from the decoded instance.

palpatim (changes requested):

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSHTTPURLResponse.swift
lawmicha: Makes sense, I have to override the `encode(NSCoder)` to encode with keys that I expect to decode from the `init(NSCoder)` to decode correctly. updated code accordingly

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSHTTPURLResponse.swift
```diff
@@ -0,0 +1,88 @@
...
+    /// The body of the response, if available
+    public let body: Data?
+
+    private let response: HTTPURLResponse
+
+    init?(response: HTTPURLResponse, body: Data?) {
+        self.body = body
+        self.response = response
+
+        // Call the super class initializer with dummy values to satisfy the requirement of the inheritance.
```
palpatim, line 37: Clarification nit:

```suggestion
        // Call the super class initializer with dummy values to satisfy the requirement of the inheritance.
        // Subsequent access to any properties of this instance (including `url`) will be delegated to
        // the `response`.
```

PR #1100, opened 2021-03-12
fix(datastore): fix selective sync expressions to run DDB query when possible

diegocstn:
*Description of changes:*
When a filter is provided on a sync operation, AppSync will attempt to optimize the request by performing a DynamoDB query, rather than a scan. Query operations enable a highly efficient and cost-effective data retrieval.

This PR introduces a change to wrap provided DataStore SyncExpressions in an `and` operator in order to retrieve items from DynamoDB with a query operation instead of a scan.
AppSync resolver requires a presence of a top level "and" group  https://github.com/aws-amplify/amplify-cli/blob/018548a772cab14a1f99fb3d2413eb720142a562/packages/graphql-key-transformer/src/KeyTransformer.ts#L1105
and the JavaScript library implementation is already handling this.

Android PR: https://github.com/aws-amplify/amplify-android/pull/1225

*Check points:*

- [x] All unit tests pass


lawmicha:
> This PR introduces a change to wrap provided DataStore SyncExpressions in an and operator in order to retrieve items from DynamoDB with a query operation instead of a scan.

do you mean with a scan instead of a query?

lawmicha (approved):

richardmcclellan (approved):
Currently, the `DataStoreSyncExpression` is passed as the filter on sync requests, and is used for client side filtering of subscriptions.  In the future, we will replace the client side filtering of subscriptions with server side filtering, likely by passing the filter argument when starting the subscription. 

For the Android implementation of this, I am wrapping with an `and` group right before the sync request is sent, so there is no change to how subscriptions are filtered.  In this current PR, it looks like you are wrapping with `and` at a higher level, so this will impact subscriptions as well.  It probably doesn't actually matter that much since wrapping with an `and` will still evaluate to the same result, but just wanted to point out this difference.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/DataStoreSyncExpression.swift
```diff
@@ -14,6 +14,25 @@ public struct DataStoreSyncExpression {
...
+            // Wrapping the predicate with a group AND enables
+            // AppSync to optimize the request by performing a DynamoDB query instead of a scan.
+            // If the provided syncPredicate is already a QueryPredicateGroup, this is not needed.
+            // If the provided group is of type AND, the optimization will occur.
+            // If the top level group is OR or NOT, the optimization is not possible anyway.
+            if predicate as? QueryPredicateGroup != nil {
+                return predicate
+            } else if let predicate = predicate as? QueryPredicateConstant,
+                      predicate == .all {
+                return predicate
```
palpatim, line 30: Will AppSync auto-optimize in the constant case? Is it possible or desirable to optimize by wrapping even constants in an `.and` group?

palpatim (approved):
LGTM. I had one question about applying the optimization to constant predicate expressions, but nothing blocking.

richardmcclellan (approved):

PR #1104, opened 2021-03-16
feat(datastore): process OperationDisabled error

diegocstn:
*Description of changes:*
This PR introduces a set of changes to let DataStore process a new type of error _OperationDisabled_ without halting the sync engine.
Developers will be able to mark an operation as "disabled" by throwing an error directly from an AppSync resolver
```JavaScript
$util.error("Subscription are disabled.", "OperationDisabled")
```

*Check points:*

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Core/AWSPluginsCore/API/AppSyncErrorType.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/API/AppSyncErrorType.swift
@@ -13,6 +13,7 @@ public enum AppSyncErrorType: Equatable {
     private static let conditionalCheckFailedErrorString = "ConditionalCheckFailedException"
     private static let conflictUnhandledErrorString = "ConflictUnhandled"
     private static let unauthorizedErrorString = "Unauthorized"
+    private static let operationDisabledErrorString = "OperationDisabled"
 
     /// Conflict detection finds a version mismatch and the conflict handler rejects the mutation.
     /// See https://docs.aws.amazon.com/appsync/latest/devguide/conflict-detection-and-sync.html for more information
@@ -21,6 +22,8 @@ public enum AppSyncErrorType: Equatable {
     case conditionalCheck
 
     case unauthorized
+    
+    case operationDisabled
 
     case unknown(String)
 
@@ -32,6 +35,8 @@ public enum AppSyncErrorType: Equatable {
             self = .conflictUnhandled
         case AppSyncErrorType.unauthorizedErrorString:
             self = .unauthorized
+        case AppSyncErrorType.operationDisabledErrorString:
+            self = .operationDisabled
         default:
             self = .unknown(value)
         }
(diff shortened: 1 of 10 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
```diff
@@ -92,6 +92,9 @@ class ProcessMutationErrorFromCloudOperation: AsynchronousOperation {
                 // TODO: dispatch Hub event
                 log.debug("Unauthorized mutation \(errorType)")
                 finish(result: .success(nil))
+            case .operationDisabled:
```
richardmcclellan, line 95: On Android, I've only added logic to ignore `OperationDisabled` errors in the `SubscriptionProcessor`, not for mutations or sync.   What was thinking for adding it here as well?

diegocstn: I believe an "OperationDisabled" error can be also thrown from a mutation resolver, so I thought it'd better handle it as we do for subscriptions

richardmcclellan: It's not part of the current set of requirements, but I could imagine it is possible someone could want that in the future.  
 I don't feel strongly either way.  I didn't add it for Android, but I think it's fine to add if you want.

AmplifyPlugins/Core/AWSPluginsCore/API/AppSyncErrorType.swift
```diff
@@ -22,6 +23,8 @@ public enum AppSyncErrorType: Equatable {
 
     case unauthorized
 
+    case operationDisabled
```
richardmcclellan, line 26: Can you add some documentation around this to indicate that customers should not use this unless they have consulted with AWS?   Here's the similar blurb I added on Android: https://github.com/aws-amplify/amplify-android/blob/f51cc5b5eea477a2ad0b34c2012de065c6051884/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncExtensions.java#L147-L151

diegocstn: thanks, updated

richardmcclellan (approved):

palpatim (approved):
Change LGTM overall, but please double check the codecov messages. I see you added tests, so I'm not sure whether your tests don't actually cover the logic codecov is noting, or whether the tests are misconfigured in some way, or whether codecov is just straight up wrong. If it's the latter, understanding why it's wrong will be important for a fast follow-up, since we don't want spurious codecov failures to degrade our confidence in our test coverage.

PR #1133, opened 2021-04-01
feat(datastore): support for readonly fields

diegocstn:
### Issue
Amplify CLI generates `createdAt` and `updatedAt` fields on every `@model`, and customer can also specify custom field name by defining models as follows
```
type Todo @model(timestamps:{ createdAt: "createdOn", updatedAt: "updatedOn" }) {
  id: ID!
  name: String!
  description: String
}
```
These fields are not currently included in autogenerated models thus are not accessible to customers.
Codegen is being updated to generate these fields (https://github.com/aws-amplify/amplify-codegen/pull/114) but, as they're fully managed by AppSync, they need a special treatment. To avoid AppSync throwing an error, they must NOT be included in the mutation _input_.

### Solution
Since we can't rely on their name to omit `createdAt` and `updatedAt` from the mutation input, codegen will mark them as "read-only". This PR introduces necessary changes to support read-only fields.

Below an example of an updated model and schema
```swift
public struct Todo: Model {
    public let id: String
    public var name: String
    public var done: Bool
    public var priority: Priority?
    public var description: String?
    public let createdAt: Temporal.DateTime?
    public let updatedAt: Temporal.DateTime?
    public init(name: String,
                done: Bool,
                priority: Priority? = nil,
                description: String? = nil)
    {
        self.init(name: name,
                  done: done,
                  priority: priority,
                  description: description,
                  createdAt: nil,
                  updatedAt: nil)
    }
    internal init(id: String = UUID().uuidString,
                  name: String,
                  done: Bool,
                  priority: Priority? = nil,
                  description: String? = nil,
                  createdAt: Temporal.DateTime? = nil,
                  updatedAt: Temporal.DateTime? = nil)
    {
        self.id = id
        self.name = name
        self.done = done
        self.priority = priority
        self.description = description
        self.createdAt = createdAt
        self.updatedAt = updatedAt
    }
}
```
```swift
extension Todo {
...
  public static let schema = defineSchema { model in
      let todo = Todo.keys
  
      model.pluralName = "Tods"
  
      model.fields(
          .id(),
          .field(todo.name, is: .required, ofType: .string),
          .field(todo.description, is: .optional, ofType: .string),
          .field(todo.createdAt, is: .optional, isReadOnly: true, ofType: .date),
          .field(todo.updatedAt, is: .optional, isReadOnly: true, ofType: .date)
          )
  }
}
```

### References
Android: https://github.com/aws-amplify/amplify-android/pull/1249
CodeGen changes: https://github.com/aws-amplify/amplify-codegen/pull/114

*Check points:*

- [x] All unit tests pass

### DIFF
--- a/Amplify/Categories/DataStore/Model/Internal/Schema/ModelSchema+Definition.swift
+++ b/Amplify/Categories/DataStore/Model/Internal/Schema/ModelSchema+Definition.swift
@@ -121,6 +121,22 @@ public enum ModelFieldNullability {
 
 /// - Warning: Although this has `public` access, it is intended for internal & codegen use and should not be used
 ///   directly by host applications. The behavior of this may change without warning.
+public enum ModelFieldWritability {
+    case readOnly
+    case readWrite
+
+    var isReadOnly: Bool {
+        switch self {
+        case .readOnly:
+            return true
+        case .readWrite:
+            return false
+        }
+    }
+}
+
+/// - Warning: Although this has `public` access, it is intended for internal & codegen use and should not be used
+///   directly by host applications. The behavior of this may change without warning.
 public struct ModelSchemaDefinition {
 
     internal let name: String
@@ -167,12 +183,14 @@ public enum ModelFieldDefinition {
     case field(name: String,
                type: ModelFieldType,
                nullability: ModelFieldNullability,
+               writability: ModelFieldWritability,
                association: ModelAssociation?,
                attributes: [ModelFieldAttribute],
                authRules: AuthRules)
 
     public static func field(_ key: CodingKey,
                              is nullability: ModelFieldNullability = .required,
+                             access writability: ModelFieldWritability = .readWrite,
                              ofType type: ModelFieldType = .string,
                              attributes: [ModelFieldAttribute] = [],
                              association: ModelAssociation? = nil,
@@ -180,6 +198,7 @@ public enum ModelFieldDefinition {
         return .field(name: key.stringValue,
                       type: type,
                       nullability: nullability,
+                      writability: writability,
                       association: association,
                       attributes: attributes,
                       authRules: authRules)
@@ -193,6 +212,7 @@ public enum ModelFieldDefinition {
         return .field(name: name,
                       type: .string,
                       nullability: .required,
+                      writability: .readWrite,
                       association: nil,
                       attributes: [.primaryKey],
                       authRules: [])
@@ -200,32 +220,38 @@ public enum ModelFieldDefinition {
 
     public static func hasMany(_ key: CodingKey,
                                is nullability: ModelFieldNullability = .required,
+                               access writability: ModelFieldWritability = .readWrite,
                                ofType type: Model.Type,
                                associatedWith associatedKey: CodingKey) -> ModelFieldDefinition {
         return .field(key,
                       is: nullability,
+                      access: writability,
                       ofType: .collection(of: type),
                       association: .hasMany(associatedWith: associatedKey))
     }
 
     public static func hasOne(_ key: CodingKey,
                               is nullability: ModelFieldNullability = .required,
+                              access writability: ModelFieldWritability = .readWrite,
                               ofType type: Model.Type,
                               associatedWith associatedKey: CodingKey,
                               targetName: String? = nil) -> ModelFieldDefinition {
         return .field(key,
                       is: nullability,
+                      access: writability,
                       ofType: .model(type: type),
                       association: .hasOne(associatedWith: associatedKey, targetName: targetName))
     }
 
     public static func belongsTo(_ key: CodingKey,
                                  is nullability: ModelFieldNullability = .required,
+                                 access writability: ModelFieldWritability = .readWrite,
                                  ofType type: Model.Type,
                                  associatedWith associatedKey: CodingKey? = nil,
                                  targetName: String? = nil) -> ModelFieldDefinition {
         return .field(key,
                       is: nullability,
+                      access: writability,
                       ofType: .model(type: type),
                       association: .belongsTo(associatedWith: associatedKey, targetName: targetName))
     }
@@ -234,6 +260,7 @@ public enum ModelFieldDefinition {
         guard case let .field(name,
                               type,
                               nullability,
+                              writability,
                               association,
                               attributes,
                               authRules) = self else {
@@ -242,6 +269,7 @@ public enum ModelFieldDefinition {
         return ModelField(name: name,
                           type: type,
                           isRequired: nullability.isRequired,
+                          isReadOnly: writability.isReadOnly,
                           isArray: type.isArray,
                           attributes: attributes,
                           association: association,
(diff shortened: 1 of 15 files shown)


Amplify/Categories/DataStore/Model/Internal/Schema/ModelSchema+Definition.swift
```diff
@@ -119,6 +119,22 @@ public enum ModelFieldNullability {
     }
 }
 
+/// - Warning: Although this has `public` access, it is intended for internal & codegen use and should not be used
+///   directly by host applications. The behavior of this may change without warning.
+public enum ModelFieldWritability {
```
diegocstn, line 124: Naming suggestions are more than welcome :)

brennanMKE (approved):

Amplify/Categories/DataStore/Model/Internal/Schema/ModelSchema+Definition.swift
```diff
@@ -119,6 +119,22 @@ public enum ModelFieldNullability {
...
+    case readOnly
+    case readWrite
+
+    var isReadOnly: Bool {
+        switch self {
```
palpatim, line 129: If we keep the enum, we could simplify this:
```swift
var isReadOnly: Bool {
    self == .readOnly
}
```

Amplify/Categories/DataStore/Model/Internal/Schema/ModelSchema+Definition.swift
palpatim: We've already got a "Nullability", this name seems reasonable. But is it needed? If we only have two cases, neither one with associated values, we're basically writing an enum to model a boolean. Why not just use a Bool?

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/GraphQLMutation.swift
```diff
@@ -29,7 +29,8 @@ public struct GraphQLMutation: SingleDirectiveGraphQLDocument {
     }
 
     public init(modelSchema: ModelSchema) {
-        self.selectionSet = SelectionSet(fields: modelSchema.graphQLFields)
+        let fields = modelSchema.graphQLFields.filter { !$0.isReadOnly }
+        self.selectionSet = SelectionSet(fields: fields)
```
palpatim, line 33: Why wouldn't we return the readonly fields in the selection set? If I understand this correctly, in addition to not serving use cases that might actually need to see the readonly field, this means I would receive a mutation response that is incomplete--what if my readonly field is required?

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLDocument/GraphQLCreateMutationTests.swift
```diff
@@ -235,4 +235,42 @@ class GraphQLCreateMutationTests: XCTestCase {
         }
         XCTAssertEqual(input["commentPostId"] as? String, post.id)
     }
+
```
palpatim, line 238: General test note: make sure we're covering all permutations of schemas, including mutability and required.

palpatim (commented):
Need to understand behavior of `required`, `readonly` fields better

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLDocument/GraphQLMutation.swift
richardmcclellan: @palpatim, you are correct - the selection set should not be affected by this change.  That is, readonly fields should still be requested in the selection set.

palpatim (changes requested):
See previous comment--flagging my review as "request changes" to increase visibility

Amplify/Categories/DataStore/Model/Internal/Schema/ModelSchema+Definition.swift
diegocstn: The only reason I opted for an enum is to be able to represent other values in the future (like `[writeOnly](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.4)`) without needed to change type. But as I'm not sure if/when/how/why we would need other values, I'm more than happy to switch to a `Bool` if we don't feel is useful

diegocstn (commented):
> Need to understand behavior of `required`, `readonly` fields better

@palpatim agreed with codegen team that `required` and `readonly` can't coexist, so a `readonly` fields are **always** optionals.

Amplify/Categories/DataStore/Model/Internal/Schema/ModelSchema+Definition.swift
```diff
@@ -193,39 +207,46 @@ public enum ModelFieldDefinition {
...
 
     public static func hasOne(_ key: CodingKey,
                               is nullability: ModelFieldNullability = .required,
+                              access writability: ModelFieldWritability = .readWrite,
                               ofType type: Model.Type,
                               associatedWith associatedKey: CodingKey,
                               targetName: String? = nil) -> ModelFieldDefinition {
         return .field(key,
                       is: nullability,
+                      access: writability,
```
diegocstn, line 236: WIP for exercising this API in unit tests

Amplify/Categories/DataStore/Model/Internal/Schema/ModelSchema+Definition.swift
palpatim: If we've got a current use case in mind, feel free to keep it, otherwise, it seems like we're adding a bit of complexity we don't need and I'd vote to remove it. I'm not saying "writeOnly" isn't a valid use case, only that we don't have a current need for it.

If we're attempting to conform to JSONSchema or some other standard, we probably have more work to do than just adding an enum case :)

palpatim (approved):

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -17,7 +17,7 @@ extension Model {
     /// used as the `input` of GraphQL related operations.
     func graphQLInput(_ modelSchema: ModelSchema) -> GraphQLInput {
         var input: GraphQLInput = [:]
-        modelSchema.fields.forEach {
+        modelSchema.fields.filter { !$0.value.isReadOnly }.forEach {
```
lawmicha, line 20: could we move this to line 23? It would keep move the logic out into the body of the for loop, and make the placement of a comment easier

```swift
// When the field is a read-only, do not add it to the GraphQL input object
            if modelField.isReadOnly {
                return
            }
```


I'm not sure if it is better to put it in the ModelDecorator as how the code is right now. In ModelDecorator, it's getting the GraphQL Input from the model, and then modifying it based on the schema's auth rules, such as removing "ownerFields". Since the read-only field is on the field, and not the schema level, i think it's probably better to keep the logic in this computed property with the rest of the field related logic

diegocstn: thanks, applied that suggestion and renamed `graphQLInput` to `graphQLInputForMutation`

PR #1137, opened 2021-04-01
fix(DataStore): remove reconcile operations serially

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/1136

*Description of changes:*

This change fixes a bug introduced in the previous PR (https://github.com/aws-amplify/amplify-ios/pull/1128) which added the `ReconcileAndSaveQueue` that manages `ReconcileAndLocalSaveOperation`'s by storing references to it so that they can be removed by `modelName`. The bug is that we were performing the removals from the dictionary that is not thread safe from another thread (not under the serial queue). The fix adds removals under the existing `serialQueue` so that the operations performed on the dictionary are performed one after another.

Why did the PR introduce the bug?
Although the PR had unit test, integration testing, and "load test" scenario performed from this app (https://github.com/lawmicha/datastoreLoadTesting) It was not caught until customer raised this issue: https://github.com/aws-amplify/amplify-ios/issues/1136 

Going forward, we can do the following when making changes to datastore. 
1. Run with thread safety checks (looking into this and will update some instructions to do so) to see if we could have caught this earlier. -Update: we do have thread sanitizer enabled on the Test schema
2. Improve datastore integration tests with a larger enough data set that represents a typical use case. (load testing effort)

This change adds the removal of operations under the `serialQueue`

*Check points:*

- [ ] All unit tests pass


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveQueue.swift
```diff
@@ -71,13 +71,15 @@ class ReconcileAndSaveQueue: ReconcileAndSaveOperationQueue {
...
     func addOperation(_ operation: ReconcileAndLocalSaveOperation, modelName: String) {
+
         serialQueue.async {
             var reconcileAndLocalSaveOperationSink: AnyCancellable?
-
             reconcileAndLocalSaveOperationSink = operation.publisher.sink { _ in
-                self.reconcileAndLocalSaveOperationSinks.with { $0.remove(reconcileAndLocalSaveOperationSink) }
-                self.modelReconcileAndSaveOperations[modelName]?[operation.id] = nil
-                self.reconcileAndSaveQueueSubject.send(.operationRemoved(id: operation.id))
+                self.serialQueue.async {
```
drochetti, line 78: since this fixes a regression, do you mind putting a paragraph in the PR description explaining why the regression happened and how/why this call fixes it? So we capture it in our history.

drochetti (approved):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveQueue.swift
```diff
@@ -71,13 +71,15 @@ class ReconcileAndSaveQueue: ReconcileAndSaveOperationQueue {
     }
 
     func addOperation(_ operation: ReconcileAndLocalSaveOperation, modelName: String) {
+
         serialQueue.async {
             var reconcileAndLocalSaveOperationSink: AnyCancellable?
-
             reconcileAndLocalSaveOperationSink = operation.publisher.sink { _ in
```
dsanvita, line 77: This code requires Swift 5.3 in the the latest version of XCode 12.  To restore compatibility with XCode 11.4, which is the minimum version indicated in the README, you can change it to:

`
            reconcileAndLocalSaveOperationSink = operation.publisher.sink( receiveCompletion: { _ in
                self.serialQueue.async {
                    self.reconcileAndLocalSaveOperationSinks.with { $0.remove(reconcileAndLocalSaveOperationSink) }
                    self.modelReconcileAndSaveOperations[modelName]?[operation.id] = nil
                    self.reconcileAndSaveQueueSubject.send(.operationRemoved(id: operation.id))
                }
            }, receiveValue: { _ in } )
`

PR #1139, opened 2021-04-02
fix(API): Subscriptions with IAM match signed headers

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/1101

*Description of changes:*

A subscription connection with IAM auth is established with the following logic ([code](https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/SubscriptionInterceptor/IAMAuthInterceptor.swift#L65))

1. An NSMutableURLRequest is created
```swift
let mutableRequest = NSMutableURLRequest(url: endpoint)
```
2. The request is populated with known headers (accept, x-amz-date, content-encoding, content-type)

```swift
mutableRequest.addValue(RealtimeProviderConstants.iamAccept,
                                forHTTPHeaderField: RealtimeProviderConstants.acceptKey)
mutableRequest.addValue(date, forHTTPHeaderField: RealtimeProviderConstants.amzDate)
mutableRequest.addValue(RealtimeProviderConstants.iamEncoding,
                                forHTTPHeaderField: RealtimeProviderConstants.contentEncodingKey)
mutableRequest.addValue(RealtimeProviderConstants.iamConentType,
                                forHTTPHeaderField: RealtimeProviderConstants.contentTypeKey)
```
3. A signature is generated by signing the request using AWSCore's Sigv4 signer. *note* this uses all header values from the request [here](https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSCore/Authentication/AWSSignature.m#L332). The contents (signature, signedHeaders, etc) is added to the Authorization header of the request.
```
signer.interceptRequest(mutableRequest)
// SignedHeaders=accept;content-encoding;content-type;host;x-amz-date;x-amz-security-token

```
5. The headers (known headers, `host`, `X-Amz-Security-Token`, and `Authorization`) is then added to the request's query string parameter "header"
```swift
let authHeader = IAMAuthenticationHeader(authorization: authorization,
                                                 host: host,
                                                 token: securityToken,
                                                 date: date,
                                                 accept: RealtimeProviderConstants.iamAccept,
                                                 contentEncoding: RealtimeProviderConstants.iamEncoding,
                                                 contentType: RealtimeProviderConstants.iamConentType)
let base64Auth = try jsonEncoder.encode(authHeader)
// json value of the encoded header object looks like:
/*
{
    "accept": "application/json, text/javascript",
    "content-encoding": "amz-1.0",
    "content-type": "application/json; charset=UTF-8",
    "host": "xxx.appsync-api.region.amazonaws.com",
    "x-amz-date": "20210316T055843Z",
    "X-Amz-Security-Token": "xxx",
    "Authorization": "AWS4-HMAC-SHA256 Credential=xxxxxx/aws4_request, SignedHeaders=accept;content-encoding;content-type;host;x-amz-date;x-amz-security-token, Signature=xxxxxx"
}
*/
// added to the "header" query string paramter
let headerQuery = URLQueryItem(name: RealtimeProviderConstants.header, value: base64Auth)

```

The changes in this PR accounts for additional header values in the `NSMutableURLRequest` by encoding the known headers, `host`, `X-Amz-Security-Token`, `Authorization`, **and** additional headers from the request. The reasoning behind this is that since the signer accounts for all header values, then the calling code should also account for all header values when creating the header object. 

An [issue](https://github.com/aws-amplify/amplify-ios/issues/1101) was raised by the developer when using third party libraries with Amplify. When they try to estasblish the connection, SigV4 signer created a signature and SignedHeaders with the following:
```
// SignedHeaders=accept;content-encoding;content-type;host;x-amz-date;x-amz-security-token,[ADDITIONAL_HEADERS]
SignedHeaders=accept;content-encoding;content-type;host;x-amz-date;x-amz-security-token,x-dynatrace
```
thus, because the SignedHeaders do not match the header object created from `IAMAuthenticationHeader`, the request fails with an authorization error since the service is unable to verify that the signature is correct. The header is missing the `x-dynatrace` header while the the SignedHeaders indicate there is a `x-dynatrace` header used in signing.

The impact of not doing this is that the customer is blocked from using Amplify with other SDKs like dynatrace which have NSURLRequest interception logic.

**How does dynatrace intercept the request?**

With the dynatrace SDK added as a dependency, whenever I create a `NSMutableURLRequest`, it will be created with the x-dynatrace header value:
```
let mutableRequest = NSMutableURLRequest(url: URL(string: "http://placeholder")!)
print("\(mutableRequest.allHTTPHeaderFields)")
// Optional(["x-dynatrace": "MT_3_7_152628227745248_3-0_bb7d650e-2518-4928-841b-16931791b2bc_0_1_2"])
```

*Check points:*

- [x] All unit tests pass


AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/SubscriptionInterceptor/IAMAuthInterceptor.swift
```diff
@@ -94,61 +113,82 @@ class IAMAuthInterceptor: AuthInterceptor {
...
-private class IAMAuthenticationHeader: AuthenticationHeader {
+class IAMAuthenticationHeader: AuthenticationHeader {
     let authorization: String
     let securityToken: String
-    let date: String
+    let amzDate: String
     let accept: String
     let contentEncoding: String
     let contentType: String
+    let remainingHeaders: [String: String]?
```
diegocstn, line 142: nit: naming, maybe `additionalHeaders`/`extraHeaders`/`customHeaders` would be more appropriate? I'd also add some documentation around this new parameter

AmplifyPlugins/API/AWSAPICategoryPluginTests/Interceptor/SubscriptionInterceptor/IAMAuthInterceptorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+//
+
+import XCTest
+import Amplify
+import AppSyncRealTimeClient
+import AWSCore
+@testable import AWSAPICategoryPlugin
+@testable import AmplifyTestCommon
+
+class IAMAuthInterceptorTests: XCTestCase {
```
diegocstn, line 15: nit: do we also want to exercise `IAMAuthenticationHeader` APIs without any additional header?

AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/SubscriptionInterceptor/IAMAuthInterceptor.swift
```diff
@@ -94,61 +113,82 @@ class IAMAuthInterceptor: AuthInterceptor {
...
+                             forKey: DynamicCodingKeys(stringValue: RealtimeProviderConstants.amzDate)!)
+        try container.encode(accept,
+                             forKey: DynamicCodingKeys(stringValue: RealtimeProviderConstants.acceptKey)!)
+        try container.encode(contentEncoding,
+                             forKey: DynamicCodingKeys(stringValue: RealtimeProviderConstants.contentEncodingKey)!)
+        try container.encode(contentType,
+                             forKey: DynamicCodingKeys(stringValue: RealtimeProviderConstants.contentTypeKey)!)
+        if let headers = remainingHeaders {
+            for (key, value) in headers {
+                try container.encode(value, forKey: DynamicCodingKeys(stringValue: key)!)
```
diegocstn, line 189: looks like we never return `nil` as long as we instantiate `DynamicCodingKeys` using the string-based initializer, but I'd add a comment to help read and understand why force-unwrapping is relatively safe here

AmplifyPlugins/API/AWSAPICategoryPlugin/Interceptor/SubscriptionInterceptor/IAMAuthInterceptor.swift
```diff
@@ -94,61 +113,82 @@ class IAMAuthInterceptor: AuthInterceptor {
...
-        case date = "x-amz-date"
-        case securityToken = "x-amz-security-token"
+    private struct DynamicCodingKeys: CodingKey {
+        var stringValue: String
+        init?(stringValue: String) {
+            self.stringValue = stringValue
+        }
+        var intValue: Int?
+        init?(intValue: Int) {
+            return nil
```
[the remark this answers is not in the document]
lawmicha: Adding comment in code
```
// We are not using this, thus just return nil. If we don't return nil, then it is expected all of the
// stored properties are initialized, forcing the implementation to have logic that maintains the two
// properties `stringValue` and `intValue`. Since we don't have a string representation of an int value
// and aren't using int values for determining the coding key, then simply return nil since the encoder
// will always pass in the header key string.
```
Returning nil seems to keep it more simple. As a reference, i learned about this from [here](https://swiftsenpai.com/swift/decode-dynamic-keys-json/)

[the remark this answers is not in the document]
diegocstn: thanks!

diegocstn (approved):
LGTM

PR #1150, opened 2021-04-07
fix: throw error when plugin added post-configure

haverchuck:
*Description of changes:*
Ensures that `add` method for Auth, Analytics, Storage, Datastore, API and Predictions categories throw an error when attempt is made to add a plugin after configuration.

*Check points:*

- [x] All unit tests pass

### DIFF
--- a/Amplify/Categories/API/AmplifyAPICategory+APICategory.swift
+++ b/Amplify/Categories/API/AmplifyAPICategory+APICategory.swift
@@ -22,8 +22,8 @@ extension AmplifyAPICategory: APICategory {
         guard !isConfigured else {
             let pluginDescription = String(describing: plugin)
             let error = ConfigurationError.amplifyAlreadyConfigured(
-                "\(pluginDescription) has already been configured.",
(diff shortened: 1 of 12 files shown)


Amplify/Categories/API/AmplifyAPICategory+APICategory.swift
```diff
@@ -19,6 +19,15 @@ extension AmplifyAPICategory: APICategory {
             throw error
         }
 
+        guard !isConfigured else {
+            let pluginDescription = String(describing: plugin)
+            let error = ConfigurationError.amplifyAlreadyConfigured(
+                "\(pluginDescription) has already been configured.",
+                "Remove the duplicate call to `Amplify.configure().`"
```
palpatim, line 26: Here and throughout: error message doesn't match the condition
```suggestion
                "Do not add plugins after calling `Amplify.configure().`"
```

palpatim (changes requested):
I think the error message isn't matched to the failure condition, but otherwise LGTM

palpatim (approved):

PR #1167, opened 2021-04-14
fix(API): Reachability resolve to GraphQL API

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/1159
Part of https://github.com/aws-amplify/amplify-ios/issues/1103

*Description of changes:*
This improves the endpoint resolution logic to return the GraphQL endpoint whenever it is possible to. This makes is so that when using DataStore, the APIPlugin's reachability by default will return the first GraphQL endpoint configured, even when there is more than one API. For example
1. one REST and one graphQL API will return the graphQL API
2. one REST and two graphQL APIs will return the REST API
3. two REST APIs and one GraphQL API wil return the GraphQL API
4. two REST APIs and two GraphQL APIs will not be able to resolve

*Check points:*

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
@@ -17,7 +17,16 @@ extension AWSAPIPlugin {
 
     @available(iOS 13.0, *)
     public func reachabilityPublisher(for apiName: String?) throws -> AnyPublisher<ReachabilityUpdate, Never>? {
-        let hostName = try determineHostName(apiName: apiName)
+        let endpoint = try pluginConfig.endpoints.getConfig(for: apiName)
+        guard let hostName = endpoint.baseURL.host else {
+            let error = APIError.invalidConfiguration("Invalid endpoint configuration",
+                """
+                baseURL does not contain a valid hostname
+                """
+            )
+            throw error
+        }
+
         if let networkReachability = reachabilityMap[hostName] {
             return networkReachability.publisher
         }
@@ -32,51 +41,4 @@ extension AWSAPIPlugin {
             return nil
         }
     }
-
-    private func determineHostName(apiName: String?) throws -> String {
-        if let apiName = apiName {
-            guard let baseUrl = pluginConfig.endpoints[apiName]?.baseURL,
-                let host = baseUrl.host else {
-                    let error = APIError.invalidConfiguration("Invalid endpoint configuration for \(apiName)",
-                                                              """
-                                                              baseURL does not contain a valid hostname
-                                                              for apiName: \(apiName)
-                                                              """
-                    )
-                    throw error
-            }
-            return host
-        }
-
-        if pluginConfig.endpoints.count > 1 {
-            let error = APIError.invalidConfiguration("Unable to determine which endpoint configuration",
-                                                      """
-                                                      Pass in the apiName to disambiguate between which endpoint
-                                                      you are requesting reachability for
-                                                      """
-            )
-            throw error
-        }
-
-        guard let configEntry = pluginConfig.endpoints.first else {
-            let error = APIError.invalidConfiguration("No API configurations found",
-                                                      """
-                                                      Review how the API category is being instantiated and
-                                                      configured.
-                                                      """
-            )
-            throw error
-        }
-
-        guard let host = configEntry.value.baseURL.host else {
-            let error = APIError.invalidConfiguration("Invalid endpoint configuration",
-                """
-                baseURL does not contain a valid hostname
-                """
-            )
-            throw error
-        }
-        return host
-    }
-
 }
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
@@ -27,6 +27,10 @@ extension AWSAPIPlugin {
             throw error
         }
 
+        reachabilityMapLock.lock()
+        defer {
+            reachabilityMapLock.unlock()
+        }
         if let networkReachability = reachabilityMap[hostName] {
             return networkReachability.publisher
         }
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
@@ -58,7 +58,7 @@ final public class AWSAPIPlugin: NSObject, APICategoryPlugin {
     }
 
     /// Lock used for performing operations atomically when getting and setting `reachabilityMap`.
-    let reachabilityMapLock = NSLock()
+    let reachabilityMapLock: NSLock
 
     public init(
         modelRegistration: AmplifyModelRegistration? = nil,
@@ -67,7 +67,8 @@ final public class AWSAPIPlugin: NSObject, APICategoryPlugin {
     ) {
         self.mapper = OperationTaskMapper()
         self.queue = OperationQueue()
-        self.authProviderFactory =  apiAuthProviderFactory ?? APIAuthProviderFactory()
+        self.authProviderFactory = apiAuthProviderFactory ?? APIAuthProviderFactory()
+        self.reachabilityMapLock = NSLock()
         super.init()
 
         modelRegistration?.registerModels(registry: ModelRegistry.self)
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
@@ -57,6 +57,9 @@ final public class AWSAPIPlugin: NSObject, APICategoryPlugin {
         }
     }
 
+    /// Lock used for performing operations atomically when getting and setting `reachabilityMap`.
+    let reachabilityMapLock = NSLock()
+
     public init(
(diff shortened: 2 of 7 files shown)


brennanMKE (commented):
Just a single change would be needed from what I can see in these changes. See the comment.

AmplifyPlugins/API/AWSAPICategoryPluginTests/AWSAPICategoryPlugin+ReachabilityTests.swift
```diff
@@ -0,0 +1,95 @@
...
+            apiPlugin.configure(using: dependencies)
+        } catch {
+            XCTFail("Failed to create endpoint config")
+        }
+
+        let publisher = try apiPlugin.reachabilityPublisher()
+        XCTAssertNotNil(publisher)
+        XCTAssertEqual(apiPlugin.reachabilityMap.count, 1)
+        guard let reachability = apiPlugin.reachabilityMap.first else {
+            return
```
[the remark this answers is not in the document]
lawmicha: nice catch, yes there should be

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
```diff
@@ -17,7 +17,16 @@ extension AWSAPIPlugin {
...
+        guard let hostName = endpoint.baseURL.host else {
+            let error = APIError.invalidConfiguration("Invalid endpoint configuration",
+                """
+                baseURL does not contain a valid hostname
+                """
+            )
+            throw error
+        }
+
         if let networkReachability = reachabilityMap[hostName] {
```
palpatim, line 30: I don't see where `reachabilityMap` is declared, but is this thread safe? What happens if different threads invoke this method at the same time?

AmplifyPlugins/API/AWSAPICategoryPlugin/Configuration/AWSAPICategoryPluginConfiguration+EndpointConfig.swift
```diff
@@ -281,13 +281,46 @@ public extension AWSAPICategoryPluginConfiguration {
...
+            throw error
+
+        }
 
+        return endpointConfig
+    }
+
+    // Retrieve the endpoint configuration when there is only one endpoint of the specified `endpointType`
+    private func getConfig(for endpointType: AWSAPICategoryPluginEndpointType) throws ->
+    AWSAPICategoryPluginConfiguration.EndpointConfig {
```
palpatim, line 323: nit: indentation. Did we run swiftFormat & swiftLint on this before submitting the PR?

AmplifyPlugins/API/AWSAPICategoryPlugin/Configuration/AWSAPICategoryPluginConfiguration+EndpointConfig.swift
```diff
@@ -302,24 +335,31 @@ extension Dictionary where Key == String, Value == AWSAPICategoryPluginConfigura
...
+        let restEndpoints = filter { (_, endpointConfig) -> Bool in
+            return endpointConfig.endpointType == .rest
+        }
 
+        if restEndpoints.count == 1, let endpoint = restEndpoints.first {
+            return endpoint.value
         }
 
-        return endpointConfig
+        throw APIError.invalidConfiguration("Unable to determine which endpoint configuration",
```
palpatim, line 359: grammar nit:
```suggestion
        throw APIError.invalidConfiguration("Unable to resolve endpoint configuration",
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Configuration/AWSAPICategoryPluginConfiguration+EndpointConfig.swift
```diff
@@ -302,24 +335,31 @@ extension Dictionary where Key == String, Value == AWSAPICategoryPluginConfigura
...
+                                            """
+                                            Pass in the apiName to disambiguate between which endpoint
+                                            you are requesting for
```
palpatim, line 362: grammar nit
```suggestion
                                            Pass in the apiName to specify the endpoint you are
                                            retrieving the config for
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Configuration/AWSAPICategoryPluginConfiguration+EndpointConfig.swift
lawmicha: i believe it runs as part of build, but i can manually add indentation to the return type, and then increase the body's indentation by one more, then it will look like this:
```
private func getConfig(for endpointType: AWSAPICategoryPluginEndpointType) throws ->
        AWSAPICategoryPluginConfiguration.EndpointConfig {
            let apiForEndpointType = filter { (_, endpointConfig) -> Bool in
  
```

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+Reachability.swift
lawmicha: nice catch, reachability is declared [here](https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift#L43-L58) with an underlying `iReachabilityMap` to work around having `@available(iOS 13.0, *)` on a stored property. (we can have a computed property, but not stored property with `@available`)

I just tested this with a `DispatchQueue.concurrentPerform` and is not thread safe. The code looks like it is retrieving the underlying map, and setting it, so when two threads try to access the dictionary, i was able to reproduce a data race

updated PR to perform the work on reachabilityMap under a NSLock, i think the cleanest solution so far. my attempt to convert it to AtomicValue had data races in various places due the access performed in the computed property: https://github.com/aws-amplify/amplify-ios/pull/1200

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
```diff
@@ -57,6 +57,9 @@ final public class AWSAPIPlugin: NSObject, APICategoryPlugin {
         }
     }
 
+    /// Lock used for performing operations atomically when getting and setting `reachabilityMap`.
+    let reachabilityMapLock = NSLock()
```
palpatim, line 61: nit: prefer to assign properties in the same place; since we have an `init` for this type, let's assign the value there instead of making it hard to find initial value assignments
```suggestion
    let reachabilityMapLock: NSLock
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Configuration/AWSAPICategoryPluginConfiguration+EndpointConfig.swift
```diff
@@ -281,31 +281,26 @@ public extension AWSAPICategoryPluginConfiguration {
...
-                "Use the apiName to specify which API to call")
+        if let endpointType = endpointType {
+            return try getConfig(for: endpointType)
         }
 
-        return endpointConfig.value
+        return try getConfig()
     }
 
+    // Retrieve the endpoint configuration for the specified `apiName`
```
palpatim, line 303: nit: doc comment
```suggestion
    /// Retrieve the endpoint configuration for the specified `apiName`
```

But do we need it? The method is pretty self-documenting

AmplifyPlugins/API/AWSAPICategoryPlugin/Configuration/AWSAPICategoryPluginConfiguration+EndpointConfig.swift
```diff
@@ -322,4 +317,49 @@ extension Dictionary where Key == String, Value == AWSAPICategoryPluginConfigura
 
         return endpointConfig
     }
+
+    // Retrieve the endpoint configuration when there is only one endpoint of the specified `endpointType`
```
palpatim, line 321: nit: doc comment
```suggestion
    /// Retrieve the endpoint configuration when there is only one endpoint of the specified `endpointType`
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Configuration/AWSAPICategoryPluginConfiguration+EndpointConfig.swift
```diff
@@ -322,4 +317,49 @@ extension Dictionary where Key == String, Value == AWSAPICategoryPluginConfigura
...
+
+            if apiForEndpointType.count > 1 {
+                throw APIError.invalidConfiguration(
+                    "More than one \(endpointType) API configured. Could not infer which API to call",
+                    "Use the apiName to specify which API to call")
+            }
+            return endpointConfig.value
+    }
+
+    // Retrieve the endpoint only if there is a single one, with GraphQL taking precedent over REST.
```
palpatim, line 341: nit: doc comment

```suggestion
    /// Retrieve the endpoint only if there is a single one, with GraphQL taking precedent over REST.
```

palpatim (approved):
Approved w/changes

AmplifyPlugins/API/AWSAPICategoryPlugin/Configuration/AWSAPICategoryPluginConfiguration+EndpointConfig.swift
lawmicha: yes for this one i'll remove, the other ones below do provide a bit more context, such as taking GraphQL over REST

PR #1182, opened 2021-04-24
fix: update SDK to 2.23.4 with AWSServiceConfiguration init override

lawmicha:
*Issue #, if available:*
- https://github.com/aws-amplify/amplify-flutter/issues/539
- Also tested in a sample app with Amplify and CognitoAuthPlugin at 1.8.1 breaks with SDK 2.23.4

*Description of changes:*
SDK update (PR https://github.com/aws-amplify/amplify-ios/pull/1181) will not fix the problem without the overriding the init from AWSServiceConfiguration, and the override cannot be added until the AmplifyPlugins depends on the updated SDK.

I'd opt to merge this PR with both commits since it's relatively small, and i've verified with a sample app that it works.

*Check points:*

- [x] All unit tests pass


AWSPluginsCore.podspec
```diff
@@ -10,7 +10,7 @@
 $AMPLIFY_VERSION = '1.8.1'
```
drochetti, line 10: Is the amplify version changing too with the release of this change? I'm a little unclear on the actual root cause here, why a patch version on the SDK breaking consumers of Amplify?

lawmicha:
We discussed this and the path forward is to 
1. Discard this PR
2. Revert the SDK feature that is released incorrectly as a patch version bump
3. Release a new patch so consumers of SDK 2.23.x can pick up a non breaking version. This is released as 2.23.5
4. Add the change back in, and release a minor version 2.24.0 for the SDK

PR #1192, opened 2021-04-28
Increase `maxConcurrentOperationCount` to 20

gavynriebau:
*Description of changes:*

Increase the `maxConcurrentOperationCount` for initial DataStore sync from `1` to `20`.

This fixes an issues where an application I am developing for work is slow to start due to the current DataStore behaviour of serially loading model data.


palpatim:
This change is not safe--we intentionally limit the concurrency of sync operations in order to invoke model sync operations according to their dependency order. Without this limit, it would be possible for the system to sync a model with an unfulfilled dependency, resulting in consistency errors.

As noted on https://github.com/aws-amplify/amplify-ios/pull/1192/files#R43, we do want to parallelize the operations to an extent, but the right way to do this is by parallelizing groups of sync operations according to their "dependency root", so if we have 5 models (A->B->C D->E), we would have 2 parallel sync threads, one syncing the models C, B, A; the other syncing E, D.

PR #1208, opened 2021-05-06
fix(Analytics): identifyUser add attributes to AWSPinpointEndpointPro…

lawmicha:
…fileUser

*Issue #, if available:*
Android PR: https://github.com/aws-amplify/amplify-android/pull/1306

*Description of changes:*
This is a POC reflecting one option for plugin to pass `properties` to the pinpoint specific type

*Check points:*

- [ ] All unit tests pass


rjuliano:
Would this implementation add all attributes both as a custom as well as user attribute?

lawmicha:
> Would this implementation add all attributes both as a custom as well as user attribute?

At the time of this draft, i was thinking we should be backwards compatible with the existing implementation so if users expect attributes are being added to `AWSPinpointEndpointProfile ` then the library should continue to do so. removing would be a breaking behavior change, unless it wasn't useful and was a bug

rjuliano:
> > Would this implementation add all attributes both as a custom as well as user attribute?
> 
> At the time of this draft, i was thinking we should be backwards compatible with the existing implementation so if users expect attributes are being added to `AWSPinpointEndpointProfile ` then the library should continue to do so. removing would be a breaking behavior change, unless it wasn't useful (for example, if developer has no way to use this information from the pinpoint console) and was a bug, and should have been inside `AWSPinpointEndpointProfile.user`, then we could remove it (remove adding to `AWSPinpointEndpointProfile` and only add it to `AWSPinpointEndpointProfile.user`) and treat it like a fix

Understood. The idea is to give callers a way to specify whether they want to add custom attributes or user attributes. Custom attributes are related to the endpoint being identified, while user attributes are associated with the user linked to that endpoint. I agree that anything passed in via the `properties` parameter of `AnalyticsUserProfile` should continue to work as is. The goal here is to give them an option to pass attributes via a separate parameter: `userAttributes` for example. 

The distinction affects how these attributes show up in the Pinpoint console when the user is creating audiences. In the screenshot below, `sub` and `name` are attributes that were duplicated under both custom and user attributes. `myproperty` is an example of a property added via the existing functionality, while `CognitoClientId` is an example of a property added as a user attribute.

image

lawmicha:
> The idea is to give callers a way to specify whether they want to add custom attributes or user attributes

this makes sense, developers should have control over what they want to achieve and see in the Pinpoint console for a particular user. 

> The goal here is to give them an option to pass attributes via a separate parameter: userAttributes for example.

I think this is a valid option:
AnalyticsUserProfile.properties -> Pinpoint's custom attributes
AnalyticsUserProfile.userAttributes -> Pinpoint's user attributes

Does this still make sense for other implementations of Analytics plugins? Are we making it simple for the customer but exposing too much of the Pinpoint term "user attributes"?

If "User attributes" is what developers actually want, the solution i imagine is
```
   // 1. Endpoint attributes
   let location = AnalyticsUserProfile.Location(latitude: 47.606209,
                                                 longitude: -122.332069,
                                                 postalCode: "98122",
                                                 city: "Seattle",
                                                 region: "WA",
                                                 country: "USA")
    // 2. properties should really go to Pinpoint's "user attributes"
    let properties: AnalyticsProperties = ["phoneNumber": "+11234567890", "age": 25]

    let userProfile = AnalyticsUserProfile(name: username,
                                           email: "name@example.com",
                                           location: location,
                                           properties: properties)

    // 3. custom attributes are pinpoint specific
    let pluginOptions = AWSPinpointPluginOptions(customAttributes: ["myProperty": "value"])
    let options = AnalyticsIdentifyUserOptions(pluginOptions: pluginOptions)
    Amplify.Analytics.identifyUser(user.userId, withProfile: userProfile, options: customAttributes) 
```

rjuliano:
> Does this still make sense for other implementations of Analytics plugins? Are we making it simple for the customer but exposing too much of the Pinpoint term "user attributes"?

I see what you're saying there. I like the solution you proposed but it definitely means a major version upgrade I think. 

The other possibility that comes to mind would be to introduce option property prefixes. If property names are not prefixed (or prefixed with `custom:`), we treat them as custom endpoint properties. If they have the prefix `user:` we treat them as user attributes.

Not super elegant, but it does unblock the customer and it's a non-breaking change.

rjuliano:
Here's the implementation using the string prefix. 

https://github.com/aws-amplify/amplify-android/pull/1395

lawmicha:
> Here's the implementation using the string prefix.
> 
> [aws-amplify/amplify-android#1395](https://github.com/aws-amplify/amplify-android/pull/1395)

great idea!

lawmicha:
closing this for tracking issue: https://github.com/aws-amplify/amplify-ios/issues/1416

PR #1217, opened 2021-05-11
fix(DataStore): redundant local metadata query in ReconcileAndLocalSaveOperation

lawmicha:
*Issue #, if available:*

*Description of changes:*
This removes a redundant query performed in ReconcileAndLocalSaveOperation by extending the logic in  `RemoteSyncReconciler` to produce a disposition `.applyRemoteModel` containing the mutationType. A new behavior has also been introduced: When there is no local metadata, drop the remote model if it has been deleted (`remoteModel.syncMetadata.deleted == true`). Originally, the deleted remote model would result in an apply- calling storageAdapter.delete and saving the metadata containing the deleted version. In this change, the remote model will not result in the two SQL write attempts (DELETE model and INSERT metadata), but instead will be dropped. If a model is deleted and does not exist locally, it is unnecessary to store that metadata locally, since a deleted model cannot be undeleted.

*Current system - ReconcileAndLocalSaveOperation*:

1. *Query* for the local metadata by model.id
2. *Query* for the pending mutation events by model.id
3. When do we apply the remote model to the local store?
    1. If there are pending mutations, drop the remote model
    2. If there are no pending mutations, and..
        1. If there is no local metadata, _apply the remote model_
        2. If the remote model is greater in version than local metadata, _apply the remote model_
        3. there is a local model that has a higher version, drop the remote model
4. Applying remote model
    1. if remote model is deleted, *delete* the model from local store, and save metadata(5)
    2. or if remote model is not deleted, *save* the remote model (first *query* then *insert or update* the model), and save metadata(5)
5. Saving Metadata
    1. *query* for local metadata, and if there is local metadata, notify that applied model was a "create"
    2. save the metadata, which does a *query* and then an *insert/update*

*Proposed change*:

Remove step 5.1

*Check points:*

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
@@ -217,7 +217,7 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
 
         // TODO: Wrap this in a transaction
         if mutationType == .delete {
-            saveDeleteMutation(storageAdapter: storageAdapter, remoteModel: remoteModel, mutationType: mutationType)
+            saveDeleteMutation(storageAdapter: storageAdapter, remoteModel: remoteModel)
         } else {
             saveCreateOrUpdateMutation(storageAdapter: storageAdapter,
                                        remoteModel: remoteModel,
@@ -226,8 +226,7 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
     }
 
     private func saveDeleteMutation(storageAdapter: StorageEngineAdapter,
-                                    remoteModel: RemoteModel,
-                                    mutationType: MutationEvent.MutationType) {
+                                    remoteModel: RemoteModel) {
         log.verbose(#function)
 
         guard let modelType = ModelRegistry.modelType(from: modelSchema.name) else {
@@ -250,7 +249,7 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
             case .success:
                 self.saveMetadata(storageAdapter: storageAdapter,
                                   inProcessModel: remoteModel,
-                                  mutationType: mutationType)
+                                  mutationType: .delete)
             }
         }
     }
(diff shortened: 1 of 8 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
```diff
@@ -218,15 +216,18 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
         }
 
         // TODO: Wrap this in a transaction
-        if remoteModel.syncMetadata.deleted {
-            saveDeleteMutation(storageAdapter: storageAdapter, remoteModel: remoteModel)
+        if mutationType == .delete {
+            saveDeleteMutation(storageAdapter: storageAdapter, remoteModel: remoteModel, mutationType: mutationType)
```
palpatim, line 220: We know it's a delete mutation; do we need to pass the mutationType?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
```diff
@@ -218,15 +216,18 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
...
+                                       remoteModel: remoteModel,
+                                       mutationType: mutationType)
         }
-
     }
 
-    private func saveDeleteMutation(storageAdapter: StorageEngineAdapter, remoteModel: RemoteModel) {
+    private func saveDeleteMutation(storageAdapter: StorageEngineAdapter,
+                                    remoteModel: RemoteModel,
+                                    mutationType: MutationEvent.MutationType) {
```
palpatim, line 230: This will always be `.delete` right? Any reason we need it passed in?

palpatim (approved):
Approved w/minor comments

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
lawmicha: the requirement is to notify the hub of `syncReceived` with the mutationType inside a MutationEvent. it's done after the two SQL operations (delete or save/update and save metadata) are done successfully, which is why it's passed down. I see now that `saveDeleteMutation` can just call `saveMetadata` with `.delete` but saveCreateOrUpdateMutation will need it. I can make this change now, and work on making the methods clearer, ie. use callbacks instead of calling each method directly

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
lawmicha: looks like we don't need to, removing

PR #1218, opened 2021-05-12
fix(DataStore): optimize dropping remote model when pending mutations in ReconcileAndLocalSaveOperation

lawmicha:
*Issue #, if available:*

*Description of changes:*
This refactors the state machine in ReconcileAndLocalSaveOperation to optimize in dropping the remote model when there are pending mutations, so to avoid an extra query for local metadata

*Current system - ReconcileAndLocalSaveOperation*:

1. *Query* for the local metadata by model.id
2. *Query* for the pending mutation events by model.id
3. When do we apply the remote model to the local store?
    1. If there are pending mutations, drop the remote model
    2. If there are no pending mutations, and..
        1. If there is no local metadata, _apply the remote model_
        2. If the remote model is greater in version than local metadata, _apply the remote model_
        3. there is a local model that has a higher version, drop the remote model
4. Applying remote model
    1. if remote model is deleted, *delete* the model from local store, and save metadata(5)
    2. or if remote model is not deleted, *save* the remote model (first *query* then *insert or update* the model), and save metadata(5)
5. Saving Metadata, which does a *query* and then an *insert/update*

*Proposed change*:

Perform step 2 and check 3.1, before performing step 1, to look like the following:

1. *Query* for the pending mutation events by model.id
2. If there are pending mutations, drop the remote model
3. *Query* for the local metadata by model.id
     1. If there is no local metadata, _apply the remote model_
     2. If the remote model is greater in version than local metadata, _apply the remote model_
     3. there is a local model that has a higher version, drop the remote model
4. Applying remote model
    1. if remote model is deleted, *delete* the model from local store, and save metadata(5)
    2. or if remote model is not deleted, *save* the remote model (first *query* then *insert or update* the model), and save metadata(5)
5. Saving Metadata, which does a *query* and then an *insert/update*

**NOTE**

This PR is draft because the unit tests have not been written against the new state machine, and some of the older unit tests were commented out

*Check points:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation+Action.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation+Action.swift
@@ -15,15 +15,21 @@ extension ReconcileAndLocalSaveOperation {
         /// Operation has been started by the queue
         case started(RemoteModel)
 
-        /// Operation has retrieved RemoteModel's corresponding sync metadata from local database
-        case queried(RemoteModel, LocalMetadata?)
+        /// Operation has retrieved pending mutations
+        case queriedPendingMutations(RemoteModel, [MutationEvent])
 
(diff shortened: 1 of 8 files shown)


lawmicha:
This PR is superseded by https://github.com/aws-amplify/amplify-ios/pull/1237 because PR 1237 once checked in will perform the same optimization as this PR, it's unlikely we'll review and check this one one before that one

PR #1228, opened 2021-05-13
feat(datastore): support mutation (deletion) with custom primary keys

thisisabhash:
*Issue #, if available:* https://github.com/aws-amplify/amplify-ios/issues/1165

*Description of changes:*  Implemented changes for adding custom primary keys to "input" variable in a delete mutation request

*Check points:*

- [x] All unit tests pass


AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelDeleteDecorator.swift
```diff
@@ -0,0 +1,49 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+/// Decorate the GraphQLDocument with the value of `Model.Identifier` for a "delete" mutation with custom primary keys
```
diegocstn, line 11:
```suggestion
/// Decorate the GraphQLDocument with either a single id of `Model.Identifier` or with all fields part of a custom primary key for a "delete" mutation
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelDeleteDecorator.swift
```diff
@@ -0,0 +1,49 @@
...
+                for key in customPrimaryKeys {
+                    objectMap[key] = graphQLInput[key]
+                }
+                inputs["input"] = GraphQLDocumentInput(type: "\(document.name.pascalCased())Input!",
+                                                       value: .object(objectMap))
+
+            } else {
+                inputs["input"] = GraphQLDocumentInput(type: "\(document.name.pascalCased())Input!",
+                                                       value: .object(["id": model.id]))
+            }
```
diegocstn, line 42: maybe you can just build the input map in this if/else block and eventually instantiate and assign the `GraphQLDocumentInput`?
```
if let customPrimaryKeys = modelSchema.customPrimaryIndexFields {
....
} else {
  objectMap["id"] = model.id
}

inputs["input"] = GraphQLDocumentInput(type: "\(document.name.pascalCased())Input!", value: .object(objectMap))                                          
```

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelDeleteDecorator.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+/// Decorate the GraphQLDocument with either a single id of `Model.Identifier`
+/// or with all fields part of a custom primary key for a "delete" mutation
+public struct ModelDeleteDecorator: ModelBasedGraphQLDocumentDecorator {
```
lawmicha, line 13: as discussed you can just reuse `ModelIDDecorator`

lawmicha: rename ideas, `ModelIdsDecorator` `ModelPrimaryIdDecorator` or keep the same? `ModelIdDecorator`

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLRequest/GraphQLRequestAnyModelWithSyncTests.swift
```diff
@@ -10,11 +10,13 @@ import XCTest
 @testable import AmplifyTestCommon
 @testable import AWSPluginsCore
 
+// swiftlint:disable type_body_length
 class GraphQLRequestAnyModelWithSyncTests: XCTestCase {
 
     override func setUp() {
         ModelRegistry.register(modelType: Comment.self)
         ModelRegistry.register(modelType: Post.self)
+        ModelRegistry.register(modelType: CustomerOrder.self)
```
lawmicha, line 19: add a new test file with CustomerOrder, `GraphQLRequestSyncCustomPrimaryTests`
- get, list, create, update, delete.

lawmicha: keep in mind `get` query may pass but will fail AppSync service call, we would need to fix for successfully creating API integ test

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelDeleteDecorator.swift
```diff
@@ -0,0 +1,48 @@
...
+                }
+            } else {
+                objectMap["id"] = model.id
+            }
+
+            inputs["input"] = GraphQLDocumentInput(type: "\(document.name.pascalCased())Input!",
+                                                   value: .object(objectMap))
+        } else if case .query = document.operationType {
+            inputs["id"] = GraphQLDocumentInput(type: "ID!", value: .scalar(model.id))
+        }
```
lawmicha, line 44: inside `ModelIdDecorator`,
- else if query and customPrimaryKeys exists, update the `input[key]` instead of "input["id"]`
-

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelIdDecorator.swift
```diff
@@ -11,10 +11,15 @@ import Amplify
 /// Decorate the GraphQLDocument with the value of `Model.Identifier` for a "delete" mutation or "get" query.
 public struct ModelIdDecorator: ModelBasedGraphQLDocumentDecorator {
 
-    private let id: Model.Identifier
+    private let input: ModelDecoratorInput
 
-    public init(id: Model.Identifier) {
-        self.id = id
```
lawmicha, line 17: So I just realized that we're changing the public interface, so even though it's public from AWSPluginsCore, we still have to be careful to keep things backwards compatible. Developers could have been using these public decorators themselves to create a valid GraphQL document. we need to go back and figure out how we should address this

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelIdDecorator.swift
```diff
@@ -27,10 +39,23 @@ public struct ModelIdDecorator: ModelBasedGraphQLDocumentDecorator {
...
-            value: .object(["id": id]))
+                                                   value: .object(objectMap))
         } else if case .query = document.operationType {
             inputs["id"] = GraphQLDocumentInput(type: "ID!", value: .scalar(id))
+
+            if let fields = fields {
+                for (fieldName, value) in fields {
+                    inputs[fieldName] = GraphQLDocumentInput(type: "String!", value: .scalar(value))
+                }
+            }
```
diegocstn, line 58: if `id` is included in the provided `fields`, are we overriding its value in `inputs` set on line 52 but with a different type? (`String!` instead of `ID!`)

lawmicha: good point, that's what would happen in the current code. passing in `fields: ["id": "123"]` will have unintended behavior. Since this is a public facing type, we should make sure the constructor can be used in different ways without causing problems. I'm not sure what's a better interface that also keeps it backwards compatible. perhaps the best we can do right now is to make it work for the customer even when they pass in parameters in diferent ways

although this can be passed in, what is the expected behavior? `init(id: "111", fields: ["id": "222"])`` do they want the id to be "111" or "222". We should simply ignore the "id" keys in fields if it exists, and add doc comment to the initializer to provide an example of usage, indicating that fields is to contain additional index values

thisisabhash: Added changes to ignore `id` in `fields` when decoration is being performed.

lawmicha (approved):

lawmicha:
related PR fix for Flutter: https://github.com/aws-amplify/amplify-ios/pull/1292

PR #1230, opened 2021-05-14
fix: multiple models in transaction with state machine

lawmicha:
*Issue #, if available:*

*Description of changes:*

*Check points:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/StateMachine.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/StateMachine.swift
@@ -12,12 +12,15 @@ import Foundation
 @available(iOS 13.0, *)
 class StateMachine<State, Action> {
     typealias Reducer = (State, Action) -> State
+    typealias Responder = (State) -> Void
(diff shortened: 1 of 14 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/StateMachine.swift
```diff
@@ -36,12 +39,28 @@ class StateMachine<State, Action> {
...
+    ///
+    /// Call this method when you want to perform the state machine's `responder` closure with the new state resolved
+    /// from the current state and action.
+    func respond(action: Action) {
+        guard let responder = responder else {
+            log.error("Use of `StateMachine.respond` without an attached responder is a programming error.")
+            return
+        }
+        log.verbose("Responding: \(action)")
+        let newState = resolve(currentState: state, action: action)
```
lawmicha, line 59: maybe i can make this thread safe using a DispatchSemaphore, but acquiring and releasing a semaphore in a single ReconileAndSaveOperation would occur for each `stateMachine.respond` call, which is about 5 or so calls

PR #1237, opened 2021-05-17
feat(DataStore): Multiple models ReconcileAndLocalSave transaction

lawmicha:
*Issue #, if available:*

*Description of changes:*
The baseline performance test done in the previous PR: https://github.com/aws-amplify/amplify-ios/pull/1215 shows time to sync 10k items is around 100s, (+/- 20s)

This change convert ReconcileAndLocalSaveOperation to reconcile multiple models of the same model schema. When the sync engine starts and a full sync is required, sync query is performed for each model type and paginated based on a default (and service maximum limit) of 1000 items per page. When the sync query results are returned, a ReconcileAndLocalSaveOperation is created with up to 1000 items, before ReconcileAndLocalSaveOperation's would be created per model instance.

By reconciling 1000 items at once, performed in a SQL transaction, this speed things up to sync 10k items around 15s (+/- 5 s).

*Check points:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
@@ -289,12 +289,29 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
     }
 
     func queryMutationSyncMetadata(for modelId: Model.Identifier) throws -> MutationSyncMetadata? {
+        let results = try queryMutationSyncMetadata(forModelIds: [modelId])
+        return try results.unique()
+    }
+
+    func queryMutationSyncMetadata(forModelIds modelIds: [Model.Identifier]) throws -> [MutationSyncMetadata] {
         let modelType = MutationSyncMetadata.self
-        let statement = SelectStatement(from: modelType.schema, predicate: field("id").eq(modelId))
-        let rows = try connection.prepare(statement.stringValue).run(statement.variables)
-        let result = try rows.convert(to: modelType,
-                                      using: statement)
-        return try result.unique()
+        let fields = MutationSyncMetadata.keys
+        var results = [MutationSyncMetadata]()
+        let maxNumberOfPredicates = 950
+        let chunkedModelIdsArr = modelIds.chunked(into: maxNumberOfPredicates)
+        for chunkedModelIds in chunkedModelIdsArr {
+            var queryPredicates: [QueryPredicateOperation] = []
+            for id in chunkedModelIds {
+                queryPredicates.append(QueryPredicateOperation(field: fields.id.stringValue, operator: .equals(id)))
+            }
+            let groupedQueryPredicates =  QueryPredicateGroup(type: .or, predicates: queryPredicates)
+            let statement = SelectStatement(from: modelType.schema, predicate: groupedQueryPredicates)
+            let rows = try connection.prepare(statement.stringValue).run(statement.variables)
+            let result = try rows.convert(to: modelType,
+                                          using: statement)
+            results.append(contentsOf: result)
+        }
+        return results
     }
 
     func queryModelSyncMetadata(for modelSchema: ModelSchema) throws -> ModelSyncMetadata? {
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
@@ -18,6 +18,15 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
     private var dbFilePath: URL?
     static let dbVersionKey = "com.amazonaws.DataStore.dbVersion"
 
+    // TODO benchmark whether a SELECT FROM FOO WHERE ID IN (1, 2, 3...) performs measurably
+    // better than SELECT FROM FOO WHERE ID = 1 OR ID=2 OR ID=3
+    //
+    // SQLite supports up to 1000 expressions per SQLStatement. We have chosen to use 50 expressions
+    // less (equaling 950) than the maximum because it is possible that our SQLStatement already has
+    // some expressions.  If we encounter performance problems in the future, we will want to profile
+    // our system and find an optimal value.
+    var maxNumberOfPredicates: Int = 950
+
     convenience init(version: String,
                      databaseName: String = "database",
                      userDefaults: UserDefaults = UserDefaults.standard) throws {
@@ -289,22 +298,21 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
     }
 
     func queryMutationSyncMetadata(for modelId: Model.Identifier) throws -> MutationSyncMetadata? {
-        let results = try queryMutationSyncMetadata(forModelIds: [modelId])
+        let results = try queryMutationSyncMetadata(for: [modelId])
         return try results.unique()
     }
 
-    func queryMutationSyncMetadata(forModelIds modelIds: [Model.Identifier]) throws -> [MutationSyncMetadata] {
+    func queryMutationSyncMetadata(for modelIds: [Model.Identifier]) throws -> [MutationSyncMetadata] {
         let modelType = MutationSyncMetadata.self
         let fields = MutationSyncMetadata.keys
         var results = [MutationSyncMetadata]()
-        let maxNumberOfPredicates = 950
         let chunkedModelIdsArr = modelIds.chunked(into: maxNumberOfPredicates)
         for chunkedModelIds in chunkedModelIdsArr {
             var queryPredicates: [QueryPredicateOperation] = []
             for id in chunkedModelIds {
                 queryPredicates.append(QueryPredicateOperation(field: fields.id.stringValue, operator: .equals(id)))
             }
-            let groupedQueryPredicates =  QueryPredicateGroup(type: .or, predicates: queryPredicates)
+            let groupedQueryPredicates = QueryPredicateGroup(type: .or, predicates: queryPredicates)
             let statement = SelectStatement(from: modelType.schema, predicate: groupedQueryPredicates)
             let rows = try connection.prepare(statement.stringValue).run(statement.variables)
             let result = try rows.convert(to: modelType,
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
@@ -25,7 +25,7 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
     // less (equaling 950) than the maximum because it is possible that our SQLStatement already has
     // some expressions.  If we encounter performance problems in the future, we will want to profile
     // our system and find an optimal value.
-    var maxNumberOfPredicates: Int = 950
+    static var maxNumberOfPredicates: Int = 950
 
     convenience init(version: String,
                      databaseName: String = "database",
@@ -306,7 +306,7 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
         let modelType = MutationSyncMetadata.self
         let fields = MutationSyncMetadata.keys
         var results = [MutationSyncMetadata]()
-        let chunkedModelIdsArr = modelIds.chunked(into: maxNumberOfPredicates)
+        let chunkedModelIdsArr = modelIds.chunked(into: SQLiteStorageEngineAdapter.maxNumberOfPredicates)
         for chunkedModelIds in chunkedModelIdsArr {
             var queryPredicates: [QueryPredicateOperation] = []
             for id in chunkedModelIds {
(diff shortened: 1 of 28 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
```diff
@@ -137,186 +116,269 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
...
-    /// - errored
-    private func apply(remoteModel: RemoteModel, mutationType: MutationEvent.MutationType) {
-        if log.logLevel == .verbose {
-            log.verbose("\(#function): remoteModel")
-        }
+    func queryLocalMetadata(_ remoteModels: [RemoteModel]) -> Future<([RemoteModel], [LocalMetadata]), DataStoreError> {
+        Future<([RemoteModel], [LocalMetadata]), DataStoreError> { promise in
+            guard !self.isCancelled else {
+                self.log.info("\(#function) - cancelled, aborting")
+                return
```
palpatim, line 208: as above, need to fulfill the promise

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
```diff
@@ -137,186 +116,269 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
...
+            notifyDropped(modelName: remoteModel.model.modelName)
         }
+
+        return dispositions
     }
 
-    private func saveDeleteMutation(storageAdapter: StorageEngineAdapter,
-                                    remoteModel: RemoteModel) {
-        log.verbose(#function)
+    func applyRemoteModels(dispositions: [RemoteSyncReconciler.Disposition]) -> Future<Void, DataStoreError> {
```
palpatim, line 246: As discussed, there's a log going on in this method, add a TODO to refactor in a future PR. Although there is a bit of code dupe in the body of the `switch`, I see its purpose and it's probably not worth the effort to abstract it. It would be nice to move each of the publisher constructions to its own utility method though for readability of the `switch`, and a single method that you can invoke in the `map`:

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
```diff
@@ -137,186 +116,269 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
...
+        Future<Void, DataStoreError> { promise in
+            guard !self.isCancelled else {
+                self.log.info("\(#function) - cancelled, aborting")
+                return
```
palpatim, line 250: As above, need to fulfill the promise

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
```diff
@@ -137,186 +116,269 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
...
+                                   Future<ReconcileAndLocalSaveOperation.RemoteModel, DataStoreError>> in
+
+                switch disposition {
+                case .create(let remoteModel):
+                    let publisher = self.save(storageAdapter: storageAdapter,
+                                              remoteModel: remoteModel)
+                        .flatMap { inProcessModel in
+                            self.saveMetadata(storageAdapter: storageAdapter,
+                                                    inProcessModel: inProcessModel,
+                                                    mutationType: .create)
```
palpatim, line 273: nit here & below: indentation

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/RemoteSyncReconciler.swift
```diff
@@ -12,39 +12,102 @@ import Amplify
...
+
+        return remoteModels.filter { remoteModel in
+            !pendingMutationModelIds.contains(remoteModel.model.id)
+        }
+    }
 
-        guard pendingMutations.isEmpty else {
-            return .dropRemoteModel(remoteModel.model.modelName)
+    /// Reconciles the incoming `remoteModels` against the local metadata.
+
```
palpatim, line 46: nit
```suggestion
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/RemoteSyncReconciler.swift
```diff
@@ -12,39 +12,102 @@ import Amplify
...
+            localMetadataMap.updateValue(localMetadata, forKey: localMetadata.id)
+        }
+
+        remoteModels.forEach { remoteModel in
+            if let disposition = reconcile(remoteModel, localMetadata: localMetadataMap[remoteModel.model.id]) {
+                dispositions.append(disposition)
+            }
+        }
+
+        return dispositions
```
palpatim, line 80: Can simplify this with std lib array methods:
```suggestion
        guard !remoteModels.isEmpty else {
            return []
        }

        guard !localMetadatas.isEmpty else {
            let dispositions = remoteModels.compactMap { reconcile($0, localMetadata: nil) }
            return dispositions
        }

        let metadataByID = localMetadatas.reduce(into: [:]) { $0[$1.id] = $1 }

        let dispositions = remoteModels.compactMap { reconcile($0, localMetadata: metadataByID[$0.model.id]) }

        return dispositions
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/RemoteSyncReconciler.swift
```diff
@@ -12,39 +12,102 @@ import Amplify
...
+    /// `.update` disposition
+    /// if there is local metadata for the corresponding remote model, and the remote model is deleted, apply a
+    /// `.delete` disposition
+    ///
+    /// - Parameters:
+    ///   - remoteModel: model retrieved from the remote store
+    ///   - localMetadata: metadata corresponding to the remote model
+    /// - Returns: disposition of the model, `nil` if to be dropped
+    static func reconcile(_ remoteModel: RemoteModel, localMetadata: LocalMetadata?) -> Disposition? {
         guard let localMetadata = localMetadata else {
```
palpatim, line 97: Can clarify & simplify this whole method:
```swift
        guard let localMetadata = localMetadata else {
            return remoteModel.syncMetadata.deleted ? nil : .create(remoteModel)
        }

        guard remoteModel.syncMetadata.version >= localMetadata.version else {
            return nil
        }

        return remoteModel.syncMetadata.deleted ? .delete(remoteModel) : .update(remoteModel)
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Query.swift
```diff
@@ -6,17 +6,57 @@
...
                                       storageAdapter: StorageEngineAdapter,
                                       completion: DataStoreCallback<[MutationEvent]>) {
+
+        pendingMutationEvents(forModelIds: [modelId], storageAdapter: storageAdapter, completion: completion)
+    }
+
+    static func pendingMutationEvents(forModelIds modelIds: [Model.Identifier],
+                                      storageAdapter: StorageEngineAdapter,
+                                      completion: DataStoreCallback<[MutationEvent]>) {
         let fields = MutationEvent.keys
```
palpatim, line 22: As discussed, let's move `maxNumberOfPredicates` into a constant on `StorageAdapter` so we can reuse across our code. We'll also explore making this asynchronous, or moving to the `run` idiom we use above, so we're not using multiple ways of doing synchronous database calls in our code.

As above, **_after_** we get this working, let's benchmark whether an `IN` query would performs better than explicitly declaring multiple `=` conditions. (Again, it may not since I think the `IN` query basically translates to multiple equals behind the scenes.)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/SQLiteStorageEngineAdapterTests.swift
```diff
@@ -564,4 +564,8 @@ class SQLiteStorageEngineAdapterTests: BaseDataStoreTests {
 
         _ = UserDefaults.removeObject(userDefaults)
     }
+
+    func testQueryMutationSyncMetadata() {
+        // TODO
```
palpatim, line 569: as discussed:
```suggestion
        XCTFail("Test not yet written")
```

lawmicha: added missing tests

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
```diff
@@ -137,186 +116,269 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
...
+        let remoteModelIds = remoteModels.map { remoteModel in
+            remoteModel.model.id
+        }
 
-        if log.logLevel == .debug {
-            log.debug("query local metadata: \(stopwatch.lap())s")
+        do {
+            try storageAdapter.transaction {
+                queryPendingMutations(forModelIds: remoteModelIds)
+                    .flatMap { mutationEvents -> Future<([RemoteModel], [LocalMetadata]), DataStoreError> in
```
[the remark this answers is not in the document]
lawmicha: they're not at the same index because they may not be the same size, a remote model may not have local metadata. the tuple is representing: `(remoteModelsToApply, localMetadatas)` like "these are the remaining remote models we might need to, and here's the queried local metadatas (if exists) for those models"
```swift
queryPendingMutations(forModelIds: remoteModelIds) // get the pending mutations
                    .flatMap { mutationEvents -> Future<([RemoteModel], [LocalMetadata]), DataStoreError> in
                        let remoteModelsToApply = self.reconcile(remoteModels, pendingMutations: mutationEvents)
                        return self.queryLocalMetadata(remoteModelsToApply)
                    }
                    .flatMap { (remoteModelsToApply, localMetadatas) -> Future<Void, DataStoreError> in
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -18,6 +18,15 @@ final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
...
     static let dbVersionKey = "com.amazonaws.DataStore.dbVersion"
 
+    // TODO benchmark whether a SELECT FROM FOO WHERE ID IN (1, 2, 3...) performs measurably
+    // better than SELECT FROM FOO WHERE ID = 1 OR ID=2 OR ID=3
+    //
+    // SQLite supports up to 1000 expressions per SQLStatement. We have chosen to use 50 expressions
+    // less (equaling 950) than the maximum because it is possible that our SQLStatement already has
+    // some expressions.  If we encounter performance problems in the future, we will want to profile
+    // our system and find an optimal value.
+    var maxNumberOfPredicates: Int = 950
```
palpatim, line 28:
```suggestion
    static let maxNumberOfPredicates = 950
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/RemoteSyncReconciler.swift
```diff
@@ -19,15 +19,15 @@ struct RemoteSyncReconciler {
         case delete(RemoteModel)
     }
 
-    /// Reconciles the incoming `remoteModels` against the pending mutations.
+    /// Filter  the incoming `remoteModels` against the pending mutations.
```
palpatim, line 22:
```suggestion
    /// Filter the incoming `remoteModels` against the pending mutations.
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Query.swift
palpatim: > in regards to the async pattern, what do you mean by the run idiom above?

Above, rather than using a `DispatchSemaphore`, we drop down to the underlying `run` method that synchronously executes arbitrary SQL. My suggestion is that we pick a pattern to do synchronous data access, and use that pattern everywhere, rather than jumping around between patterns.

That said, there may be good reasons to keep them separate: maybe the place we use `run` above requires custom SQL, and isn't supported via the adaptor. Maybe the things we're trying to express in the code where we use `DispatchSemaphore` is clearer by using the StorageAdaptor methods, which aren't synchronous. So this is just a suggestion that we explore it. If it makes the code more understandable to have different patterns, keep them (but maybe document at the call site why we depart from the standard pattern)

palpatim (approved):
Minor nits, LGTM overall

PR #1241, opened 2021-05-20
feat: increase sync concurrecny count based on associations

lawmicha:
*Issue #, if available:*

*Description of changes:*
There is a single thread allocated for two different queues, the Sync queue and the Reconcilliation queue. When the sync operations produce a reconcillliation operation containing 1000 items in memory, the resource spent on saving to disk causes the sync thread to get deprioitized. The result of this is most of the time spent writing to disk while network traffic trickles in.

This PR increases the thread count when we can determine that it can be optimized, for example when the models have no associations among them, then they can be retrieved and reconciled in any order. If it cannot be optimized (models with associations) then it falls back to the 1 thread per sync and recon queue. With the optimization, setting the thread to the number of models will immedately kick off one API call per sync operation, getting the results in memory (a reconcilliation operation) in the FIFO reconcilliation queue. Because they are reconcilled in order of the insertion into the queue, models with little data will not be blocked by models with a lot of data.

Baseline testing (TBD):
- AModel, 200k items
- BModel, 100k
- CModel, 10k
- DModel, 1k
- EModel, 100

total ModelSync time for each

*Check points:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Model+SQLite.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Model+SQLite.swift
@@ -182,4 +182,8 @@ extension Array where Element == ModelSchema {
         return sortedKeys.map { sortMap[$0]! }
     }
 
+    func containsAssociations() -> Bool {
+        // TODO: actually check for associations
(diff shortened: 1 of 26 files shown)


lawmicha:
this has been merged in over in https://github.com/aws-amplify/amplify-ios/pull/1267

PR #1244, opened 2021-05-25
chore(docs): cleanup API docs

ameter:
*Description of changes:*

Modified the jazzy config to use symbol declarations as documentation for all uncommented public symbols, remove documentation coverage metric from the generated docs, add a docset icon, and include build tool settings in the config so they no longer have to be passed on the command line. Corrected a typo in a documentation comment in Amplify.swift.

Screen Shot 2021-05-25 at 5 49 40 PM

*Check points:*

- [x] All unit tests pass


.jazzy.yaml
```diff
@@ -3,9 +3,19 @@ author: Amazon Web Services
...
 skip_undocumented: false
+undocumented_text: ""
+hide_documentation_coverage: true
 sdk: iphone
 theme: apple
+build_tool_arguments:
+  - -workspace
+  - Amplify.xcworkspace
+  - -scheme
+  - Amplify
```
lawmicha, line 18: Do these changes update the command we should be using to generate the docs? 

```
// Before
jazzy --output='./docs' --min-acl=public -x -workspace,./Amplify.xcworkspace,-scheme,Amplify
// With this change? Is it possible to put the `output` and `min-acl` over in this yaml file as well?
jazzy --output='./docs' --min-acl=public 
```

.jazzy.yaml
```diff
@@ -3,9 +3,19 @@ author: Amazon Web Services
...
+
+docset_icon: readme-images/icon.png
```
lawmicha, line 20: where did this come from? how about `amplify-icon.png` for clarity?

.jazzy.yaml
ameter: Yes, these changes make all command line arguments unnecessary.  --output='./docs' and --min-acl=public are the defaults, so they were never actually required unless they were overriding something from the config.

You can now generate the docs simply by running `jazzy` with no args.  However, in order to make the readme pictures show up in the docs, we need to symlink the directory containing them in the docs directory, so something like the following: `jazzy && ln -s ../readme-images docs`.

We should really update the Fastlane release process to generate the docs so that it's not a manual step at all, but in the meantime, I will update the release process wiki once we get these changes merged.

.jazzy.yaml
ameter: This is the favicon from the main amplify docs site (https://docs.amplify.aws), which you can bounce back and forth between as you navigate the main docs and api docs, so including it here would improve consistency.  With that said, I'm not sure this will show as the favicon because it doesn't show up when testing locally and I don't have a readily available test web server to test it on.  It doesn't hurt to include it though.  I used icon.png because that's the traditional name for a page's favicon, but it can be anything.  I'll rename it.

lawmicha (approved):

PR #1256, opened 2021-06-01
Integration test in Datastore for custom primary keys

thisisabhash:
*Issue #, if available:* None

*Description of changes:* Integration tests in Datastore for custom primary key

*Check points:*

- [x] All unit tests pass


lawmicha:
to get the correct diff with only the datastore integ tests, we'll have to set base branch to `custom-primary-key-datastore-api-integ` since this builds on top of the existing commits with the APi integ tests https://github.com/aws-amplify/amplify-ios/pull/1246

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreCustomPrimaryKeyTests.swift
```diff
@@ -0,0 +1,109 @@
...
+                }
+                XCTAssertEqual(order.id, customerOrder.id)
+                XCTAssertEqual(order.orderId, customerOrder.orderId)
+                XCTAssertEqual(order.email, customerOrder.email)
+                queryBeforeDeleteExpectation.fulfill()
+            case .failure(let error):
+                print("Error : \(error)")
+            }
+        }
+        wait(for: [queryBeforeDeleteExpectation], timeout: networkTimeout)
```
lawmicha, line 91: can we incorporate another DataStore.save to get an update mutation going? Ensuring update mutation works is important even though currently the update mutation has the same API signature of the create mutation

The other reason why we shoud start testing update mutations as well is looking forward to the work for "per changed fields model updates". When we do that work, it will be closely tied to custom primary keys since a model with only changed fields will include both the `id` and `orderId` regardless of which fields were changed.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreCustomPrimaryKeyTests.swift
```diff
@@ -0,0 +1,109 @@
...
+class DataStoreCustomPrimaryKeyTests: SyncEngineIntegrationTestBase {
+
+    /// - Given: API has been setup with CustomerOrder model registered
+    /// - When: A new customer order is
+    ///     - saved and queried - should return the saved model
+    ///     - deleted
+    ///     - queried
+    /// - Then: The model should be deleted finally and the sync events should be received in order
+
+    func testCreateQueryDelete() throws {
```
lawmicha, line 26: suggestion `func testDeleteModelWithCustomPrimaryKey()`

lawmicha (approved):

PR #1258, opened 2021-06-02
feat(datastore): support for @auth provider

diegocstn:
*Description of changes:*
This PR introduces changes to capture the value of `provider` in @auth directives as an `AuthRule` metadata.

*Check points:*

- [x] All unit tests pass

### DIFF
--- a/Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
+++ b/Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
@@ -25,6 +25,14 @@ public enum ModelOperation {
 
 /// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly
 ///   by host applications. The behavior of this may change without warning.
+public enum AuthRuleProvider {
+    case apiKey
+    case oidc
+    case custom(name: String)
+}
+
+/// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly
+///   by host applications. The behavior of this may change without warning.
 public typealias AuthRules = [AuthRule]
 
 /// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly
@@ -37,6 +45,7 @@ public struct AuthRule {
     public let groups: [String]
     public let groupsField: String?
     public let operations: [ModelOperation]
+    public let provider: AuthRuleProvider?
 
     public init(allow: AuthStrategy,
                 ownerField: String? = nil,
@@ -44,6 +53,7 @@ public struct AuthRule {
                 groupClaim: String? = nil,
                 groups: [String] = [],
                 groupsField: String? = nil,
+                provider: AuthRuleProvider? = nil,
                 operations: [ModelOperation] = []) {
         self.allow = allow
         self.ownerField = ownerField
@@ -51,6 +61,7 @@ public struct AuthRule {
         self.groupClaim = groupClaim
         self.groups = groups
         self.groupsField = groupsField
+        self.provider = provider
         self.operations = operations
     }
 }
--- a/Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
+++ b/Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
@@ -28,7 +28,8 @@ public enum ModelOperation {
 public enum AuthRuleProvider {
     case apiKey
     case oidc
-    case custom(name: String)
+    case iam
+    case userPools
 }
 
 /// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly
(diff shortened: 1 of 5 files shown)


Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
```diff
@@ -23,6 +23,14 @@ public enum ModelOperation {
     case read
 }
 
+/// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly
+///   by host applications. The behavior of this may change without warning.
+public enum AuthRuleProvider {
+    case apiKey
+    case oidc
+    case custom(name: String)
```
lawmicha, line 31: is the plan for codegen to generate strings like `.apiKey`, `oidc`, `.custom("iam")`, and `.custom("userPools")` in the Model+Schema file?

diegocstn: The plan is to generate enum cases and/or static method calls like `.apiKey`, `.oidc`, `.iam` `.userPools`.
`AuthRuleProvider` contains only generic providers, and `AWSAuthRuleProvider` adds AWS specific providers via static methods but at the call site won't make any difference

lawmicha: would the generate code need `import AWSPluginsCore`?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/AuthRule+Extension.swift
```diff
@@ -7,6 +7,44 @@
...
+    public func toAWSAuthorizationType() throws -> AWSAuthorizationType {
+        var authType: AWSAuthorizationType
+        switch self {
+        case .apiKey:
+            authType = .apiKey
+        case .oidc:
+            authType = .openIDConnect
+        case .custom(name: let name) where name == AuthRuleProvider.valueForIAM():
+            authType = .awsIAM
+
```
lawmicha, line 36:
```suggestion
```

lawmicha (approved):

Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
drochetti: > would the generated code need `import AWSPluginsCore`?

that's a good question... if we create a dependency between the generated models and `AWSPluginsCore` I wonder if the abstraction is worth it, given I don't think it's realistic to drift away from a non-AWS permission model (a security model that doesn't depend on Cognito would require the entire system to change).

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/AuthRule+Extension.swift
```diff
@@ -7,6 +7,43 @@
...
+    public static var iam: AuthRuleProvider {
+        .custom(name: valueForIAM())
+    }
+
+    public static var userPools: AuthRuleProvider {
+        .custom(name: valueForCognitoUserPool())
+    }
+
+    public func toAWSAuthorizationType() throws -> AWSAuthorizationType {
```
drochetti, line 27: what is the difference between `AWSAuthorizationType` and `AuthRuleProvider`? Is it the latter constrained to the metadata and the former used in the auth implementation?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/AuthRule+Extension.swift
```diff
@@ -7,6 +7,43 @@
...
+        case .oidc:
+            authType = .openIDConnect
+        case .custom(name: let name) where name == AuthRuleProvider.valueForIAM():
+            authType = .awsIAM
+        case .custom(name: let name) where name == AuthRuleProvider.valueForCognitoUserPool():
+            authType = .amazonCognitoUserPools
+        case .custom(name: let name):
+            throw DataStoreError.unknown(
+                "Invalid AWS authorization provider in schema @auth rule: \(name)",
+                "Verify your schema.")
```
drochetti, line 41: thinking out loud... given the need to handle unknown cases and the `import AWSPluginsCore`, I wonder if we should drop the AWS abstraction and just define all the supported cases in the `AuthRuleProvider` enum itself.

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/AuthRule+Extension.swift
diegocstn: Yes, that's correct. `AuthRuleProvider` defines values allowed by `@auth` rule specs

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/AuthRule+Extension.swift
diegocstn: yeah, I agree it's probably easier to drop the AWS abstraction and simplify things a bit

PR #1273, opened 2021-06-11
fix(DataStore): do not send completion event to publisher on DataStore.clear() and DataStore.stop()

thisisabhash:
*Issue #, if available:* https://github.com/aws-amplify/amplify-flutter/issues/395

*Description of changes:* `.finished` events are not sent to datastore plugin subscribers when `Datastore.clear()`  and `Datastore.stop` is called 

*Check points:*

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -157,7 +157,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         case .failure(let dataStoreError):
             dataStorePublisher.send(dataStoreError: dataStoreError)
         case .finished:
-            dataStorePublisher.sendFinished()
+            break
         }
     }
 
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -155,7 +155,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         }
         switch completed {
         case .failure(let dataStoreError):
-            dataStorePublisher.send(dataStoreError: dataStoreError)
+            log.error("Error: \(dataStoreError)")
         case .finished:
             break
         }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -149,13 +149,9 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
 
     @available(iOS 13.0, *)
     private func onReceiveCompletion(completed: Subscribers.Completion<DataStoreError>) {
-        guard let dataStorePublisher = self.dataStorePublisher else {
-            log.error("Data store publisher not initalized")
-            return
-        }
         switch completed {
         case .failure(let dataStoreError):
-            log.error("Error: \(dataStoreError)")
+            log.error("StorageEngine completed with error: \(dataStoreError)")
         case .finished:
             break
         }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
@@ -369,8 +369,8 @@ class AWSAPICategoryPluginTests: XCTestCase {
 
             let mockModel = MockSynced(id: "12345")
             try plugin.dataStorePublisher?.send(input: MutationEvent(model: mockModel,
-                                                                 modelSchema: mockModel.schema,
-                                                                 mutationType: .create))
+                                                                     modelSchema: mockModel.schema,
+                                                                     mutationType: .create))
 
             waitForExpectations(timeout: 1.0)
             sink.cancel()
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
@@ -386,7 +386,6 @@ class AWSAPICategoryPluginTests: XCTestCase {
     ///     - plugin.stop() is called
     ///     - a mutation event is sent
     /// - Then: The subscriber to plugin's publisher should receive the mutation
-
     func testStorageEngineStartStopSend() {
         let startExpectation = expectation(description: "Start Sync should be called with start")
         let stopExpectation = expectation(description: "Stop should be called")
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
@@ -6,6 +6,7 @@
 //
 
 import XCTest
+import AmplifyTestCommon
 
 @testable import Amplify
 @testable import AWSDataStoreCategoryPlugin
@@ -106,21 +107,34 @@ class AWSAPICategoryPluginTests: XCTestCase {
                                         dataStorePublisher: dataStorePublisher,
                                         validAPIPluginKey: "MockAPICategoryPlugin",
                                         validAuthPluginKey: "MockAuthCategoryPlugin")
+
+        let finishReceived = expectation(
+            description: "publisher should receive .finished completion event after stop() is called")
+
         do {
             try plugin.configure(using: nil)
-            XCTAssert(plugin.storageEngine == nil)
+            XCTAssertNil(plugin.storageEngine)
+
+            let sink = plugin.publisher.sink { completion in
+                switch completion {
+                case .finished:
+                    finishReceived.fulfill()
+                case .failure(let error):
+                    XCTFail("Error \(error)")
+                }
+            } receiveValue: { _ in }
 
             let semaphore = DispatchSemaphore(value: 0)
             plugin.start(completion: {_ in
-                XCTAssert(plugin.storageEngine != nil)
-                XCTAssert(plugin.dataStorePublisher != nil)
+                XCTAssertNotNil(plugin.storageEngine)
+                XCTAssertNotNil(plugin.dataStorePublisher)
                 semaphore.signal()
             })
             semaphore.wait()
 
             plugin.stop(completion: { _ in
-                XCTAssert(plugin.storageEngine == nil)
-                XCTAssert(plugin.dataStorePublisher == nil)
+                XCTAssertNil(plugin.storageEngine)
+                XCTAssertNil(plugin.dataStorePublisher)
                 semaphore.signal()
             })
             semaphore.wait()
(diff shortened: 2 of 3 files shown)


lawmicha:
- can you update the title? `observe()` API is related to other platforms and the API on iOS is "publisher()`. please update title to reflect the changes in the PR like "fix(DataStore): do not send publisher completion on DataStore.clear()`
- Can you link to the related issue for this fix? the flutter issue should be fine

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -157,7 +157,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         case .failure(let dataStoreError):
             dataStorePublisher.send(dataStoreError: dataStoreError)
         case .finished:
-            dataStorePublisher.sendFinished()
+            break
```
lawmicha, line 160: when/where causes this which we then call `dataStorePublisher.sendFinished()`? and then why change it to not do this?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -175,7 +175,7 @@ class AWSAPICategoryPluginTests: XCTestCase {
 
             plugin.clear(completion: { _ in
                 XCTAssert(plugin.storageEngine == nil)
-                XCTAssert(plugin.dataStorePublisher == nil)
+                XCTAssert(plugin.dataStorePublisher != nil)
```
lawmicha, line 178: nit: perfer `XCTAssertNil` and `XCTAssertNotNil`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -230,7 +230,7 @@ class AWSAPICategoryPluginTests: XCTestCase {
...
```
lawmicha, line 233: nit: perfer XCTAssertNil and XCTAssertNotNil

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -256,4 +256,67 @@ class AWSAPICategoryPluginTests: XCTestCase {
         }
         return count
     }
+
+    /// - Given: Datastore plugin
+    /// - When: Plugin is configured, started and cleared
+    /// - Then: Plugin's publisher should not received .finished and dataStorePublisher should not be nil
+
```
lawmicha, line 263: remove empty line

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -256,4 +256,67 @@ class AWSAPICategoryPluginTests: XCTestCase {
...
+        let finishNotReceived = expectation(description: "Finish should not be received")
+        finishNotReceived.isInverted = true
+        _ = plugin.publisher.sink { completion in
+            switch completion {
+            case .finished:
+                print("Finish received")
+                finishNotReceived.fulfill()
+            case .failure(let error):
+                print("Error \(error)")
```
lawmicha, line 294: should this be an XCTFail?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -256,4 +256,67 @@ class AWSAPICategoryPluginTests: XCTestCase {
...
+                                        dataStorePublisher: dataStorePublisher,
+                                        validAPIPluginKey: "MockAPICategoryPlugin",
+                                        validAuthPluginKey: "MockAuthCategoryPlugin")
+
+        let finishNotReceived = expectation(description: "Finish should not be received")
+        finishNotReceived.isInverted = true
+        _ = plugin.publisher.sink { completion in
+            switch completion {
+            case .finished:
+                print("Finish received")
```
lawmicha, line 291: please remove this print statement, and in general i think we want to keep print statements out because... to keep the console logging clean? TBD i'm not too sure what is the best practice here. but anyways in this case, the inverted fulfill should print out the message from the expectation anyways.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -256,4 +256,67 @@ class AWSAPICategoryPluginTests: XCTestCase {
...
+            return storageEngine
+        }
+        let dataStorePublisher = DataStorePublisher()
+        let plugin = AWSDataStorePlugin(modelRegistration: TestModelRegistration(),
+                                        storageEngineBehaviorFactory: storageEngineBehaviorFactory,
+                                        dataStorePublisher: dataStorePublisher,
+                                        validAPIPluginKey: "MockAPICategoryPlugin",
+                                        validAuthPluginKey: "MockAuthCategoryPlugin")
+
+        let finishNotReceived = expectation(description: "Finish should not be received")
```
lawmicha, line 286: somehing like 
```
description: "publisher should not receive `.finished` completion event after `clear()` is called"
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -256,4 +256,67 @@ class AWSAPICategoryPluginTests: XCTestCase {
         }
         return count
     }
+
+    /// - Given: Datastore plugin
+    /// - When: Plugin is configured, started and cleared
+    /// - Then: Plugin's publisher should not received .finished and dataStorePublisher should not be nil
+
+    func testStorageEngineClear() throws {
```
lawmicha, line 264: i was originally thinking whether this test should even exist and if the additional assert should just live in `testStorageEngineStartClearStart`. I think it should because the sequence start -> clear -> start should ensure that the publisher is not finished. so this test is a subset of that one

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -256,4 +256,67 @@ class AWSAPICategoryPluginTests: XCTestCase {
...
+        _ = plugin.publisher.sink { completion in
+            switch completion {
+            case .finished:
+                print("Finish received")
+                finishNotReceived.fulfill()
+            case .failure(let error):
+                print("Error \(error)")
+            }
+        } receiveValue: { event in
+            print(event)
```
lawmicha, line 297: nit: perfer not printing, perhaps we can remove the print if it doesn't matter in this test or replace it with an assert

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -230,7 +230,7 @@ class AWSAPICategoryPluginTests: XCTestCase {
```
lawmicha, line 230: please add inverted expectation on the publisher completion event in this test as well.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -175,7 +175,7 @@ class AWSAPICategoryPluginTests: XCTestCase {
```
lawmicha, line 175: please add the expectation that the publisher is finished for the test `testStorageEngineStartStopStart` (not the test at this comment, but above it)

lawmicha (commented):
I think we can ensure that the bug is fixed by making sure data can be observed after the clear() has completed. currently the test only makes sure that the published doesn't get a completion event, but we don't ensure that the publisher actually gets data back in this scenario:
```
// 1. set up observer
// 2. save
// 3. observer gets data
// 4. clear - observer is not finished.
// 5. save
// 6. observer gets data
```

I think we are missing steps 5 and 6 and this can probably be in a unit test against a real local DB or in the integration test (local or sync enabled).

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -155,7 +155,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
         }
```
lawmicha, line 155: i don't think we need the check for publisher at line 152 if we are no longer using it

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -155,7 +155,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
...
         switch completed {
         case .failure(let dataStoreError):
-            dataStorePublisher.send(dataStoreError: dataStoreError)
+            log.error("Error: \(dataStoreError)")
```
lawmicha, line 158: suggestion: log.error("StorageEngine completed with error: \(dataStoreError)")

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSAPICategoryPluginTests.swift
```diff
@@ -283,40 +329,53 @@ class AWSAPICategoryPluginTests: XCTestCase {
...
+                XCTAssertNil(plugin.storageEngine)
+                XCTAssertNotNil(plugin.dataStorePublisher)
                 semaphore.signal()
             })
             semaphore.wait()
+
+            let mockModel = MockSynced(id: "12345")
+            try plugin.dataStorePublisher?.send(input: MutationEvent(model: mockModel,
+                                                                 modelSchema: mockModel.schema,
+                                                                 mutationType: .create))
```
lawmicha, line 373: nit: fix indentation (CTRL+i)

lawmicha (commented):
i think the last thing to do is to sync up with Flutter and Android team on the expected behavior of the publishers when DataStore.stop is called. Even though the bug on Flutter side (https://github.com/aws-amplify/amplify-flutter/issues/395) was "The iOS plugin side, however, tears all of that down when DS.clear() is called." should apply to DataStore.stop as well. The question to answer is, 

1. Given iOS plugin will tear down the subscribers for DataStore.stop(), is it impacting Flutter's plugin for platform parity like clear() was?
2. What is Android's publisher behavior? https://docs.amplify.aws/lib/datastore/real-time/q/platform/android I see there is a message like "Observation complete." in one of the closures, what happens when DataStore.clear() and DataStore.stop() when using the Android plugin?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
lawmicha: As discussed, the datastore publisher was being finished twice, once in the callback of storageEngine.stopSync/storageEngine.clear and once in this OnReceviedCompletion block. There isn't a need to do this twice. By doing nothing to the `dataStorePublisher` at this point in the code, there is no impact to the behavior of DataStore.clear or stop, so we have decided to remove it

lawmicha (commented):
1. flutter code can remove the patch added: https://github.com/aws-amplify/amplify-flutter/pull/399 please coordinate with flutter team. if needed, open issue linking to this issue and them to track the change needed in flutter code
2. do iOS docs need to be updated to reflect the expected behavior of this API? Since not all platforms have the same behavior, it's worth describing the behavior on iOS in the docs

lawmicha (approved):

PR #1296, opened 2021-06-23
fix(DataStore): dynamic model support for cascade delete

lawmicha:
*Issue #, if available:*

*Description of changes:*
DataStore.delete API performs cascade delete by first deleting the model from local store. Before performing the delete transaction, it collects all the associated models and it's children recursively, and delete those as well. All local deletes (the parent and child objects) are then synced to the cloud.

The problem can be seen in Flutter when performing a cascade delete. There are various places in the flow where the system gets a model name from the wrong place. Wrong places to get the model name:
- modelType.modelName
- model.modelName (Similar PR fix here: https://github.com/aws-amplify/amplify-ios/pull/1292)

The correct place to get the model name is from the schema `modelSchema.name`. This is difficult to do in the cascade delete logic as currently the `recurseQueryAssociatedModels` method will return an array of `Model` objects, losing the schema in which it relates to. To avoiding accessing the name from the model instance, we store `[ModelName: [Model]]` instead of `[Model]` to perserve the model name that is needed to create the MutationEvent to sync to cloud.

*Check points:*

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+AnyModel.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+AnyModel.swift
@@ -10,20 +10,20 @@ import SQLite
 import Foundation
 
 extension Statement {
-    func convert(toUntypedModel modelSchema: ModelSchema,
-                 using statement: SelectStatement) throws -> [Model] {
+    func convertToUntypedModel(using modelSchema: ModelSchema,
+                               statement: SelectStatement) throws -> [Model] {
         var models = [Model]()
 
         for row in self {
-            let modelValues = try convert(row: row, to: modelSchema, using: statement)
-            let untypedModel = try convert(toAnyModel: modelSchema, modelDictionary: modelValues)
+            let modelValues = try convert(row: row, withSchema: modelSchema, using: statement)
+            let untypedModel = try convertToAnyModel(using: modelSchema, modelDictionary: modelValues)
             models.append(untypedModel)
         }
 
         return models
     }
 
-    private func convert(toAnyModel modelSchema: ModelSchema, modelDictionary: ModelValues) throws -> Model {
+    private func convertToAnyModel(using modelSchema: ModelSchema, modelDictionary: ModelValues) throws -> Model {
         let data = try JSONSerialization.data(withJSONObject: modelDictionary)
         guard let jsonString = String(data: data, encoding: .utf8) else {
             let error = DataStoreError.decodingError(
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+AnyModel.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+AnyModel.swift
@@ -10,20 +10,20 @@ import SQLite
 import Foundation
 
 extension Statement {
-    func convert(toUntypedModel modelType: Model.Type,
+    func convert(toUntypedModel modelSchema: ModelSchema,
                  using statement: SelectStatement) throws -> [Model] {
         var models = [Model]()
 
         for row in self {
-            let modelValues = try convert(row: row, to: modelType, using: statement)
-            let untypedModel = try convert(toAnyModel: modelType, modelDictionary: modelValues)
+            let modelValues = try convert(row: row, to: modelSchema, using: statement)
+            let untypedModel = try convert(toAnyModel: modelSchema, modelDictionary: modelValues)
             models.append(untypedModel)
         }
 
         return models
     }
 
-    private func convert(toAnyModel modelType: Model.Type, modelDictionary: ModelValues) throws -> Model {
+    private func convert(toAnyModel modelSchema: ModelSchema, modelDictionary: ModelValues) throws -> Model {
         let data = try JSONSerialization.data(withJSONObject: modelDictionary)
         guard let jsonString = String(data: data, encoding: .utf8) else {
             let error = DataStoreError.decodingError(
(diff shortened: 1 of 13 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+AnyModel.swift
```diff
@@ -36,7 +36,7 @@ extension Statement {
             throw error
         }
 
-        let instance = try ModelRegistry.decode(modelName: modelType.modelName, from: jsonString)
+        let instance = try ModelRegistry.decode(modelName: modelSchema.name, from: jsonString)
```
lawmicha, line 39: `modelType.modelName` to `modelSchema.name`.  I think this was done because the callers no longer passed the modelType, so this had to change because we are now passing modelSchema. I don't think it was broken before. Question for Dynamic Model types aka Flutter: does `modelType` have modelName?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -75,7 +78,7 @@ extension Statement: StatementModelConvertible {
             guard let (schema, field) = columnMapping[column] else {
                 logger.debug("""
                 A column named \(column) was found in the result set but no field on
-                \(modelType.modelName) could be found with that name and it will be ignored.
+                \(modelSchema.name) could be found with that name and it will be ignored.
```
lawmicha, line 81: This appears also not to be broken, but changed due to the fact that we are moving to passing in the schema instead of the type.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/StorageEngineAdapter+UntypedModel.swift
```diff
@@ -43,13 +43,13 @@ extension SQLiteStorageEngineAdapter {
         }
     }
 
-    func query(untypedModel modelType: Model.Type,
+    func query(modelSchema: ModelSchema,
                predicate: QueryPredicate? = nil,
                completion: DataStoreCallback<[Model]>) {
         do {
-            let statement = SelectStatement(from: modelType.schema, predicate: predicate)
+            let statement = SelectStatement(from: modelSchema, predicate: predicate)
```
lawmicha, line 50: this was broken because we were getting schema from modelType

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
```diff
@@ -109,57 +109,76 @@ extension StorageEngine {
...
+        var associatedModelsMap = [ModelName: [Model]]()
+        for associatedModel in associatedModels {
+            let modelName = associatedModel.0
+            let model = associatedModel.1
+            if var models = associatedModelsMap[modelName] {
+                models.append(model)
+                associatedModelsMap.updateValue(models, forKey: modelName)
+            } else {
+                associatedModelsMap.updateValue([model], forKey: modelName)
+            }
```
royjit, line 167:
```suggestion
           associatedModelsMap[modelName, default: []].append(model)
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
```diff
@@ -32,10 +32,10 @@ extension StorageEngine {
 
     func queryAndDeleteTransaction<M: Model>(_ modelType: M.Type,
```
royjit, line 33: where is this function used?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -66,7 +69,7 @@ extension Statement: StatementModelConvertible {
     }
 
     func convert(row: Element,
-                 to modelType: Model.Type,
```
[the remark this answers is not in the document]
lawmicha: updating to `withSchema` instead of `to`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine+DeleteTransaction.swift
lawmicha: this is used by the StorageEngine's delete API, so whenever a model is deleted, we perform a cascade delete by finding the associated children objects as well, deleting all models locally in a transaction, and then syncing them to the cloud

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Storage/StorageEngineTestsBase.swift
```diff
@@ -95,7 +95,7 @@ class StorageEngineTestsBase: XCTestCase {
     func deleteModelSynchronous<M: Model>(modelType: M.Type,
                                           withId id: String,
                                           where predicate: QueryPredicate? = nil,
-                                          timeout: TimeInterval = 1) -> DataStoreResult<M?> {
+                                          timeout: TimeInterval = 10) -> DataStoreResult<M?> {
```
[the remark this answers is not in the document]
lawmicha: this change isn't needed and was because i was stepping through tests that increasing it by a factor helped debug, since i would continue on breakpoints within 10 seconds. I opted to keep this here so that next time we have to debug, we don't have to change it again, and revert it back, unless there's a good reason to keep it 1 second. Keeping it 1s is probably better representation of the time it takes to perform this method though. let me know if u prefer reverting this change

diegocstn (approved):

PR #1299, opened 2021-06-24
fix(datastore): initalSync should be successful in case of unauthorized errors

diegocstn:
*Description of changes:*
As we support different @auth rules on models, if a model initial sync returns an `unauthorized` error we should consider the sync successful so that DataStore can reach a "ready" state.

*Check points:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
@@ -154,19 +154,16 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
     }
 
     private func makeCompletionResult() -> Result<Void, DataStoreError> {
-        if syncErrors.allSatisfy(isUnauthorizedError) {
+        if syncErrors.isEmpty || syncErrors.allSatisfy(isUnauthorizedError) {
             return .successfulVoid
         }
 
-        guard syncErrors.isEmpty else {
-            let allMessages = syncErrors.map { String(describing: $0) }
-            let syncError = DataStoreError.sync(
-                "One or more errors occurred syncing models. See below for detailed error description.",
-                allMessages.joined(separator: "\n")
-            )
-            return .failure(syncError)
-        }
-        return .successfulVoid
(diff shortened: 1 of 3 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -180,3 +188,38 @@ extension AWSInitialSyncOrchestrator: Resettable {
...
+        if case let .api(apiError, _) = datastoreError,
+           let responseError = apiError as? GraphQLResponseError<ResponseType>,
+           let graphQLError = graphqlErrors(from: responseError)?.first,
+           let errorTypeValue = errorTypeValueFrom(graphQLError: graphQLError),
+           case .unauthorized = AppSyncErrorType(errorTypeValue) {
+            return true
+        }
+        return false
+    }
+}
```
diegocstn, line 225: This is logic is a copy of what's in `AWSModelReconciliationQueue`. It should be moved to an extension of one of the errors (.i.e. `DataStoreError`) or to a shareable static method

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -150,6 +154,10 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
     }
 
     private func makeCompletionResult() -> Result<Void, DataStoreError> {
+        if syncErrors.allSatisfy(isUnauthorizedError) {
+            return .successfulVoid
+        }
```
lawmicha, line 159: what happens if syncErrors is empty, does it return true for `syncErrors.allSatisfy(isUnauthorizedError`? seems like logically this can also be inside the guard statement like
```
guard syncErrors.isEmpty else {
  if syncErrors.allSatisfy(isUnauthorizedError) {
            return .successfulVoid
     } else {
    // return failure
   }
} 
```

diegocstn: `allSatisfy` returns true for an empty array, refactored to return early

lawmicha (approved):

PR #1128, opened 2021-03-26
fix(DataStore): Reconcile and save serially

lawmicha:
*Issue #, if available:*

*Description of changes:*
problem: `ReconcileAndSaveOperation`

solution:

performance metrics:

optimizations:

*Check points:*

- [x] Added new tests to cover change, if needed
- [x] All unit tests pass
- [x] All integration tests pass
- [ ] Documentation update for the change if required
- [x] PR title conforms to conventional commit style

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1001, opened 2021-01-02
DataStore sync throws foreign key constraint errors


dyang:
**Describe the bug**
My iOS app uses DataStore for local data storage. I notice that whenever there's a need to do a full-sync, e.g. when a user logs in, or when the schemas are changed during development, the sync process might throw some "FOREIGN KEY constraint failed" errors. As a result, no data gets displayed in the app. Users have to terminate the app and re-launch it, then the sync seems to work. 

**To Reproduce**

In my case, I do:

- sign out, which runs `DataStore.clear` to clear local user data
- sign back in, and then runs `DataStore.start` to kick off a full-sync

I can reproduce this 100% with my app. 

**Observed Behavior**

No data gets displayed. Neither the `syncQueriesReady` event nor the `ready` event get sent. 

**Expected Behavior**

Sync completes without the above errors and send out the ready event(s). (I have some code that displays spinners when there's a potentially long sync)

**Stack Trace**

Here's what an error might look like:

```
2021-01-02 10:31:16.186053+0100 App [11981:8265777] [StateMachine<State, Action>] Notifying: errored(DataStoreError: The operation couldn’t be completed. (SQLite.Result error 0.)
Recovery suggestion: The operation couldn’t be completed. (SQLite.Result error 0.)
Caused by:
FOREIGN KEY constraint failed (code: 19))
```
Upon closer inspection, I find that the error was thrown at `SQLiteStorageEngineAdapter.save` method. Whenever this happens, it's because the failing object has a foreign key dependency on another object. In my case I have a Model called `Episode`, which depends on another object `Media`. The simplified model schema looks like this:

```
type Episode @model @auth(rules: [{ allow: owner}]) {
  id: ID!
  title: String!
  media: Media! @connection
}

type Media @model @auth(rules: [{ allow: owner}]) {
  id: ID!
  audioUrl: String
  s3Key: String
  mimeType: String
}
```

Most of the time the corresponding media record gets saved before the sync engine attempts to save its containing episode, but this is not always the case. I put a break point at the catch clause in the above mentioned `save` method, and when the foreign key error happens, I run a query against the local sqlite db with the media id, and the media record is not there. 

It's interesting that later in the sync process, that "missing" media record does eventually get saved, so this looks more like a race condition than data lost.

**Areas of the SDK you are using (AWSMobileClient, Cognito, Pinpoint, IoT, etc)?**

DataStore, Storage, Authentication

**Environment(please complete the following information):**
[...]

PR #1321, opened 2021-07-15
fix(DataStore): drop failed constraint violation reconciliations

lawmicha:
*Issue #, if available:*

https://github.com/aws-amplify/amplify-ios/issues/1157 -> cascade delete fails for various reasons causing remote store to be missing the parent object. datastore full sync will fail in ReconcileAndLocalSaveOperation with a foreign key constraint violation when saving the child object. The ready event is not sent due to this failure.

https://github.com/aws-amplify/amplify-ios/issues/1310 -> also related to cascade delete failing, possible a bug in the cascade delete, however this fix will still be applicable to get the system in a ready state.

https://github.com/aws-amplify/amplify-ios/discussions/1097#discussioncomment-471914 -> additional context from @anatomybook. Sounds like "children without parents" scenario is bound to happen, not just in cascade delete. references #1001 

https://github.com/aws-amplify/amplify-ios/issues/1001 -> there are two bugs here. the first one is related to sync-ing models in the correct order but reconciling them concurrently which results in being done in the incorrect order. The first fix was done in #1128 with `ReconcileAndSaveQueue`. The first bug was fixed but the second bug still exists. This PR fix would address the second bug- now that models are reconcilled in the correct order, if there is a missing parent, there wil still be a constraint violation error

https://github.com/aws-amplify/amplify-android/pull/1058 -> Android PR for additional context

*Description of changes:*
`ReconcileAndLocalSaveOperation` will reconcile remote models from different sources (sync query results or subscription events). When the item is reconciled to be applied locally and the SQL operation fails with a constraint violation, the entire operation is fails. As a result of this, the `ready` event is never fired. By dropping the constraint violation errors and emitting a dropped model event, the `ready` event calculation will eventually be emitted once all models have been reconciled successfully (whether it was dropped or applied).

*Check points:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/ModelStorageBehavior.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/ModelStorageBehavior.swift
@@ -44,3 +44,13 @@ protocol ModelStorageBehavior {
                          completion: DataStoreCallback<[M]>)
 
 }
+
+protocol ModelStorageErrorBehavior {
+    func shouldIgnoreError(error: DataStoreError) -> Bool
+}
+
+extension ModelStorageErrorBehavior {
+    func shouldIgnoreError(error: DataStoreError) -> Bool {
+        return false
+    }
+}
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/SQLiteResultError.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/SQLiteResultError.swift
@@ -24,7 +24,7 @@ enum SQLiteResultError {
     /// - statement: the statement which produced the error
     case error(message: String, code: Int32, statement: Statement?)
 
-    init?(dataStoreError: DataStoreError) {
+    init?(from dataStoreError: DataStoreError) {
         guard case let .invalidOperation(error) = dataStoreError,
               let resultError = error as? Result,
               case .error(let message, let code, let statement) = resultError else {
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/SQLiteResultError.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/SQLiteResultError.swift
@@ -1,0 +1,28 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import SQLite
+import Amplify
+
+/// Checks for specific SQLLite error codes
+/// See https://sqlite.org/rescode.html#primary_result_code_list for more details
+struct SQLiteResultError {
+
+    /// Constraint Violation, such as foreign key constraint violation, occurs when trying to process a SQL statement
+    /// where the insert/update statement is performed for a child object and its parent does not exist.
+    /// See https://sqlite.org/rescode.html#constraint for more details
+    static func isConstraintViolation(_ dataStoreError: DataStoreError) -> Bool {
+        guard case let .invalidOperation(error) = dataStoreError,
+              let resultError = error as? Result,
+              case .error(_, let code, _) = resultError,
+              code == SQLITE_CONSTRAINT else {
+            return false
+        }
+
+        return true
+    }
+}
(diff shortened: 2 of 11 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/SQLiteResultError.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import SQLite
+import Amplify
+
+/// Checks for specific SQLLite error codes
+/// See https://sqlite.org/rescode.html#primary_result_code_list for more details
+struct SQLiteResultError {
```
diegocstn, line 13: have you considered representing `SQLiteResultError` with an enum? And we can have a factory method to bridge a `DataStoreError` with it?

diegocstn:
I think that we may need a level of abstraction/indirection between an `SQLiteError` and the `ReconcileAndLocalSaveOperation`. 
`ReconcileAndLocalSaveOperation` at that point doesn't have any knowledge about the implementation of the `StorageAdapter` (which in this case happens to be a SQLite database). 

Can we perhaps have an associated error type on `StorageEngineAdapter` or a `shouldIgnoreError` method (or similar) that storage classes conforming to `StorageEngineAdapter` need to implement?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/SQLiteResultError.swift
lawmicha: example
```
extension DataStoreError
  static func toSQLiteError() -> SQLiteError? 
}

enum SQLiteError {
  case constraintViolation
  case other(message, code)

  init(error: DataStoreError) throw {
  }
}
```

anatomybook:
Hello,

could you kindly provide some insight as to when this will go onto main?

Thanks so much!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/ModelStorageBehavior.swift
```diff
@@ -44,3 +44,13 @@ protocol ModelStorageBehavior {
                          completion: DataStoreCallback<[M]>)
 
 }
+
+protocol ModelStorageErrorBehavior {
+    func shouldIgnoreError(error: DataStoreError) -> Bool
```
diegocstn, line 49: do you think conforming classes may also need the operation type as additional parameter here? 
I wonder that we may need to treat an error differently based on the context where it originated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/SQLiteResultError.swift
```diff
@@ -9,20 +9,33 @@ import SQLite
...
+    case constraintViolation(statement: Statement?)
+
+    /// Represents a SQLite specific [error code](https://sqlite.org/rescode.html)
+    ///
+    /// - message: English-language text that describes the error
+    /// - code: SQLite [error code](https://sqlite.org/rescode.html#primary_result_code_list)
+    /// - statement: the statement which produced the error
+    case error(message: String, code: Int32, statement: Statement?)
+
+    init?(dataStoreError: DataStoreError) {
```
diegocstn, line 27: nit: maybe `init?(from dataStoreError: DataStoreError) {`?

diegocstn (approved):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/ModelStorageBehavior.swift
lawmicha: what's the operation type you are referring to here?

diegocstn: I was referring to the "operation types" defined by `ModelStorageBehavior` like `query`, `save`, `delete` (but I don't think we have an actual Swift type for those)

lawmicha: ah got it, im not sure if the decision to ignore an error differs based on operation type. we can keep this in mind if the use case extends beyond constraint violation errors

diegocstn (approved):

PR #1355, opened 2021-08-04
fix(DataStore): Fix race condition in sync mutation to cloud

palpatim:
*Issue #, if available:* During some investigation into an unrelated issue, I encountered race conditions in the SyncMutationToCloudOperation. My original implementation incorrectly subclassed it as an `Operation` rather than an `AsynchronousOperation`, which meant that after I fixed retain cycles in that class, it was being released before completing the callback. Digging into that fix led to another issue where `ModelSyncedEmitter` wasn't properly invoking the completion callbacks during initial sync when the sync included dropped events.

*Description of changes:*

*Check points:*

- [X] All unit tests pass
- [X] All integration tests pass _(Excepting known failures as noted in https://github.com/aws-amplify/amplify-ios/issues/1354)_

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
@@ -14,17 +14,20 @@ import AWSPluginsCore
 /// not a retriable error. If it is, attempts a retry until either success or terminal failure. Upon success or
 /// terminal failure, publishes the event response to the appropriate ModelReconciliationQueue subject.
 @available(iOS 13.0, *)
-class SyncMutationToCloudOperation: Operation {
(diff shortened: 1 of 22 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
```diff
@@ -245,18 +285,13 @@ class SyncMutationToCloudOperation: Operation {
...
-    private func scheduleRetry(advice: RequestRetryAdvice) {
-        log.verbose("\(#function) scheduling retry for mutation")
-        mutationRetryNotifier = MutationRetryNotifier(advice: advice,
-                                                      networkReachabilityPublisher: networkReachabilityPublisher) {
-            self.sendMutationToCloud()
-            self.mutationRetryNotifier = nil
-        }
-        currentAttemptNumber += 1
-    }
-
```
palpatim, line 257: This was redundant. Removed in favor of the one below.

diegocstn (approved):

PR #1406, opened 2021-09-01
fix(Datastore): deadlock in sqlite Connection.sync

ashiemke:
Fix deadlock in sqlite Connection.sync by moving publishers off of "SQLite.Database" queue as described in https://github.com/aws-amplify/amplify-ios/issues/1405

*Description of changes:*
use `subscribe(on:)` in the transaction completion handler to move work off of the sqlite synchronous queue to the datastore's work queue.

A `receive(on:)` before the sink would probably work as well, but it seems like this work should be on the work queue rather than sqlite's queue.

I've validated this fixes my issue, but have no idea how to construct a test this would resolve.

*Check points:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
@@ -126,6 +126,7 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
         do {
             try storageAdapter.transaction {
                 queryPendingMutations(forModelIds: remoteModelIds)
+                    .subscribe(on: workQueue)
                     .flatMap { mutationEvents -> Future<([RemoteModel], [LocalMetadata]), DataStoreError> in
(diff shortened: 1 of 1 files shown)


palpatim:
Note to @aws-amplify/ios-team -- we could probably test this by writing a test that sends a bunch of `save()`s with `DispatchQueue.concurrentPerform`, where each `save` callback is matched with an individual expectation. If we don't get all expectations fulfilled within a short time, we're probably hitting the deadlock.

ashiemke:
@palpatim can I help get this merged? I'd like to get this into our app to fix issues we're seeing now in production.

ameter:
@ashiemke Thanks for reporting this issue and submitting this PR.  We are investigating this change and will provide an update soon.

lawmicha (approved):

lawmicha:
Hi @ashiemke, thanks for opening this PR. I added a PR with an integration test that performs a bunch of save's concurrently, with your fix included: https://github.com/aws-amplify/amplify-ios/pull/1415 . From my investigation, scheduling on the internal queue instead of defaulting to the SQL queue would definitely reduce the consuming the SQL thread. Although I couldn't reproduce the deadlock with the integ test, I don't think it should block this from merging.

yazoonic:
@lawmicha  is this fixed in v13.4 main branch? I still can reproduce the issue!

lawmicha:
this fix is on 1.14.0 branch

Issue #1206, opened 2021-05-06
Datastore delete immediately after modify doesn't work


bgomberg:
**Describe the bug**
Using datastore, if I perform a delete immediately after calling save() to modify an existing record, the result of the delete is `.success`, but the record still exists.

**To Reproduce**
1. Create a new record (in this case a `Post`) via `Amplify.DataStore.save()`
2. Wait for the new record to sync with the backend
3. Modify the new record (saving it via `Amplify.DataStore.save()`)
4. Immediately after the completion handler for the `save()` is called, delete the new record
5. Wait for things to sync with the backend
6. Observe that the new record was not deleted (but the modification was persisted)

**Expected behavior**
I would expect the row to have been deleted.

**Environment(please complete the following information):**
 - Amplify Framework Version: 1.9.0
 - Amplify Categories: API + DataStore
 - Dependency Manager: Cocoapods
 - Swift Version: 5.4
 - CLI Version: 4.50.0
 - Xcode Version: 12.5

**Device Information (please complete the following information):**
 - Device: iPhone SE
 - iOS Version: 14.5

**Additional context**
Here's an example unit test which reproduces this issue (based on the types from the getting started guide):
```swift
    func testSaveModifyDelete() throws {
        print("Clearing local DB")
        let clearExpectation = expectation(description: "Clear")
        Amplify.DataStore.clear() { result in
            switch result {
            case .success:
                clearExpectation.fulfill()
            case .failure(let error):
                XCTFail("Error saving post: \(error)")
            }
        }
        wait(for: [clearExpectation], timeout: 10)

        var post = Post(title: UUID().uuidString, status: .published, rating: 5)
        print("Creating post (id=\(post.id), title=\(post.title))")
        let saveExpectation = expectation(description: "Save")
        Amplify.DataStore.save(post) { result in
            switch result {
            case .success:
                saveExpectation.fulfill()
            case .failure(let error):
                XCTFail("Error saving post: \(error)")
            }
        }
        wait(for: [saveExpectation], timeout: 10)

        print("Syncing to cloud")
        sleep(10)

        // Reading posts
        print("Querying posts")
        let readExpectation = expectation(description: "Read")
        Amplify.DataStore.query(Post.self) { result in
            switch result {
            case .success(let posts):
                XCTAssertEqual(posts.count, 1)
                XCTAssertEqual(posts.first!.id, post.id)
                readExpectation.fulfill()
            case .failure(let error):
[...]

PR #1407, opened 2021-09-01
fix(Datastore): Consecutive Updates (Save, Sync, Update and Immediately Delete Scenario)

thisisabhash:
*Issue #, if available:* #1206

*Description of changes:* This PR contains fix for Consecutive Update (Save, Sync, Update and Immediately Delete Scenario). The head of mutation event table for a given model id is reconciled with the version of the model in the API response after the models in the API request and response are compared for equality.

*Check points:*

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
@@ -17,7 +17,7 @@ extension ModelSchema {
     /// Returns true if equal, false otherwise
     /// Currently, schemas where system timestamps fields (createdAt, updatedAt)
     /// are renamed using with `@model`'s `timestamps` attribute and explicitly
-    /// added to the input schema are not supported by this check since they are
+    /// added to the input schema are not supported by this check since they are not
     /// marked as "read-only" fields and will fail the check when the service generates
     /// and returns the value of `createdAt` or `updatedAt`.
     /// for e.g.
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
@@ -14,11 +14,9 @@ extension ModelSchema {
 
     /// Compare two `Model` based on a given `ModelSchema`
     /// Returns true if equal, false otherwise
-    public static func isEqual(_ model1: Model,
-                               _ model2: Model,
-                               _ modelSchema: ModelSchema) -> Bool {
-        for (fieldName, modelField) in modelSchema.fields {
-            // skip read only fields
+    public func isEqual(_ model1: Model, _ model2: Model) -> Bool {
+        for (fieldName, modelField) in fields {
+            // read only fields are skipped for eqaulity check as they are created by the service
             if modelField.isReadOnly {
                 continue
             }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
@@ -30,21 +30,21 @@ extension ModelSchema {
                 guard let value1Optional = value1 as? String?, let value2Optional = value2 as? String? else {
                     return false
                 }
-                if !modelCompareHelper(value1Optional, value2Optional) {
+                if !isEqual(value1Optional, value2Optional) {
                     return false
                 }
             case .int:
                 guard let value1Optional = value1 as? Int?, let value2Optional = value2 as? Int? else {
                     return false
                 }
-                if !modelCompareHelper(value1Optional, value2Optional) {
+                if !isEqual(value1Optional, value2Optional) {
                     return false
                 }
             case .double:
                 guard let value1Optional = value1 as? Double?, let value2Optional = value2 as? Double? else {
                     return false
                 }
-                if !modelCompareHelper(value1Optional, value2Optional) {
+                if !isEqual(value1Optional, value2Optional) {
                     return false
                 }
             case .date:
@@ -52,7 +52,7 @@ extension ModelSchema {
                       let value2Optional = value2 as? Temporal.Date? else {
                     return false
                 }
-                if !modelCompareHelper(value1Optional, value2Optional) {
+                if !isEqual(value1Optional, value2Optional) {
                     return false
                 }
             case .dateTime:
@@ -60,7 +60,7 @@ extension ModelSchema {
                       let value2Optional = value2 as? Temporal.DateTime? else {
                     return false
                 }
-                if !modelCompareHelper(value1Optional, value2Optional) {
+                if !isEqual(value1Optional, value2Optional) {
                     return false
                 }
             case .time:
@@ -68,21 +68,21 @@ extension ModelSchema {
                       let value2Optional = value2 as? Temporal.Time? else {
                     return false
                 }
-                if !modelCompareHelper(value1Optional, value2Optional) {
+                if !isEqual(value1Optional, value2Optional) {
                     return false
                 }
             case .timestamp:
                 guard let value1Optional = value1 as? String?, let value2Optional = value2 as? String? else {
                     return false
                 }
-                if !modelCompareHelper(value1Optional, value2Optional) {
+                if !isEqual(value1Optional, value2Optional) {
                     return false
                 }
             case .bool:
                 guard let value1Optional = value1 as? Bool?, let value2Optional = value2 as? Bool? else {
                     return false
                 }
-                if !modelCompareHelper(value1Optional?.intValue, value2Optional?.intValue) {
+                if !isEqual(value1Optional?.intValue, value2Optional?.intValue) {
                     return false
                 }
             case .enum:
@@ -95,7 +95,7 @@ extension ModelSchema {
                 }
                 let enumValue1Optional = (value1Optional as? EnumPersistable)?.rawValue
                 let enumValue2Optional = (value2Optional as? EnumPersistable)?.rawValue
-                if !modelCompareHelper(enumValue1Optional, enumValue2Optional) {
+                if !isEqual(enumValue1Optional, enumValue2Optional) {
                     return false
                 }
             case .embedded, .embeddedCollection:
@@ -104,7 +104,7 @@ extension ModelSchema {
                        let encodable2 = value2 as? Encodable {
                         let json1 = try SQLiteModelValueConverter.toJSON(encodable1)
                         let json2 = try SQLiteModelValueConverter.toJSON(encodable2)
-                        if !modelCompareHelper(json1, json2) {
+                        if !isEqual(json1, json2) {
                             return false
                         }
                     }
@@ -123,8 +123,8 @@ extension ModelSchema {
         return true
     }
 
-    private func modelCompareHelper<T: Comparable>(_ optional1: T?, _ optional2: T?) -> Bool {
-        switch (optional1, optional2) {
+    private func isEqual<T: Comparable>(_ value1: T?, _ value2: T?) -> Bool {
+        switch (value1, value2) {
         case(nil, nil):
             return true
         case(nil, .some):
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
@@ -16,9 +16,10 @@ extension ModelSchema {
     /// Compare two `Model` based on a given `ModelSchema`
     /// Returns true if equal, false otherwise
     /// Currently, schemas where system timestamps fields (createdAt, updatedAt)
-    /// are changed with `timestamps` attribute to model directive and also added
-    /// to model schema are not covered. This will cause `isEqual` to fail in
-    /// such cases.
+    /// are renamed using with `@model`'s `timestamps` attribute and explicitly
+    /// added to the input schema are not supported by this check since they are
+    /// marked as "read-only" fields and will fail the check when the service generates
+    /// and returns the value of `createdAt` or `updatedAt`.
     /// for e.g.
     /// type Post @model(timestamps:{createdAt: "createdOn", updatedAt: "updatedOn"}) {
     ///  id: ID!
@@ -27,7 +28,14 @@ extension ModelSchema {
     ///  createdOn: AWSDateTime
     ///  updatedOn: AWSDateTime
     /// }
-    func isEqual(_ model1: Model, _ model2: Model) -> Bool {
+    func compare(_ model1: Model, _ model2: Model) -> Bool {
+        let modelType1 = ModelRegistry.modelType(from: model1.modelName)
+        let modelType2 = ModelRegistry.modelType(from: model2.modelName)
+        if modelType1 != modelType2 {
+            // no need to compare models as they have different type
+            return false
+        }
+
         for (fieldName, modelField) in fields {
             // read only fields or fields updated from the service are skipped for equality check
             // examples of such fields include `createdAt`, `updatedAt` and `coverId` in `Record`
@@ -44,21 +52,21 @@ extension ModelSchema {
                 guard let value1Optional = value1 as? String?, let value2Optional = value2 as? String? else {
                     return false
                 }
-                if !isEqual(value1Optional, value2Optional) {
+                if !compare(value1Optional, value2Optional) {
                     return false
                 }
             case .int:
                 guard let value1Optional = value1 as? Int?, let value2Optional = value2 as? Int? else {
                     return false
                 }
-                if !isEqual(value1Optional, value2Optional) {
+                if !compare(value1Optional, value2Optional) {
                     return false
                 }
             case .double:
                 guard let value1Optional = value1 as? Double?, let value2Optional = value2 as? Double? else {
                     return false
                 }
-                if !isEqual(value1Optional, value2Optional) {
+                if !compare(value1Optional, value2Optional) {
                     return false
                 }
             case .date:
@@ -66,7 +74,7 @@ extension ModelSchema {
                       let value2Optional = value2 as? Temporal.Date? else {
                     return false
                 }
-                if !isEqual(value1Optional, value2Optional) {
+                if !compare(value1Optional, value2Optional) {
                     return false
                 }
             case .dateTime:
@@ -74,7 +82,7 @@ extension ModelSchema {
                       let value2Optional = value2 as? Temporal.DateTime? else {
                     return false
                 }
-                if !isEqual(value1Optional, value2Optional) {
+                if !compare(value1Optional, value2Optional) {
                     return false
                 }
             case .time:
@@ -82,21 +90,21 @@ extension ModelSchema {
                       let value2Optional = value2 as? Temporal.Time? else {
                     return false
                 }
-                if !isEqual(value1Optional, value2Optional) {
+                if !compare(value1Optional, value2Optional) {
                     return false
                 }
             case .timestamp:
                 guard let value1Optional = value1 as? String?, let value2Optional = value2 as? String? else {
                     return false
                 }
-                if !isEqual(value1Optional, value2Optional) {
+                if !compare(value1Optional, value2Optional) {
                     return false
                 }
             case .bool:
                 guard let value1Optional = value1 as? Bool?, let value2Optional = value2 as? Bool? else {
                     return false
                 }
-                if !isEqual(value1Optional?.intValue, value2Optional?.intValue) {
+                if !compare(value1Optional?.intValue, value2Optional?.intValue) {
                     return false
                 }
             case .enum:
@@ -109,7 +117,7 @@ extension ModelSchema {
                 }
                 let enumValue1Optional = (value1Optional as? EnumPersistable)?.rawValue
                 let enumValue2Optional = (value2Optional as? EnumPersistable)?.rawValue
-                if !isEqual(enumValue1Optional, enumValue2Optional) {
+                if !compare(enumValue1Optional, enumValue2Optional) {
                     return false
                 }
             case .embedded, .embeddedCollection:
@@ -118,7 +126,7 @@ extension ModelSchema {
                        let encodable2 = value2 as? Encodable {
                         let json1 = try SQLiteModelValueConverter.toJSON(encodable1)
                         let json2 = try SQLiteModelValueConverter.toJSON(encodable2)
-                        if !isEqual(json1, json2) {
+                        if !compare(json1, json2) {
                             return false
                         }
                     }
@@ -137,7 +145,7 @@ extension ModelSchema {
         return true
     }
 
-    private func isEqual<T: Comparable>(_ value1: T?, _ value2: T?) -> Bool {
+    private func compare<T: Comparable>(_ value1: T?, _ value2: T?) -> Bool {
         switch (value1, value2) {
         case(nil, nil):
             return true
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
@@ -11,7 +11,7 @@ import Amplify
 // swiftlint:disable cyclomatic_complexity
 extension ModelSchema {
 
-    private static let serviceUpdatedFields: Set = ["updatedAt"]
+    private static let serviceUpdatedFields: Set = ["updatedAt", "createdAt"]
 
     /// Compare two `Model` based on a given `ModelSchema`
     /// Returns true if equal, false otherwise
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
@@ -15,6 +15,18 @@ extension ModelSchema {
 
     /// Compare two `Model` based on a given `ModelSchema`
     /// Returns true if equal, false otherwise
+    /// Currently, schemas where system timestamps fields (createdAt, updatedAt)
+    /// are changed with `timestamps` attribute to model directive and also added
+    /// to model schema are not covered. This will cause `isEqual` to fail in
+    /// such cases.
+    /// for e.g.
+    /// type Post @model(timestamps:{createdAt: "createdOn", updatedAt: "updatedOn"}) {
+    ///  id: ID!
+    ///  title: String!
+    ///  tags: [String!]!
+    ///  createdOn: AWSDateTime
+    ///  updatedOn: AWSDateTime
+    /// }
     func isEqual(_ model1: Model, _ model2: Model) -> Bool {
         for (fieldName, modelField) in fields {
             // read only fields or fields updated from the service are skipped for equality check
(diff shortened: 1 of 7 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,144 @@
...
+// swiftlint:disable syntactic_sugar
+extension ModelSchema {
+
+    /// Compare two `Model` based on a given `ModelSchema`
+    /// Returns true if equal, false otherwise
+    public static func isEqual(_ model1: Model,
+                               _ model2: Model,
+                               _ modelSchema: ModelSchema) -> Bool {
+        for (fieldName, modelField) in modelSchema.fields {
+            // skip read only fields
```
lawmicha, line 21: good place to add comment to why we skip read-only fields; like `updatedAt` is created/updated by the service so for the purpose of this equality check, we ignore it. what are the read-only fields other than updatedAt? does it hold true for those other fields as well?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,144 @@
...
+
+            let value1 = model1[fieldName] ?? nil
+            let value2 = model2[fieldName] ?? nil
+
+            // check equality for different `ModelFieldType`
+            switch modelField.type {
+            case .string:
+                guard let value1Optional = value1 as? String?, let value2Optional = value2 as? String? else {
+                    return false
+                }
```
lawmicha, line 34: what if we have a `modelField.type` and then `value1` and `value2` are both `nil`- meaning both values do not exist on either models. it should continue because there is nothing to check but this returns false and says the model's aren't equal

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Extensions.swift
```diff
@@ -30,21 +29,34 @@ extension MutationEvent {
...
-                        """)
-                    existingEvent.version = mutationSync.syncMetadata.version
-                    storageAdapter.save(existingEvent, condition: nil) { result in
-                        switch result {
-                        case .failure(let dataStoreError):
-                            completion(.failure(dataStoreError))
-                        case .success:
-                            completion(.success(()))
+                // check if version returned from API is not nil and <= version sent in the request
+                if existingEvent.version != nil && mutationSync.syncMetadata.version <= existingEvent.version! {
```
lawmicha, line 33: 1. if the `mutationSync.syncMetadata.version` is == to `existingEvent.version` then if the model equality check checks out, we replace it anyways- so i think it should just execute the below logic for `<`.
2. if `existing.version == nil` it continues to the logic below. before it would replace it regardless of what the response version is, but now we check the model. do we want to do this? if the existing.version is `nil`, can we avoid the model check and simply replace it?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Extensions.swift
```diff
@@ -30,21 +29,34 @@ extension MutationEvent {
...
+                    return
+                }
+
+                do {
+                    let responseModel = mutationSync.model.instance
+                    let requestModel = try mutationEvent.decodeModel()
+
+                    // check if the data sent in the request is the same as the response
+                    // if it is, update the pending mutation event version to the response version
+                    if let modelSchema = ModelRegistry.modelSchema(from: mutationEvent.modelName),
```
lawmicha, line 44: nit: inverse with `guard` instead of if else

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreConsecutiveUpdatesTests.swift
```diff
@@ -292,9 +285,8 @@ class DataStoreConsecutiveUpdatesTests: SyncEngineIntegrationTestBase {
 
             if mutationEvent.mutationType == GraphQLMutationType.delete.rawValue {
                 XCTAssertEqual(post, updatedPost)
-                // this needs to be commented out once the bug is fixed
-                // deleteSyncReceived.fulfill()
-                // XCTAssertEqual(mutationEvent.version, 3)
+                deleteSyncReceived.fulfill()
+                XCTAssertEqual(mutationEvent.version, 3)
```
lawmicha, line 289: nice!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Extensions.swift
```diff
@@ -30,21 +29,34 @@ extension MutationEvent {
...
-                            \(mutationSync.syncMetadata.version)
-                        """)
-                    existingEvent.version = mutationSync.syncMetadata.version
-                    storageAdapter.save(existingEvent, condition: nil) { result in
-                        switch result {
-                        case .failure(let dataStoreError):
-                            completion(.failure(dataStoreError))
-                        case .success:
-                            completion(.success(()))
+                // check if version returned from API is not nil and <= version sent in the request
```
lawmicha, line 32: `// check if the pending mutation version is not nil and <= version from the response``

lawmicha:
```
guard let existingVersion = existingEvent.version, existingVersion <= mutationSync.syncMetadata.version else {
completion(.success(()))
   return
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Extensions.swift
thisisabhash: `if existingEvent.version!=nil && existingEvent.version >= mutationSync.syncMetadata.version`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Extensions.swift
```diff
@@ -30,21 +29,34 @@ extension MutationEvent {
...
+                }
+
+                do {
+                    let responseModel = mutationSync.model.instance
+                    let requestModel = try mutationEvent.decodeModel()
+
+                    // check if the data sent in the request is the same as the response
+                    // if it is, update the pending mutation event version to the response version
+                    if let modelSchema = ModelRegistry.modelSchema(from: mutationEvent.modelName),
+                       ModelSchema.isEqual(responseModel, requestModel, modelSchema) {
```
lawmicha, line 45: `modelSchema.isEqual` since there is an extension on modelSchema instances

lawmicha:
```
guard let modelSchema = ModelRegistry.modelSchema(from: mutationEvent.modelName), modelSchema.isEqual(model1, model1) else {
  complete(.success())
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,144 @@
...
+
+// swiftlint:disable cyclomatic_complexity
+// swiftlint:disable syntactic_sugar
+extension ModelSchema {
+
+    /// Compare two `Model` based on a given `ModelSchema`
+    /// Returns true if equal, false otherwise
+    public static func isEqual(_ model1: Model,
+                               _ model2: Model,
+                               _ modelSchema: ModelSchema) -> Bool {
```
lawmicha, line 19: `public func isEqual`

lawmicha: my mistake, this probably doesn't need to be `public`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/Support/MutationEventExtensionsTests.swift
```diff
@@ -199,4 +239,317 @@ class MutationEventExtensionsTest: BaseDataStoreTests {
...
+        wait(for: [updatingVersionExpectation], timeout: 1)
+
+        // query for head of mutation event table for given model id and check if it has the correct version
+        MutationEvent.pendingMutationEvents(for: post1.id,
+                                            storageAdapter: storageAdapter) { result in
+            switch result {
+            case .failure(let error):
+                XCTFail("Error : \(error)")
+            case .success(let mutationEvents):
+                guard !mutationEvents.isEmpty, let updatedEvent = mutationEvents.first else {
```
lawmicha, line 439: if it wasn't supposed to be updated, `updatedEvent` is confusing, perhaps just `pendingMutationEvent` or `mutationEventHead`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Extensions.swift
```diff
@@ -11,15 +11,14 @@ import AWSPluginsCore
...
-    // if it has a `nil` version, with syncMetadata version in `mutationSync`
-    // and saves it in the mutation event table
-    static func updatePendingMutationEventVersionIfNil(for modelId: Model.Identifier,
-                                                       mutationSync: MutationSync<AnyModel>,
-                                                       storageAdapter: StorageEngineAdapter,
-                                                       completion: @escaping DataStoreCallback<Void>) {
+    // Updates the version of the head of pending mutation event queue for a given model `id`
+    // with syncMetadata version in `mutationSync` and saves it in the mutation event table
+    static func reconcilePendingMutationEventsVersion(mutationEvent: MutationEvent,
+                                                      mutationSync: MutationSync<AnyModel>,
```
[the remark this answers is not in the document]
thisisabhash: the test `testRecordCoversAreEqualWithDifferentReadOnlyFields` tests the read only fields case where `createdAt` is a read only field.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/Support/ModelCompareTests.swift
```diff
@@ -0,0 +1,304 @@
...
+import SQLite
+import XCTest
+
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+@testable import AWSPluginsCore
+
+// swiftlint:disable type_body_length
+class ModelCompareTests: BaseDataStoreTests {
```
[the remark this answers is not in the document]
thisisabhash: the test `testRecordCoversAreEqualWithDifferentReadOnlyFields` tests the read only fields case where `createdAt` is a read only field.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/Support/ModelCompareTests.swift
```diff
@@ -0,0 +1,317 @@
...
+        let color1 = Color(name: "Color1", red: 100, green: 100, blue: 100)
+        let color2 = Color(name: "Color2", red: 200, green: 200, blue: 200)
+        let category1 = Category(name: "Category1", color: color1)
+        let category2 = Category(name: "Category1", color: color2)
+        let todo1 = Todo(id: id, name: name, categories: [category1])
+        let todo2 = Todo(id: id, name: name, categories: [category2])
+        XCTAssertFalse(Todo.schema.isEqual(todo1, todo2))
+    }
+
+    func testRecordCoversAreEqualWithDifferentReadOnlyFields() {
```
lawmicha, line 305: perhaps a test comment that "createdAt" is a read-only field

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,142 @@
...
+                            continue
+                        }
+                    }
+                } catch {
+                    continue
+                }
+            case .model:
+                continue
+            case .collection:
+                continue
```
lawmicha, line 137: explain why these are skipped

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/Support/ModelCompareTests.swift
```diff
@@ -0,0 +1,317 @@
...
+import SQLite
+import XCTest
+
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+@testable import AWSPluginsCore
+
+// swiftlint:disable type_body_length
+class ModelCompareTests: BaseDataStoreTests {
```
lawmicha, line 18: can you add test for when a Post has a `nil` field value and the other Post has a value? they are not equal, right?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,142 @@
...
+            // check equality for different `ModelFieldType`
+            switch modelField.type {
+            case .string:
+                guard let value1Optional = value1 as? String?, let value2Optional = value2 as? String? else {
+                    return false
+                }
+                if let value1 = value1Optional, let value2 = value2Optional, value1 != value2 {
+                    return false
+                } else {
+                    continue
```
lawmicha, line 36: if `value1Optional` cannot be unwrapped because it's `nil` but value2Optional could have been unwrapped, then we should return false, right? since one side is nil and the other side is valid, but this logic would continue and move onto the next field

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/Support/MutationEventExtensionsTests.swift
```diff
@@ -199,4 +239,317 @@ class MutationEventExtensionsTest: BaseDataStoreTests {
...
+                                                        modelId: modelId,
+                                                        modelName: Post.modelName,
+                                                        json: try post2.toJSON(),
+                                                        mutationType: .update,
+                                                        version: 2)
+
+        let metadata = MutationSyncMetadata(id: modelId,
+                                            deleted: false,
+                                            lastChangedAt: Int(Date().timeIntervalSince1970),
+                                            version: 1)
```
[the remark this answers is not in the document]
thisisabhash: This test is for when we receive an older version in the response and the version in pending mutation event table is newer.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/Support/MutationEventExtensionsTests.swift
```diff
@@ -187,16 +227,320 @@ class MutationEventExtensionsTest: BaseDataStoreTests {
...
+        }
+
+        let pendingMutationEventExpectation = expectation(description: "save pendingMutationEvent success")
+        storageAdapter.save(pendingMutationEvent) { result in
+            guard case .success = result else {
+                XCTFail("Failed to save metadata")
+                return
+            }
+            pendingMutationEventExpectation.fulfill()
+        }
```
lawmicha, line 289: this can be reduced to a set up step that invovles saving an array of pending mutation events, and verifying that the head is as expected:
```
setUpPendingMutationQueue([requestMutationEvent, pendingMutationEvent])
```


```
func setUpPendingMutationQueue(_ mutationEvents: [MutationEvent]) {
   // for each mutationEvent, perform storageAdapter.save

  // verify the head of the queue is as expected
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Extensions.swift
```diff
@@ -30,11 +35,25 @@ extension MutationEvent {
...
+
+                    // check if the data sent in the request is the same as the response
+                    // if it is, update the pending mutation event version to the response version
+                    guard let modelSchema = ModelRegistry.modelSchema(from: mutationEvent.modelName),
+                          modelSchema.isEqual(responseModel, requestModel) else {
+                        completion(.success(()))
+                        return
+                    }
+
                     existingEvent.version = mutationSync.syncMetadata.version
```
lawmicha, line 57: can this be factored out so it can be called like this? (and easier to test)

```swift
guard let reconciledEvent = reconcile(pendingMutationEvent: existingEvent, with: mutationEvent, responseMutationSync: mutationSync) else {
  complete(.success())
 return
}

 storageAdapter.save(reconciledEvent)
```
```swift
static func reconcile(pendingMutationEvent: MutationEvent, 
         with requestMutationEvent: MutationEvent, 
          responseMutationSync: MutationSync<AnyModel>) -> MutationEvent? {


 // return if version of the pending mutation event is not nil and
                // is >= version contained in the response
                if pendingMutationEvent.version != nil && pendingMutationEvent.version! >= mutationSync.syncMetadata.version {
                    return nil
                }

                do {
                    let responseModel = mutationSync.model.instance
                    let requestModel = try mutationEvent.decodeModel()

                    // check if the data sent in the request is the same as the response
                    // if it is, update the pending mutation event version to the response version
                    guard let modelSchema = ModelRegistry.modelSchema(from: mutationEvent.modelName),
                          modelSchema.isEqual(responseModel, requestModel) else {
                        return nil
                    }

                    pendingMutationEvent.version = mutationSync.syncMetadata.version
                 return pendingMutationEvent
          } catch {
               Amplify.log.verbose("Error decoding models: \(error)")
               return nil
          }
}


existing unit tests can be extended with one line:
```swift
// currently we unit test `reconcilePendingMutationEventsVersion` like so, followed by verifying the pending mutation queue
MutationEvent.reconcilePendingMutationEventsVersion(sent: request,
                                                             received: response,

// you can add an additional check like so:
let reconciledEvent = reconcile(pendingMutationEvent: head, with: mutationEvent, responseMutationSync: response)

// depending on the situation you would either
XCTAssertNil(reconciledEvent) // no version update
// or
XCTAssertEqual(reconciledEvent.version, response.version)
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/Support/MutationEventExtensionsTests.swift
```diff
@@ -14,26 +14,49 @@ import XCTest
 @testable import AWSDataStoreCategoryPlugin
 @testable import AWSPluginsCore
 
+// swiftlint:disable cyclomatic_complexity
+// swiftlint:disable type_body_length
+// swiftlint:disable file_length
```
lawmicha, line 19: added a few suggestions below to reduce file length

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,138 @@
...
+                if !modelCompareHelper(value1Optional, value2Optional) {
+                    return false
+                }
+            case .int:
+                guard let value1Optional = value1 as? Int?, let value2Optional = value2 as? Int? else {
+                    return false
+                }
+                if !modelCompareHelper(value1Optional, value2Optional) {
+                    return false
+                }
```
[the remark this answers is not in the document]
thisisabhash: it can't be a return statement as in true case, we need to continue checking through other fields

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Extensions.swift
```diff
@@ -10,16 +10,21 @@ import Dispatch
...
-
-    // Updates the head of pending mutation event queue for a given model `id`
-    // if it has a `nil` version, with syncMetadata version in `mutationSync`
-    // and saves it in the mutation event table
-    static func updatePendingMutationEventVersionIfNil(for modelId: Model.Identifier,
-                                                       mutationSync: MutationSync<AnyModel>,
-                                                       storageAdapter: StorageEngineAdapter,
-                                                       completion: @escaping DataStoreCallback<Void>) {
+    // Reconciles the version of pending mutation events in mutation event table with the
+    // API response version for the following consecutive update scenarios :
```
lawmicha, line 14: why do we reconcile? Consecutive operations that modify a model results in a sequence of pending mutation events that have the current version of the model. The first mutation event has the correct version of the model, while the subsequent events will have lower versions if the first mutation event is synced to the cloud. By checking reconciling the pending mutation events after syncing the first mutation event, we attempt to update the pending version to the latest version of the object.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/MutationEvent+Extensions.swift
```diff
@@ -10,16 +10,21 @@ import Dispatch
...
+    //  - Save, then update
```
lawmicha, line 15: "Save, then immediately update" makes more sense, for each scenario there are versions involves, what does the mutation queue look like and what are the versions for each operation? what does reconcile do in this scenario?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/Support/MutationEventExtensionsTests.swift
```diff
@@ -88,69 +123,115 @@ class MutationEventExtensionsTest: BaseDataStoreTests {
...
-                                                        version: nil)
-        let metadata = MutationSyncMetadata(id: modelId,
-                                            deleted: false,
-                                            lastChangedAt: Int(Date().timeIntervalSince1970),
-                                            version: 1)
+        let post1 = Post(id: modelId, title: "title1", content: "content1", createdAt: .now())
+        let post2 = Post(id: modelId, title: "title2", content: "content2", createdAt: .now())
+        let requestMutationEvent =
+            try createMutationEvent(model: post1, mutationType: .create, version: 2, inProcess: true)
+        let pendingMutationEvent = try createMutationEvent(model: post2, mutationType: .update, version: 2)
```
lawmicha, line 150: nit: remove mutationType parameter if it doesn't matter

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/Support/MutationEventExtensionsTests.swift
```diff
@@ -88,69 +123,115 @@ class MutationEventExtensionsTest: BaseDataStoreTests {
...
+        let responseMutationSync = createMutationSync(model: post1, deleted: false, version: 1)
```
lawmicha, line 151: nit: remove deleted parameter if it doesn't matter

lawmicha (approved):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/Support/MutationEventExtensionsTests.swift
thisisabhash: type is needed for `testSentModelWithNilVersion_SecondPendingEventNotReconciled` where we have `update` and `delete` events in queue and only `update` event is modified. Will let it be as the tests model the different scenarios like "Save, then update" etc.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,140 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+// swiftlint:disable cyclomatic_complexity
+extension ModelSchema {
+
+    private static let serviceUpdatedFields: Set = ["updatedAt"]
```
lawmicha, line 14: - can use a `Set` for faster look up
- is it possible to add `createdAt` as an explicit and optional field? resulting in `createdAt` not a read-only and optionally passed in when creating the model. if this case is true, then should add `createdAt` into the Set as well

thisisabhash: it seems logical to add `createdAt` to the list - will do.
is there a faster `Set` implementation?

lawmicha (approved):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,140 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+// swiftlint:disable cyclomatic_complexity
+extension ModelSchema {
+
+    private static let serviceUpdatedFields: Set = ["updatedAt", "createdAt"]
```
lawmicha, line 14: add comment regarding this

lawmicha: which cases we fix, which we don't

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,152 @@
...
+
+    /// Compare two `Model` based on a given `ModelSchema`
+    /// Returns true if equal, false otherwise
+    /// Currently, schemas where system timestamps fields (createdAt, updatedAt)
+    /// are changed with `timestamps` attribute to model directive and also added
+    /// to model schema are not covered. This will cause `isEqual` to fail in
+    /// such cases.
```
lawmicha, line 21:
```suggestion
    /// Currently, schemas where system timestamps fields (createdAt, updatedAt)
    /// are renamed using with `@model`'s `timestamps` attribute and explicitly
    /// added to the input schema are not supported by this check since they are
    /// marked as "read-only" fields and will fail the check when the service generates
    /// and returns the value of `createdAt` or `updatedAt`.
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,160 @@
...
+extension ModelSchema {
+
+    private static let serviceUpdatedFields: Set = ["updatedAt", "createdAt"]
+
+    /// Compare two `Model` based on a given `ModelSchema`
+    /// Returns true if equal, false otherwise
+    /// Currently, schemas where system timestamps fields (createdAt, updatedAt)
+    /// are renamed using with `@model`'s `timestamps` attribute and explicitly
+    /// added to the input schema are not supported by this check since they are
+    /// marked as "read-only" fields and will fail the check when the service generates
```
lawmicha, line 21:
```suggestion
    /// added to the input schema are not supported by this check since they are not
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/Support/Model+Compare.swift
```diff
@@ -0,0 +1,160 @@
...
+// swiftlint:disable cyclomatic_complexity
+extension ModelSchema {
+
+    private static let serviceUpdatedFields: Set = ["updatedAt", "createdAt"]
+
+    /// Compare two `Model` based on a given `ModelSchema`
+    /// Returns true if equal, false otherwise
+    /// Currently, schemas where system timestamps fields (createdAt, updatedAt)
+    /// are renamed using with `@model`'s `timestamps` attribute and explicitly
+    /// added to the input schema are not supported by this check since they are
```
lawmicha, line 20:
```suggestion
    /// added to the input schema are not supported by this check since they are not
```

lawmicha (approved):

PR #1411, opened 2021-09-03
fix(Datastore) : Use throwing iterator to avoid hidden force try

ashiemke:
*Issue #, if available:* None

*Description of changes:* Using the default iterator on a statement uses a force try, which crashes (although I'm not sure the circumstances). Replace that with a long-form while rather than for-in to re-throw exception rather than crashing.

See https://github.com/stephencelis/SQLite.swift/pull/1030 for discussion

*Check points:*

~~- [ ] Added new tests to cover change, if needed~~
- [x] All unit tests pass


thisisabhash:
Hello, Could you please add information about when a situation would arise that'd cause a crash and possible ways to test it?

thisisabhash (approved):
Existing unit tests and integration tests are passing with the changes. LGTM.

thisisabhash:
SQLite codebase uses a force `try` in default iterator which causes a crash.
I tried to reproduce by calling `Datastore.clear()` on `Auth` `signedOut` events but couldn't reproduce.
Merging this in because it is difficult to replicate the crash in an integration test having a clear use case.
A unit test to cover the change is difficult as it is difficult to mock `Statement` and `Connection` classes because they are marked `final` and part of SQLite codebase.

Verified that existing integration tests and unit tests are passing for this PR.

PR #1422, opened 2021-09-14
Feat/datastore observe query

lawmicha:
*Issue #, if available:*

*Description of changes:*

*Check points:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
@@ -102,10 +102,10 @@ extension ObserveQuerySubscription: DefaultLogger { }
 ///     - Subscribe to Item changes
 ///     - Perform initial query to set up the internal state of the items
 ///     - Generate first snapshot based on the internal state
-///     When the operation receives item changs
+///     When the operation receives item changes
 ///     - Batch them into batches of up to 1000 items or when 2 seconds have elapsed (`.collect(2s,1000)`)`
 ///     - Update internal state of items based on the changed items
-///     - Generate new snapshot based on lates state of the items, with the items changed dedupped.
+///     - Generate new snapshot based on latest state of the items, with the items changed dedupped.
 @available(iOS 13.0, *)
 public class AWSDataStoreObserveQueryOperation<M: Model>: AsynchronousOperation, DataStoreObserveQueryOperation {
 
@@ -166,6 +166,10 @@ public class AWSDataStoreObserveQueryOperation<M: Model>: AsynchronousOperation,
     override public func cancel() {
         if let itemsChangedSink = itemsChangedSink {
             itemsChangedSink.cancel()
+        }
+
+        if let dataStoreEventSink = dataStoreEventSink {
+            dataStoreEventSink.cancel()
         }
         passthroughPublisher.send(completion: .finished)
         super.cancel()
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
@@ -9,7 +9,7 @@ import Amplify
 import AWSPluginsCore
 import Combine
 
-protocol DataStoreObseverQueryOperation {
+protocol DataStoreObserveQueryOperation {
     func resetState()
     func startObserveQuery()
 }
@@ -19,9 +19,9 @@ public class ObserveQueryPublisher<M: Model>: Publisher {
     public typealias Output = DataStoreQuerySnapshot<M>
     public typealias Failure = DataStoreError
 
-    weak var operation: AWSDataStoreObseverQueryOperation<M>?
+    weak var operation: AWSDataStoreObserveQueryOperation<M>?
 
-    func configure(operation: AWSDataStoreObseverQueryOperation<M>) {
+    func configure(operation: AWSDataStoreObserveQueryOperation<M>) {
         self.operation = operation
     }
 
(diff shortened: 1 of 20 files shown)


lawmicha:
in the [first commit since last PR](https://github.com/aws-amplify/amplify-ios/pull/1422/commits/2d3def6e85b7c4cf166a093f9616732289132bbc)

- Update QuerySnapshot to remove `itemsChanged`
- before sync is true: process batches at time/count of 2s/pageSize
- after sync is true: process one at a time
- improved isSynced reliability by using internal state so the operation can check “is this model synced?” at any time. Imagine `modelSyncedEvent` being sent before the observeQuery is established. Checking state allows this to work properly, since it would have missed the event. This is aligned with JS logic. By using the state, the logic has also been simplified to a computed property `isSynced` that can be used at the time of the snapshot being generated

diegocstn (approved):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/Support/Model+Sort.swift
```diff
@@ -0,0 +1,161 @@
...
+extension Array where Element: Model {
+    mutating func sortModels(by sortBy: QuerySortDescriptor, modelSchema: ModelSchema) {
+        sort { modelSchema.comparator(model1: $0, model2: $1, sortBy: sortBy) }
+    }
+}
+
+enum ModelValueCompare<T: Comparable> {
+    case bothNil
+    case leftNil(value2: T)
+    case rightNil(value1: T)
```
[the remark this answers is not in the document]
lawmicha: if i remember correctly, `leftNil` case means the left side "value1" is nil so the associated value is the "right side". it may clearer if
`case leftNil(rightValue: T)`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
```diff
@@ -0,0 +1,359 @@
...
+        }
+        passthroughPublisher.send(currentSnapshot)
+    }
+
+    func updateCurrentItems(with itemsChanged: [MutationEvent]) {
+        for item in itemsChanged {
+            do {
+                let model = try item.decodeModel(as: modelType)
+                if item.mutationType == MutationEvent.MutationType.delete.rawValue {
+                    if let index = currentItems.firstIndex(where: { $0.id == model.id }) {
```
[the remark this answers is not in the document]
lawmicha: updated

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
```diff
@@ -0,0 +1,359 @@
...
+            }
+
+            self.generateQuerySnapshot(itemsChanged: mutationEvents)
+        }
+    }
+
+    func generateQuerySnapshot(itemsChanged: [MutationEvent] = []) {
+        updateCurrentItems(with: itemsChanged)
+        if let sort = sortInput {
+            sort.forEach { currentItems.sortModels(by: $0, modelSchema: modelSchema) }
```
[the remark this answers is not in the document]
lawmicha: added

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
```diff
@@ -137,22 +135,37 @@ public class AWSDataStoreObserveQueryOperation<M: Model>: AsynchronousOperation,
         return observeQueryPublisher.eraseToAnyPublisher()
     }
 
+    var isSynced: Bool {
```
thisisabhash, line 138: is there a possibility that this requires being thread safe? same for `currentSnapshot`?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
```diff
@@ -198,52 +214,23 @@ public class AWSDataStoreObserveQueryOperation<M: Model>: AsynchronousOperation,
...
-
     // MARK: - Query
 
     func initialQuery() {
         if isCancelled || isFinished {
             finish()
             return
         }
-
+        log.verbose("Initial Query")
```
thisisabhash, line 233: nit: if `initialQuery` is called multiple times, we can remove the log statement

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
```diff
@@ -302,6 +291,10 @@ public class AWSDataStoreObserveQueryOperation<M: Model>: AsynchronousOperation,
         }
     }
 
+    func onItemChange(mutationEvent: MutationEvent) {
+        onItemChanges(mutationEvents: [mutationEvent])
+    }
+
     func onItemChanges(mutationEvents: [MutationEvent]) {
```
thisisabhash, line 298: nit: i think the naming should be `onItemsChange`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
```diff
@@ -318,21 +311,27 @@ public class AWSDataStoreObserveQueryOperation<M: Model>: AsynchronousOperation,
...
     }
 
     func updateCurrentItems(with itemsChanged: [MutationEvent]) {
         for item in itemsChanged {
             do {
                 let model = try item.decodeModel(as: modelType)
                 if item.mutationType == MutationEvent.MutationType.delete.rawValue {
-                    currentItemsMap.removeValue(forKey: model.id)
+                    if let index = currentItems.firstIndex(where: { $0.id == model.id }) {
+                        currentItems.remove(at: index)
```
thisisabhash: i think `currentItems` should be thread safe to avoid duplicates. and a unit test to check for duplicates case.

lawmicha: It may not be obvious since this method itself is not thread safe but it's run under the serial DispatchQueue which will perform each task under `serialQueue.async` serially, so it doesn't have to be thread safe. As long as the entry points put the work on the serialQueue, then the work will run one after another and no two threads will be accessing `currentItems` at the same time. There's a `testConcurrent` test DataStoreObserveQueryOperationTests which tests this

thisisabhash (approved):
LGTM

PR #1425, opened 2021-09-16
test(DataStore): Integration tests to verify eager loading in belongs-to relationship

thisisabhash:
*Issue #, if available:* None

*Description of changes:*  
1. add integration tests for verifying eager loading in belongs to relationship between models
2. move `Amplify.reset()` to `teardown()` in `SyncEngineTestBase` for avoid data race on line 47 and 48 in the unit test `ModelReconciliationDeleteTests.testUpdateAfterDelete()`

*Check points:*

- [x] All unit tests pass
~~- [] Documentation update for the change if required~~

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario6Tests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario6Tests.swift
@@ -90,6 +90,119 @@ class DataStoreConnectionScenario6Tests: SyncEngineIntegrationTestBase {
         }
     }
 
+    func testGetCommentThenFetchPostThenFetchBlog() throws {
+        try startAmplifyAndWaitForSync()
+        guard let blog = saveBlog(name: "name"),
+              let post = savePost(title: "title", blog: blog),
+              let comment = saveComment(post: post, content: "content") else {
+            XCTFail("Could not create blog, post, and comment")
+            return
+        }
+
+        // fetch comment
+        let getCommentCompleted = expectation(description: "get comment complete")
+        var resultPost: Post6?
+        Amplify.DataStore.query(Comment6.self, byId: comment.id) { result in
+            switch result {
+            case .success(let queriedCommentOptional):
+                guard let queriedComment = queriedCommentOptional else {
+                    XCTFail("Could not get comment")
+                    return
+                }
+                XCTAssertEqual(queriedComment.id, comment.id)
+                resultPost = queriedComment.post
+                getCommentCompleted.fulfill()
+            case .failure(let response):
+                XCTFail("Failed with: \(response)")
+            }
+        }
+        wait(for: [getCommentCompleted], timeout: TestCommonConstants.networkTimeout)
+
+        guard let fetchedPost = resultPost else {
+            XCTFail("Could not get post")
+            return
+        }
+
+        guard let fetchedBlog = fetchedPost.blog else {
+            XCTFail("Could not get blog")
+            return
+        }
+
+        // check if post was eagerly loaded
+        XCTAssertEqual(fetchedPost.id, post.id)
+        XCTAssertEqual(fetchedPost.title, post.title)
+
+        // check if blog was eagerly loaded
+        XCTAssertEqual(fetchedBlog.id, blog.id)
+        XCTAssertEqual(fetchedBlog.name, blog.name)
+    }
+
+    func testGetPostThenFetchBlogAndComment() throws {
+        try startAmplifyAndWaitForSync()
+        guard let blog = saveBlog(name: "name"),
+              let post = savePost(title: "title", blog: blog),
+              let comment = saveComment(post: post, content: "content") else {
+            XCTFail("Could not create blog, post, and comment")
+            return
+        }
+
+        // fetch post
+        let getPostCompleted = expectation(description: "get post complete")
+        var resultComments: List<Comment6>?
+        var resultBlog: Blog6?
+        Amplify.DataStore.query(Post6.self, byId: post.id) { result in
+            switch result {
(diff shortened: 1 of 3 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario6Tests.swift
```diff
@@ -90,6 +90,119 @@ class DataStoreConnectionScenario6Tests: SyncEngineIntegrationTestBase {
...
+
+        guard let fetchedBlog = resultBlog else {
+            XCTFail("Could not get blog")
+            return
+        }
+
+        guard let fetchedComments = resultComments else {
+            XCTFail("Could not get comments")
+            return
+        }
```
lawmicha, line 178: calling them fetched is a bit confusing since they weren't really the result of the above async call. they may have been fetched if they were eagarly loaded, but we're not sure at this point in the code. With above suggestion, this becomes one guard statement:
```
guard let fetchedPost = resultPost else {
  XCTFail("Could not get post")
  return
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario6Tests.swift
```diff
@@ -90,6 +90,119 @@ class DataStoreConnectionScenario6Tests: SyncEngineIntegrationTestBase {
...
+
+        // check if blog was eagerly loaded
+        XCTAssertEqual(fetchedBlog.id, blog.id)
```
lawmicha, line 181: with above, then this becomes
```
guard let eagerlyLoadedBlog = fetchedPost.blog else {
  XCTFail("Blog is nil, should be loaded")
  return
}
```
The message in the XCTFail would also communicate enough info about the expectation that the comment can be removed

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario6Tests.swift
```diff
@@ -90,6 +90,119 @@ class DataStoreConnectionScenario6Tests: SyncEngineIntegrationTestBase {
...
+        guard let blog = saveBlog(name: "name"),
+              let post = savePost(title: "title", blog: blog),
+              let comment = saveComment(post: post, content: "content") else {
+            XCTFail("Could not create blog, post, and comment")
+            return
+        }
+
+        // fetch comment
+        let getCommentCompleted = expectation(description: "get comment complete")
+        var resultPost: Post6?
```
lawmicha, line 104: this is confusing because the expectation is one thing and the result is something else, keeping it `resultComment` makes it clearer later especially since the test is making sure models are loaded, eagarly/lazily, etc. for example, is the `fetchedPost` not there because it is not eagarly loaded? or is it because the async call failed?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/TestSupport/SyncEngineTestBase.swift
```diff
@@ -41,11 +41,14 @@ class SyncEngineTestBase: XCTestCase {
 
     // MARK: - Setup
 
+    override func setUp() {
+        Amplify.Logging.logLevel = .verbose
+    }
+
```
lawmicha, line 47: according to https://developer.apple.com/documentation/xctest/xctest/1500341-setup XCTest calls setUpWithError(), followed by setUp().. this seems like hidden information, unless we have a reason to have two set up methods, would be clearer to move the logLevel set to the last line of `setUpWithError`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario6Tests.swift
```diff
@@ -90,6 +90,124 @@ class DataStoreConnectionScenario6Tests: SyncEngineIntegrationTestBase {
...
+    func testGetPostThenFetchBlogAndComment() throws {
+        try startAmplifyAndWaitForSync()
+        guard let blog = saveBlog(name: "name"),
+              let post = savePost(title: "title", blog: blog),
+              let comment = saveComment(post: post, content: "content") else {
+            XCTFail("Could not create blog, post, and comment")
+            return
+        }
+
+        // fetch post
```
lawmicha, line 152: nit: comment seems unnecessary, remove or add testcomment?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/Connection/DataStoreConnectionScenario6Tests.swift
```diff
@@ -90,6 +90,124 @@ class DataStoreConnectionScenario6Tests: SyncEngineIntegrationTestBase {
...
+        }
+
+        guard case .notLoaded = lazilyLoadedComments.loadedState else {
+            XCTFail("Should not be in loaded state")
+            return
+        }
+        XCTAssertEqual(lazilyLoadedComments.count, 1)
+        XCTAssertTrue(lazilyLoadedComments.contains(where: { (commentIn) -> Bool in
+            commentIn.id == comment.id
+        }))
```
lawmicha, line 204: if there's only one and we are sure of it from the XCTAssertEqual, then we can probably make due with 
```
XCTAssertEqual(lazilyLoadedComments.count, 1)
XCTAssertTrue(lazilyLoadedComments[0], comment.id)
```

lawmicha (approved):

PR #1454, opened 2021-10-06
fix(DataStore): Avoid model name from mutation sync in ModelSyncedEventEmitter

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/1446

*Description of changes:*
When running the iOS code from Flutter, the model name cannot be retrieved from the model. I'm not sure what is the exact failing comparison, it could be that `model.modelName` returns empty string and the string comparsion fails silently. This PR prioritizes the fix for now. 

We have other efforts that is working towards addressing the Flutter integration with iOS and ensuring there are no regressions when new features are being developed, like https://github.com/aws-amplify/amplify-ios/pull/1440 . The issue is  that iOS code allows the modelName to be accessed from the model and there's no context when the iOS developer developing the plugin code to know when it cannot be, so we should audit the code for other incidences of this usage or mark it as something that cannot be used in the code but still allow iOS developers building apps to have the convenience to it. 

The integation tests for `FlutterSerializedModel` can assert that the `modelSyncedEvent` contains expected number of items added when in fullSync. This may deem difficult when written since the integration tests aren't set up to clear the remote database and a full sync will always return more items on each run.

*Check points:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOperation.swift
@@ -185,7 +185,7 @@ final class InitialSyncOperation: AsynchronousOperation {
 
         reconciliationQueue.offer(items, modelSchema: modelSchema)
         for item in items {
-            initialSyncOperationTopic.send(.enqueued(item))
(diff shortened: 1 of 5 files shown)


lawmicha:
test failed (transient)
```
AWSDataStoreCategoryPluginTests:
		ReconcileAndLocalSaveOperationTests.testReconcile()
```

PR #1490, opened 2021-10-26
fix(DataStore): Allow different model types with the same ID

lawmicha:
*Issue #, if available:*

*Description of changes:*
This PR contains two changes

1.  Use modelName in MutationSyncMetadata.id as the new key format "modelName|modelId"
2. Migration scripts that run before the sync engine starts to migrate the `ids` to the new key format

Migrating to the new key format is required to support the use case of having duplicate Id's across different modelTypes. For apps that do not have this use case (they just have unique Id's across all models of different model types), then the migration will be to backfill the table with the new id formats before the sync process starts. 

If the apps are in the scenario where there are duplicate model Id's, it would be impossible to backfill because we are missing one or more version/deleted information about the duplicate models. In this case, migration will clear the metadata table and the "model sync" table. Clearing the model sync table will allow the sync process to do a full sync. Clearing the metadata table will allow the full sync to re-populate the table with new metadata items that contain the new key format.

**Callout**: For all apps that migrate to the new version of the library with this change, the local metadata table will be modified. So if during development, the developer decides to roll back to a previous version of the library, `DataStore.clear()` is required to be called to clear out the metadata table that has the new key format.

- [x] All unit tests pass


AmplifyPlugins/Core/AWSPluginsCore/Sync/MutationSync/MutationSyncMetadata.swift
```diff
@@ -8,9 +8,31 @@
...
     public var version: Int
+    
+    public var modelId: String {
+        id.components(separatedBy: "|").last ?? ""
+    }
+    public var modelName: String {
+        id.components(separatedBy: "|").first ?? ""
+    }
+    
+    public init(id: Model.Identifier, deleted: Bool, lastChangedAt: Int, version: Int) {
```
royjit, line 23: is `id` a type of `Model.Identifier`? Instead should we call it `String` or better create a MutationSyncMetadata.Identifier?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/AddModelNameToMutationSyncMetadataMigration.swift
```diff
@@ -0,0 +1,108 @@
...
+        return dropStatement
+    }
+
+    @discardableResult func createMutationSyncMetadataCopyTable() throws -> String {
+        let createTableStatement = CreateTableStatement(modelSchema: MutationSyncMetadataCopy.schema).stringValue
+        try connection.execute(createTableStatement)
+        return createTableStatement
+    }
+
+    @discardableResult  func backfillMutationSyncMetadata() throws -> String {
```
diegocstn, line 78: nit: extra space
```suggestion
    @discardableResult func backfillMutationSyncMetadata() throws -> String {
```

diegocstn (commented):
General comment:
I believe `ModelMigrations` shouldn't be directly coupled to a specific storage adapter (in this case the SQLite adapter)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/AddModelNameToMutationSyncMetadataMigration+Validation.swift
```diff
@@ -0,0 +1,87 @@
...
+import AWSPluginsCore
+
+extension AddModelNameToMutationSyncMetadataMigration {
+
+    // MARK: - Validation
+
+    /// Retrieve the record count from the MutationSyncMetadata table for
+    ///     1. the total number of records
+    ///     2. the total number of records that have the `id` match `<modelName>|<modelId>`
+    func selectMutationSyncMetadataRecords() throws -> (metadataCount: Int64, metadataIdMatchNewKeyCount: Int64) {
```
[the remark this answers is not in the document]
lawmicha: I tried this as well, I had some troubles getting the value out at this line if i change `Int64` to `Int`
```
if let metadataCount = row[0] as? Int64, let metadataIdMatchNewKeyCount = row[1] as? Int64 {
```

Will try this out again to better understand the record casting

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
```diff
@@ -0,0 +1,119 @@
...
+    @discardableResult func dropMutationSyncMetadataCopyIfExists() throws -> String {
+        guard let storageAdapter = storageAdapter else {
+            log.debug("Missing SQLiteStorageEngineAdapter")
+            throw DataStoreError.unknown("Missing storage adapter for model migration", "", nil)
+        }
+
+        return try storageAdapter.removeStore(for: MutationSyncMetadataCopy.schema)
+    }
+
+    @discardableResult func createMutationSyncMetadataCopyTable() throws -> String {
```
diegocstn, line 73: if we decide to take the approach described in the above comment, some examples of storage agnostic names:
- `createMutationSyncMetadataCopyStorage`
- `createMutationSyncMetadataCopyCollection`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataCopy.swift
```diff
@@ -0,0 +1,45 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Amplify
+
+public struct MutationSyncMetadataCopy: Model {
```
diegocstn, line 11: to avoid confusion, can we make this an internal struct of the `MutationSyncMetadataMigration` migration?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLiteValidation.swift
```diff
@@ -0,0 +1,137 @@
...
+        guard MutationSyncMetadataCopy.schema.fields.count == 4 else {
+            throw DataStoreError.internalOperation("MutationSyncMetadataCopy schema has changed from 4 fields", "", nil)
+        }
+
+        guard MutationSyncMetadataCopy.schema.fields["id"] != nil,
+              MutationSyncMetadataCopy.schema.fields["deleted"] != nil,
+              MutationSyncMetadataCopy.schema.fields["lastChangedAt"] != nil,
+              MutationSyncMetadataCopy.schema.fields["version"] != nil else {
+            throw DataStoreError.internalOperation("MutationSyncMetadataCopy schema missing expected fields", "", nil)
+        }
```
diegocstn, line 40: nit: would checking that `MutationSyncMetadata` fields and `MutationSyncMetadataCopy` fields are exactly the same achieve the same result here?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
```diff
@@ -0,0 +1,119 @@
...
+
+    // MARK: - Migration
+
+    func migrate() throws {
+        log.debug("Modifying and backfilling MutationSyncMetadata")
+        try dropMutationSyncMetadataCopyIfExists()
+        try createMutationSyncMetadataCopyTable()
+        try backfillMutationSyncMetadata()
+        try dropMutationSyncMetadata()
+        try renameMutationSyncMetadataCopy()
```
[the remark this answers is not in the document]
diegocstn: it can even be a single method similar to `func applyMigrationStep(_ step: MutationSyncMetadataMigrationStep)` where `MutationSyncMetadataMigrationStep` is an enum 
```swift
enum MutationSyncMetadataMigrationStep {
  case dropMutationSyncMetadataCopy
  .....
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLiteValidation.swift
lawmicha: i think the current check is more simple as they are explicit string checks that aren't referencing their actual field names. If we check the fields are exactly the same, even though it would be rare, but both structs could change at the same time and then this would pass (but should have failed). we could keep the explicit field count == 4 but still run into an even more rare scenario where the fields were changed in both models and count is still 4.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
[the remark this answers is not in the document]
lawmicha: > all these methods listed in MutationSyncMetadataMigrationDelegate

let me give this a try and add them back to the Delegate protocol, I hastily removed them earlier

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLiteValidation.swift
lawmicha: i pulled out the strings as local variables to the this method as a way to at least reuse them for both models. let me know if you prefer a different check or if i misinterpreted what you meant by checking MutationSyncMetadata fields and MutationSyncMetadataCopy fields are exactly the same

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
lawmicha: over in the storage adapter, now conforms to
```
protocol StorageEngineMigrationAdapter {

    @discardableResult func removeStore(for modelSchema: ModelSchema) throws -> String

    @discardableResult func createStore(for modelSchema: ModelSchema) throws -> String

    @discardableResult func emptyStore(for modelSchema: ModelSchema) throws -> String

    @discardableResult func renameStore(from: ModelSchema, toModelSchema: ModelSchema) throws -> String
}
```

so will name these accordingly, like `createMutationSyncMetadataCopyStore`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLiteValidation.swift
diegocstn: makes sense! I thought we only cared about `MutationSyncMetadata` and `MutationSyncMetadataCopy` being in sync

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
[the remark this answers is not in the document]
lawmicha: updated the delegate to have `applyMigrationStep` and defined the steps in the enum. the delegate just passes the call to the internal method of the same name as the enum case, which may be defined/implemented different for different storage adapters

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
```diff
@@ -0,0 +1,123 @@
...
+    init(storageAdapter: SQLiteStorageEngineAdapter, modelSchemas: [ModelSchema]) {
+        self.storageAdapter = storageAdapter
+        self.modelSchemas = modelSchemas
+    }
+
+    func transaction(_ basicClosure: BasicThrowableClosure) throws {
+        try storageAdapter?.transaction(basicClosure)
+    }
+
+    func applyMigration(_ step: MutationSyncMetadataMigrationStep) throws {
```
diegocstn, line 27: Nice! 
nit: I'd either call this `applyMigration(step:)` or `applyMigrationStep(_ step:)` just make clear we're not applying the migration multiple times

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
```diff
@@ -0,0 +1,123 @@
...
+        var sql = ""
+        for modelSchema in modelSchemas {
+            let modelName = modelSchema.name
+
+            if sql != "" {
+                sql += " UNION ALL "
+            }
+            sql += "SELECT id, \'\(modelName)\' as tableName FROM \(modelName)"
+        }
+        sql = "INSERT INTO \(MutationSyncMetadataMigration.MutationSyncMetadataCopy.modelName) (id,deleted,lastChangedAt,version) " +
```
diegocstn, line 96: nit: can we perhaps get the list of field names from the schema here?

diegocstn (approved):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
lawmicha: updating with `applyMigrationStep(_ step:)`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
lawmicha: from the initial glance, there is more than one place that uses the fields as a hard coded string: `(id,deleted,lastChangedAt,version)` and `select models.tableName || '|' || mm.id, mm.deleted, mm.lastChangedAt, mm.version`  so even if we got the fields from the schema in the first spot, we're then hardcoding them again in the next line, so we should it for both. Seems like it will get more obfuscated by doing so

PR #1494, opened 2021-10-27
fix(DataStore): ModelSyncedEventEmitter event order

lawmicha:
*Issue #, if available:*
Previous PR that was meant to fix the issue https://github.com/aws-amplify/amplify-ios/pull/1472
The issue: https://github.com/aws-amplify/amplify-ios/issues/1463#issuecomment-953030220

*Description of changes:*

This PR currently does two things
1. ModelSyncedEventEmitter was incorrectly sending out the ModelSyncedEvent on the hub before sending the last mutation event. The first commit fixes this by sending it in the right order from ModelSyncedEventEmitter

ModelSyncedEventEmitter sends the mutation event at `0140` -> SyncedEventEmitter `0630` -> RemoteSyncEngine `0809` -> publisher `0930` -> subscriber received at `1030`. The modelSynced is dispatched to the Hub at `0508` -> subscriber received at `1720`. The problem is the subscriber receiving the hub event could have been any time after `0508` and the subscriber to the publisher API could have received the mutation event any time after `0930`.

2. The second part of this PR will funnel all events up to AWSDataStorePlugin instance which is now responsible for sending events back on the publisher or to the hub. The events from ModelSyncedEventEmitter (as they are sent in the correct order) are received by AWSDataStorePlugin in the correct order. At this point, the event stream is processed one at a time, receiving the last mutation event before the modelSyncedEvent, before the syncQueries event, before the ready event, etc. As they are processed one at a time, the mutation event is sent to the publisher, and the modelSyncedEvent is sent to the hub

- [ ] All unit tests pass


thisisabhash:
I tested the bug where `isSynced` was being sent as `true` for snapshots beside the last snapshot and this PR addresses it.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/TestSupport/BaseDataStoreTests.swift
```diff
@@ -70,6 +71,7 @@ class BaseDataStoreTests: XCTestCase {
             try Amplify.add(plugin: apiPlugin)
             try Amplify.add(plugin: dataStorePlugin)
             try Amplify.configure(amplifyConfig)
+            XCTAssertEqual(dataStorePlugin.dispatchedModelSyncedEvents.count, 7)
```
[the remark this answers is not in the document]
lawmicha: `TestModelRegistration()` contains 7 models that are registered on configure. to make it more clear, i can assert against ModelRegistry.modelSchemas.count rather than a hardcoded value here

lawmicha:
> I tested the bug where `isSynced` was being sent as `true` for snapshots beside the last snapshot - this PR addresses it.

Thank you for testing this! @thisisabhash

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/RequestRetryablePolicyTests.swift
```diff
@@ -232,7 +232,7 @@ class RequestRetryablePolicyTests: XCTestCase {
             XCTAssertGreaterThanOrEqual(milliseconds, greaterThan)
             XCTAssertLessThanOrEqual(milliseconds, lessThan)
         default:
-            XCTFail("Expected milliseconds, but recieved \(retryInterval)")
+            XCTFail("Expected milliseconds, but receivedd \(retryInterval)")
```
thisisabhash, line 235: minor typo *received*

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -98,6 +102,9 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
     /// them to `StorageEngine.setUp(modelSchemas:)`
     public func configure(using amplifyConfiguration: Any?) throws {
         modelRegistration.registerModels(registry: ModelRegistry.self)
+        for modelSchema in ModelRegistry.modelSchemas {
```
thisisabhash, line 105: do we need to reset the `dispatchedModelSyncedEvents` map on `configure()` and `reset()` ?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/ModelSyncedEventEmitterTests.swift
```diff
@@ -86,7 +142,14 @@ class ModelSyncedEventEmitterTests: XCTestCase {
             XCTFail("Should not have completed")
         } receiveValue: { value in
             switch value {
-            case .modelSyncedEvent:
+            case .modelSyncedEvent(let modelSyncedEvent):
+                let expectedModelSyncedEventPayload = ModelSyncedEvent(modelName: "Post",
+                                                                       isFullSync: true,
+                                                                       isDeltaSync: false,
+                                                                       added: 2,
```
thisisabhash, line 149: should the number of added mutations be 5? as number of applied mutations are 5 at line 168-172?

thisisabhash (approved):
Approved with minor comments to consider

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/ModelSyncedEventEmitterTests.swift
lawmicha: the reason why it's 2 is because the initial sync only synced 2 models, so models that were synced were 2. the mutation events coming from he reconciliation whether it's a applied or dropped) from 3rd and onwards are all due to subscription events- unrelated to the two that were synced from the initial sync

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
lawmicha: `reset` is used internally for testing, so it looks okay that we do not set `dispatchedModelSyncedEvents = [:]` since configure will just add new ones/replace existing entries. when the model registry changes, then `dispatchedModelSyncedEvents` may have extra entries if this happens, which is just held in memory won't get used by the plugin. so, it make sense to set it back to `[:]`  on reset

PR #1502, opened 2021-11-01
feat(AWSPluginsCore): Add non-blocking methods to AWSAuthServiceBehavior / deprecate existing blocking methods  

atierian:
*Issue #, if available:* 
443

*Description of changes:*
Add new non-blocking APIs to AWSAuthServiceBehavior and deprecate existing blocking APIs.

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthService.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthService.swift
@@ -16,18 +16,7 @@ public class AWSAuthService: AWSAuthServiceBehavior {
     public func getCredentialsProvider() -> AWSCredentialsProvider {
         return AmplifyAWSCredentialsProvider()
     }
-    
-    #if compiler(>=5.5) && canImport(_Concurrency)
-    @available(iOS 13, *)
-    /// Retrieves the identity identifier for this authentication session from Cognito
-    /// - Returns: The identity token or an error.
-    public func getIdentityId() async -> Result<String, AuthError> {
-        await withCheckedContinuation {
-            getIdentityId(completion: $0.resume(returning:))
-        }
-    }
-    #endif
-    
+
     /// Retrieves the identity identifier for this authentication session from Cognito.
     /// - Parameter completion: Completion handler defined for the input `Result<String, AuthError>`
     public func getIdentityId(completion: @escaping (Result<String, AuthError>) -> Void) {
@@ -41,7 +30,7 @@ public class AWSAuthService: AWSAuthServiceBehavior {
                     Did not receive a valid response from fetchAuthSession for identityId.
                     """)))
                 }
-                
+
                 completion(identityID)
             case .failure(let error):
                 completion(.failure(error))
@@ -49,11 +38,7 @@ public class AWSAuthService: AWSAuthServiceBehavior {
         }
     }
 
-    @available(*, deprecated, message: """
-            Use getIdentityId(completion: (Result<String, AuthError>) -> Void)
-            -- or --
-            getIdentityId() async -> Result<String, AuthError> instead.
-            """)
+    @available(*, deprecated, message: "Use getIdentityId(completion: (Result<String, AuthError>) -> Void) instead")
     public func getIdentityId() -> Result<String, AuthError> {
         var result: Result<String, AuthError>?
         let semaphore = DispatchSemaphore(value: 0)
@@ -120,18 +105,7 @@ public class AWSAuthService: AWSAuthServiceBehavior {
         }
         return .success(convertedDictionary)
     }
-    
-    #if compiler(>=5.5) && canImport(_Concurrency)
-    @available(iOS 13, *)
-    /// Retrieves the Cognito token from the AuthCognitoTokensProvider
-    /// - Returns: The Cognito token or an error.
-    public func getToken() async -> Result<String, AuthError> {
-        await withCheckedContinuation {
-            getToken(completion: $0.resume(returning:))
-        }
-    }
-    #endif
-    
+
     /// Retrieves the Cognito token from the AuthCognitoTokensProvider
     /// - Parameter completion: Completion handler defined for the input `Result<String, AuthError>`
     public func getToken(completion: @escaping (Result<String, AuthError>) -> Void) {
@@ -152,11 +126,7 @@ public class AWSAuthService: AWSAuthServiceBehavior {
         }
     }
 
-    @available(*, deprecated, message: """
-            Use getToken(completion: (Result<String, AuthError>) -> Void)
-            -- or --
-            getToken() async -> Result<String, AuthError> instead.
-            """)
+    @available(*, deprecated, message: "Use getToken(completion: (Result<String, AuthError>) -> Void) instead")
     public func getToken() -> Result<String, AuthError> {
         var result: Result<String, AuthError>?
         let semaphore = DispatchSemaphore(value: 0)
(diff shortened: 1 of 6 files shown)


lawmicha:
are there a lot of internal usages of the now deprecated API? I know storage plugin internally uses .getIdentityId() . perhaps we can make a list in this PR, then reference the list in an issue for us to track updating internal usages of the deprecated API.

lawmicha:
title suggestion: `fix(AWSPluginsCore): provide new async methods for getIdentityId` ?

lawmicha:
> are there a lot of internal usages of the now deprecated API? I know storage plugin internally uses .getIdentityId() . perhaps we can make a list in this PR, then reference the list in an issue for us to track updating internal usages of the deprecated API.

https://github.com/aws-amplify/amplify-ios/search?q=getIdentityId

follow up to remove usage of the deprecated API in APIPlugin, Storage, etc

atierian:
This PR addresses [Issue 443](https://github.com/aws-amplify/amplify-ios/issues/443).
The protocol `AWSAuthServiceBehavior` has two methods, `getIdentityId() -> Result<String, AuthError>` and `getToken() -> Result<String, AuthError>`, that do need to do asynchronous work, but return synchronously. As a consequence, the implementation by `AWSAuthService` blocks the calling thread out of necessity. 

This fix deprecates the two blocking methods mentioned above, and introduces completion handlers alternatives:
- `getIdentityId(completion: @escaping (Result<String, AuthError>) -> Void)`
- `getToken(completion: @escaping (Result<String, AuthError>) -> Void)`

In order to prevent a breaking change through adding a new requirement to the public protocol `AWSAuthServiceBehavior`, we have added default implementations for the two new methods. These default implementations call the existing blocking methods, but on a queue with a `utility` quality of service. These default implementations shall be removed when we're ready to remove the now deprecated methods.

A list of internal blocking method callees is including in the code documentation in `AWSAuthServiceBehavior`. 
For ease of future reference, here it is:

     `StorageAccessLevelAwarePrefixResolver`
        - File Path: `AmplifyPlugins/Storage/AWSS3StoragePlugin/Configuration/AWSS3PluginPrefixResolver.swift`
        - Uses: `getIdentityId()`
     
     `IncomingAsyncSubscriptionEventPublisher`
        - File Path: `AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthServiceBehavior.swift`
        - Uses: `getToken()`
     
     `BasicUserPoolTokenProvider`
        - File Path: `AmplifyPlugins/Core/AWSPluginsCore/Auth/Provider/AuthTokenProvider.swift`
        - Uses: `getToken()`
     
     `AWSOIDCAuthProvider`
        - File Path: `AmplifyPlugins/API/AWSAPICategoryPlugin/SubscriptionFactory/AWSOIDCAuthProvider.swift`
        - Uses: `getToken()`
     
     `AWSAuthServiceBehavior` protocol extension. Default implementation of new completion handler APIs to prevent breaking change.
         - File Path: `AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthServiceBehavior.swift`
         - Uses: `getToken()`, `getIdentityId()`

Additionally the callees can be searched with the following links:
https://github.com/aws-amplify/amplify-ios/search?q=getIdentityId
https://github.com/aws-amplify/amplify-ios/search?q=getToken

The recommended course of action is for Amplify to switch to the new APIs module by module to prevent any issues.
This fix should address the issues we're seeing with thread explosion when making several requests.

atierian:
One note to explicitly call out here, this PR adds two Swift Concurrency `async` method overloads to `AWSAuthService`.

AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthServiceBehavior.swift
```diff
@@ -12,9 +12,67 @@ public protocol AWSAuthServiceBehavior: AnyObject {
...
+         - Uses: `getToken()`, `getIdentityId()`
+     */
+    
+    
+    /// Retrieves the identity identifier of for the Auth service
+    /// - Parameter completion: Completion handler defined for the input `Result<String, AuthError>`
+    /// - Note: This default implementation was added to prevent a breaking change,
+    /// and will be removed when the blocking API versions are removed.
+    public func getIdentityId(completion: @escaping (Result<String, AuthError>) -> Void) {
+        DispatchQueue.global(qos: .utility).async { [weak self] in
```
brennanMKE, line 62: I would remove `qos: .utility` so that it can inherit priority.

AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthServiceBehavior.swift
```diff
@@ -12,9 +12,67 @@ public protocol AWSAuthServiceBehavior: AnyObject {
...
+            completion(identityIdResult)
+        }
+    }
+
+    /// Retrieves the token from the Auth token provider
+    /// - Parameter completion: Completion handler defined for the input `Result<String, AuthError>`
+    /// - Note: This default implementation was added to prevent a breaking change,
+    ///  and will be removed when the blocking API versions are removed.
+    public func getToken(completion: @escaping (Result<String, AuthError>) -> Void) {
+        DispatchQueue.global(qos: .utility).async { [weak self] in
```
brennanMKE, line 73: Same as the comment above.

brennanMKE (commented):
See comments.

AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthServiceBehavior.swift
atierian: Would it inherit priority? I believe it would use the [`default`](https://developer.apple.com/documentation/dispatch/dispatchqos/qosclass/default) QOS ([`global` method doc](https://developer.apple.com/documentation/dispatch/dispatchqueue/2300077-global)).
Do we want to be using `default` QOS here?

atierian: In agreement after further conversation. Specifically, the insight that these requests generally aren't making network requests.
Addressed in afa08aa.

AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthServiceBehavior.swift
atierian: Addressed in afa08aa.

brennanMKE (approved):
LGTM

AmplifyPlugins/Core/AWSPluginsCore/Auth/AWSAuthService.swift
```diff
@@ -84,6 +106,27 @@ public class AWSAuthService: AWSAuthServiceBehavior {
         return .success(convertedDictionary)
     }
 
+    /// Retrieves the Cognito token from the AuthCognitoTokensProvider
+    /// - Parameter completion: Completion handler defined for the input `Result<String, AuthError>`
+    public func getToken(completion: @escaping (Result<String, AuthError>) -> Void) {
```
royjit, line 111: Should we rename this as `getUserPoolAccessToken`? `getToken` is too generic; cognito User Pool returns accessToken, idToken and refreshToken.

harsh62: +1

atierian: I'll defer to your Auth knowledge for proper naming here. Since we're renaming, my suggestion would be to drop the `get` prefix. So the new methods would be `userPoolAccessToken(…)` and `identityID(…)` (notice the capitalization change from `Id` to `ID`).

Thoughts?

atierian: I recant dropping the `get` prefix.
Method names are now `getUserPoolAccessToken(completion:)` and `getIdentityID(completion:)`.
Addressed in 7f71335.
Tests also added in eb1fe4c.

royjit (approved):

Issue #1508, opened 2021-11-05
Support SwiftUI Previews from code in Swift packages


brennanMKE:
### Is your feature request related to a problem? Please describe.

When code is in a Swift package it doe not have the `.json` files required when `Amplify.configure` is called. It results in an exception which causes Previews to crash. This is because the app has these files included as resources but the Swift package does not.

### Describe the solution you'd like

The `Amplify.configure` command should return silently when it is called while running for SwiftUI Previews.

```swift
var isRunningForPreviews: Bool {
    ProcessInfo.processInfo.environment["XCODE_RUNNING_FOR_PREVIEWS"] != nil
}
```

### Describe alternatives you've considered

I attempted to configure AWS with a dummy dictionary but that did not work.

kierancrown:
Is there any update on this? I'm unable to preview with Amplify installed.

Pedrosolisg:
Hi @brennanMKE Thanks for actioning this change! I've seen that it has been merged already a while ago, and yet I am experiencing the crashes on my previews.

Is there anything I need to configure additionally so the flag you implemented has effect when rendering previews?

Thanks a lot!!

harsh62:
@Pedrosolisg can you share any logs for the crash that you are seeing?

Pedrosolisg:
> @Pedrosolisg can you share any logs for the crash that you are seeing?

Thanks for replying to quick! @harsh62 :)

Below the crash report by Xcode when trying to render the Preview on the canvas:

`-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Incident Identifier: 4DED7D1A-83C1-409F-9F7A-9467283D906B
CrashReporter Key:   B17AFC89-7248-A1EE-5B40-4A5F667C3E76
Hardware Model:      MacBookPro16,1
Process:             Kalisten [67794]
Path:                /Users/USER/Library/Developer/Xcode/UserData/Previews/Simulator Devices/74D1B278-0DF5-4A83-AB10-35363A4E4ED3/data/Containers/Bundle/Application/FC5DFDC3-367B-45DE-A0F9-9D869287AA82/Kalisten.app/Kalisten
Identifier:          vilhon.Kalisten
Version:             1.03 (1)
Code Type:           X86-64 (Native)
Role:                Foreground
Parent Process:      launchd_sim [7496]
[...]

PR #1509, opened 2021-11-05
feat:(Core) prevents `Amplify.configure` while running for SwiftUI Previews

brennanMKE:
Previews crash when it is not possible to load `awsconfiguration.json` when `Amplify.configure` is run which prevents this feature from working in Xcode.

*Issue #, if available:*

#1508 

*Description of changes:*

- [x] All unit tests pass

### DIFF
--- a/Amplify/Core/Configuration/AmplifyConfiguration.swift
+++ b/Amplify/Core/Configuration/AmplifyConfiguration.swift
@@ -95,6 +95,10 @@ extension Amplify {
     ///
     /// - Parameter configuration: The AmplifyConfiguration for specified Categories
     public static func configure(_ configuration: AmplifyConfiguration? = nil) throws {
+        guard !isRunningForSwiftUIPreviews else {
+            log.info("Running for SwiftUI reviews, skipping configuration.")
+            return
+        }
         log.info("Configuring")
         log.debug("Configuration: \(String(describing: configuration))")
         guard !isConfigured else {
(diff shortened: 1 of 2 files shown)


brennanMKE:
This Swift package was created to test this change. It includes `setup.sh` to clone the required repo.

[1508-previews.zip](https://github.com/aws-amplify/amplify-ios/files/7489261/1508-previews.zip)

harsh62 (approved):

lawmicha:
i don't think `feat:(Core)` will get picked up by the release scripts as a **Core** feature, since `(Core)` is after the semicolon, just FYI we may need to manually modify the changelog later

hervenivon:
I never had issue with my projects' using Amplify to run previews - until this release 😬

Do you have any guidance to condition the use of plugins methods as the `Amplify.isConfigured` is inaccessible due to 'internal' protection level?

I would love to avoid use of target environment compiler's directive `#if targetEnvironment(simulator)`

Best,

harsh62:
@hervenivon .. It would be nice if you could comment on the issue https://github.com/aws-amplify/amplify-ios/issues/1508 with the exact problem you are facing for tracking purposes. May be provide some examples of why you think the solution in the PR should be avoided. It will help us to prioritize and resolve your issue. Thanks.

pierceglennie:
@harsh62 I believe I'm facing the same issue as @hervenivon.

Before the update to 1.17.0, previews worked fine. This is presumably because `Amplify.configure` worked because I had a `awsconfiguration.json` file in the project, unlike the problem case where the code was in a Swift Packages (i.e. issue #1508). Use of any following methods such as `Amplify.Auth.fetchAuthSession` then worked, since Amplify was configured.

With 1.17.0, I now find that the previews all crash. This is because the changes in this PR mean that Amplify has not been configured before `Amplify.Auth.fetchAuthSession` is used by the preview process, and hence there is a fatal error.

What's the intended approach to avoid this crash? I don't believe there is any method to check whether Amplify is configured so the only solution I have been able to use is checking for `ProcessInfo.processInfo.environment["XCODE_RUNNING_FOR_PREVIEWS"] != "1"` before calling any Amplify methods. 

I'm assuming there is a better solution though. Unfortunately, the documentation doesn't seem to address this anywhere so far as I can see.

evan-templeton:
Any update on this?

PR #1516, opened 2021-11-10
fix(DataStore): support predicate evaluation on model with Enum

lawmicha:
*Issue #, if available:*

*Description of changes:*
Predicate evaluation on a model for a predicate with Enum value was not returning the correct result. For example, a predicate "not equals" to PostStatus.published (`Post.keys.status.eq(PostStatus.published)`) used to evaluate a model with status == published would evaluate to false when it should be true. Debugging the code looks like it was falling into the default case of return false. This PR adds the code necessary to extract the string value from the model correctly so that it can be compared to the predicate's value as a String comparison.

- [x] All unit tests pass

### DIFF
--- a/Amplify/Categories/DataStore/Query/QueryPredicate.swift
+++ b/Amplify/Categories/DataStore/Query/QueryPredicate.swift
@@ -130,6 +130,7 @@ public class QueryPredicateOperation: QueryPredicate {
         guard let fieldValue = target[field] else {
             return false
         }
+
         guard let value = fieldValue else {
             return false
         }
@@ -145,8 +146,13 @@ public class QueryPredicateOperation: QueryPredicate {
         if let intValue = value as? Int {
             return self.operator.evaluate(target: intValue)
         }
+
         if let timeValue = value as? Temporal.Time {
             return self.operator.evaluate(target: timeValue)
+        }
+
+        if let enumValue = value as? EnumPersistable {
+            return self.operator.evaluate(target: enumValue.rawValue)
         }
 
         return self.operator.evaluate(target: value)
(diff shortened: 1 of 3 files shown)


AmplifyPlugins/Core/AWSPluginsCoreTests/Query/QueryPredicateEvaluateGeneratedEnumTests.swift
```diff
@@ -0,0 +1,67 @@
...
+        let predicate = Post.keys.status.ne(PostStatus.published)
+        let instance = Post(title: "title",
+                            content: "content",
+                            createdAt: .now(),
+                            updatedAt: .now(),
+                            status: .draft)
+
+        let evaluation = try predicate.evaluate(target: instance.eraseToAnyModel().instance)
+
+        XCTAssert(evaluation)
```
thisisabhash, line 39: nit suggestion:  XCTAssertTrue for readability

Amplify/Categories/DataStore/Query/QueryPredicate.swift
```diff
@@ -145,10 +146,15 @@ public class QueryPredicateOperation: QueryPredicate {
         if let intValue = value as? Int {
             return self.operator.evaluate(target: intValue)
         }
+
         if let timeValue = value as? Temporal.Time {
             return self.operator.evaluate(target: timeValue)
         }
 
+        if let enumValue = value as? EnumPersistable {
+            return self.operator.evaluate(target: enumValue.rawValue)
```
thisisabhash, line 155: in terms of other predicates like `greaterThan` or `lessThan`, the comparison will check for lexicographical ordering of `rawValue` of `Enum` right? It is not a valid usecase, but just wanted to understand the how it'd work

thisisabhash (approved):
LGTM, left minor comments

Amplify/Categories/DataStore/Query/QueryPredicate.swift
lawmicha: yes that's correct, for example, the developer could take the string version of the enum and create a predicate like this for that check
```
let predicate = Post.keys.status.gt(PostStatus.draft.rawValue)
```

All posts with `.published` or `private` would match and is greater than `.draft`

lawmicha: Added some more unit tests in here as "EnumToString" use cases

thisisabhash: LGTM

PR #1522, opened 2021-11-16
fix(DataStore): Remove from snapshot when item no longer matches predicate

lawmicha:
*Issue #, if available:*

*Description of changes:*
This PR addresses the issue where an item that is originally added to the snapshot (given a predicate with the observeQuery), won't get removed when it is updated to no longer match the predicate.

- [ ] All unit tests pass


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/Support/SortedList.swift
```diff
@@ -9,42 +9,52 @@ import Amplify
...
-                sortedModels.remove(at: index)
+            return remove(model)
+        }
+
+        guard let sortInputs = sortInput else {
+            // If there is no sort order, check if it exists to replace, otherwise add it to the end
+            if modelIds.contains(model.id), let index = sortedModels.firstIndex(where: { $0.id == model.id }) {
+                sortedModels[index] = model
+            } else {
+                sortedModels.append(model)
```
thisisabhash, line 38: can we move 
```
sortedModels.append(model)
modelIds.insert(model.id)
``` 
in a separate function, to make sure both are happening together to keep parity?
like `remove()` function

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreObserveQueryTests.swift
```diff
@@ -235,4 +522,65 @@ class DataStoreObserveQueryTests: SyncEngineIntegrationTestBase {
...
+                receivedPost.fulfill()
+            }
+        }
+        _ = Amplify.DataStore.save(post)
+        wait(for: [receivedPost], timeout: 100)
+        sink.cancel()
+    }
+
+    func deletePostAndWaitForSync(_ post: Post) {
+        let receivedPost = expectation(description: "received Post")
```
thisisabhash, line 549: nit description : deleted post

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreObserveQueryTests.swift
```diff
@@ -235,4 +522,65 @@ class DataStoreObserveQueryTests: SyncEngineIntegrationTestBase {
         wait(for: [startCompleted, secondSnapshot], timeout: 10)
         sink.cancel()
     }
+
+    // MARK: - Helpers
+
+    func savePostAndWaitForSync(_ post: Post) {
+        let receivedPost = expectation(description: "received Post")
+        receivedPost.expectedFulfillmentCount = 2
```
thisisabhash, line 530: we can add a small note here about `expectedFulfillmentCount` being 2

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreObserveQueryTests.swift
```diff
@@ -104,11 +179,223 @@ class DataStoreObserveQueryTests: SyncEngineIntegrationTestBase {
...
+    ///     - Delete a model that does NOT match the predicate. No snapshot is emitted
+    func testPredicateWithCreateUpdateDelete() throws {
+        let started = expectation(description: "Amplify started")
+        try startAmplify {
+            started.fulfill()
+        }
+        wait(for: [started], timeout: 2)
+        Amplify.Logging.logLevel = .info
+
+        let testId = UUID().uuidString
```
thisisabhash, line 217: nit rename to `testContent`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreObserveQueryTests.swift
```diff
@@ -104,11 +179,223 @@ class DataStoreObserveQueryTests: SyncEngineIntegrationTestBase {
...
+                XCTAssertEqual(items.count, 1)
+                XCTAssertEqual(items[0].title, "xyz 1")
+            case 4:
+                // See (2), should not be added to the snapshot.
+                // See (3), should be removed from the snapshot. So the resulting snapshot is empty.
+                self.log.info("4. \(querySnapshot)")
+                XCTAssertEqual(items.count, 0)
+            case 5, 6:
+                // See (4). the post that now matches the snapshot should be added
+                self.log.info("5/6. \(querySnapshot)")
```
thisisabhash, line 250: for update and predicate match, snapshots are received twice `case 5,6`
for update and predicate not matching, snapshot is received once `case 4` 
is this the case?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreObserveQueryTests.swift
```diff
@@ -104,11 +179,223 @@ class DataStoreObserveQueryTests: SyncEngineIntegrationTestBase {
...
+                    XCTAssertEqual(items[1].title, "m")
+                case 6, 7:
+                    XCTAssertEqual(items.count, 3)
+                    XCTAssertEqual(items[0].title, "a")
+                    XCTAssertEqual(items[1].title, "m")
+                    XCTAssertEqual(items[2].title, "z")
+                case 8, 9:
+                    XCTAssertEqual(items.count, 3)
+                    XCTAssertEqual(items[0].title, "m")
+                    XCTAssertEqual(items[1].title, "n")
```
thisisabhash, line 350: can we assert on one more field like `content` to highlight that the actual `Post` object changed positions in the snapshot? 
for e.g. create posts with content like "1", "2", "3" and assert on content

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
```diff
@@ -310,31 +313,77 @@ public class AWSDataStoreObserveQueryOperation<M: Model>: AsynchronousOperation,
...
-        updateCurrentItems(with: itemsChanged)
-        passthroughPublisher.send(currentSnapshot)
-        if log.logLevel >= .debug {
-            let time = stopwatch.stop()
-            log.debug("Time to generate snapshot: \(time) seconds")
+    // Item changes after sync is more elaborate than item changes during sync because the item was never filtered out
+    // by the predicate (unlike during sync). An item that no longer matches the predicate may already exist in the
+    // snapshot and now needs to be removed. The evaluation is done here under the serial queue since checking to
+    // remove the item requires that check on `currentItems` and is required to be performed under the serial queue.
+    func onItemChangeAfterSync(mutationEvent: MutationEvent) {
```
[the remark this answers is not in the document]
lawmicha: After sync, we only process a single item at a time so i think it should be `onItemChange..`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/DataStoreObserveQueryOperation.swift
```diff
@@ -310,31 +313,77 @@ public class AWSDataStoreObserveQueryOperation<M: Model>: AsynchronousOperation,
...
+                    if self.currentItems.apply(model: model, mutationType: mutationType) {
+                        self.sendSnapshot()
+                    }
+                    return
+                }
+
+                // 2. When there is a predicate, evaluate further
+                let modelMatchesPredicate = predicate.evaluate(target: model)
+
+                guard !modelMatchesPredicate else {
```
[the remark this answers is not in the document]
lawmicha: the guard forcses us to return in the body when the model matches the predicate, so i think it is better to keep it as the guard

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/Support/SortedList.swift
lawmicha: i'm not sure exactly what the name of the separate function you are thinking of since these two lines are only ever called once, however to keep things in parity I extracted the additional logic above these two lines into a method called `appendOrReplace(model)`, which has a similar structure to `remove` so there's some parity here now

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreObserveQueryTests.swift
lawmicha: the intent here was to use a predicate on the `testId` for this particular test run, so data is isolated to local models of the test. the fact that it's stored in content is not really relevant and would prefer to keep this as `testId` or `testIdentifier`, let me know if you have a preference or still think it should be `testContent`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreObserveQueryTests.swift
lawmicha: > for update and predicate match, snapshots are received twice case 5,6

Yup because the update is added twice due to the locally source mutation and remote event

> for update and predicate not matching, snapshot is received once case 4

Yes because the item has been removed from the snapshot (due to locally source mutation), and the remote event will no longer be applied to the snapshot (since it does not exist in the snapshot anymore) so the second event does not cause a snapshot to be emitted a second time

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreObserveQueryTests.swift
lawmicha: currently, the `content` is used to isolate data to the locally scoped models that were produced from this test, and then to only operate on the isolated data, this predicate is used `where: Post.keys.content == testId,` . If we modify the content then we won't have consistent count of items (items from other test runs will impact the latest test run). I can set the `rating` to 1.0, 2.0, 3.0 to achieve what you're asking but not sure why the assert on "m" and "n" is not sufficient?

thisisabhash: it was just to highlight that actual object(whose `title` was updated) changed places after sorting as other fields like `content` are identical for all posts. asserting on any other distinguishing field(like `id` or `rating`) would be a good to have. i would leave it up to you.

thisisabhash (approved):

PR #1539, opened 2021-11-19
feat(Storage): Add async resolvePrefix

lawmicha:
*Issue #, if available:*
Related to the original feature: https://github.com/aws-amplify/amplify-ios/pull/1277

*Description of changes:*

This is a non-breaking change (minor version bump) to add an asynchronous `resolvePrefix` method for the plugin to use and for developers to override.

**Why this is non-breaking**

Scenario 1. Developer is not overriding any prefix resolver and relies on the plugin to resolve the prefix.
When upgrading, the plugin will use the built in `StorageAccessLevelAwarePrefixResolver` which now calls the `resolvePrefix(completion)`. We can later improve this (draft https://github.com/aws-amplify/amplify-ios/pull/1544) by calling the async `authService.getIdentityId()` within the `resolvePrefix(completion)` call once this PR is merged: https://github.com/aws-amplify/amplify-ios/pull/1502

Scenario 2. Developer has implemented their own `AWSS3PluginPrefixResolver` with synchronous `resolvePrefix -> Result` method.
When upgrading..
1. The code compiles because the plugin provides a default implementation of the async resolvePrefix, which calls the sync resolvePrefix.
2. Developers can continue to keep this code but is advised to change the implementation to the async resolvePrefix. There is a deprecated flag on the sync resolvePrefix method. If developer decides to take no action, it will continue to work because the plugin will get their resolver, call the async resolvePrefix, which defaults to calling their sync resolvePrefix.
3. If developer changes their implementation to the async resolvePrefix then it will work because the plugin gets their resolver, calls the async resolvePrefix, which is their updated implementation.

Scenario 3. New Developers with the use case to override the prefix for each access level:
The developer will see that the sync resolvePrefix is deprecated and should implement the async resolvePrefix. The related docs and examples: https://github.com/aws-amplify/docs/pull/3321
- If the developer only implements only the sync resolvePrefix, this continues to work since the plugin calls the async resolvePrefix and its default implementation calls the sync resolvePrefix.
- If the developer only implements the async resolve prefix, the conformance is valid and will work since there's a default implementation provided by the library for the sync resolvePrefix
- If the developer implements both sync and async resolvePrefix, the plugin will use the async resolvePrefix and sync resolvePrefix will never be used.

- [x] All unit tests pass
- [x] All integration tests pass  (https://github.com/lawmicha/amplify-ios/actions/runs/1491935488)


lawmicha:
Ran integration tests locally and they are passing, trying to perform them in using GitHub actions to see how consistent they are: https://github.com/lawmicha/amplify-ios/runs/4289928769?check_suite_focus=true

brennanMKE (approved):
LGTM

Issue #1547, opened 2021-11-23
Apple Silicon Support


brennanMKE:
### Describe the bug

A dependency for testing, CwlPreconditionTesting, requires changes to run on a modern Apple Silicon Mac.

The following links provide further details.

https://github.com/mattgallagher/CwlPreconditionTesting/issues/21
https://github.com/wwt/SwiftCurrent/pull/107/files

### Steps To Reproduce

```swift
Steps to reproduce the behavior:

* Build and run tests on M1 Mac
```

### Expected behavior

Should be and run tests on M1 Mac

### Amplify Framework Version

1.16.1

### Amplify Categories

Analytics, API, DataStore

### Dependency manager

Cocoapods

### Swift version

5.0

### CLI version

4.4.25

### Xcode version

13.2

### Relevant log output

```shell
...
```

### Is this a regression?

No

### Regression additional context

...

### Device

iPhone Simulator

### iOS Version

12.0

brennanMKE:
Merged

PR #1548, opened 2021-11-23
Supports building and testing on Apple Silicon

brennanMKE:
*Issue #, if available:*

#1547 

A special testing dependency is able catch fatal errors but it requires changes to work with Apple Silicon. It has been updated to support the new M1 chips and 2 new test functions have been added to AmplifyTestCommon.

* `XCTAssertThrowFatalError`
* `XCTAssertNoThrowFatalError`

These functions use CwlPreconditionTesting and provide an API which matches XCTest so that it is familiar. All direct imports of CwlPreconditionTesting have been removed so it is only used from AmplifyTestCommon to support these 2 functions. Apple's API will show in a different color in Xcode than custom functions which are added in this code base.

## Flaky Tests

There were several tests which were failing sporadically and each were using `DispatchSemaphore` instead of an `XCTestExpectation` and so these were changed out and now the tests run reliably.

- [x] All unit tests pass


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 ## Unreleased
+
+### Build Fixes
+
+- **Apple Silicon**: Adds support to build and run tests on M1 Macs
+
```
diegocstn, line 7: this might not be necessary as the changelog is automatically updated by the release process

diegocstn (approved):

CHANGELOG.md
brennanMKE: I could take it out.

AmplifyTestCommon/Helpers/AmplifyAssertions.swift
```diff
@@ -0,0 +1,43 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import XCTest
+
+@testable import CwlPreconditionTesting
+
+public func XCTAssertThrowFatalError(_ expression: @escaping () -> Void,
```
royjit, line 13: Is this different from https://developer.apple.com/documentation/xctest/1500795-xctassertthrowserror

brennanMKE: Yes, it is meant to look like those XCTest functions. This one catches fatal errors like the CwlPreconditionTesting package does.

PR #1582, opened 2022-01-03
feat (auth): add deleteUser API

ameter:
*Issue #, if available:*

*Description of changes:*

- [x] All unit tests pass


AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
```diff
@@ -0,0 +1,44 @@
...
+#if COCOAPODS
+import AWSMobileClient
+#else
+import AWSMobileClientXCF
+#endif
+
+extension AuthenticationProviderAdapter {
+
+    func deleteUser(request: AuthDeleteUserRequest, completionHandler: @escaping (Result<Void, AuthError>) -> Void) {
+        awsMobileClient.deleteUser(signOut: false) { [weak self] error in
```
palpatim, line 18: Why is `signOut` false rather than true? It's not clear what relationship this argument has to the `globalSignOut` call below on L20, so a comment is in order to explain the apparent discrepancy

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
```diff
@@ -0,0 +1,44 @@
...
+                    switch result {
+                    case .success:
+                        completionHandler(.success(()))
+                        return
+                    case .failure(let error):
+                        completionHandler(.failure(AuthErrorHelper.toAuthError(error)))
+                        return
+                    }
+                }
+                return
```
palpatim, line 32: Looks like we wouldn't invoke `completionHandler` if `self` is nil for some reason. That may be expected, but wanted to call it out.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Support/Constants/AuthPluginErrorConstants.swift
```diff
@@ -59,6 +59,10 @@ struct AuthPluginErrorConstants {
         "There is no user signed in to retreive user sub",
         "Call Auth.signIn to sign in a user and then call Auth.fetchSession")
 
+    static let deleteUserSignOutError: AuthPluginErrorString = (
+        "There is no user signed in to delete an account",
+        "Call Auth.signIn to sign in a user and then call Auth.deleteUser")
```
palpatim, line 64: Is it necessary to call `fetchSession` first, or is `signIn` sufficient?

AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/AuthDeleteUserTests/AuthDeleteUserTests.swift
```diff
@@ -0,0 +1,139 @@
...
+        sleep(2)
+    }
+
+    /// Test deleteUser in authenticated state
+    ///
+    /// - Given: An authenticated state
+    /// - When:
+    ///    - I invoke `deleteUser`
+    /// - Then:
+    ///    - I should get successul result, the user should no longer exist, and the auth session should be signedout
```
palpatim, line 33:
```suggestion
    ///    - I should get successful result, the user should no longer exist, and the auth session should be signedOut
```

Amplify/Categories/Auth/Request/AuthDeleteUserRequest.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+/// Request for delete user
+public struct AuthDeleteUserRequest: AmplifyOperationRequest {
+
+    /// Options is unused for AuthDeleteUserRequest.  It is included for conformance
```
royjit, line 13:
```suggestion
    /// Options is unused for AuthDeleteUserRequest. It is included for conformance
```

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
```diff
@@ -0,0 +1,44 @@
...
+            }
+            if case .notSignedIn = error as? AWSMobileClientError {
+                let authError = AuthError.signedOut(AuthPluginErrorConstants.deleteUserSignOutError.errorDescription,
+                                                    AuthPluginErrorConstants.deleteUserSignOutError.recoverySuggestion,
+                                                    error)
+                completionHandler(.failure(authError))
+                return
+            }
+            completionHandler(.failure(AuthErrorHelper.toAuthError(error)))
```
royjit, line 41: Does this give the right error message if the user is signed in by federating to Cognito Identity Pool?

We should test:
1. Sign in using username/password SRP (should delete user )
2. Sign in using hostedUI (should delete user )
3. Sign in via federation to Identity Pool ( should return error )

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
ameter: By default, ASWMobileClient calls signOut internally during deleteUser. For Amplify, we instead call Amplify's signOut function, which contains some higher level logic. This has the advantage of keeping all the Amplify signOut/cleanup code in one place. I added a comment to document this.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
ameter: Correct. The instance of AuthenticationProvider that is referenced by self is retained during plugin configuration and cannot be nil if the plugin is successfully configured. It the plugin configuration is invalid, configuration will fail and we will never reach this point. An alternative would be to use `guard let self = self else { return }` to make the intent explicit. With a single call, both styles are used elsewhere in the plugin.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Support/Constants/AuthPluginErrorConstants.swift
ameter: signIn is sufficient for deleteUser.  I believe the only time fetchAuthSession is necessary (other than to check signIn status) is to access the underlying credentials directly.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
ameter: Updated to handle signout in AWSMobileClient in order to better handle hosted UI sessions.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
ameter: Tested with all 3.  1 and 2 delete the user and I updated to error message for 3 to state that there is not a user signed in to Cognito User Pool.

royjit (approved):
LGTM

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
```diff
@@ -0,0 +1,42 @@
...
+import AWSMobileClient
+#else
+import AWSMobileClientXCF
+#endif
+
+extension AuthenticationProviderAdapter {
+
+    func deleteUser(request: AuthDeleteUserRequest, completionHandler: @escaping (Result<Void, AuthError>) -> Void) {
+        // By default, ASWMobileClient calls signOut internally during deleteUser.
+        // For Amplify, we instead call Amplify's signOut function, which contains some higher level logic.
```
royjit, line 19: Remove this comment? Amplify is not calling signOut right?

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
```diff
@@ -0,0 +1,42 @@
...
+        awsMobileClient.deleteUser(signOut: true) { [weak self] error in
+            guard let error = error else {
+                if let self = self, self.userdefaults.isPrivateSessionPreferred() {
+                    // Reset the user defaults.
+                    self.userdefaults.storePreferredBrowserSession(privateSessionPrefered: false)
+                }
+
+                completionHandler(.success(()))
+                Amplify.Hub.dispatch(to: .auth, payload: HubPayload(eventName: HubPayload.EventName.Auth.signedOut))
```
royjit, line 28: Should we dispatch deletedUser event here instead? I am not sure what we have decided across team to be send here when a delete succeed

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/HubEvents/AuthHubEventHandler.swift
```diff
@@ -69,6 +73,13 @@ class AuthHubEventHandler: AuthHubEventBehavior {
                 }
                 self?.sendUserSignedOutEvent()
 
+            case HubPayload.EventName.Auth.deleteUserAPI:
+                guard let event = payload.data as? AWSAuthDeleteUserOperation.OperationResult,
+                    case .success = event else {
+                        return
+                }
+                self?.sendUserDeletedEvent()
```
royjit, line 81: Regarding my previous comment above, it looks like we are sending both events? signedOut and userDeleted? Is this approved across platform?

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
ameter: Yes, we discussed this between all three platforms and emitting both userDeleted and signedOut hub events was the agreed behavior.  The rational for firing both was that in many cases, for top-level state within the app, they may not care whether the user deleted their account any only want to update state on signedIn vs signedOut.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/HubEvents/AuthHubEventHandler.swift
ameter: yes, see my previous comment.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/Dependency/AuthenticationProviderAdapter+DeleteUser.swift
ameter: done

PR #1610, opened 2022-01-26
chore: Added attribution notice for 3P libraries [ci skip]

royjit:
*Issue #, if available:*

*Description of changes:*

- [ ] All unit tests pass

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,2 +1,29 @@
 Amplify for iOS
 Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved. 
+
+**********************
+THIRD PARTY COMPONENTS
+**********************
+
+** SQLite.swift; version 0.12.2 -- https://github.com/stephencelis/SQLite.swift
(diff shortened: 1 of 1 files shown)


NOTICE
```diff
@@ -1,2 +1,29 @@
 Amplify for iOS
 Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved. 
+
+**********************
+THIRD PARTY COMPONENTS
+**********************
+
+** SQLite.swift; version 0.12.2 -- https://github.com/stephencelis/SQLite.swift
```
ameter, line 8:
```suggestion
** SQLite.swift -- https://github.com/stephencelis/SQLite.swift
```
Unless the license changed in a different version, it would likely be better not to include the version here because it would need to be updated every time we update the dependency version and could easily be missed.

ameter (approved):
Approved with comment / suggestion.

NOTICE
royjit: we need to keep this and update attribution when version changes.

PR #1612, opened 2022-01-27
fix(datastore): Add DateTime, Date, Time, Int, Float, Enum field in indexes to GraphQL input correctly for mutations

thisisabhash:
*Issue #, if available:* None 

*Description of changes:* Add DateTime, Date, Time, Int, Float, Enum field in indexes to GraphQL input correctly for mutations when it is a part of custom primary key. For e.g. In the schema below, `id`, `dob`, `date`, `time`, `phoneNumber`, `priority`, `height` are part of Primary Key.
```
type Customer5
@model {
  id: ID! @primaryKey(sortKeyFields: ["dob","date","time","phoneNumber","priority","height"])
  dob: AWSDateTime!
  date: AWSDate!
  time: AWSTime!
  phoneNumber: Int!
  priority: Priority!
  height: Float!
  firstName: String
  lastName: String
}
```

- [x] All unit tests pass
~~- [ ] All integration tests pass~~


AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLRequest/GraphQLRequestSyncCustomPrimaryKeyTests.swift
```diff
@@ -220,4 +221,65 @@ class GraphQLRequestSyncCustomPrimaryKeyTests: XCTestCase {
...
+                    dob
+                    firstName
+                    lastName
+                    updatedAt
+                    __typename
+                    _version
+                    _deleted
+                    _lastChangedAt
+                  }
+                }
```
lawmicha, line 249: For conflict resolution enabled API, we care mostly about create/update/delete mutations for sync to cloud and subscriptions for initial sync process. Looks like we have the delete mutation with version tested, can we also add subscription like onCreate test as a minimum? A stretch goal would be to add test for non-conflict resolution enabled API

lawmicha: + syncQuery test as well!

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLRequest/GraphQLRequestSyncCustomPrimaryKeyTests.swift
```diff
@@ -220,4 +221,153 @@ class GraphQLRequestSyncCustomPrimaryKeyTests: XCTestCase {
...
+            XCTAssertEqual(input["_version"] as? Int, expectedInput["_version"] as? Int)
+        } catch {
+            XCTFail(error.localizedDescription)
+        }
+    }
+
+    func testOnCreateSubscriptionGraphQLRequestWithDateInPK() throws {
+        do {
+            let datetime = try Temporal.DateTime(iso8601String: "2020-01-20T08:00")
+            let customer = CustomerWithDateInPK(dob: datetime, firstName: "John", lastName: "Doe")
```
lawmicha, line 289: can we pass in model name using the schema instead of the object? The method we're testing `subscription(to:subscriptionType)` does not require the a customer object to be passed in so the set up code should be set up the same way

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelIdDecorator.swift
```diff
@@ -18,7 +18,15 @@ public struct ModelIdDecorator: ModelBasedGraphQLDocumentDecorator {
...
             for key in customPrimaryKeys {
-                fields[key] = model[key] as? String
+                if let date = model[key] as? Temporal.Date {
+                    fields[key] = date.iso8601String
+                } else if let dateTime = model[key] as? Temporal.DateTime {
+                    fields[key] = dateTime.iso8601String
+                } else if let time = model[key] as? Temporal.Time {
+                    fields[key] = time.iso8601String
+                } else {
+                    fields[key] = model[key] as? String
```
lawmicha, line 28: the previous code assumed that the primary key can only be a string. 
1. are all values of the primary key just the ones you've listed here? (string, datetime, date, time). are there anything else? It doesn't make sense to fix it for one field type but not for others. 
2. based on 1. if it expands to more than (string, datetime, date, time), then is there already some code that allows you to extract the graphQL value out of the field? ie. can we reduce the code down to `fields[keys] = model[key].toGraphQLValue()`

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLRequest/GraphQLRequestSyncCustomPrimaryKeyTests.swift
lawmicha: looks like ur doing this below, use `CustomerWithDateInPK.modelName` instead

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLRequest/GraphQLRequestSyncCustomPrimaryKeyTests.swift
```diff
@@ -14,6 +14,7 @@ class GraphQLRequestSyncCustomPrimaryKeyTests: XCTestCase {
 
     override func setUp() {
         ModelRegistry.register(modelType: CustomerOrder.self)
+        ModelRegistry.register(modelType: CustomerWithDateInPK.self)
```
lawmicha, line 17: nit: would prefer this to be a in a different file like `GraphQLRequestSyncCustomPrimaryKeyWithDateTests`

lawmicha: CustomerOrder was written at a very different time, whereas CustomerWithDateInPK was written with V2 transform. Even though this is a unit test, there's an aspect of this where we will most likely provision the backend to continue verifying integration tests. also if we want to add another CPK use case, most likely it will be in another test class so it scales

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelIdDecorator.swift
```diff
@@ -61,3 +63,49 @@ public struct ModelIdDecorator: ModelBasedGraphQLDocumentDecorator {
         return document.copy(inputs: inputs)
     }
 }
+
+extension Model {
+
+    func graphQLInputForPrimaryKey(modelFieldName: ModelFieldName) -> String? {
+
```
lawmicha, line 70: let's move this over to `Model+GraphQL` since the other similar method also lives there and then we're not duplicating `getFieldValue` in two places

AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/ModelIdDecorator.swift
```diff
@@ -61,3 +63,49 @@ public struct ModelIdDecorator: ModelBasedGraphQLDocumentDecorator {
         return document.copy(inputs: inputs)
     }
 }
+
+extension Model {
+
+    func graphQLInputForPrimaryKey(modelFieldName: ModelFieldName) -> String? {
```
lawmicha, line 69: comment like
```
/// Retrieve the custom primary key's value used for the GraphQL input.
    /// Only a subset of model types are applicable as custom indexes such as
    /// `date`, `dateTime`, `time`, `enum`, `string`, `double`, and `int`.
```

AmplifyPlugins/Core/AWSPluginsCoreTests/Model/GraphQLRequest/GraphQLRequestSyncCustomPrimaryKeyWithMultipleFieldsTests.swift
```diff
@@ -0,0 +1,194 @@
...
+
+    func testDeleteMutationGraphQLRequestWithDateInPK() throws {
+        let customer = CustomerWithMultipleFieldsinPK(dob: Temporal.DateTime.now(),
+                                 date: Temporal.Date.now(),
+                                 time: Temporal.Time.now(),
+                                 phoneNumber: 1_234_567,
+                                 priority: Priority.high,
+                                 height: 6.1,
+                                 firstName: "John",
+                                 lastName: "Doe")
```
lawmicha, line 32: indentation

lawmicha (approved):
LGTM, great coverage on the possible sort key fields. minor comment to move the method over to existing extension method

PR #1640, opened 2022-02-21
fix(DataStore): Handle .all predicate for create/update

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/1636

*Description of changes:*
This change the logic to handle when a predicate that is passed in for `save` is equal to `.all`. When `.all` is passed in, ignore it.

### Android's logic
Android's logic https://github.com/aws-amplify/amplify-android/blob/main/aws-datastore/src/main/java/com/amplifyframework/datastore/storage/sqlite/SQLiteStorageAdapter.java#L339-L351

- if model does not exist, predicate does not equal to .all -> throw
- if model does exist , predicate == .all, generate patch item for update.

### iOS Logic
Since iOS doesn't generate patch items for the update, the logic is slightly different:
- if model does not exist, there is a condition and != .all -> invalidCondition error
- if model does exist, there is a condition and != .all, continue with UpdateStatement(condition: condition). when the condition is `.all` for the UpdateStatement, it will translate it to SQL as `or 1 = 1`

- [x] All unit tests pass


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/QueryPredicate+SQLite.swift
```diff
@@ -50,3 +50,13 @@ extension QueryOperator {
...
     }
 }
+
+extension QueryPredicate {
+    var isAll: Bool {
+        if let allPredicate = self as? QueryPredicateConstant, allPredicate == .all {
+            return true
+        } else {
+            return false
+        }
```
royjit, line 60:
```suggestion
        guard let allPredicate = self as? QueryPredicateConstant else {
            return false
        } 
        return allPredicate == .all
```


Will the above code cover if the query passed in is a `QueryPredicateGroup` with an `or QueryPredicateConstant.all` ?

lawmicha: the code won't cover if the `.all` predicate is within a group like OR or AND. This change is added purely for the use case of explicitly passing in just `.all`. This is in discussion with @fjnoyp to match the same logic as Android, and so that allows Flutter's integration with native to work easily, since both native platforms have the same behavior. Passing in `.all` is the same as not passing anything in. From Flutter's prespective and according to Android's code, this is a valid use case. `.all` is a special predicate and I don't think it makes sense to do something like "condition = rating >5 or .all" since that reads "rating is greater than 5 or apply no condition", so that behavior is untested/undefined. Ideally we have compile time check on valid predicates as it would be hard to validate all invalid usages of predicates

royjit (approved):

PR #1663, opened 2022-03-04
fix(DataStore): ObserveQuery InitialQuery empty modelIds set

lawmicha:
*Issue #, if available:* https://github.com/aws-amplify/amplify-ios/issues/1627

*Description of changes:*

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/Support/SortedList.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/Support/SortedList.swift
@@ -8,8 +8,8 @@
 import Amplify
 
 class SortedList<ModelType: Model> {
-    var sortedModels: [ModelType]
-    var modelIds: Set<Model.Identifier>
+    private(set) var sortedModels: [ModelType]
+    private(set) var modelIds: Set<Model.Identifier>
     private let sortInput: [QuerySortDescriptor]?
     private let modelSchema: ModelSchema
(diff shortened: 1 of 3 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Subscribe/Support/SortedList.swift
```diff
@@ -20,10 +20,19 @@ class SortedList<ModelType: Model> {
         self.modelSchema = modelSchema
     }
 
+    /// Sets the internal data structures to the list of models. `sortedModels` should already be sorted to efficiently
+    /// update the internal instructures of `SortedList`. If `sortedModels` is not sorted, then the subsequent
+    /// methods used on the data structure will have adverse consequences, for example adding a model assumes the list
+    /// is already sorted (since it performs a binary search insertion).
```
diegocstn, line 26: what would be an example of adverse consequence? Something acceptable and/or recoverable (like a temporary degrade in performance) or something more severe?

lawmicha: the binary search insertion would not accurately find the correct index since the list is no longer sorted. So the items returned will be out of order. This only impacts the caller if they decided to pass in a sort order

lawmicha:
this has been released in https://github.com/aws-amplify/amplify-ios/releases/tag/v1.22.0

PR #1705, opened 2022-03-23
chore(DataStore): Remove @available attribute for iOS 13.0

thisisabhash:
*Issue #, if available:* None

*Description of changes:* Developer preview bumps the minimum target to iOS 13 which removes the redundant `@available(iOS 13.0, *)` checks.

~~- [ ] Added new tests to cover change, if needed~~
- [x] All unit tests pass
~~- [ ] All integration tests pass~~
~~- [ ] Security oriented best practices and standards are followed (e.g. using input sanitization, principle of least privilege, etc)~~

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
@@ -44,7 +44,6 @@ final public class AWSAPIPlugin: NSObject, APICategoryPlugin, AWSAPIAuthInformat
 
     var authProviderFactory: APIAuthProviderFactory
 
-    @available(iOS 13.0, *)
     var reachabilityMap: [String: NetworkReachabilityNotifier] {
         get {
             if iReachabilityMap == nil {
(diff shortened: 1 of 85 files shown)


atierian (approved):
Looks good. It's obvious now why we're doing this, but do you mind writing a sentence in PR outlining the reason for the removal for future reference. 

Something like: "Dev preview has a minimum target of iOS 13 - this removes the redundant `@available(iOS 13.0, *)` checks."

harsh62 (approved):

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin.swift
```diff
@@ -44,7 +44,6 @@ final public class AWSAPIPlugin: NSObject, APICategoryPlugin, AWSAPIAuthInformat
```
lawmicha, line 44: the comment above for `iReachabilityMap` no longer makes sense
```
/// The underlying map that retains instances of NetworkReachabilityNotifier.  Using a computed property
    /// to work around @available for use on stored properties
    var iReachabilityMap: [String: Any]?
```

if we no longer need `@available`, then we also don't need to use `iReachabilityMap`. can you capture this work, as a follow up if needed since this PR does have a lot of files change?

thisisabhash: I will create a new PR to remove `iReachabilityMap` and link it here.

PR #1706, opened 2022-03-24
fix(DataStore): Reconcile locally sourced mutations while subscriptions are disabled

lawmicha:
*Issue #, if available:*

*Description of changes:*
Reconcile locally sourced mutations to save/update/delete the model and metadata into local store. The locally sourced mutation response is identical to the subscription event that would be received on subscriptions, so reuse the reconciliation queue to process it. 

A concern regarding this is the real-time events behavior of the additional event. Currently, locally sourced mutations with subscriptions enabled will emit 2 events, one from the operation on local store and one from the subscription event being reconciled to the local store. By handling the locally sourced mutation event, the subscription event would attribute to a third real-time event. The reconciliation logic has been updated to to ignore the remote model when the local version is equal to the incoming remote model, which will dedup the subscription event.

**RemoteSyncEngine** new requirement is to provide `outgoingMutationQueue` with the `reconciliationQueue`. Both are now created at `start`, first the `reconciliationQueue` then the `outgoingMutationQueue` with the `reconciliationQueue`. Before, creating the `reconciliationQueue` in `RemoteSyncEngine.initializeSubscription()` also established the subscriptions. A new method is exposed from the reconciliation queue to maintain the same behavior.

**OutgoingMutationQueue** gets the `reconciliationQueue`, so upon processing a MutationEvent (`SyncToCloudOperation`), it will pass the successfully mutation response to the reconciliation queue to process, much like as if it were an subscription event.

**ReconcileAndLocalSave**'s `RemoteSyncReconciler` determines the disposition of the remote model, when compared to the local metadata. If the local metadata's version is equal to the remote model's then skip the remote model.

*Manual Testing done*
The manual testing done is by disabling subscriptions, and performing a save. Two events are received by the publisher for that model. Disabling subscriptions can be done via the data resolver for all 3 subscription operations of a model (onCreate, onUpdate, onDelete) with `$util.error("Subscription are disabled.", "OperationDisabled")` on the last line in *Before mapping template.*

The second manual test done is to ensure that there are only two events dispatched to the caller for locally sourced mutations with subscriptions enabled. Two events are observed, since the reconciliation of the subscription event determines it to be dropped.

- [ ] All unit tests pass


lawmicha:
This PR is no longer needed, a different implementation done to RemoteSyncEngine is implemented in https://github.com/aws-amplify/amplify-ios/pull/1712

PR #1715, opened 2022-03-31
Auth plugin: Add support of custom endpoint for CognitoUserPool

Rusik:
*Description of changes:*
Android SKD supports custom endpoint for Cognito User Pool, while iOS version isn't. This PR adds support of this.

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
@@ -66,7 +66,7 @@ extension AWSCognitoAuthPlugin {
 
     func awsMobileClientAdapter(from authConfiguration: JSONValue) throws -> AWSMobileClientBehavior {
         let identityPoolConfig = identityPoolServiceConfiguration(from: authConfiguration)
-        let userPoolConfig = userPoolServiceConfiguration(from: authConfiguration)
+        let userPoolConfig = try userPoolServiceConfiguration(from: authConfiguration)
 
         // Auth plugin require atleast one of the Cognito service to work. Throw an error if both the service
         // configuration are nil.
@@ -100,7 +100,7 @@ extension AWSCognitoAuthPlugin {
         return AmplifyAWSServiceConfiguration(region: region, credentialsProvider: anonymousCredentialProvider)
     }
 
-    func userPoolServiceConfiguration(from authConfiguration: JSONValue) -> AmplifyAWSServiceConfiguration? {
+    func userPoolServiceConfiguration(from authConfiguration: JSONValue) throws -> AmplifyAWSServiceConfiguration? {
         let regionKeyPath = "CognitoUserPool.Default.Region"
         guard case .string(let regionString) = authConfiguration.value(at: regionKeyPath) else {
             Amplify.Logging.warn("Could not read Cognito user pool information from the configuration.")
@@ -108,12 +108,32 @@ extension AWSCognitoAuthPlugin {
         }
         let region = (regionString as NSString).aws_regionTypeValue()
 
-        let endpointKeyPath = "CognitoUserPool.Default.Endpoint"
-        if case .string(let endpointString) = authConfiguration.value(at: endpointKeyPath) {
-            return AmplifyAWSServiceConfiguration(region: region, endpoint: .init(urlString: endpointString))
+        if  let endpoint = try resolveCognitoOverrideEndpoint(using: authConfiguration, region: region) {
+            return AmplifyAWSServiceConfiguration(region: region, endpoint: endpoint)
         } else {
             return AmplifyAWSServiceConfiguration(region: region)
         }
+    }
+
+    func resolveCognitoOverrideEndpoint(
+        using authConfiguration: JSONValue,
+        region: AWSRegionType) throws -> AWSEndpoint? {
+
+        let endpointKeyPath = "CognitoUserPool.Default.Endpoint"
+        guard case .string(let endpointString) = authConfiguration.value(at: endpointKeyPath) else {
+            return nil
+        }
+            guard let components = URLComponents(string: endpointString),
+                  let url = components.url,
+                    components.scheme == "https" || components.scheme == "http" else {
+            let amplifyError = AuthError.configuration(
+                "Error configuring \(String(describing: self))",
+                """
+                Invalid Endpoint value \(endpointString). Expected a fully-qualified hostname.
+                """)
+            throw amplifyError
+        }
+        return AWSEndpoint(region: region, service: .CognitoIdentityProvider, url: url)
     }
 
     // MARK: Internal
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
@@ -119,22 +119,22 @@ extension AWSCognitoAuthPlugin {
         using authConfiguration: JSONValue,
         region: AWSRegionType) throws -> AWSEndpoint? {
 
-        let endpointKeyPath = "CognitoUserPool.Default.Endpoint"
-        guard case .string(let endpointString) = authConfiguration.value(at: endpointKeyPath) else {
-            return nil
-        }
+            let endpointKeyPath = "CognitoUserPool.Default.Endpoint"
+            guard case .string(let endpointString) = authConfiguration.value(at: endpointKeyPath) else {
+                return nil
+            }
             guard let components = URLComponents(string: endpointString),
                   let url = components.url,
-                    components.scheme == "https" || components.scheme == "http" else {
-            let amplifyError = AuthError.configuration(
-                "Error configuring \(String(describing: self))",
+                  components.scheme == "https" || components.scheme == "http" else {
+                let amplifyError = AuthError.configuration(
+                    "Error configuring \(String(describing: self))",
                 """
                 Invalid Endpoint value \(endpointString). Expected a fully-qualified hostname.
                 """)
-            throw amplifyError
+                throw amplifyError
+            }
+            return AWSEndpoint(region: region, service: .CognitoIdentityProvider, url: url)
         }
-        return AWSEndpoint(region: region, service: .CognitoIdentityProvider, url: url)
-    }
 
     // MARK: Internal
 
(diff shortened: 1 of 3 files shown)


atierian:
Hey @Rusik, thanks for opening this pull request! We'll review it and get back to you with any questions / comments.

jracollins:
Hi @atierian 👋  

I work with @Rusik and just wanted to add some additional context/reasoning for the PR: 

The endpoint functionality is claimed/documented along with its use case in the following AWS post:  https://aws.amazon.com/blogs/security/protect-public-clients-for-amazon-cognito-by-using-an-amazon-cloudfront-proxy/

```
If you have a mobile application that uses the Amplify mobile SDK, you can override the endpoint in your configuration as follows (don't include AppClientSecret parameter in your configuration).
Note that the Endpoint value contains the domain name only, not the full URL. 
This feature is available in the latest releases of the iOS and Android SDKs.

"CognitoUserPool": {
  "Default": {
    "AppClientId": "<APP-CLIENT-ID>",
    "Endpoint": "<CF-DISTRIBUTION-DOMAIN>",
    "PoolId": "<USER-POOL-ID>",
    "Region": "<REGION>"
  }
}
```

This also links to the release of the `aws-sdk-ios` with said functionality not in this `amplify-ios` library:  https://github.com/aws-amplify/aws-sdk-ios/releases/tag/2.24.0

We can confirm that this functionality works in the current Android Amplify library "out of the box" too.

Let us know if you need anything else from us to help get this merged.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
```diff
@@ -107,7 +107,13 @@ extension AWSCognitoAuthPlugin {
...
         }
         let region = (regionString as NSString).aws_regionTypeValue()
-        return AmplifyAWSServiceConfiguration(region: region)
+
+        let endpointKeyPath = "CognitoUserPool.Default.Endpoint"
+        if case .string(let endpointString) = authConfiguration.value(at: endpointKeyPath) {
+            return AmplifyAWSServiceConfiguration(region: region, endpoint: .init(urlString: endpointString))
+        } else {
+            return AmplifyAWSServiceConfiguration(region: region)
+        }
```
royjit, line 116: Thank you for the PR, overall looks good. Can we add a validation for the endpoint similar to what we have in AWS SDK for iOS.


```suggestion

        if  let endpoint = resolveCognitoOverrideEndpoint(using: authConfiguration, region: region) {
            return AmplifyAWSServiceConfiguration(region: region, endpoint: endpoint)
        } else {
            return AmplifyAWSServiceConfiguration(region: region)
        }
    }

    // Uses the same logic as resolveEndpointOverrideFromServiceInfo from
    // PR: https://github.com/aws-amplify/aws-sdk-ios/pull/3556/
    func resolveCognitoOverrideEndpoint(using authConfiguration: JSONValue, region: AWSRegionType) -> AWSEndpoint? {
        let endpointKeyPath = "CognitoUserPool.Default.Endpoint"
        guard case .string(let endpointString) = authConfiguration.value(at: endpointKeyPath) else {
            return nil
        }
        var components = URLComponents()
        components.scheme = "https"
        components.percentEncodedHost = endpointString;

        guard let endpointURL = components.url else {
            let message = """
            Invalid Endpoint value \(endpointString). Expected a fully-qualified hostname.
            """
            fatalError(message)
        }
        return AWSEndpoint(region: region, service: .CognitoIdentityProvider, url: endpointURL)
    }
```

Rusik: Hi @royjit 

I inspected the code from PR you sent in your suggestion https://github.com/aws-amplify/aws-sdk-ios/pull/3556/
It doesn't work as I expected and can't be used to validate urls like `https://example.org`

For example if we run this code
```swift
var comp = URLComponents()
comp.scheme = "https"
comp.percentEncodedHost = "https://example.org"
``` 
we will see obj-с exception in the log
```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** +[NSURLComponents setPercentEncodedHost:]: invalid characters in percentEncodedHost'
```
Seems like it is ok for obj-с code in [AWSCognitoIdentityUserPool.m](https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSCognitoIdentityProvider/AWSCognitoIdentityUserPool.m) but not useful for our case in swift code

So I implemented check like this
```swift
guard let components = URLComponents(string: endpointString), 
           let url = components.url,
           components.scheme == "https" || components.scheme == "http" else {
    let amplifyError = AuthError.configuration(
        "Error configuring \(String(describing: self))",
        """
        Invalid Endpoint value \(endpointString). Expected a fully-qualified hostname.
        """)
    throw amplifyError
}
```
As you can see I used exception throwing strategy instead of `fatalError` because it already used elsewhere in the code

royjit: I think the code mentioned in the PR is to avoid cases like that, the hostname should not include the scheme. The configuration should look something like this:

```
"CognitoUserPool": {
  "Default": {
    "AppClientId": "xxx",
    "Endpoint": "example.org",
    "PoolId": "xxxxx",
    "Region": "xx-xxx-1"
  }
}

```

@palpatim can you take a look?

Rusik: @royjit @palpatim  I see what you mean guys, however we already using amplify SDK on android and we have endpoint with `https://` scheme in our configuration json file and it works absolutely fine

I deep dived into android sdk and it is 100% supported urls with schemes, there is no checks there for that

* It starts [here](https://github.com/aws-amplify/amplify-android/blob/662ed22cabfe2751be94d08dcb30b8ad42ce716c/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/AWSCognitoAuthPlugin.java#L175) when we initialize `awsMobileClient` with json configuration
```java
awsMobileClient.initialize(
    context,
    new AWSConfiguration(mobileClientConfig), 
    ...
``` 

* Then [here](https://github.com/aws-amplify/aws-sdk-android/blob/848912f023d072bcdd2c79ed946b2b7319799bf1/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java#L592) we get custom endpoint as string without any checks
```java
final String cognitoUserPoolCustomEndpoint = userPoolJSON.optString(COGNITO_USERPOOL_CUSTOM_ENDPOINT);
```

* [Then](https://github.com/aws-amplify/aws-sdk-android/blob/848912f023d072bcdd2c79ed946b2b7319799bf1/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java#L605) create `userpool` with this string enpdoint
```java
userpool = new CognitoUserPool(mContext, mUserPoolPoolId, clientId, clientSecret, userpoolLL, pinpointEndpointId, cognitoUserPoolCustomEndpoint);
```

* Inside [CognitoUserPool.java](https://github.com/aws-amplify/aws-sdk-android/blob/848912f023d072bcdd2c79ed946b2b7319799bf1/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUserPool.java#L349) we set this enpoint to `client` 
```java
if(cognitoUserPoolCustomEndpoint != null && !cognitoUserPoolCustomEndpoint.isEmpty()) {
    this.client.setEndpoint(cognitoUserPoolCustomEndpoint);
}
```

* `client`  conforms to `AmazonCognitoIdentityProvider` interface where in the docs of `setEndpoint` we can see [this](https://github.com/aws-amplify/aws-sdk-android/blob/848912f023d072bcdd2c79ed946b2b7319799bf1/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/services/cognitoidentityprovider/AmazonCognitoIdentityProvider.java#L46)
```
* Callers can pass in just the endpoint (ex:
* "cognito-idp.us-east-1.amazonaws.com") or a full URL, including the
* protocol (ex: "https://cognito-idp.us-east-1.amazonaws.com"). If the
* protocol is not specified here, the default protocol from this client's
* {@link ClientConfiguration} will be used, which by default is HTTPS.
```

* Also `client` itself is `AmazonWebServiceClient` where we [can see the same](https://github.com/aws/aws-sdk-java/blob/4f416d58e07c15875cf73ac873d2fe20e89ad6df/aws-java-sdk-core/src/main/java/com/amazonaws/AmazonWebServiceClient.java#L293) in the docs for `setEndpoint`
```
* Callers can pass in just the endpoint (ex: "ec2.amazonaws.com") or a full
* URL, including the protocol (ex: "https://ec2.amazonaws.com"). If the
* protocol is not specified here, the default protocol from this client's
* {@link ClientConfiguration} will be used, which by default is HTTPS.
```

So we can see that android sdk supports endpoints with and without schemes and there are no additional checks around it. What kind of strategy are we going to have on iOS side?

royjit: Will inform @aws-amplify/android-team to make the necessary changes without breaking existing users. Since Android takes endpoint without the protocol, we could have amplify-ios the same behavior as aws-sdk-ios like I explained above. 
You could use the same configuration in android and iOS like shown below because android supports it already:
```
"CognitoUserPool": {
  "Default": {
    "AppClientId": "xxx",
    "Endpoint": "example.org",
    "PoolId": "xxxxx",
    "Region": "xx-xxx-1"
  }
}
```

Rusik: @royjit I left support only for url without schemes and paths, so `https://example.org` or `example.org/path` throw and exception.

I didn't use the code with `components.percentEncodedHost = endpointString` because it throws obj-с exception that can't be handled and doesn't contain enough information about what's going wrong for end user

Could you take a look please

royjit (approved):
LGTM

Rusik:
@palpatim could you also take a look please

palpatim (approved):
LGTM, thanks very much for filing this!

Rusik:
Nice, thanks guys!

Rusik:
@palpatim @royjit @atierian hey guys!

I tried to use custom endpoint on iOS but faced an issue. Seems it's on aws-sdk-ios side, not amplify. But as you are working on both sdks could you please take a look and help with it please

https://github.com/aws-amplify/aws-sdk-ios/issues/4153

atierian:
We'll take a look and respond on the new issue you opened. Thanks for linking it here.

AmplifyPlugins/Auth/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
Reonarudo: Currently if I need to add a path is there a way to modify the Endpoint before maiking a call?

PR #1724, opened 2022-04-05
chore(DataStore): Internal Delete API rename (predicate to condition and filter)

lawmicha:
*Issue #, if available:*

*Description of changes:*
Delete API has the following use cases
1. Delete model instance / delete model type by id (`DataStore.delete(model)` / `DataStore.delete(modelType:withId)`)
2. Delete model instance conditionally / delete model type by id and condition (`DataStore.delete(model:where)` / `DataStore.delete(modelType:withId:where)`)
3. Delete model type with filter (`DataStore.delete(modelType:where)`)

This internally renames `predicate` to `condition` for use case 2 and `filter` for use case 3 for better clarity on the type of `predicate` the parameter is. Also refactor some the caller code that calls `queryAndDeleteTransaction` 

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
@@ -238,115 +238,106 @@ final class StorageEngine: StorageEngineBehavior {
     func delete<M: Model>(_ modelType: M.Type,
                           modelSchema: ModelSchema,
                           withId id: Model.Identifier,
-                          predicate: QueryPredicate? = nil,
+                          condition: QueryPredicate? = nil,
                           completion: @escaping (DataStoreResult<M?>) -> Void) {
         var deleteInput = DeleteInput.withId(id: id)
-        if let predicate = predicate {
-            deleteInput = .withIdAndPredicate(id: id, predicate: predicate)
+        if let condition = condition {
+            deleteInput = .withIdAndCondition(id: id, condition: condition)
         }
         let transactionResult = queryAndDeleteTransaction(modelType,
                                                           modelSchema: modelSchema,
                                                           deleteInput: deleteInput)
-        let modelsFromTransactionResult = collapseMResult(transactionResult)
-        let associatedModelsFromTransactionResult = resolveAssociatedModels(transactionResult)
-
-        let deletedModel: M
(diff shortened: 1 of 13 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -238,115 +238,106 @@ final class StorageEngine: StorageEngineBehavior {
...
-        let modelsFromTransactionResult = collapseMResult(transactionResult)
-        let associatedModelsFromTransactionResult = resolveAssociatedModels(transactionResult)
-
-        guard modelSchema.isSyncable, let syncEngine = self.syncEngine else {
-            if !modelSchema.isSystem {
-                log.error("Unable to sync model (\(modelSchema.name)) where isSyncable is false")
+                                                          deleteInput: .withFilter(filter))
+        switch transactionResult {
+        case .success(let queryAndDeleteResult):
+            guard !queryAndDeleteResult.deletedModels.isEmpty else {
```
thisisabhash, line 304: Should this block be at line 318, before `guard #available...`? Compared to the previous code

lawmicha: the previous logic
1. if the model is not syncable, then return the results (will be either success of failure) [L319]
2. if the model is syncable, return if failure [L325]
3. if the model is syncable, and success, check if iOS13, otherwise return the success [L349]
4. if the model is syncable, and success, check if iOS13, then sync the models if there are any. [L340] and return the results [L330]
5. 4. if the model is syncable, and success, check if iOS13, no models to sync, return the success [L338]

The current logic
1. if the result is failure, return it [L340]
2. if the result is success, and there are no models to sync, return empty result [L305]
3. if the result is success, and is not syncable, return results [L316]
4. if the result is success, and is syncable, and is not iOS 13, return [L321]
6. if the result is success, and is syncable, and is iOS 13, then sync the models [L325]

the first difference between the previous and current logic is the current logic checks the transactionResult first before checking whether it should sync or not. Then it checks if it should sync or not based on the three conditions
- are there any models? [L304]
- is the model syncable? [L309]
- is it IOS 13? [L320]

the reason why i pulled this check up [previously L337] is because if we know there are no models to sync, there is no point in checking if sync is required

PR #1757, opened 2022-04-20
fix(DataStore): Nesting DataStore integration tests

lawmicha:
*Issue #, if available:*

*Description of changes:*
Improve the DataStore integration tests runnable across multiple test classes. 

- ~~Some API plugin updates were made in this PR to remove force cast of properties to optionals.~~
- DataStore model registration is done with a subset of the models for what's needed for the test. This will reduce the sync process for syncing unnecessarily model data.
- Split all the tests into different schemas, makes a re-run much easier since it's a shorter run time, and addressing failures more isolated to a test class.
- There are some race conditions w.r.t. Amplify.reset() that can't be solved without some changes to the API signature, but for now the changes in this PR is to try and work around that with `sleep(2)`. 

Latest run (At the time of pasting this link):
https://github.com/aws-amplify/amplify-ios/actions/runs/2220808219

image

- [ ] All unit tests pass


.github/workflows/integ_test.yml
```diff
@@ -1,7 +1,7 @@
 name: IntegrationTest
 on:
   push:
-    branches: [main]
```
lawmicha, line 4: this change should be reverted once this PR is ready to merge in

.github/workflows/integ_test.yml
```diff
@@ -271,6 +270,501 @@ jobs:
...
+          aws_region: ${{ secrets.AWS_REGION }}
+          aws_s3_bucket: ${{ secrets.AWS_S3_BUCKET_INTEG }}
+
+      - name: Run Integration test
+        uses: ./.github/composite_actions/run_xcodebuild_test
+        with:
+          project_path: ./AmplifyPlugins/DataStore/
+          workspace: DataStoreCategoryPlugin.xcworkspace
+          scheme: AWSDataStoreIntegrationV1Scenario1Tests
+  datastore-integration-V1-S2-test:
```
royjit, line 352: What does Vx Sx mean? Is there an easy way to understand what failed.

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+AuthInformation.swift
```diff
@@ -15,6 +15,9 @@ extension AWSAPIPlugin {
     }
 
     public func defaultAuthType(for apiName: String?) throws -> AWSAuthorizationType {
-        try pluginConfig.endpoints.getConfig(for: apiName).authorizationType
```
royjit, line 18: This is a public api and you are changing the error thrown which will be a breaking change.

AmplifyPlugins/DataStore/DataStoreCategoryPlugin.xcodeproj/xcshareddata/xcschemes/AWSDataStoreIntegrationLocalTests.xcscheme
```diff
@@ -0,0 +1,156 @@
...
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "2149E61C23886CEE00873955"
+               BuildableName = "AWSDataStoreCategoryPluginIntegrationTests.xctest"
+               BlueprintName = "AWSDataStoreCategoryPluginIntegrationTests"
+               ReferencedContainer = "container:DataStoreCategoryPlugin.xcodeproj">
+            </BuildableReference>
+            <SkippedTests>
```
royjit, line 24: Are these intentional? I see skipped tests in all xcsheme

.github/workflows/integ_test.yml
royjit: Can we group them under a common parent?

AmplifyPlugins/DataStore/DataStoreCategoryPlugin.xcodeproj/xcshareddata/xcschemes/AWSDataStoreIntegrationLocalTests.xcscheme
lawmicha: yup, most tests in AWSDataStoreCategoryPluginIntegrationTests are skipped, and split into individual schemas that enable one or more test classes from the original target

AmplifyPlugins/API/AWSAPICategoryPlugin/AWSAPIPlugin+AuthInformation.swift
lawmicha: I'm debating removing this change from this PR all together as it does have some behavior changes to APIPlugin when we removed the force cast to `pluginConfig` and various other properties in APIPlugin. A few details regarding this change, take `pluginConfig` for example, it was originally force cast and set when `Amplify.configure()` is called. If the developer does not call Amplify.configure and calls into one of the plugin methods like `defaultAuthType` it will crash the app. By turning it into optional, the app gracefully fails with the error indiciating that the plugin configuration not yet configured. This may happen in two cases, at development time when the developer does not call Amplify.configure or during testing when Amplify.reset is called to reset the plugins. At development time, the behavior changes from app crashing to an error thrown- which in this case is fairly similar DX, a non-recoverable error which is needs to be fixed. During testing, Amplify.reset() is called between test cases (like at the beginning of setUp or in tearDown) and i've observed that some operations are in-progress will attempt to use pluginConfig, and crash the host app. Since this PR is fairly large already, it will be easier to reason about these changes in separate PRs. This PR to only contain more granular testing done, and another PR for changes made to APIPlugin

lawmicha: Removed the APIPlugin changes from this PR

.github/workflows/integ_test.yml
lawmicha: Vx is the version of the transformer and Sx is the scenario, this is just following the test class that it's enabling as close as possible, so changing the test class name should drive the change for these workflow names. One example is we could be more verbose if we want to to something like `TransformerV1ScenarioOneDirectionalHasOneTests`. (perhaps drop `AWSDataStoreIntegration` completely, and be explicit about the Transformer version and verbose about the scenario as closely as possible). I updated the nesting better as they all have one common parent which is the original `AWSDataStoreCategoryPluginIntegrationTests` that contains E2E tests

AmplifyPlugins/DataStore/DataStoreCategoryPlugin.xcodeproj/xcshareddata/xcschemes/AWSDataStoreIntegrationLocalTests.xcscheme
royjit: nit: if they are used in other scheme, should we then delete it from the scheme where it is skipped?

lawmicha: went over to try this, and don't think we can in the current structure
image

Each scheme is under a particular target, which inherits all of the test classes of that target. Having multiple targets is another option, and then 1-1 schema to target would clean it up. I believe this should be what we should aim for, but each scheme only contains a few tests, it's much more maintainable at the moment to keep them all under a single target. Once we improve on each schema's test and add more coverage, then we can migrate them to its own separate target (like we do with the other two targets MultiAuth and Flutter)

Issue #1734, opened 2022-04-08
Improve performance in DataStore where DataFormatter is costly to initialize


brennanMKE:
### Describe the bug

Optimize the use of DateFormatter which is initializing instances repeatedly.

https://github.com/aws-amplify/amplify-ios/blob/02ac2ba5c417883ea90f3d59b3cd28e63d2a1d39/Amplify/Categories/DataStore/Model/Temporal/Temporal.swift#L88

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Expected behavior

It should cache DateFormatter instances.

### Amplify Framework Version

1.22.4

### Amplify Categories

DataStore

### Dependency manager

Swift PM

### Swift version

5.6

### CLI version

8.0.0

### Xcode version

13.3

### Is this a regression?

No

### Device

iPhone 13

### iOS Version

iOS 15

Issue #1583, opened 2022-01-04
DataStore Temporal cannot persist precision of micro seconds and nano seconds


HuiSF:
### Describe the bug

Using schema

```graphql
type Blog @model {
  name: String!
  awsDateTime: AWSDateTime
}
```

Run below sample code

```swift
let models = [
    try Blog(name: "1", awsDateTime: Temporal.DateTime.init(iso8601String: "2020-01-01T19:30:45.000000000Z")),
    try Blog(name: "2", awsDateTime: Temporal.DateTime.init(iso8601String: "2020-01-01T19:30:45.100000000Z")),
    try Blog(name: "3", awsDateTime: Temporal.DateTime.init(iso8601String: "2020-01-01T19:30:45.100250000Z")),
    try Blog(name: "4", awsDateTime: Temporal.DateTime.init(iso8601String: "2020-01-01T19:30:45.1000Z")),
    try Blog(name: "5", awsDateTime: Temporal.DateTime.init(iso8601String: "2020-01-01T19:30:45.111Z")),
    try Blog(name: "6", awsDateTime: Temporal.DateTime.init(iso8601String: "2020-01-01T19:30:45.111+00:00")),
    try Blog(name: "7", awsDateTime: Temporal.DateTime.init(iso8601String: "2020-01-01T19:30:45.111+01:00")),
    try Blog(name: "7", awsDateTime: Temporal.DateTime.init(iso8601String: "2020-01-01T19:30:45.111222333Z")),
]

for model in models {
    Amplify.DataStore.save(model) { result in
        switch(result) {
        case .success(let savedItem):
            print("Saved item: \(savedItem.name)")
        case .failure(let error):
            print("Could not save item to DataStore: \(error)")
        }
    }
}
```

The `awsDateTime` values stored in local SQLite DB drops the micro seconds and nano seconds specified in the input.
<img width="165" alt="image-20220104105222805" src="https://user-images.githubusercontent.com/10602282/148110508-e4b290c8-dc5e-4a8e-b829-5235ae46f250.png">

The fractional seconds dropping seems to be done by this format string:

https://github.com/aws-amplify/amplify-ios/blob/2c57ccce43d60789183f3737e44226a453eae969/Amplify/Categories/DataStore/Model/Temporal/TemporalFormat%2BDateTime.swift#L21

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Run the sample code
2. Observe the records inserted in SQLite DB
```

[...]

Issue #1579, opened 2021-12-28
Question regarding decoding depth


anatomybook:
Hello everyone,

as I've said [before](https://github.com/aws-amplify/amplify-ios/issues/1428), I'm currently struggling with the performance of DataStore. After some profiling, I determined that the reason for the sync being horribly slow is decoding. The syncing bottleneck is not the bandwidth (far from it), but the CPU time decoding the result.

For me, this is  significant, as I've many n:m relationships. So given that a certain item can belong to many others, a given item is decoded many, many times, burning up the CPU and resulting in unacceptable syncing times.

Now, I've run some tests and figured that I'm _much_ faster establishing the relations later, i.e. downloading all the items from all tables standalone and then in a separate step hookup the relations by ID - which is typically done on the server side. I know this is against the idea of DataStore, but at least it'll decode an item exactly once, drastically reducing the decoding inferno.

So this raises the question: I'd rather avoid removing the dependencies in the GraphQL data model, but set the DataStore decoding depth to zero, so it'll not attempt the decode parents or children. Is there a way to do that?

Any help is greatly appreciated! Thanks!

brennanMKE:
@anatomybook We will be making an update which could improve performance for syncing. See below.

https://github.com/aws-amplify/amplify-ios/issues/1428
https://github.com/aws-amplify/amplify-ios/issues/1734

anatomybook:
Thanks for the update. I look forward!

anatomybook:
I've now resorted to setting the decoding depth to "zero", i.e. decouple the tables in the GraphQL scheme. That's painful to do, as it's against the idea and basically shifts responsibility from the library to the app, which hurts to do.

However, by doing that, the bottleneck moved from processing to drive throughput, greatly accelerating the sync process. I now tap into the event `HubPayload.EventName.DataStore.syncReceived` where I assemble the table contents to the finished models. Any orphaned table entry is deleted, thereby eliminating any sync not completing and / or errors showing up, which I've still seen in the recent past.

Anyhow, I'll pay close attention and monitor performance. I can't wait to revert the changes I made. But for now, it works much better for me.

Issue #1428, opened 2021-09-18
Very bad performance due to date decoding


anatomybook:
### Describe the bug

Technically speaking this is not really a bug, but merely an observation. Nevertheless, it's very annoying as in the presence of many items, operations take forever.

In my case initial sync takes about a minute - which is user waiting for that long for everything to be ready?

For example consider this screenshot I obtained by just randomly pausing initial sync. Two threads busy doing date operations?! My point is that taking screenshots like this isn't hard, because it's almost always busy decoding dates.

![Screenshot 2021-09-18 at 14 23 33](https://user-images.githubusercontent.com/67764399/133889097-bf29d06f-6d9c-4ff4-a9c5-90e62740135a.png)

Another screenshot, that I obtained when going through a list of items. Again, it's almost always busy decoding the date. Given that the decoding happens in the inner loop, I wonder whether something can be done to accelerate the process?

![Screenshot 2021-09-18 at 14 46 14](https://user-images.githubusercontent.com/67764399/133889279-4b9a5d06-5224-48b3-8987-69978dcc6540.png)

### Steps To Reproduce

Load up a database with many items and observe where the time is spent during initial sync or general operation, such as going through lists.

Another good one is sorting. Make a simple model containing a string and AWSDateTime. Make several hundred objects and sort them. Observe that sorting according to the string is much faster than using AWSDateTime.

### Expected behavior

I expect the date decoding not to slow down my app significantly.

### Amplify Framework Version

6dca41c8620c27d9da2749187957808a19cdafb3

### Amplify Categories

DataStore

### Dependency manager

Cocoapods

### Swift version

Apple Swift version 5.4.2 (swiftlang-1205.0.28.2 clang-1205.0.19.57)

### CLI version

5.3.0

### Xcode version

12.5.1

### Is this a regression? (i.e. was this working before a version upgrade)

Performance was always bad, as far as I recall.

### Device

iPhone 12 Pro Max, but same everywhere

### iOS Version

iOS 14.8

### Specific to simulators

no

lawmicha:
Thanks for bringing this to our attention @anatomybook , we'll look into testing the performance of decoding DateTime and seeing if we can optimized this. In JS, the datetime's are stored as strings in the local store and are not decoded until the caller queries for them. We'll look into taking a similar approach after getting a baseline of the performance. Every model would be affected by this since it's each model contains `createdAt` and `updatedAt` system level fields.

brennanMKE:
[...]

PR #1760, opened 2022-04-22
feat(DataStore): Temporal performance enhancements

atierian:
*Issue #, if available:*
**Addresses:**
 #1734 

**Partially addresses:**
#1579 
#1428

*Description of changes:*
Performance Improvements

Re-write of `Temporal` implementation details relating to `String <-> Date` conversion to cache `DateFormatter`s. The public API surface remains mostly the same (with a few minor exceptions documented below).

**Current situation:**
Every `String` -> `Date` conversion currently creates up to 4 separate `DateFormatter`s. This is due to this API supporting the conversion of `String` -> `Date` when the exact format isn't known, but still limited to a maximum of 4 different formats per `TemporalSpec` conforming type (`Temporal.Date`, `Temporal.DateTime`, `Temporal.Time`). 
`DateFormatter` instantiation isn't a particularly cheap process. 

**With this change:**
These changes cache `DateFormatter`s for reuse in a thread safe way, up to a maximum of 14, using a Random Replacement (RR) eviction policy. It also adds lower level utility converters (`Date` -> `String` and `String` -> `Date`) leveraged by the new implementation.

**API Changes**
- `Temporal` was previously a namespace `struct` with a private init. It's now a namespace (caseless) `enum`. This change has no potential to break client code.
- ~The `TemporalSpec` protocol requirement `init(iso8601String: String) throws` is now `init(iso8601String: String, format: Format) throws`. Every concrete type conforming to `TemporalSpec` (`Temporal.Date`, `Temporal.DateTime`, `Temporal.Time`) has a default argument for `format` to prevent any possibility of breakage. This is an improvement in the API, because while there are cases where we don't know the exact `dateFormat`, there are other cases where we do know it. This eliminates the need to attempt to convert based on various different formats if the format is known.~ Added a separate `init(iso8601String:format:)` with a default implementation, so no breaking. 
- `TemporalFormat` in the previous implementation was an `enum`. This has been changed to a `struct` with static members. **This change does have the possibility to break client code**, although the likelihood is exceedingly low. The only situation this could happen is if client code is switching over all cases exhaustively; the change to a `struct` means that this check would no longer be exhaustive. Switching over the formats as they exist today would serve no purpose. Even if this is happening somewhere, the fix is adding a `default` case to the end of the `switch`. The main reason behind this change is to support the future expansion of further date formats as documented in #1583 and #1713, without breaking client code.
- Deprecates `TemporalUnit`, a public protocol that was not used anywhere within `Amplify`.
- ~Adds `AnyTemporalSpec`, a type-erasing existential that provides access to `iso8601String` and is needed due to the introduction of an `associatedtype` in the `TemporalSpec` protocol.~ Change reverted
## **Performance Wins:**
Micro and Macro performance testing was conducted to accurately test and improvements or regressions these changes introduce. 

Micro testing consisted of calling `Temporal.Date(iso8601String:)`, `Temporal.DateTime(iso8601String:)`, and `Temporal.Time(iso8601String:)` 50 times for each of the applicable `iso8601String` supported formats.

Macro testing consisted of decoding an array of 100 `Model` conforming objects, each with 50 child models. The parent contained (among other primitives) two `Temporal.x` properties, and each child contained one.

### **Results:**
*All times displayed in seconds.

**Micro**

| `Temporal.Date` | short <br />(yyyy-MM-dd) | medium / long / full <br />(yyyy.MM.ddZZZZ) |
| -----: | :-----: | :-----: |
| Old | 0.027 | 0.007 |
| New | 0.005 | 0.003 |
| Δ Absolute | 0.022 | 0.004 |
| Δ % | 440% | 133% |

| `Temporal.DateTime` | short <br />(yyyy-MM-dd'T'HH:mm) | medium <br />(yyyy-MM-dd'T'HH:mm:ss) | long <br />(yyyy-MM-dd'T'HH:mm:ssZZZZZ) | full <br />(yyyy-MM-dd'T'HH:mm:ssSSSZZZZZ) |
| -----: | :-----: | :-----: | :-----: | :-----: |
| Old | 0.027 | 0.021 | 0.013 | 0.007 |
| New | 0.009 | 0.007 | 0.004 | 0.002 |
| Δ Absolute | 0.018 | 0.014 | 0.009 | 0.005 |
| Δ % | 200% | 200% | 225% | 250% |

| `Temporal.Time` | short <br />(HH:mm) | medium <br />(HH:mm:ss) | long <br />(HH:mm:ss.SSS) | full <br />(HH:mm:ss.SSSZZZZZ) |
| -----: | :-----: | :-----: | :-----: | :-----: |
| Old | 0.025 | 0.020 | 0.015 | 0.010 |
| New | 0.007 | 0.006 | 0.005 | 0.004 |
| Δ Absolute | 0.018 | 0.014 | 0.010 | 0.006 |
| Δ % | 257% | 233% | 200% | 150% |

**Macro**
| | |
|  -----: | :-----: |
| Old | 1.517 |
| New | 0.553 |
| Δ Absolute | 0.964 | 
| Δ % | 174% | 

- [x] All unit tests pass


Amplify/Categories/DataStore/Model/Temporal/TemporalFormat.swift
```diff
@@ -7,37 +7,23 @@
 
 import Foundation
 
-/// `Enum` that represent the different options for formatting ISO8601 values.
-/// Each implementation of a `TemporalSpec` is responsible for converting the enum value
-/// to a corresponding format string.
-public enum TemporalFormat: CaseIterable {
+public protocol TemporalSpecValidFormatRepresentable: Equatable {
```
royjit, line 10: name change in public protocol is a breaking change right? Even though developers might not be using this, it goes against the semver versioning.

Amplify/Categories/DataStore/Model/Temporal/TemporalSpec.swift
```diff
@@ -0,0 +1,9 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
```
royjit, line 7: Empty file?

Amplify/Categories/DataStore/Model/Temporal/Time.swift
```diff
@@ -8,46 +8,64 @@
...
-    public struct Time: TemporalSpec, TimeUnitOperable {
+    ///  * `.short` => `HH:mm`
+    ///  * `.medium` => `HH:mm:ss`
+    ///  * `.long` => `HH:mm:ss.SSS`
+    ///  * `.full` => `HH:mm:ss.SSSZZZZZ`
+    public struct Time: TemporalSpec {
+        // Inherits documentation from `TemporalSpec`
+        public let foundationDate: Foundation.Date
 
-        public static var iso8601DateComponents: Set<Calendar.Component> {
```
royjit, line 19: Breaking change? I see that it is not used, should we deprecate this and make an internal one for  the init method below?

Amplify/Categories/DataStore/Model/Temporal/Temporal.swift
```diff
@@ -7,23 +7,17 @@
...
-    /// - seealso: Temporal.TimeStamp
-    private init() {}
-}
-
-/// The `TemporalSpec` protocol defines an [ISO-8601](https://www.iso.org/iso-8601-date-and-time-format.html)
-/// formatted Date value. Types that conform to this protocol are responsible for providing
-/// the parsing and formatting logic with the correct granularity.
 public protocol TemporalSpec {
-
+    associatedtype Format: TemporalSpecValidFormatRepresentable
```
royjit, line 20: For some reason, I do not see the implementation of this protocol `TemporalSpecValidFormatRepresentable` in this PR. Will post the comments I have so far, will review it after we figure out if we are missing some files in the PR

Amplify/Categories/DataStore/Model/Temporal/Temporal.swift
```diff
@@ -7,23 +7,17 @@
...
-    /// not be directly instantiated.
-    ///
-    /// - seealso: Temporal.Date
-    /// - seealso: Temporal.DateTime
-    /// - seealso: Temporal.Time
-    /// - seealso: Temporal.TimeStamp
-    private init() {}
-}
-
-/// The `TemporalSpec` protocol defines an [ISO-8601](https://www.iso.org/iso-8601-date-and-time-format.html)
```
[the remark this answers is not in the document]
atierian: It wasn't removed, it was moved to the type level so it can be viewed in quick help. The documentation here on the `private init` wasn't visible anywhere. Also the `- seealso: Temporal.TimeStamp` was removed because `Temporal.TimeStamp` doesn't exist.

Amplify/Categories/DataStore/Model/Temporal/Temporal.swift
atierian: It's in `TemporalFormat.swift`

Amplify/Categories/DataStore/Model/Temporal/TemporalFormat.swift
atierian: This is a new protocol, so it's not a name change. This is the protocol that defines structure of formats. 

Your point is valid one though. `TemporalFormat` was a public enum and doesn't exist anymore after these changes. However, we don't provide any public APIs that make use of the `TemporalFormat` enum. It only provided any value in combination with a `TemporalSpec.Type` through the **internal** `getFormat(for:) -> String` API. 

Removing can't break behavior because it has no public behavior. I don't believe this constitutes a semver breaking change. If you still have concerns despite this information, we can leave the type and deprecate it.

Amplify/Categories/DataStore/Model/Temporal/TemporalSpec.swift
atierian: This was moved into `Temporal.swift` to help with the diff. Deleting it.

Amplify/Categories/DataStore/Model/Temporal/Time.swift
atierian: Yes, good catch.

atierian: Addressed in: [cdace94](https://github.com/aws-amplify/amplify-ios/pull/1760/commits/cdace94592eca4ee9c74d5ac7b52811187b179c1)

Amplify/Categories/DataStore/Model/Temporal/TemporalSpec.swift
atierian: Addressed in [cdace94](https://github.com/aws-amplify/amplify-ios/pull/1760/commits/cdace94592eca4ee9c74d5ac7b52811187b179c1)

Amplify.xcodeproj/project.pbxproj
```diff
@@ -349,6 +349,36 @@
 		76C0F4EC26797F4500F0E6AB /* AuthModeStrategyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 76C0F4EA26797F3C00F0E6AB /* AuthModeStrategyTests.swift */; };
 		76FFFC91267966D9001EC92B /* RetryableGraphQLOperationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 76FFFC90267966D9001EC92B /* RetryableGraphQLOperationTests.swift */; };
 		8484A17AAAFE78545A932849 /* Pods_Amplify.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 7A1769ED39B45A981FD1D8AC /* Pods_Amplify.framework */; };
+		902AE05B281304B800CD12CA /* SpecBasedDateConverting.swift in Sources */ = {isa = PBXBuildFile; fileRef = 902AE04C281304B800CD12CA /* SpecBasedDateConverting.swift */; };
```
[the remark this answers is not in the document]
atierian: Addressed in [cdace94](https://github.com/aws-amplify/amplify-ios/pull/1760/commits/cdace94592eca4ee9c74d5ac7b52811187b179c1)

Amplify/Categories/DataStore/Model/Temporal/Temporal.swift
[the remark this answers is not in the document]
atierian: Ah, I see what you mean now. Re-added the `TemporalSpec` documentation in [22f0216](https://github.com/aws-amplify/amplify-ios/pull/1760/commits/22f02161a91baab1365c3a61058f0f9e1ab53078)

Amplify/Categories/DataStore/Model/Temporal/Temporal+Cache.swift
```diff
@@ -0,0 +1,172 @@
...
+            // Since we just confirmed the cache isn't empty, and we're guaranteeing exclusive access
+            // through the lock, we can safely force unwrap here.
+            let keyToEvict = formatterCache.randomElement()!.key
+            formatterCache[keyToEvict] = nil
+            // This can likely be improved at a later time by
+            // using a LFU, or potentially LRU, eviction policy.
+        }
+
+        // unlock if no early return
+        os_unfair_lock_unlock(&lock)
```
royjit, line 86: This is to let other waiting calls to acquire the lock and move forward? And below code is expensive.

Amplify/Categories/DataStore/Model/Temporal/TemporalFormat.swift
royjit: I am not sure whether it is being used by developers, I see that we had to update the test below to make it work. 

Also I see here https://github.com/aws-amplify/amplify-ios/issues/1048 that there are developers who might be using the format to get a specific date format? 

@lawmicha might have more context here.

atierian: The changed test uses the format with an internal API.

That linked issue discusses using `TemporalFormat` in a fork of Amplify, not as part of the Amplify public API surface.

Amplify/Categories/DataStore/Model/Temporal/Temporal+Cache.swift
atierian: That's correct. It would also be fine to keep this locked while instantiating the new `DateFormatter`. Happy to make that change if there's a good reason to.

Amplify/Categories/DataStore/Model/Temporal/Temporal+Cache.swift
```diff
@@ -0,0 +1,172 @@
...
+        in timeZone: TimeZone = .utc
+    ) throws -> Foundation.Date {
+        for format in formats {
+            let formatter = formatter(for: format, in: timeZone)
+            if let date = formatter.date(from: string) {
+                return date
+            }
+        }
+        throw DataStoreError
+            .invalidDateFormat(formats.joined(separator: " | "))
```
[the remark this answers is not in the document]
atierian: The variadic arg overload should actually just be removed at this point. It was added as a public convenience option. But since we've made these internal, there's no need for it anymore.

Amplify/Categories/DataStore/Model/Temporal/TemporalFormat.swift
atierian: Actually what I said a few comments up is incorrect. There is a public API that takes a `TemporalFormat` as part of the `TemporalSpec` protocol 😕.
This would only matter if the format was being created before being used as a function argument.

```swift
// ✅ this would be fine 
_ = Temporal.DateTime.iso8601FormattedString(format: .full)

// ❌ this would break
let format = TemporalFormat.full.dateTime
_  = Temporal.DateTime.iso8601FormattedString(format: format) 
```

I'm going to take a look restructuring format to not break in this case. Might be able to even drop the new associatedtype requirement in `TemporalSpec` as well 🤞

atierian:
@lawmicha can you provide any insights into the failing DataStore tests? I can't seem to reproduce them failing locally and from what I can tell the tests have nothing to do with these temporal changes.

atierian:
Despite the tests passing, there's an underlying issue here. In local tests, running `testUpdatedModelNoLongerMatchesPredicateRemovedFromSnapshot` repeatedly 100 times results in failures. `main` passes, as does the commit before these changes. So this requires some more investigation before approving / merging.

royjit (dismissed):
LGTM

Amplify/Categories/DataStore/Model/Temporal/TemporalFormat.swift
```diff
@@ -7,37 +7,92 @@
...
-public enum TemporalFormat: CaseIterable {
+public struct TemporalFormat {
+    let dateFormat: String
+    let dateTimeFormat: String
+    let timeFormat: String
 
-    case short
-    case medium
-    case long
-    case full
```
lawmicha, line 18: does changing the enum case to the static properties impact customers? are we thinking along the lines that TemporalFormat wasn't meant to be used publicly?

lawmicha (commented):
Since we are changing public facing properties by replacing or adding new ones, thoughts on the versioning? can we mark this as a `feat(DataStore):` in the PR title to indicate a minor version bump?

lawmicha:
> @lawmicha can you provide any insights into the failing DataStore tests? I can't seem to reproduce them failing locally and from what I can tell the tests have nothing to do with these temporal changes.

I added a commit onto this PR https://github.com/aws-amplify/amplify-ios/pull/1760/commits/c384882a1ff99050c6f5ea756faf0ec838a03d26 for fixing the test that failed on "run repeatedly". The race condition was adding the operation to the operation queue while publishing events to the publisher. Not really sure why it was not appearing before the changes in this PR, as the only code path that changed was the initialization of `createdAt` field in MutationEvent using `.now()`. The test started failing when the events were published first, then a thread picked up the operation from the operation queue. Since the observe query operation started running after the events that were published from the test logic, it would not receive them.

lawmicha:
Locally I'm going to run a few API integration tests to exercise the decoding of strings to Temporal types, and also enabled the integration test workflow in this separate PR to get more visibility into the test results: https://github.com/aws-amplify/amplify-ios/pull/1827

lawmicha:
Will investigate the flaky test https://app.circleci.com/pipelines/github/aws-amplify/amplify-ios/4953/workflows/23e03ca6-e27c-4a90-9f43-7b5aa9c13d93/jobs/42782 `testSentModelWithNilVersion_Reconciled`

Amplify/Categories/DataStore/Model/Temporal/TemporalFormat.swift
atierian: Yes, but only in one specific case that is very improbable because there would be no reason for a customer to do so. It would only break if they're exhaustively switching over `TemporalFormat` without a `default` case. 
```swift
switch temporalFormat {
case .short: ...
case .medium: ...
case .long: ...
case .full: ...
}
```

If that occurred, they would have to add a `default` case to satisfy exhaustiveness.

However, we will almost certainly have to add new cases in the future. This protects us against those changes being breaking changes. The description in the original PR comment above discusses the rationale behind this change.

> TemporalFormat in the previous implementation was an enum. This has been changed to a struct with static members. This change does have the possibility to break client code, although the likelihood is exceedingly low. The only situation this could happen is if client code is switching over all cases exhaustively; the change to a struct means that this check would no longer be exhaustive. Switching over the formats as they exist today would serve no purpose. Even if this is happening somewhere, the fix is adding a default case to the end of the switch. The main reason behind this change is to support the future expansion of further date formats as documented in https://github.com/aws-amplify/amplify-ios/issues/1583 and https://github.com/aws-amplify/amplify-ios/discussions/1713, without breaking client code.

lawmicha: thanks for explaining, extendability sounds like it will help a lot with platform parity across Android/Flutter/JS since right now most platforms i believe do not change the user's input, while iOS normalizes to these 4 formats. A potential solution for platform parity is to find the lowest common denominator and if iOS can increase that denominator then we'll have a wider range of coverage

Amplify/Categories/DataStore/Model/Temporal/Temporal+Cache.swift
atierian: Great catch and callout, thanks! That's almost certainly what's causing the intermittent DataStore test failures. Just pushed [b2062c2](https://github.com/aws-amplify/amplify-ios/pull/1760/commits/b2062c2cc8709fe839982b5209df9d9a2d5bea78) after tests passed locally with the change.

lawmicha (approved):

PR #1770, opened 2022-04-28
bugfix(Auth): Fixing the token expiring logic

harsh62:
*Issue #, if available:*

*Description of changes:*
Fixing token expiring logic to use JWT tokens instead of the API response expiration date. 

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSAuthCognitoSession.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSAuthCognitoSession.swift
@@ -7,6 +7,7 @@
 
 import Amplify
 import AWSPluginsCore
+import Foundation
 
 public struct AWSAuthCognitoSession: AuthSession,
                                      AuthAWSCredentialsProvider,
@@ -75,6 +76,25 @@ public struct AWSAuthCognitoSession: AuthSession,
         }
     }
 
+}
+
+/// Internal Helpers for managing session tokens
+internal extension AWSAuthCognitoSession {
+    func areTokensExpiring(in seconds: TimeInterval? = nil) -> Bool {
+        
+        guard let tokens = try? cognitoTokensResult.get(),
+              let idTokenClaims = try? AWSAuthService().getTokenClaims(tokenString: tokens.idToken).get(),
+              let accessTokenClaims = try? AWSAuthService().getTokenClaims(tokenString: tokens.idToken).get(),
+              let idTokenExpiration = idTokenClaims["exp"]?.doubleValue,
+              let accessTokenExpiration = accessTokenClaims["exp"]?.doubleValue else {
+            return true
+        }
+        
+        // let refreshInterval = AuthPluginConstants.sessionRefreshInterval
+        // If the session expires < X minutes return it
+        return (Date(timeIntervalSince1970: idTokenExpiration).compare(Date(timeIntervalSinceNow: seconds ?? 0)) == .orderedDescending &&
+                Date(timeIntervalSince1970: accessTokenExpiration).compare(Date(timeIntervalSinceNow: seconds ?? 0)) == .orderedDescending)
+    }
 }
 
 /// Helper method to update specific results if needed
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSAuthCognitoSession.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSAuthCognitoSession.swift
@@ -78,24 +78,6 @@ public struct AWSAuthCognitoSession: AuthSession,
 
 }
 
-/// Internal Helpers for managing session tokens
-internal extension AWSAuthCognitoSession {
-    func areTokensExpiring(in seconds: TimeInterval? = nil) -> Bool {
-        
-        guard let tokens = try? cognitoTokensResult.get(),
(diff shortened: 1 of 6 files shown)


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSAuthCognitoSession.swift
```diff
@@ -77,6 +78,24 @@ public struct AWSAuthCognitoSession: AuthSession,
 
 }
 
+/// Internal Helpers for managing session tokens
+internal extension AWSAuthCognitoSession {
+    func areTokensExpiring(in seconds: TimeInterval? = nil) -> Bool {
+        
+        guard let tokens = try? cognitoTokensResult.get(),
```
royjit, line 85: Can we add these under "cognitoUserPoolTokens" struct? So that we do not need to do the condition like below

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSAuthCognitoSession.swift
```diff
@@ -77,6 +78,24 @@ public struct AWSAuthCognitoSession: AuthSession,
...
+              let idTokenClaims = try? AWSAuthService().getTokenClaims(tokenString: tokens.idToken).get(),
+              let accessTokenClaims = try? AWSAuthService().getTokenClaims(tokenString: tokens.idToken).get(),
+              let idTokenExpiration = idTokenClaims["exp"]?.doubleValue,
+              let accessTokenExpiration = accessTokenClaims["exp"]?.doubleValue else {
+            return true
+        }
+        
+        // If the session expires < X minutes return it
+        return (Date(timeIntervalSince1970: idTokenExpiration).compare(Date(timeIntervalSinceNow: seconds ?? 0)) == .orderedDescending &&
```
royjit, line 94: We used to compare against current time + 2 minute or something to expire early to avoid returning a session that will expire in next minute. Should we add that logic here?

harsh62: Its happening in the action ATM, so I kept this generic if we want to know something different in the future.. And the other thing that came to my mind was that action should drive that logic..

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSAuthCognitoSession.swift
harsh62: we can do it, in that case the method would need to be public, because the tokens model exists in the `AmplifyPluginsCore` .. I wasn't sure if we really needed that logic to be public.. Thats the reason I had it here..

royjit: we could add an internal extension for the struct in the plugin right?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AuthCognitoTokens+Validation.swift
```diff
@@ -0,0 +1,28 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import AWSPluginsCore
+
+/// Internal Helpers for managing session tokens
+extension AuthCognitoTokens {
+    
+    func areTokensExpiring(in seconds: TimeInterval? = nil) -> Bool {
```
royjit, line 14: Do we need this value to be optional? Also add a comment for the function, mention that we consider the token to be expired if we cannot retrieve a valid expiration value from the tokens. 

```suggestion
    func areTokensExpiring(in seconds: TimeInterval = 0) -> Bool {
```

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AuthCognitoTokens+Validation.swift
```diff
@@ -0,0 +1,28 @@
...
+        guard let idTokenClaims = try? AWSAuthService().getTokenClaims(tokenString: idToken).get(),
+              let accessTokenClaims = try? AWSAuthService().getTokenClaims(tokenString: accessToken).get(),
+              let idTokenExpiration = idTokenClaims["exp"]?.doubleValue,
+              let accessTokenExpiration = accessTokenClaims["exp"]?.doubleValue else {
+            return true
+        }
+        
+        // If the session expires < X minutes return it
+        return (Date(timeIntervalSince1970: idTokenExpiration).compare(Date(timeIntervalSinceNow: seconds ?? 0)) == .orderedDescending &&
+                Date(timeIntervalSince1970: accessTokenExpiration).compare(Date(timeIntervalSinceNow: seconds ?? 0)) == .orderedDescending)
```
royjit, line 25:
```suggestion
        let idTokenExpiry = Date(timeIntervalSince1970: idTokenExpiration)
        let accessTokenExpiry = Date(timeIntervalSince1970: accessTokenExpiration)

        // If the session expires < X minutes return it
        let expiryTime = Date(timeIntervalSinceNow: seconds)
        return (idTokenExpiry.compare(expiryTime) == .orderedDescending &&
                accessTokenExpiry.compare(expiryTime) == .orderedDescending)
```

royjit (approved):
LGTM, added some nit

PR #1773, opened 2022-05-03
fix(DataStore): retry initial sync network failures from RemoteSyncEngine

lawmicha:
### Description:

This PR fixes **RemoteSyncEngine** to accurately determine whether it should retry based on the URLError. Previously, when errors were handled, they were ignored and a hardcoded `URLError.notConnected` is passed to the `retryRequestAdvice` method to get the advice to retry. This is a bug when the error is a non-retryable error, since the RemoteSyncEngine's retry attempt will immediately fall into the same errored state. For example, data decoding issues which should be non-retryable but the bug caused the sync engine to restart a number of times up to the limit determine by the algorithm in **RequestRetryablePolicy**.

### Changes made
- RemoteSyncEngine's error handling path to stop using the hardcoded `URLError` and start checking against a real URLError, retrieved from the underlying error. All components, specifically the sync and subscription components, are responsible for populating the underlying error that it receives from `Amplify.API`. 
- AWSInitialSyncOrchestrator: If any of the syncQuery requests fail with a network error, pick the first one to be the underlying error. This is returned as a `DataStoreError.sync` error with the URLError as the underlying error.

### Testing Initial Sync - AWSInitialSyncOrchestrator
An end-to-end flow is when the component propagates the URLError back up to RemoteSyncEngine and RemoteSyncEngine accurately determines that it should retry. 
1. network on, DataStore.start(), then breakpoint into InitialSyncOperation's `query(lastSyncTime:, nextToken:)`
2. network off, See that API returns a failure and AWSInitialSyncOrchestrator pulls it out and into the DataStoreError.sync error.

### Testing data decoding issue
1. A schema with a model containing enum cases like
```
type Todo @model {
  id: ID!
  name: String!
  description: String
  status: Status!
}

enum Status {
  notStarted
  inProgress
  completed
  blocked
}
```
2. save some data, and then update the Status enum to remove one of the cases

```swift
public enum Status: String, EnumPersistable {
  case notStarted
  case inProgress
  case completed
  //case blocked
}
```
3. DataStore.start() should sync the data and receive the error
```
dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "items", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "status", intValue: nil)], debugDescription: "Cannot initialize Status from invalid String value blocked", underlyingError: nil))
```
4. Observe that no retries are made.

### Subscriptions

Amplify.API.subscription API will use the AppSyncRealTimeClient library for subscriptions. When there is a network status changes, it will internally disconnect and reconnect with exponential back-ff. From Amplify.API's perspective, it doesn't receive an error until all retry attempts have been exhausted, and the error is an `.operationError`, not a URLError. See https://github.com/aws-amplify/aws-appsync-realtime-client-ios/pull/58 for more details

### Sync to Cloud Mutations
OutgoingMutationQueue will create a SyncMutationtoCloudOperation for the mutation event, which contains retry logic for network failures. See https://github.com/aws-amplify/amplify-ios/pull/1722 for more details

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
@@ -159,8 +159,8 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
         }
 
         var underlyingError: Error?
-        if syncErrors.contains(where: isNetworkError) {
-            underlyingError = getFirstUnderlyingNetworkError(errors: syncErrors)
+        if let error = syncErrors.first(where: isNetworkError(_:)) {
+            underlyingError = getFirstUnderlyingNetworkError(error)
         }
 
         let allMessages = syncErrors.map { String(describing: $0) }
@@ -242,22 +242,19 @@ extension AWSInitialSyncOrchestrator {
         return false
     }
 
-    private func getFirstUnderlyingNetworkError(errors: [DataStoreError]) -> Error? {
-        for error in errors {
-            guard case let .sync(_, _, underlyingError) = error,
-                  let datastoreError = underlyingError as? DataStoreError
-                  else {
-                continue
-            }
+    private func getFirstUnderlyingNetworkError(_ error: DataStoreError) -> Error? {
+        guard case let .sync(_, _, underlyingError) = error,
+              let datastoreError = underlyingError as? DataStoreError
+              else {
+            return nil
+        }
 
-            if case let .api(amplifyError, _) = datastoreError,
-               let apiError = amplifyError as? APIError,
-               case let .networkError(_, _, underlyingError) = apiError {
-                return underlyingError
-            }
+        if case let .api(amplifyError, _) = datastoreError,
+           let apiError = amplifyError as? APIError,
+           case let .networkError(_, _, underlyingError) = apiError {
+            return underlyingError
         }
 
         return nil
-
     }
 }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
@@ -158,10 +158,16 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
             return .successfulVoid
         }
 
+        var underlyingError: Error?
+        if syncErrors.contains(where: isNetworkError) {
+            underlyingError = getFirstUnderlyingNetworkError(errors: syncErrors)
+        }
+
         let allMessages = syncErrors.map { String(describing: $0) }
         let syncError = DataStoreError.sync(
             "One or more errors occurred syncing models. See below for detailed error description.",
-            allMessages.joined(separator: "\n")
+            allMessages.joined(separator: "\n"),
+            underlyingError
         )
         return .failure(syncError)
     }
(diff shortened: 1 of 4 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -158,10 +158,16 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
             return .successfulVoid
         }
 
+        var underlyingError: Error?
+        if syncErrors.contains(where: isNetworkError) {
+            underlyingError = getFirstUnderlyingNetworkError(errors: syncErrors)
```
diegocstn, line 163: Any potential side effects on using the first network error? (ie. can different network errors have different retry policies?)

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -219,4 +225,39 @@ extension AWSInitialSyncOrchestrator {
         }
         return false
     }
+
+    private func isNetworkError(_ error: DataStoreError) -> Bool {
+        guard case let .sync(_, _, underlyingError) = error,
```
thisisabhash, line 230: is this check required if we are only looking for network error?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -219,4 +225,39 @@ extension AWSInitialSyncOrchestrator {
...
+           case .networkError = apiError {
+            return true
+        }
+
+        return false
+    }
+
+    private func getFirstUnderlyingNetworkError(errors: [DataStoreError]) -> Error? {
+        for error in errors {
+            guard case let .sync(_, _, underlyingError) = error,
```
thisisabhash, line 247: same comment as previous

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
diegocstn: nit: can we use `.first(where predicate:)` and check if `syncErrors` has any network error and grab the first in a single operation?

lawmicha: Currently, it is not ideal to have RemoteSyncEngine contain URLError based retry logic, and this is why it's forcing us to propagate one of the network errors up to the RemoteSyncEngine. Ideally, InitialSyncOperation should be responsible for checking the network error, determining that the error is retryable, and retry with the exponential back-off. This means it will cycle on the single InitialSyncOperation trying to sync that one model. Once the network is back up, then the operation will finish, and move onto the next InitialSyncOperation.

This code current assumes that the first network error is probably similar to other network errors if they were run approximately the same time. Of course this assumption could be wrong, but the side effect of it being wrong and pick up a network error that RemoteSyncEngine decides is non-retryable is that the sync engine will move to an errored state, as with all other non-retryable errors.

The retry policy i believe is determined by the implementation in `RequestRetryablePolicy` which filters down to
```
case .notConnectedToInternet,
             .dnsLookupFailed,
             .cannotConnectToHost,
             .cannotFindHost,
             .timedOut:
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
lawmicha: This is required because the InitialSyncOrchestrator wraps the error from the operations in a `DataStoreError.sync` at line 120, stores in `syncErrors`, and moves onto the next operation.  Comes to think of it, i'm not really sure why it needs to be this way, since `syncErrors` is just an array of `[DataStoreError]`. But as it is currently, it needs to unwrap the `.sync` case like it's done above in `isUnauthorizedError`
```
if case .failure(let dataStoreError) = result {
                    let syncError = DataStoreError.sync(
                        "An error occurred syncing \(modelSchema.name)",
                        "",
                        dataStoreError)
                    self.syncErrors.append(syncError)
```

thisisabhash (approved):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -159,8 +159,8 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
         }
 
         var underlyingError: Error?
-        if syncErrors.contains(where: isNetworkError) {
-            underlyingError = getFirstUnderlyingNetworkError(errors: syncErrors)
+        if let error = syncErrors.first(where: isNetworkError(_:)) {
+            underlyingError = getFirstUnderlyingNetworkError(error)
```
diegocstn, line 163: nit: `getFirstUnderlyingNetworkError` -> `getUnderlyingNetworkError`

diegocstn (approved):

diegocstn (approved):

Issue #1787, opened 2022-05-10
Replace precondition handling in tests so it works with Swift Concurrency


brennanMKE:
### Describe the bug

[CwlPreconditionTesting](https://github.com/mattgallagher/CwlPreconditionTesting) is used to intercept exits when  `preconditionFailure` is used during testing. It does not work with Swift Concurrency and an issue has been created for this dependency.

https://github.com/mattgallagher/CwlPreconditionTesting/issues/27

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Create a unit test which calls a function which calls `preconditionFailure`
2. Run that test in an async context with modern Swift Concurrency
3. Observe that the exit is not intercepted and the process exits which fails the test
```

### Expected behavior

The exit should be trapped like it has been working without the new async features in Swift.

### Amplify Framework Version

dev-preview

### Amplify Categories

Analytics, API, Auth, DataStore, Geo, Predictions, Storage

### Dependency manager

Swift PM

### Swift version

5.6

### CLI version

8.1.0

### Xcode version

13.3.1

### Is this a regression?

No

### Device

iOS Simulator

### iOS Version

iOS 15

PR #1793, opened 2022-05-11
chore: adds preconditionFailure support for testing with Swift Concurrency

brennanMKE:
Swift Concurrency is not compatible with the way `preconditionFailure` is trapped while testing. And so this alternate function supports tests which may need to run under a `Task` which calls `preconditionFailure`. See the issue for more context.

*Issue #, if available:*

#1787

*Description of changes:*

The key function is below. If it is executed while testing and the `instanceFactory` is defined it will be used to get the instance which much be returned. Normally `precondtionFailure` won't return because the return type is `Never`. This way tests can register a type which will be returned and use the closure as a part of the test.

```swift
@discardableResult
public static func preconditionFailure<T>(_ message: @autoclosure () -> String = String(),
                                          file: StaticString = #file,
                                          line: UInt = #line) -> T {
    guard let instanceFactory = AmplifyTesting.getInstanceFactory() else {
        Swift.preconditionFailure(message(), file: file, line: line)
    }
    do {
        return try instanceFactory.get(type: T.self, message: message())
    } catch {
        die(reason: "Error: \(error)", file: file, line: line)
    }
}
```

A test will work with the following code.

```swift
let registry = TypeRegistry.register(type: AuthCategoryPlugin.self) { _ in
    MockAuthCategoryPlugin()
}

_ = Amplify.Auth.update(oldPassword: "current", to: "new", listener: nil)

XCTAssertGreaterThan(registry.messages.count, 0)
```

The `TypeRegistry` is defined in AmplifyTestCommon and can be used to register a type with a closure which returns an instance of that type. The closure takes a string which is the message for `preconditionFailure` which can be seen with a breaking in the unit test. The asserts check that the precondition failed by ensuring the `message` is defined and `count` is 1.

The features defined for testing are `internal` and are available during testing due to the `@testable import` statement. The new additions are not available as a part of Amplify's public API.

- [x] All unit tests pass


AmplifyTestCommon/Helpers/TypeRegistry.swift
```diff
@@ -0,0 +1,37 @@
...
+typealias TypeFactory = (String?) -> Any
+
+class TypeRegistry {
+    enum Failure: Error {
+        case typeNotRegistered(Any.Type)
+    }
+
+    private var registry = [TypeKey: TypeFactory]()
+
+    func register<T>(type: T.Type, factory: @escaping (String?) -> T) {
```
thisisabhash, line 22: does this need to be thread safe?

Amplify/Amplify.swift
```diff
@@ -76,6 +81,24 @@ public class Amplify {
                 "Verify that the library version is correct and supports the plugin's category.")
         }
     }
+
+    static var isTesting: Bool {
+        ProcessInfo.processInfo.environment["XCTestConfigurationFilePath"] != nil
```
thisisabhash, line 86: Is the key "XCTestConfigurationFilePath" going to change later on? Or maybe can this be replaced with a swift library constant?

thisisabhash (approved):
LGTM. I added some minor comments

diegocstn (approved):

AmplifyTestCommon/Helpers/TypeRegistry.swift
brennanMKE: no, it will be called in the unit test once and not during any async activity

Amplify/Amplify.swift
brennanMKE: this is the most common way that code checks if unit tests are being run. if this does change we can update the code with that change.

AmplifyTestCommon/Helpers/TypeKey.swift
```diff
@@ -0,0 +1,24 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+struct TypeKey: Hashable {
```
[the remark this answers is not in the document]
brennanMKE: moved

PR #1807, opened 2022-05-18
Feature: add analytic local storage support

phantumcode:
*Issue #, if available:*

*Description of changes:*
* Add SQLite local storage adapter to support storing analytics event
* A temporary class `AnalyticsEventStorage` is added as a placeholder for some of the queries currently being used.  These can be migrated/updated to codable SQL queries when the events and events data model for available for consumption.
* Note that the Analytics Plugin currently doens't compile and isn't a target in Package.swift.  This work still needs to be done and is out of scope for this PR.

~- [ ] Build succeeds with all target using Swift Package Manager~
- [x] All unit tests pass
~- [ ] All integration tests pass~
~- [ ] Documentation update for the change if required~
~- [ ] If breaking change, documentation/changelog update with migration instructions~

### DIFF
--- a/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/AnalyticEventStorage.swift
+++ b/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/AnalyticEventStorage.swift
@@ -18,6 +18,45 @@ class AnalyticsEventStorage {
         self.dbAdapter = dbAdapter
     }
     
+    /// Create the Event and Dirty Event Tables
+    func createTables() throws {
+        let createEventTableStatement = """
+            CREATE TABLE IF NOT EXISTS Event (
+            id TEXT NOT NULL,
+            attributes BLOB NOT NULL,
+            eventType TEXT NOT NULL,
+            metrics BLOB NOT NULL,
+            eventTimestamp TEXT NOT NULL,
+            sessionId TEXT NOT NULL,
+            sessionStartTime TEXT NOT NULL,
+            sessionStopTime TEXT NOT NULL,
+            timestamp REAL NOT NULL,
+            dirty INTEGER NOT NULL,
+            retryCount INTEGER NOT NULL)
+        """
+        let createDirtyEventTableStatement = """
+            CREATE TABLE IF NOT EXISTS DirtyEvent (
+            id TEXT NOT NULL,
+            attributes BLOB NOT NULL,
+            eventType TEXT NOT NULL,
+            metrics BLOB NOT NULL,
+            eventTimestamp TEXT NOT NULL,
+            sessionId TEXT NOT NULL,
+            sessionStartTime TEXT NOT NULL,
+            sessionStopTime TEXT NOT NULL,
+            timestamp REAL NOT NULL,
+            dirty INTEGER NOT NULL,
+            retryCount INTEGER NOT NULL)
+        """
+
+        do {
+            try dbAdapter.createTable(createEventTableStatement)
+            try dbAdapter.createTable(createDirtyEventTableStatement)
+        } catch {
+            throw LocalStorageError.invalidOperation(causedBy: error)
+        }
+    }
+    
     /// Insert an Event into the Even table
     /// - Parameter bindings: a collection of values to insert into the Event
     func insertEvent(bindings: [Binding]) throws {
@@ -28,7 +67,7 @@ class AnalyticsEventStorage {
             sessionStopTime, timestamp, dirty, retryCount)
             VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
         """
-        _ = try dbAdapter.executeSqlQuery(insertStatement, bindings)
+        _ = try dbAdapter.executeQuery(insertStatement, bindings)
     }
     
     /// Delete events from the Events older than a specified timestamp
@@ -38,15 +77,15 @@ class AnalyticsEventStorage {
          DELETE FROM Event
          WHERE timestamp < ?"
         """
-        _ = try dbAdapter.executeSqlQuery(deleteStatement, [timeStamp])
+        _ = try dbAdapter.executeQuery(deleteStatement, [timeStamp])
     }
     
     /// Delete all dirty events from the Event and DirtyEvent tables
     func deleteDirtyEvents() throws {
         let deleteFromDirtyEventTable = "DELETE FROM DirtyEvent"
         let deleteFromEventTable = "DELETE FROM Event WHERE dirty = true"
-        _ = try dbAdapter.executeSqlQuery(deleteFromDirtyEventTable, [])
-        _ = try dbAdapter.executeSqlQuery(deleteFromEventTable, [])
+        _ = try dbAdapter.executeQuery(deleteFromDirtyEventTable, [])
+        _ = try dbAdapter.executeQuery(deleteFromEventTable, [])
     }
     
     /// Delete the oldest event from the Event table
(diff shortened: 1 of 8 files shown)


AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/LocalStorageError.swift
```diff
@@ -0,0 +1,13 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Errors interfacing with local storage
+public enum LocalStorageError: Error {
```
ruisebas, line 9: Is `public` needed or can it be `internal`?

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/AnalyticEventStorage.swift
```diff
@@ -0,0 +1,95 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import SQLite
```
ruisebas, line 8: I assume this library will need to be added to the plugin as a dependency

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/SQLiteLocalStorageAdapter.swift
```diff
@@ -0,0 +1,136 @@
...
+    }
+    
+    /// Create the Event and Dirty Event Tables
+    private func createTables() throws {
+        guard let connection = connection else {
+            throw LocalStorageError.invalidOperation(causedBy: nil)
+        }
+
+        let createEventTableStatement = """
+            CREATE TABLE IF NOT EXISTS Event (
```
ruisebas, line 53: I'm wondering if these table definitions should be owned by this adapter, or if they should be provided by its consumer, i.e. the `AnalyticsEventStorage`

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/LocalStorageProtocol.swift
```diff
@@ -0,0 +1,18 @@
...
+import Foundation
+import SQLite
+
+protocol LocalStorageProtocol {
+    /// Executes a SQL statement
+    /// - Parameters:
+    ///   - statement: SQL statement
+    ///   - bindings: Collection of SQL Bindings
+    /// - Returns: A Single SQL Statement
+    func executeSqlQuery(_ statement: String, _ bindings: [Binding?]) throws -> Statement
```
diegocstn, line 17: since `LocalStorageProtocol` seems to be agnostic in respect to the underlying storage, can we perhaps change this method name such that doesn't have any reference to `SQL` and the `SQLite` library?

diegocstn: note: I'd also add a generic "initializeStorage"-like method that types conforming to this protocol should implement to inject their own logic (i.e. creating the db and tables for SQLite or something for a different adapter)

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/AnalyticEventStorage.swift
diegocstn, line 8: shouldn't the `AnalyticsEventStorage` be storage-agnostic? If so, the logic specific to a SQLite adapter (i.e. how to retrieve the actual data from the storage) should be part of `SQLiteLocalStorageAdapter` only.

diegocstn: I think not having `AnalyticsEventStorage` tied to a specific implementation of a storage will simplify testing of its consumers

ruisebas: I think it's not made agnostic because this class is the replacement for `AWSPinpointEventRecorder`, which included all the SQL queries within. 

But maybe we can improve on that, but eventually we'll need to have the SQL-queries being generated somewhere. Perhaps inside `SQLiteLocalStorageAdapter` ?

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/LocalStorageError.swift
phantumcode: Good catch, I've updated to be internal.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/AnalyticEventStorage.swift
phantumcode: yes, this will a plugin dependency

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/SQLiteLocalStorageAdapter.swift
phantumcode: That's a good suggestion, I've moved all query statements that are specific to events out of the adapter and into the storage class.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/LocalStorageProtocol.swift
phantumcode: I've re-organize the structure a little bit.  I've made the LocalStorageProtocol specific to SQL and renamed it to SQLStorageProtocol where the adapter is limited to creating table and executing SQL queries and doesn't know anything about Events or Pinpoint Events.  The agnostic abstraction is moved one layer up to the AnalyticsEventStorage and AnalyticsEventSQLStorage implementation.  I'm open to deleting and removing this protocol completely if it doesn't make sense anymore.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/AnalyticEventStorage.swift
phantumcode: The initial intent of AnalyticsEventStorage is just a placeholder class for now to hold all of the SQL queries. Based on the feedback, I've made an AnalyticsEventStorage protocol that is storage agnostic and an AnalyticsEventsSQLStorage implementation that is SQL specific.  All SQL queries and create statements should be in this SQL implimentation, I'd imagine that this is where all Pinpoint Events operations gets converted to SQL.  The SQLiteLocalStorageAdapter becomes a thin lightweight interface to the database, and all it does it executes SQL commands.  Let me know if that is a workable solution.

ruisebas (approved):

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/LocalStorageError.swift
```diff
@@ -0,0 +1,13 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Errors interfacing with local storage
+enum LocalStorageError: Error {
+    case missingConnection
+    case invalidDatabase(path: String, Error? = nil)
```
diegocstn, line 11: nit: perhaps we can rename this to `invalidStorage` since this `LocalStorageError` is now decoupled from SQLite?

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/LocalStorage/LocalStorageError.swift
```diff
@@ -0,0 +1,13 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Errors interfacing with local storage
+enum LocalStorageError: Error {
```
diegocstn, line 9: nit: have you considered moving this enum under `AnalyticsEventStorage`?

diegocstn (approved):

PR #1823, opened 2022-05-23
fix(datastore): ModelSyncedEvent dispatch consistency

lawmicha:
*Issue #, if available:*

*Description of changes:*
SyncEventEmitter takes in both InitialSyncOrchestrator and ReconciliationQueue to receive events from. InitialSyncOrchestrator is responsible for providing events to each of the ModelSyncedEventEmitter's to increment a count of models received from the syncQuery, by filtering down to the modelName, and then using the events to figure out information like whether it was a full or delta sync, number of records received. ReconciliationQueue similarly provides events to each of the ModelSyncedEventEmitter's to increment the count of models reconciled. With the two stream of events, an on-going calculation is done to check and emit the ModelSyncedEvent when the reconciled number matches the received number.

In this PR, InitialSyncOperation will always send an `.finished` event whenever the operation itself finishes. The `finished` event is an event passed to the ModelSyncedEventEmitter to ensure calculation done correctly. When the `.finished` event is not sent from InitialSyncOperation, 
 ModelSyncedEventEmitter will never emit the ModelSyncedEvent because it doesn't think it has finished receiving the received record count. This would happen if InitialSyncOperation performs a successful syncQuery API call, followed by a failed one.

The second change in this PR, ReconcileAndLocalSaveOperation will emit the mutationEventDropped event in operation error cases as well. From ModelSyncedEventEmitter's prespective, a reconciled count is when a record was either reconciled (save/updated/deleted) or dropped (smaller or duplicate version). ReconcileAndLocalSaveOperation will emit the dropped event even in the case where the model was determined to be reconciled but failed (for example, local storage adapter failed to save). This will contribute to the overall calculation of ModelSyncedEventEmitter so that the records reconciled will eventually match the records received, despite errors within those reconciliation. Errors within reconciliation is not handled by ModelSyncedEventEmitter and is out of scope of this PR. Reconciliation errors in general should be sent back on the `errorHandler` for visibility. 

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOperationTests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOperationTests.swift
@@ -451,7 +451,10 @@ class InitialSyncOperationTests: XCTestCase {
                 XCTAssertEqual(syncType, .fullSync)
                 syncStartedReceived.fulfill()
             case .finished(modelName: let modelName, error: let error):
-                XCTAssertNotNil(error)
+                guard case .api = error else {
+                    XCTFail("Should be api error")
+                    return
+                }
                 XCTAssertEqual(modelName, "MockSynced")
                 finishedReceived.fulfill()
             default:
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOrchestratorTests.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOrchestratorTests.swift
@@ -178,7 +178,10 @@ class InitialSyncOrchestratorTests: XCTestCase {
                     syncStartedReceived.fulfill()
                 case .finished(let modelName, let error):
                     if modelName == "Post" {
-                        XCTAssertNotNil(error)
+                        guard case .api(let apiError, _) = error, case .operationError = apiError as? APIError else {
+                            XCTFail("Should be api error")
+                            return
+                        }
                     } else if modelName == "Comment" {
                         XCTAssertNil(error)
                     }


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -138,7 +126,17 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
...
-    private func onReceiveCompletion() {
+    private func onReceiveCompletion(modelSchema: ModelSchema, result: Subscribers.Completion<DataStoreError>) {
+        if case .failure(let dataStoreError) = result {
+            let syncError = DataStoreError.sync(
+                "An error occurred syncing \(modelSchema.name)",
+                "",
+                dataStoreError)
+            self.syncErrors.append(syncError)
+        }
+
```
thisisabhash, line 137: is this logic no longer needed?
```
if self.isUnauthorizedError(syncError) {
      self.initialSyncOrchestratorTopic.send(.finished(modelName: modelSchema.name))
}
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOrchestratorTests.swift
```diff
@@ -96,6 +102,102 @@ class InitialSyncOrchestratorTests: XCTestCase {
...
+                case .failure:
+                    failureCompletionReceived.fulfill()
+                }
+            }, receiveValue: { value in
+                switch value {
+                case .started:
+                    syncStartedReceived.fulfill()
+                case .finished(let modelName, let error):
+                    if modelName == "Post" {
+                        XCTAssertNotNil(error)
```
thisisabhash, line 181: Can we assert on the error type also we are expecting [OperationError]

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Sync/InitialSync/InitialSyncOperationTests.swift
```diff
@@ -443,6 +450,10 @@ class InitialSyncOperationTests: XCTestCase {
                 XCTAssertEqual(modelName, "MockSynced")
                 XCTAssertEqual(syncType, .fullSync)
                 syncStartedReceived.fulfill()
+            case .finished(modelName: let modelName, error: let error):
+                XCTAssertNotNil(error)
```
thisisabhash, line 454: Can we add assert for the error type here also?

thisisabhash (dismissed):
LGTM, minor comments

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
lawmicha: yup, it's no longer needed because we will always mark the initial sync as finished regardless of the error. we can and should in the future reuse the method to adhere to the `errorHandler` behavior, ie.

```
if self.isUnauthorizedError(syncError) {
   self.errorHandler(SyncError.unauthorizedError)
}
```

I left this out to isolate the changes in this PR to be just related to the consistency of emitting the ModelSyncedEvent, and kept the method since i anticipated that we'll be using it soon in a future change

thisisabhash (approved):

PR #1826, opened 2022-05-24
feat(Analytics): Implementing PinpointContext, PinpointSession and KeychainStore

ruisebas:
*Description of changes:*
- Implemented the `PinpointContext` class and its `PinpointContextConfiguration`.
- Implemented the `PinpointSession` class and some of the `SessionTracker` functionality.
- Changed the Type used to represent milliseconds to `Int64`. `TimeInterval` by definition is always expressed in seconds, so it was wrong and misleading to use it. 
- Implemented the `KeychainStore` utility in `AWSPluginsCore`. Note that it's a copy-paste from Auth's `CredentialStore`, but located in a shared place. We'll update Auth to remove the old implementation in a future PR.


AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift
```diff
@@ -5,12 +5,46 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+import Amplify
+import AWSClientRuntime
 import AWSPinpoint
+import AWSPluginsCore
 import Foundation
 
+protocol UserDefaultsBehaviour {
```
phantumcode, line 14: Do you know what the protocol naming convention is for this project? is it `UserDefaultsProtocol`, `UserDefaultsBehavior` or `UserDefaultsBehaviour`? I've seen a mix so far, we should pick one and use it consistently.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift
```diff
@@ -5,12 +5,46 @@
...
+
+extension UserDefaults: UserDefaultsBehaviour {}
+
+struct PinpointContextConfiguration {
+    /// The Pinpoint AppId.
+    let appId: String
+    /// The session timeout in seconds. Defaults to 5 seconds.
+    let sessionTimeout: TimeInterval
+    /// The max storage size to use for event storage in MB. Defaults to 5 MB.
+    let maxStorageSize: Int
```
diegocstn, line 30: can we use a more descriptive typealias here as well?

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Session/SessionTracker.swift
```diff
@@ -5,12 +5,158 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+import Amplify
 import Foundation
+import UIKit
 
 class SessionTracker: InternalPinpointClient {
+    private var session: PinpointSession? = nil
+    private var backgroundTask: UIBackgroundTaskIdentifier = .invalid
```
diegocstn, line 14: we'll need abstractions around all these UIKit-related symbols (`UIBackgroundTaskIdentifier`, `UIApplication`) to support our cross-platforms effort.
I'd recommend having protocols and add conformance using Swift compiler `canImport` directive.

brennanMKE: in this case you do not need an abstraction because there is no equivalent outside of UIKit.

ruisebas: I will extract this behaviour to a separate class, so we keep UIKit out of this one.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Session/PinpointSession.swift
```diff
@@ -0,0 +1,101 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+class PinpointSession: Codable {
+    let sessionId: String
```
diegocstn, line 11: nit: `SessionId` typealias?

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift
```diff
@@ -19,17 +61,132 @@ class PinpointContext {
...
+              FileManager.default.fileExists(atPath: preferencesPath),
+              let preferencesJson = try? JSONSerialization.jsonObject(with: Data(contentsOf: URL(fileURLWithPath: preferencesPath)),
+                                                                      options: .mutableContainers) as? [String: String] else {
+            return nil
+        }
+        
+        return preferencesJson[Constants.Preferences.uniqueIdKey]
+    }
+    
+    private func retrieveUniqueId() -> String {
```
diegocstn, line 118: nit: let's document this method :) i.e. why we're looking in three different places etc etc

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift
```diff
@@ -19,17 +61,132 @@ class PinpointContext {
...
+        return preferencesFileUrl?.path
+    }
+    
+    private func removeLegacyPreferencesFile() {
+        guard let preferencesPath = legacyPreferencesFilePath else {
+            return
+        }
+        
+        do {
+            try FileManager.default.removeItem(atPath: preferencesPath)
```
diegocstn, line 101: in order to fully test this (and the `legacyUniqueId` and `retrieveUniqueId` methods), we should also inject an abstraction of FileManager in the initializer
We can perhaps have a protocol similar https://github.com/aws-amplify/amplify-ios/blob/main/AmplifyTools/AmplifyXcode/Sources/AmplifyXcodeCore/Environment/AmplifyFileManager.swift
and then just add conformance to Foundation `FileManager`

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift
diegocstn: Also, here and throughout these new types instead of accessing `Amplify.Analytics.log` we should add conformance to the `DefaultLogger` 
```swift
extension PinpointContext: DefaultLogger {}
```
and use the `log` method directly on this type
```swift
self.log.verbose("...")
```
By doing so log entries will show up with the `[PinpointContext]` prefix, making debug easier.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Session/SessionClient.swift
```diff
@@ -0,0 +1,160 @@
...
+        saveSession()
+        Amplify.Analytics.log.info("Session Started.")
+        activityTracker.delegate = self
+        activityTracker.beginActivityTracking()
+
+        let startEvent = context.analyticsClient.createEvent(withEventType: Constants.Events.start)
+        // Update Endpoint and record Session Start event
+        Task {
+            try? await context.targetingClient.updateEndpointProfile()
+            Amplify.Analytics.log.verbose("Firing Session Event: Start")
```
diegocstn, line 52: ^^^ ditto, `DefaultLogger` conformance

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift
ruisebas: Yep, I definitely agree. I'll add it

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift
ruisebas: > why we're looking in three different places 

I honestly have no idea, I can deduce that we've been migrating the place where we store that id through the years and kept backwards compatibility 😆 . I'll add some fancy wording at the beginning to reflect that.

> we should add conformance to the `DefaultLogger`

Nice!! I didn't know about that, but we absolutely should follow that

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Session/PinpointSession.swift
ruisebas: It can be done and I guess since it's an internal class it shouldn't be a problem, but I'm wondering: where do we draw the line when creating `typealias`es?

I mean this id is just a `String`, what's the reasoning behind favouring a typealias over the primitive?  🤔

diegocstn (approved):

Issue #1829, opened 2022-05-25
Implement get current user in Auth


brennanMKE:
### Is your feature request related to a problem? Please describe.

dev-preview

### Describe the solution you'd like

dev-preview

### Describe alternatives you've considered

dev-preview

### Is the feature request related to any of the existing Amplify categories?

Auth

PR #1830, opened 2022-05-25
feat(auth): implements get current user

brennanMKE:
*Issue #, if available:*

#1829 

*Description of changes:*

Implements `getCurrentUser` function and adds unit tests.

- [ ] All unit tests pass

### DIFF
--- a/Amplify/Categories/Auth/AuthCategory+UserBehavior.swift
+++ b/Amplify/Categories/Auth/AuthCategory+UserBehavior.swift
@@ -9,8 +9,12 @@ import Foundation
 
 extension AuthCategory: AuthCategoryUserBehavior {
 
-    public func getCurrentUser() -> AuthUser? {
-        return plugin.getCurrentUser()
+    public func getCurrentUser() async -> AuthUser? {
+        return await plugin.getCurrentUser()
+    }
+
+    public func getCurrentUser(closure: @escaping (Result<AuthUser?, Error>) -> Void) {
+        plugin.getCurrentUser(closure: closure)
     }
 
     @discardableResult
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/ClientBehavior/AWSCognitoAuthPlugin+UserBehavior.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/ClientBehavior/AWSCognitoAuthPlugin+UserBehavior.swift
@@ -111,7 +111,26 @@ public extension AWSCognitoAuthPlugin {
         return operation
     }
 
-    func getCurrentUser() -> AuthUser? {
-        fatalError("Not implemented")
+    func getCurrentUser() async -> AuthUser? {
+        let authState = await authStateMachine.currentMachineState
+        if case .configured(let authenticationState, _) = authState,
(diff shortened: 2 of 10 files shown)


Amplify/Categories/Auth/AuthCategory+UserBehavior.swift
```diff
@@ -13,6 +13,14 @@ extension AuthCategory: AuthCategoryUserBehavior {
         return plugin.getCurrentUser()
     }
 
+    public func getCurrentUser() async -> AuthUser? {
+        return await plugin.getCurrentUser()
+    }
```
harsh62, line 18: @royjit is it ok if we start providing async API's to customers?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
```diff
@@ -162,6 +162,16 @@ class StateMachine<
         }
     }
 
+    var currentMachinState: StateType {
```
harsh62, line 165: should we do it like this instead `func getCurrentState() async -> StateType`
AND
deprecate `func getCurrentState(_ completion: @escaping (StateType) -> Void)`

Amplify/Categories/Auth/AuthCategory+UserBehavior.swift
brennanMKE: Since Auth already makes this process async we don't have a choice since we cannot block in this function.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/ClientBehavior/AWSCognitoAuthPlugin+UserBehavior.swift
```diff
@@ -70,7 +70,45 @@ public extension AWSCognitoAuthPlugin {
...
+    func getCurrentUser(closure: @escaping (Result<AuthUser?, Error>) -> Void) {
+        authStateMachine.getCurrentState { authState in
+            if case .configured(let authenticationState, _) = authState,
+               case .signedIn(_, let signInData) = authenticationState {
+                let authUser = AWSCognitoAuthUser(username: signInData.userName, userId: signInData.userId)
+                closure(.success(authUser))
+            } else {
+                closure(.success(nil))
+            }
+        }
```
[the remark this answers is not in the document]
brennanMKE: I discussed this with Chris. When a function uses `async` it operates in the cooperative thread pool which has behavior which may not behave if it is mixed with code which uses Dispatch on the caller side. By offering this version with the closure it will not use the cooperative thread pool and instead just use the operation queue which Auth is using. It is best to keep these code paths separate.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
brennanMKE: We want to keep them separate for now since async functions run with different threads than what Dispatch normally uses.

PR #1849, opened 2022-06-02
fix(datastore): query for missing optional associations

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/1792
https://github.com/aws-amplify/amplify-ios/issues/1784

*Description of changes:*
This PR adds logic in the SQL results to Model decoding that fixes the problem with missing associations.
1. SQL select statement queries for all belongs-to associated models. Querying for a comment will eager load the post and post.blog. The results from the SQL query is a flat array of values.

```
// when blog exists on the post
[
 "id": <POST_ID>
 "blog.id" -> <BLOG_ID>
 "blog.description" -> "<value>"
 ...
]

// when blog does not exist on the post
[
 "id": <POST_ID>
 "blog.id" -> nil
 "blog.description" -> nil
 ...
]
```
3. The flat array of values is then transformed into a nested dictionary of type `[String: Any]`
4. The issue is that this transformation creates undecodable models to the Model types, when the associated model does not exist. 

One example is when post does not belong to any blog, the SQL results contain `nil` values for all of blog fields, and is transformed to a dictionary containing nil values for the blog object. 
```
"blog.id" = nil
"blog.notes" = nil
"blog.description" = nil

is transformed to
{
  "post":  {
    "id": <ID>,
     "blog": { 
         "id": nil,
          "notes": nil
          "description": nil
   }
 }
}
```
Any required field of blog will fail decoding, for example `id` is the primary key, is required, and fails decoding due to the value being `nil`.

5. The fix is to check for the primary key value when transforming to the dictionary, if it's `nil` then store `nil` at the root of the associated model, ie. if "blog.id" is nil then store dictionary["post"]["blog"] = nil, so that decoding will decode to a post with post.blog = nil. 

```
{
  "post":  {
    "id": <ID>,
     "blog": nil
 }
}
```

6. This logic gets a bit complicated when we consider the scenarios of nested associations, like comment belongs to post, and post belongs to blog. We only set the dictionary["post"]["blog"] = nil only when dictionary["post"] != nil, so if both are nil then we only set dictionary["post"] = nil.
```
// if post exists on comment, but blog doesn't
// desired transformation
{
  "comment":  {
    "id": <ID>,
     "post": {
        "id": <ID>
         "blog": nil
   }
 }
}
// if post and blog does not exist
// desired transformation
{
  "comment":  {
    "id": <ID>,
     "post": nil
 }
}
```
7. the dictionary is constructed in a way that is decodable to the Model types, which contains optional associations.

- [x] All unit tests pass


mikepschneider (dismissed):

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -88,9 +89,39 @@ extension Statement: StatementModelConvertible {
...
+            // path from the column, for example given "blog.id" `column` has a nil `modelValue`, then store the
+            // keypath in `skipColumns`, which will be used to set modelDictionary["blog"] to nil later on.
+            //
+            // `skipColumns` keeps track of these scenarios. At this point in the code we cannot make an assumption
+            // about the ordering of the columns, it may handle "blog.description", then "blog.id", then "blog.title".
+            // The code will perform the following:
+            //  1. set ["blog"]["description"] = `nil`, because it has not encountered `nil` primary key
+            //  2. store skipColumn["blog"], because the value is the primary key and is nil
+            //  3. skip setting modelDictionary["blog"]["title"], because `skipColumn` has been set for "blog".
+            if field.isPrimaryKey && modelValue == nil {
```
lawmicha, line 105: - primary key usage here, rebase with CPK work to make sure tests are still passing

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Storage/SQLite/StatementModelTests.swift
```diff
@@ -0,0 +1,35 @@
...
+import SQLite
+import XCTest
+
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSDataStoreCategoryPlugin
+
+class StatementModelTests: XCTestCase {
+
+    func testDropLastPath_Root() throws {
```
thisisabhash, line 18: Can there be a testcase like "post." or empty string "" ?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
thisisabhash, line 105: i am wondering how this will work in case of a composite primary key where there would be an array of fields for primary keys and a custom primary key ""key1"#"key2".." with a column @ @ primarykey

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Storage/StorageEngineTestsOptionalAssociation.swift
```diff
@@ -0,0 +1,287 @@
...
+            return
+        }
+
+        guard case let .success(queriedBlog) =
+            querySingleModelSynchronous(modelType: Blog8.self,
+                                        predicate: Blog8.keys.id == blog.id) else {
+                XCTFail("Failed to query blog")
+                return
+        }
+        XCTAssertEqual(queriedBlog.id, blog.id)
```
thisisabhash, line 100: nit: can also add assertions for custom model and nested model

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Storage/StorageEngineTestsOptionalAssociation.swift
```diff
@@ -0,0 +1,287 @@
...
+
+        guard case let .success(updatedPost) =
+            querySingleModelSynchronous(modelType: Post8.self,
+                                        predicate: Post8.keys.id == post.id) else {
+                XCTFail("Failed to query post")
+                return
+        }
+        XCTAssertEqual(updatedPost.id, post.id)
+        XCTAssertNotNil(updatedPost.blog)
+        XCTAssertEqual(updatedPost.blog?.id, blog.id)
```
thisisabhash, line 156: nit: similar comment as above

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Storage/StorageEngineTestsOptionalAssociation.swift
```diff
@@ -0,0 +1,287 @@
...
+        guard case .success = saveModelSynchronous(model: post) else {
+            XCTFail("Failed to save post")
+            return
+        }
+        let statement = SelectStatement(from: Post8.schema,
+                                        predicate: nil,
+                                        sort: nil,
+                                        paginationInput: nil)
+        let rows = try connection.prepare(statement.stringValue).run(statement.variables)
+        XCTAssertEqual(rows.columnCount, 12)
```
thisisabhash, line 247: Can we add a small comment about why we expect columnCount to be 12?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Storage/StorageEngineTestsOptionalAssociation.swift
```diff
@@ -0,0 +1,287 @@
...
+        let results: [Post8] = try rows.convert(to: Post8.self,
+                                                withSchema: Post8.schema,
+                                                using: statement)
+
+        XCTAssertNotNil(results)
+        XCTAssertEqual(results.count, 1)
```
thisisabhash, line 253: similar comment as above.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/SQLite/Statement+Model.swift
lawmicha: I compared the two PRs, this one and the CPK one, in this PR
```
public var isPrimaryKey: Bool {
        return name == "id"
    }
```

In CPK PR
```swift
@available(*, deprecated, message: "Use the primaryKey member of the schema")
    public var isPrimaryKey: Bool {
        return attributes.contains { $0 == .primaryKey }
    }
```

It appears in the CPK PR, it is still checking the deprecated isPrimaryKey from the field for creating lazy lists on the associations (Under the comment `// create lazy list for "many" relationships`, if let id = modelValue as? String, field.isPrimaryKey`). if there is one primaryKey on the 1 model then it will create a lazy list for each association. If there is a composite key, it will create a lazy list for each association for each time it finds a key that is part of the composite key. the lazy list would be overwritten by the latest lazy list on each component of the composite key, thus the associated id stored in the composite key is only part of the id that can be used for lazy laoding. I've opened a comment over in the CPK PR to fix this: https://github.com/aws-amplify/amplify-ios/pull/1752/files#r900313614

Now back to this check that we're introducing here, if any of the fields is a primary key and the value is `nil`, we will mark the entire model as should be skipped / set to `nil`. This will still work in this case of a composite key since if any of the components of the composite key has a `nil` value, the code will evaluate to true, and run the code at lines 106 and 107 twice. It still works as we expect it to. Until we fully deprecate the field, we can still continue to use it, however should migrate to the CPK's latest way to get the primary key from the schema. I can't exactly do that in this PR since that would mean we would have to ship the CPK PR first.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Storage/SQLite/StatementModelTests.swift
lawmicha: this is interesting to reason about, I'll add them for the reader to understand what is the outcome, but i don't think that input will happen

thisisabhash (approved):

thisisabhash (dismissed):

AmplifyTestCommon/Models/TransformerV2/schema.graphql
```diff
@@ -237,5 +278,4 @@ type SchemaDrift @model {
 enum EnumDrift {
    ONE
    TWO
-   THREE
-}
+   THREE
```
thisisabhash, line 281: is this missing a '}' ?

thisisabhash (approved):
LGTM, one minor comment

AmplifyTestCommon/Models/TransformerV2/schema.graphql
lawmicha: nice catch, updating

thisisabhash (approved):

PR #1858, opened 2022-06-02
feat(analytics): protect AnalyticsClient state with actor

diegocstn:
*Issue #, if available:*
This PR introduces the changes to make `AnalyticsClient` an actor to enforce synchronization for reads/writes to its inner state.
The `createAppleMonetizationEvent`, `createVirtualMonetizationEvent`, `createMonetizationEvent` and `createEvent` APIs are marked as `nonisolated` as they're just factories to instantiate new events and don't mutate the internal state.

*Description of changes:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
+++ b/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
@@ -8,7 +8,7 @@
 import Foundation
 
 class EndpointClient: InternalPinpointClient {
-    internal var context: PinpointContext
+    unowned var context: PinpointContext
     
     init(context: PinpointContext) {
         self.context = context
--- a/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
+++ b/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
@@ -7,7 +7,13 @@
 
 import Foundation
 
-class EndpointClient: InternalPinpointClient {    
+class EndpointClient: InternalPinpointClient {
+    internal var context: PinpointContext
+    
+    init(context: PinpointContext) {
+        self.context = context
+    }
+    
     func currentEndpointProfile() -> PinpointEndpointProfile {
         // TODO: Implement
         fatalError("Not yet implemented")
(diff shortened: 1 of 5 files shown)


AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
```diff
@@ -7,7 +7,13 @@
 
 import Foundation
 
-class EndpointClient: InternalPinpointClient {    
+class EndpointClient: InternalPinpointClient {
+    internal var context: PinpointContext
```
ruisebas, line 11: `context` should be `unowned` in **all** classes that conform to `InternalPinpointClient` to avoid a retain cycle, as the context has strong references to all clients. 

These relationships mirror what the Pinpoint SDK had, perhaps once we have all of them properly implemented we can check if we really do need to have a reference to `context` in each client, or if we just can pass the required values instead.

But for now, they need to be `unowned`

ruisebas (changes requested):

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
ruisebas: Also, do we need the `internal`  to be explicit?

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift
```diff
@@ -89,9 +89,9 @@ class PinpointContext {
         self.keychainStore = keychainStore
         self.userDefaults = userDefaults
         self.fileManager = fileManager
-        let pinpointConfiguration = try PinpointClient.PinpointClientConfiguration(credentialsProvider: credentialsProvider,
-                                                                                   frameworkMetadata: AmplifyAWSServiceConfiguration.frameworkMetaData(),
-                                                                                   region: region)
+        let pinpointConfiguration = try PinpointClient.PinpointClientConfiguration(region: region,
```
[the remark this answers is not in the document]
diegocstn: yeah, apparently has changed

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
diegocstn: > Also, do we need the internal to be explicit? 
Nope, thanks for catching that!

> as the context has strong references to all clients.
yikes, adding `unowned` modifier for now

ruisebas (approved):

PR #1863, opened 2022-06-04
chore: fix release merge action

ameter:
*Description of changes:*
fixes a bug with the release merge action that prevented it from being limited to PRs targeting the release branch

- ~~[ ] Added new tests to cover change, if needed~~
- [x] All unit tests pass
- ~~[ ] Documentation update for the change if required~~

### DIFF
--- a/.github/workflows/release_merge_pr.yml
+++ b/.github/workflows/release_merge_pr.yml
@@ -21,4 +21,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR: ${{ github.event.pull_request.number }}
-        run: "gh pr merge $PR --repo ameter/github-actions-test --merge --subject 'chore: kickoff release' --body ''"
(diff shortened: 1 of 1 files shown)


atierian (commented):
I see we have the `if` check to confirm that the branch is `release` in the **Merge PR** job. Should we also keep the `branches: -release` check on the `on` step to prevent the job from unnecessarily starting? Or does the `on: branches:` check not work as we need it to here?

ameter:
> I see we have the `if` check to confirm that the branch is `release` in the **Merge PR** job. Should we also keep the `branches: -release` check on the `on` step to prevent the job from unnecessarily starting? Or does the `on: branches:` check not work as we need it to here?

`on: branches` doesn't work here.  That was the major problem this PR addresses.  It turns out the `pull_request_review` trigger doesn't support the branches filter, so this would fire on all PR reviews.  Moving the filters from the step to the job level was an optimization to address it triggering on all PR reviews.

PR #1868, opened 2022-06-06
fix(datastore): Test schema drift does not restart sync engine

lawmicha:
*Issue #, if available:*

*Description of changes:*
This PR mainly adds an integration test for a schema drift scenario: 
Schema
```
type SchemaDrift @model {
  id: ID!
  enumValue: EnumDrift
}

enum EnumDrift {
   ONE
   TWO
   THREE
}
```

relevant code generated files
```
public enum EnumDrift: String, EnumPersistable {
  case one = "ONE"
  case two = "TWO"
  // case three = "THREE"
}
```
The thrid enum case has been commented out to replicate an out of date client that hasn't updated to the latest app with the updated code generated schema.

The test makes sure the backend contains a model with the third enum value. Starting DataStore should perform the sync engine process, and fail. Assert that the sync engine doesn't schedule any retries.

The reason why this is a `fix` PR is because the sync engine events is modified to emit `schedulingRestart`, currently only useful for inverted asserts in the integration test

- [x] All unit tests pass


AmplifyPlugins/DataStore/DataStoreCategoryPlugin.xcodeproj/xcshareddata/xcschemes/AWSDataStoreIntegrationV2Scenario3V2Tests.xcscheme
```diff
@@ -100,6 +100,9 @@
                <Test
                   Identifier = "DataStoreScalarTests">
                </Test>
+               <Test
+                  Identifier = "DataStoreSchemaDriftTests">
```
thisisabhash, line 104: Maybe we can have a section like < SelectedTests > which will only include the tests corresponding to one scheme, which can help making number of files less in a PR
https://stackoverflow.com/a/63879084

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/TransformerV2/DataStoreSchemaDriftTests.swift
```diff
@@ -0,0 +1,130 @@
...
+        setUp(withModels: TestModelRegistration())
+        let startSuccess = expectation(description: "start success")
+        try startAmplify {
+            startSuccess.fulfill()
+        }
+        wait(for: [startSuccess], timeout: TestCommonConstants.networkTimeout)
+        // Save some data with the missing enum case, do this by directly calling API
+        // with a custom variables object. Later, decoding will fail.
+        let saveSuccessWithTransformationError = expectation(description: "saved success with transformation error")
+        let schemaDrift = SchemaDrift(enumValue: .one)
```
thisisabhash, line 56: why are we not creating the model with enumvalue as `.three` directly here?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/TransformerV2/DataStoreSchemaDriftTests.swift
```diff
@@ -0,0 +1,130 @@
...
+        } receiveValue: { event in
+            if case .cleanedUp = event {
+                syncEngineCleanedUp.fulfill()
+            }
+            if case .schedulingRestart = event {
+                syncEngineRestarting.fulfill()
+            }
+        }.store(in: &subscriptions)
+        wait(for: [syncEngineCleanedUp, syncEngineFailed], timeout: TestCommonConstants.networkTimeout)
+        wait(for: [syncEngineRestarting], timeout: 10)
```
thisisabhash, line 128: is there a reason for keeping the timeout here different than `TestCommonConstants.networkTimeout`

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/TransformerV2/DataStoreSchemaDriftTests.swift
lawmicha: it's because we're replicating a old client which hasn't upgraded to the latest app containing the new codegenerated files, so there is no `.three` case defined in the Enum, we manually edit the value to "THREE" and send it to AppSync directly

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/TransformerV2/DataStoreSchemaDriftTests.swift
lawmicha: i don't think so, just old code, updating to use test constant

AmplifyPlugins/DataStore/DataStoreCategoryPlugin.xcodeproj/xcshareddata/xcschemes/AWSDataStoreIntegrationV2Scenario3V2Tests.xcscheme
lawmicha: I'll capture this as an issue to explore further to improve the set up of our tests. Currently it's a hassle when adding a new test, we have to disable it from all other schemas except for the schema we want it in (be that a new or existing schema). That link you posted describes using Xcode Test Plans, so I think we should explore that and extend it to other plugins as well if it makes sense.

thisisabhash (approved):

PR #1878, opened 2022-06-08
feat(datastore): custom primary key - add more tests, fix cascade delete

diegocstn:
*Description of changes:*
This PR adds more test models with a custom primary key and fix an issue with cascade delete.

- [x] All unit tests pass


AmplifyTestCommon/Models/CustomPrimaryKey/CommentWithCompositeKeyUnidirectional.swift
```diff
@@ -0,0 +1,44 @@
...
+import Amplify
+import Foundation
+
+public struct CommentWithCompositeKeyUnidirectional: Model {
+  public let id: String
+  public let content: String
+  public var createdAt: Temporal.DateTime?
+  public var updatedAt: Temporal.DateTime?
+  public var postWithCompositeKeyUnidirectionalCommentsId: String?
+  public var postWithCompositeKeyUnidirectionalCommentsTitle: String?
```
lawmicha, line 18: is this the correct code generated file? should it be operating on the optional association like `public var post: Post?` ?

AmplifyTestCommon/Models/CustomPrimaryKey/primarykey_schema.graphql
```diff
@@ -53,3 +45,26 @@ type CommentWithCompositeKey @model {
   content: String!
   post: PostWithCompositeKey @belongsTo
 }
+
+type PostWithTagsCompositeKey @model {
```
lawmicha, line 49: can we add a comment to section off between groups of types that are used together? hard to tell which types are together. the comment can as simple as just the types or description of the relationship we're testing, ie
```
# Composite Key Many-to-Many (Post has-many Tag, Tag has-many Post)
```

AmplifyTestCommon/Models/CustomPrimaryKey/CommentWithCompositeKeyUnidirectional.swift
diegocstn: Not in case of unidirectional has-many (@hasMany without a @belongsTo)

PR #1880, opened 2022-06-08
chore: fix unit tests workflow

ruisebas:
*Description of changes:*
- Fixing `.github/composite_actions/run_xcodebuild_test/action.yml` to actually run the scheme passed in the input
- Adding `AmplifyTests` to the `Amplify` scheme

### DIFF
--- a/.github/composite_actions/run_xcodebuild_test/action.yml
+++ b/.github/composite_actions/run_xcodebuild_test/action.yml
@@ -14,5 +14,5 @@ runs:
   steps:
     - name: Test ${{ inputs.scheme }}
       run: |
-        xcodebuild test -scheme AWSCognitoAuthPlugin -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 13' | xcpretty --simple --color --report junit && exit ${PIPESTATUS[0]}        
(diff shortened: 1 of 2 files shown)


phantumcode (approved):

harsh62 (approved):

.github/composite_actions/run_xcodebuild_test/action.yml
```diff
@@ -14,5 +14,5 @@ runs:
   steps:
     - name: Test ${{ inputs.scheme }}
       run: |
-        xcodebuild test -scheme AWSCognitoAuthPlugin -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 13' | xcpretty --simple --color --report junit && exit ${PIPESTATUS[0]}        
+        xcodebuild test -scheme ${{ inputs.scheme }} -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 13' | xcpretty --simple --color --report junit && exit ${PIPESTATUS[0]}
```
ameter, line 17: Should we pipe this through an environment variable?  I know we (people with write access to the repo) should be the only ones able to control the input here, but we're still using it directly in a bash substitution, so I think reading it into an environment variable and then referencing that in the bash line would be safer

ameter (changes requested):

ameter (approved):

PR #1904, opened 2022-06-14
feat(Auth): Adding UpdateUserAttribute API implementation

harsh62:
*Description of changes:*
The purpose of the PR is to add implementation of Update User Attribute API. 

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Operations/UserOperations/AWSAuthUpdateUserAttributeOperation.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Operations/UserOperations/AWSAuthUpdateUserAttributeOperation.swift
@@ -9,10 +9,10 @@ import Amplify
 import AWSPluginsCore
 import AWSCognitoIdentityProvider
 
-public class AWSAuthFetchUserAttributesOperation: AmplifyOperation<
-AuthFetchUserAttributesRequest,
-[AuthUserAttribute],
-AuthError>, AuthFetchUserAttributeOperation {
+public class AWSAuthUpdateUserAttributeOperation: AmplifyOperation<
+AuthUpdateUserAttributeRequest,
+AuthUpdateAttributeResult,
+AuthError>, AuthUpdateUserAttributeOperation {
     
     typealias CognitoUserPoolFactory = () throws -> CognitoUserPoolBehavior
     
@@ -22,7 +22,7 @@ AuthError>, AuthFetchUserAttributeOperation {
     private var statelistenerToken: AuthStateMachineToken?
     private let fetchAuthSessionHelper: FetchAuthSessionOperationHelper
     
-    init(_ request: AuthFetchUserAttributesRequest,
+    init(_ request: AuthUpdateUserAttributeRequest,
          authStateMachine: AuthStateMachine,
          credentialStoreStateMachine: CredentialStoreStateMachine,
          userPoolFactory: @escaping CognitoUserPoolFactory,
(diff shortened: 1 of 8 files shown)


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Operations/UserOperations/AWSAuthUpdateUserAttributeOperation.swift
```diff
@@ -0,0 +1,144 @@
...
+                                                  clientMetadata: clientMetaData,
+                                                  userAttributes: attributeList.map({ $0.sdkClientAttributeType() }))
+            
+            let result = try await userPoolService.updateUserAttributes(input: input)
+            
+            if self.isCancelled {
+                return
+            }
+            
+            guard let codeDeliveryDetailsList = result.codeDeliveryDetailsList else {
```
royjit, line 85: Does the service always return this value? What about cases where we do not need to verify the update?

AmplifyPlugins/Auth/Tests/AuthHostApp/AuthIntegrationTests/AuthUserAttributesTests/AuthUserAttributesTests.swift
```diff
@@ -100,6 +100,23 @@ class AuthUserAttributesTests: AWSAuthBaseTest {
...
         }
         wait(for: [updateExpectation], timeout: networkTimeout)
+        
+        let fetchUserAttributeExpectation = expectation(description: "Fetch User Attribute operation should complete")
+        
+        _ = Amplify.Auth.fetchUserAttributes(listener: { result in
+            switch result {
+            case .success(let attributes):
+                if let emailAttribute = attributes.filter({ $0.key == .email }).first {
+                    XCTAssertEqual(emailAttribute.value, updatedEmail)
```
royjit, line 110: Does this work? We did not verify the attribute correct?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Operations/UserOperations/AWSAuthUpdateUserAttributeOperation.swift
harsh62: So  the list is an optional that's why I need to place a guard around it. 
```
public var codeDeliveryDetailsList: [CognitoIdentityProviderClientTypes.CodeDeliveryDetailsType]?
```

In cases where the next step is not needed, the list would be empty and not `nil`. In those scenarios, the for loop underneath should handle it. Which is as follows. It iterates the dictionary and if the attribute doesn't exist, will update the result accordingly. 

```
            // Check if all items are added to the dictionary
            for item in attributeList where finalResult[item.key] == nil {
                let updateAttributeResult = AuthUpdateAttributeResult(isUpdated: true, nextStep: .done)
                finalResult[item.key] = updateAttributeResult
            }
```

AmplifyPlugins/Auth/Tests/AuthHostApp/AuthIntegrationTests/AuthUserAttributesTests/AuthUserAttributesTests.swift
harsh62: Yes. So if you check the entire test `func testSuccessfulUpdateEmailAttribute() throws`.
* it first sign up and sign in the user. 
* Updates the email 
* Fetches attributes to check if email was updated. (Added this additional step to reinforce the test)

royjit (approved):

royjit (approved):

PR #1956, opened 2022-06-24
feat(Analytics): Adding targeting region and cleaning up outdated code.

ruisebas:
*Description of changes:*

- I realized that Amplify supports having different `targeting` and `analytics` region on the config file, so I've made the changes to reflect that in the implementation.

- I've created a new `AWSPinpoint` struct to be used as the return type for the `getEscapeHatch()` function, in order to provide both the analytics and targeting clients without needing to either change the signature or create a new one. Also, this remains similar to how it works now with the legacy SDK.

- I removed all references to `AppSessionTracker.swift`, as it was never implemented and its feature (i.e. app session tracking) is already provided by our `SessionClient` class, so there was no point on keeping it there just collecting dust.
  -  I renamed the `sessionTimeout` config to `sessionBackgroundTimeout` to better reflect what it actually does.

- [X] All unit tests pass


phantumcode (approved):
LGTM

PR #1975, opened 2022-06-29
feat(Analytics): Fixing issues when submitting events

ruisebas:
**Description of changes:**

This PR does a couple of things, but mainly focuses in fixing some issues that prevented events from being submitted to the Pinpoint backend and/or to show correctly in the console:
- Forced the ISO8061 format to have an explicit trailing `Z` instead of the `+0000` offset, as otherwise the request was rejected. Did so by creating our own `DateFormatter` and extensions so that it's easier to use.
- Fixed the Event timestamp being calculated wrong before submission (they were stored in milliseconds but used as seconds, resulting in dates many many years in the future 😆 )
- Removed the empty string we were using for the `SdkVersion`, as it was rejected otherwise. A `nil` value is accepted for now, but I've kept the `//TODO` comment so we don't forget to follow up.
- According to the documentation for `EventRequest`:
  >   The batch of events to process. For each item in a batch, the endpoint ID acts as a key that has an EventsBatch object as its value.
 
  So i've changed it to use the `endpointId` as key, instead of the `applicationId`. This meant replacing `EndpointClient.currentPublicEndpoint` with a new `EndpointClient.convertToPublicEndpoint(:)` instead. 

---

Some optimizations were also made:
- We were sending one `putEvent` request per each event, even though the API supports to have many of them in a single request. I've made the change to submit all of them together and properly report/update the ones that succeeded/failed accordingly.
- Changed the storage type for attributes and metrics to be `Blob` instead of encoded String, in order to keep backwards compatibility for migrating customers that might have data in their local db (the legacy SDK used Blob)
- Changed the result we report on `dispatchFlushEvents` to be a `[AnalyticsEvent]` to avoid exposing internal implementations and to allow consumers to cast the result (`PinpointEvent` is an internal class)

---
Finally, some small tweaks:
- Renamed the `Date.utcTimeMillis` extension method to the much better `Date.millisecondsSince1970` name, which is also consistent with the existing `timeintervalSince1970`.
- Decided to have a `Date` instead of `Date.Millisecond` in the `PinpointEvent` class. What matters is what we store in the DB, so it was more friendly to directly work with dates.

----

- [X] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Analytics/EventRecorder.swift
+++ b/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Analytics/EventRecorder.swift
@@ -99,7 +99,7 @@ class EventRecorder: AnalyticsEventRecording {
         do {
             let response = try await pinpointClient.putEvents(input: putEventsInput)
             guard let results = response.eventsResponse?.results else {
-                log.error("Unexpected response from server when atempting to submit events.")
+                log.error("Unexpected response from server when attempting to submit events.")
                 return
             }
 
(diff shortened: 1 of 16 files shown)


AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Analytics/EventRecorder.swift
```diff
@@ -69,30 +71,76 @@ class EventRecorder: AnalyticsEventRecording {
...
+                    // Mark event as retryable
+                    log.warn("Unable to successfully deliver event with id \(eventId) to the server. It will be updated with retry count += 1.")
+                    try storage.incrementEventRetry(eventId: eventId)
+                }
+            }
+
+        } catch {
+            // This means all events were rejected, so we will update them all in the local storage accodingly
+            var isRetryable = false
+            if let sdkError = error as? SdkError<Any> {
```
diegocstn, line 138: this might not work as expected :)

diegocstn: yeah, I think the cast to `SdkError<Any>` will most likely fail

diegocstn (approved):

PR #1986, opened 2022-07-08
chore(api): clean up and enable user pool integ test

lawmicha:
*Issue #, if available:*

*Description of changes:*
Overall this enables one test class called **GraphQLWithUserPoolIntegrationTests.swift** and makes sure the tests are passing based on the latest provisioning steps. 

- [ ] All unit tests pass

### DIFF
--- a/.github/workflows/integ_test.yml
+++ b/.github/workflows/integ_test.yml
@@ -1,7 +1,7 @@
 name: Integration Tests
 on:
   push:
-    branches: [main]
+    branches: [chore/api-userpool-integ-tests]
 
 permissions:
     id-token: write
@@ -112,7 +112,6 @@ jobs:
           scheme: GraphQLWithIAMIntegrationTests
 
   api-graphqluserpool-integration-test:
-    if: ${{ false }}
     needs: [api-integration-test]
     runs-on: macos-latest
     environment: IntegrationTest
--- a/.github/workflows/integ_test.yml
+++ b/.github/workflows/integ_test.yml
@@ -1,7 +1,7 @@
 name: Integration Tests
 on:
   push:
-    branches: [chore/api-userpool-integ-tests]
+    branches: [main]
 
 permissions:
     id-token: write
--- a/AmplifyPlugins/API/APICategoryPlugin.xcodeproj/xcshareddata/xcschemes/GraphQLWithUserPoolIntegrationTests.xcscheme
+++ b/AmplifyPlugins/API/APICategoryPlugin.xcodeproj/xcshareddata/xcschemes/GraphQLWithUserPoolIntegrationTests.xcscheme
@@ -24,7 +24,7 @@
             </BuildableReference>
             <SkippedTests>
                <Test
-                  Identifier = "GraphQLWithUserPoolIntegrationTests/testSetUpOnce()">
+                  Identifier = "GraphQLAuthDirectiveIntegrationTests">
                </Test>
             </SkippedTests>
          </TestableReference>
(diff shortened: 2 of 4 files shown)


AmplifyPlugins/API/APICategoryPlugin.xcodeproj/xcshareddata/xcschemes/GraphQLWithUserPoolIntegrationTests.xcscheme
```diff
@@ -24,7 +24,7 @@
             </BuildableReference>
             <SkippedTests>
                <Test
-                  Identifier = "GraphQLWithUserPoolIntegrationTests/testSetUpOnce()">
+                  Identifier = "GraphQLAuthDirectiveIntegrationTests">
```
lawmicha, line 27: Within the UserPool integ test target, we're enabling just one of the two backends for testing, which is why GraphQLAuthDirectiveIntegrationTests is disabled here. It uses different set of backend and auth rules and will need some significant refactor such as 
- If continue to use V1, then it should be fairly straight foward to enable, then we might as well consolidate it to one backend with two models.  By consolidating it, we can remove section in the readme (in GraphQLWithUserPoolIntegrationTests/AuthDirective/README.md) that specifics a credentials file with username/password since the first backend adds presign-up hooks to confirm the user, and has runtime methods to register and sign in users.
- If migrating to V2, the operations on the auth rule is deny by default so update it to include "read" operation for the tests to pass, or update the tests.

considering these changes, will keep it disbabled while enabling one more integration test in the pipeline

.github/workflows/integ_test.yml
```diff
@@ -112,7 +112,6 @@ jobs:
           scheme: GraphQLWithIAMIntegrationTests
 
   api-graphqluserpool-integration-test:
-    if: ${{ false }}
```
lawmicha, line 115: removing this expects the test to pass and it's test result contributes to the overall success of the integration test workflow

.github/workflows/integ_test.yml
```diff
@@ -1,7 +1,7 @@
 name: Integration Tests
 on:
   push:
-    branches: [main]
+    branches: [chore/api-userpool-integ-tests]
```
lawmicha, line 4: this is for running the workflow in this PR and will be reverted before merging to `main`

AmplifyPlugins/API/AWSAPICategoryPluginIntegrationTests/GraphQL/GraphQLWithUserPoolIntegrationTests/GraphQLWithUserPoolIntegrationTests.swift
```diff
@@ -169,10 +169,10 @@ class GraphQLWithUserPoolIntegrationTests: XCTestCase {
         let completeInvoked = expectation(description: "request completed")
         let uuid = UUID().uuidString
 
-        // create a Todo mutation with a missing/invalid "description" variable value
+        // create a Todo mutation with a missing/invalid "name" variable value
```
lawmicha, line 172: the default Todo schema when generating the API has required name and optional description. This wasn't the case before.. I found it easier to maintain the steps in the provisioning as is (use default Todo schema, while displaying which schema was used), so the test code now needs to be updated to accomdate the new schema.

AmplifyPlugins/API/AWSAPICategoryPluginTestCommon/Model/Todo.swift
```diff
@@ -8,14 +8,11 @@
...
 
-/// TODO: Replace with AmplifyTestCommon's DataStore's geneated model
-/// This model corresponds to the resources created for the Todo graphQL endpoint. As a developer, this is
-/// hand-written from the codegen that created the API.swift that depends on AppSync sdk.
 struct Todo: Decodable {
    let typename: String
    let id: String
    let name: String
-   let description: String
+   let description: String?
```
lawmicha, line 15: this is because the default schema changed, and the decision made here is to go with the new default schema so the test code needs to be updated

drochetti (approved):

drochetti (approved):

PR #1989, opened 2022-07-08
fix(datastore): Emit ModelSyncedEvent on sync finished event

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/1864

*Description of changes:*
This PR assumes out of order events are received by the two streams of events, one from the sync operation and the other from the reconciliation events.

The sync operation updates "recordsReceived" while the reconciliation events update "recordsReconciled". If the last supposedly reconciliation event comes in before the sync operation sends the `.finished` event then in the previous code it would never emit the modelSynced event because `initialSyncOperationFinished` is set to true and it is checked within one of the reconciliation events.

This code adds another special case along with the "received records == 0" by checking that if this event comes after the reconciliation event and number of records received have been reconciled, then emit the modelSyncedEvent. 

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ModelSyncedEventEmitter.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ModelSyncedEventEmitter.swift
@@ -26,9 +26,7 @@ enum IncomingModelSyncedEmitterEvent {
 ///     - Then send the mutation event which was used in the check above.
 @available(iOS 13.0, *)
 final class ModelSyncedEventEmitter {
-    private let queue = DispatchQueue(label: "com.amazonaws.ModelSyncedEventEmitterQueue",
-                                      target: DispatchQueue.global())
-    private let dispatchedModelSyncedEventLock = NSLock()
+    private let queue = DispatchQueue(label: "com.amazonaws.ModelSyncedEventEmitterQueue")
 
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/InitialSync/ModelSyncedEventEmitter.swift
```diff
@@ -116,9 +116,10 @@ final class ModelSyncedEventEmitter {
         case .enqueued:
             recordsReceived += 1
         case .finished:
-            initialSyncOperationFinished = true
-            if recordsReceived == 0 {
+            if recordsReceived == 0 || recordsReceived == reconciledReceived {
                 sendModelSyncedEvent()
+            } else {
+                initialSyncOperationFinished = true
```
lawmicha, line 122: to move this PR out of draft, let's add a unit test that mimics the behavior of the ordering of sync and reconcile events like so:

- sync record received (records received = 1)
- reconcile event applied (records reconciled = 1)
- sync record finished (causes the model synced event to be emitted)

We can also run this unit test against the previous code to see that it would fail to emit the model synced event.

lawmicha:
The criticial change was checked in here: https://github.com/aws-amplify/amplify-ios/pull/1752/files#diff-dfcf37a04a86ece2e164fb480b683831a11411a5aae6185d7727419e35c24417 and merged to `main` with the CPK feature branch

PR #1994, opened 2022-07-12
feat(api): AppSyncRTC 2.0.0 upgrade - Handle unauthorized connection

lawmicha:
*Issue #, if available:*

*Description of changes:*
This PR upgrades AppSyncRTC to 2.0.0 (release notes: https://github.com/aws-amplify/aws-appsync-realtime-client-ios/releases/tag/2.0.0) which contains a breaking change adding new enum cases to ConnectionProviderError. It it did not impact APIPlugin's usage of AppSyncRTC since it never checked for any other `ConnectionProviderError` other than `.subscription`. ConnectionProviderError returns unauthorized in two different ways, on the connection it will return `.unauthorized` and on the subscription of the connection, it will return `.subscription(payload)` where payload contains unauthorized error. 

This PR checks for the newly added `ConncetionProviderError.unauthorized` and stores that information without changing the behavior of the error scenario by continuing to return `APIError.operationError`. We introduce a helper method called `APIError.unauthorized()` to consoliate code paths as an unbreaking change.

- [ ] All unit tests pass


AmplifyPlugins/API/AWSAPICategoryPlugin/APIError+Extension.swift
```diff
@@ -0,0 +1,33 @@
+//
```
drochetti, line 1: I generally avoid using `+Extension` in filename because the `Type+` pattern is supposed to described what's being added to that type through the extension (the `+` already implies extension).

So maybe a `APIError+Unauthorized.swift` is a better fit here?

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
```diff
@@ -174,18 +174,41 @@ final public class AWSGraphQLSubscriptionOperation<R: Decodable>: GraphQLSubscri
...
             return
+        } else if case ConnectionProviderError.unauthorized = error {
+            errorDescription += ": \(APIError.UnauthorizedMessageString)"
         }
 
         dispatch(result: .failure(APIError.operationError(errorDescription, "", error)))
         finish()
     }
 
+    private static func anyGraphQLErrorIsUnauthorized(_ graphQLErrors: [GraphQLError]) -> Bool {
```
drochetti, line 198: not critical, but another option could be to make this an extension of `Array where Element == GraphQLError` so you could `errors.hasUnauthorizedError()`

AmplifyPlugins/API/AWSAPICategoryPlugin/APIError+Extension.swift
```diff
@@ -0,0 +1,33 @@
...
+    public func isUnauthorized() -> Bool {
+        if case .operationError(let errorDescription, _, _) = self,
+           errorDescription.range(of: APIError.UnauthorizedMessageString, options: .caseInsensitive) != nil {
+            return true
+        } else if case .httpStatusError(let statusCode, _) = self, statusCode == 403 {
+            return true
+        }
+
+        return false
+    }
```
atierian: I see that you call out in the code documentation that this logic may not match customized responses. Can we try to allow custom rules without sacrificing DX?
e.g.
```suggestion
       private var is403: Bool {
        switch self {
        case .httpStatusError(let statusCode, _):
            return statusCode == 403
        default:
            return false
        }
    }
    
    private var containsUnauthorizedMessageString: Bool {
        switch self {
        case .operationError(let errorDescription, _, _):
            return errorDescription.range(of: "Unauthorized", options: .caseInsensitive) != nil
        default: return false
        }
    }
    
    public struct UnauthorizedDetermining {
        let isUnauthenticated: (APIError) -> Bool
        
        public init(isUnauthenticated: @escaping (APIError) -> Bool) {
            self.isUnauthenticated = isUnauthenticated
        }
        
        public static let `default` = UnauthorizedDetermining { error in
            error.containsUnauthorizedMessageString || error.is403
        }
    }
    
    public func isUnauthorized(rule: UnauthorizedDetermining = .default) -> Bool {
        rule.isUnauthenticated(self)
    }
```

lawmicha: Awesome addition of `UnauthorizedDetermining`, thanks for the code snippet, I incorporated this into the latest commit

AmplifyPlugins/API/AWSAPICategoryPlugin/APIError+Unauthorized.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// ```
+    ///
+    /// - Parameter rule: Used to determine if the `APIError` is an unauthorized error.
+    /// - Returns: `true` if unauthorized error, `false` otherwise
+    public func isUnauthorized(rule: UnauthorizedDetermining = .default) -> Bool {
+        rule.isUnauthenticated(self)
+    }
+
+    public struct UnauthorizedDetermining {
+        let isUnauthenticated: (APIError) -> Bool
```
atierian, line 36: I wrote it inconsistently in suggestion snippet, my bad.
So that needs to be changed here and where ever it's being used.
```suggestion
        let isUnauthorized: (APIError) -> Bool
```

atierian (commented):
LGTM besides the one comment.

PR #2021, opened 2022-07-18
chore(Analytics): migrate key-value Userdefaults to Keychain

phantumcode:
*Issue #, if available:*

*Description of changes:*
* add logic to migrate Analytics key-value store using Userdefaults to Keychain
* added unit tests to verify data migration

- [x] All unit tests pass


AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
```diff
@@ -225,6 +244,47 @@ actor EndpointClient: EndpointClientBehaviour {
     nonisolated private func getOptOut(from endpointProfile: PinpointEndpointProfile) -> String {
         return endpointProfile.isOptOut ? EndpointClient.Constants.OptOut.all : EndpointClient.Constants.OptOut.none
     }
+    
+    private static func migrateLegacyKeyValueStore(_ userDefaults: UserDefaultsBehaviour, _ keychain: KeychainStoreBehavior, _ archiver: AmplifyArchiverBehaviour) {
```
ruisebas, line 248: Consider renaming the function and having named arguments, otherwise you can't tell what's going on by reading the signature (`migrateLegacyKeyValueStore(:::)`)

E.g:
```
private static func migrateStoredValues(from userDefaults: UserDefaultsBehaviour, 
                                        to keychain: KeychainStoreBehavior, 
                                        using archiver: AmplifyArchiverBehaviour)
```

This way is clear that you're migrating **storedValues** _from_ UserDefaults _to_ the Keychain, and _using_ the Archiver somewhere along the way

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
```diff
@@ -47,18 +52,24 @@ actor EndpointClient: EndpointClientBehaviour {
...
+    ) {
         self.configuration = configuration
         self.pinpointClient = pinpointClient
         self.archiver = archiver
         self.currentDevice = currentDevice
         self.userDefaults = userDefaults
+        self.keychain = keychain
 
-        if let attributes = userDefaults.object(forKey: Constants.attributesKey) as? [String: [String]] {
+        EndpointClient.migrateLegacyKeyValueStore(userDefaults, keychain, archiver)
```
ruisebas, line 65: nit: You can use `Self.` instead of `EndpointClient.` and save some characters :P

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
```diff
@@ -168,7 +187,7 @@ actor EndpointClient: EndpointClientBehaviour {
         do {
             let output = try await pinpointClient.updateEndpoint(input: input)
             if let encodedData = try? archiver.encode(endpointProfile) {
-                userDefaults.save(encodedData, forKey: Constants.endpointProfileKey)
+                try keychain.set(encodedData, key: Constants.endpointProfileKey)
```
ruisebas, line 190: If this operation fails, the error will propagate to the `catch` block and be rethrown, which is not desired as the `updateEndpoint` operation was actually successful on line 188. 

Either use `try?` instead or create another `catch` block that only handles this situation and does not re-throw the error.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
```diff
@@ -124,25 +147,21 @@ actor EndpointClient: EndpointClientBehaviour {
...
-        // 2. Look for a valid PinpointEndpointProfile object stored in UserDefaults. It needs to match the current applicationId, otherwise we'll discard it.
-        if let endpointProfileData = userDefaults.data(forKey: Constants.endpointProfileKey),
+        // 2. Look for a valid PinpointEndpointProfile object stored in the Keychain. It needs to match the current applicationId, otherwise we'll discard it.
+        if let endpointProfileData = try? keychain.getData(Constants.endpointProfileKey),
            let decodedEndpointProfile = try? archiver.decode(PinpointEndpointProfile.self, from: endpointProfileData),
            decodedEndpointProfile.applicationId == configuration.appId {
             return await configure(endpointProfile: decodedEndpointProfile)
         }
 
-        userDefaults.removeObject(forKey: Constants.endpointProfileKey)
```
ruisebas, line 134: You should remove the stored profile from the Keychain here as well, as reaching this line  means it doesn't match the current `applicationId`, so it needs to be discarded.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
```diff
@@ -225,6 +244,47 @@ actor EndpointClient: EndpointClientBehaviour {
...
         return endpointProfile.isOptOut ? EndpointClient.Constants.OptOut.all : EndpointClient.Constants.OptOut.none
     }
+    
+    private static func migrateLegacyKeyValueStore(_ userDefaults: UserDefaultsBehaviour, _ keychain: KeychainStoreBehavior, _ archiver: AmplifyArchiverBehaviour) {
+        if let endpointProfileData = userDefaults.data(forKey: Constants.endpointProfileKey) {
+            do {
+                try keychain.set(endpointProfileData, key: Constants.endpointProfileKey)
+                userDefaults.removeObject(forKey: Constants.endpointProfileKey)
+            } catch {
+                log.error("Unable to migrate Analytics key-value store for key \(Constants.endpointProfileKey)")
```
ruisebas, line 254: I've just realized something: We are attempting to migrate everything to the Keychain on this function and we are removing any existing data from UserDefaults while doing so. 

However, if the `keychain.set` operations fail, we keep the data on UserDefaults (which is good), but then we only attempt to retrieve their values from the keychain (see lines 66, 71. 151 and 164).

This basically means that if we have stored values in UserDefaults and the migration to the Keychain fails, we will behave as if we didn't have any stored value and we might even end up with two different values in each storage after a while

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPluginTests/Pinpoint/EndpointClientTests.swift
```diff
@@ -17,20 +17,13 @@ class EndpointClientTests: XCTestCase {
...
     private var currentApplicationId = "applicationId"
     private var currentEndpointId = "endpointId"
-
+    private var keychain: MockKeychainStore!
     override func setUp() {
         archiver = MockArchiver()
         userDefaults = MockUserDefaults()
         pinpointClient = MockPinpointClient()
         device = MockDevice()
-        endpointClient = EndpointClient(configuration: .init(appId: currentApplicationId,
```
ruisebas, line 26: Why was this removed from the `setUp` and added to all tests instead?

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
```diff
@@ -47,18 +52,24 @@ actor EndpointClient: EndpointClientBehaviour {
          pinpointClient: PinpointClientProtocol,
          archiver: AmplifyArchiverBehaviour = AmplifyArchiver(),
          currentDevice: Device = DeviceProvider.current,
-         userDefaults: UserDefaultsBehaviour = UserDefaults.standard) {
+         userDefaults: UserDefaultsBehaviour = UserDefaults.standard,
+         keychain: KeychainStoreBehavior = KeychainStore()
```
ruisebas, line 56: Don't forget to pass the right instance when this client is created in [`PinpointContext`](https://github.com/aws-amplify/amplify-ios/blob/dev-preview/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift#L149).

Also, you might wanna add the proper `service` key to the default value:
```
KeychainStore(service: PinpointContext.Constants.Keychain.service)
```

ruisebas (changes requested):

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
phantumcode: Updated per suggestion.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
phantumcode: Updated to use `Self.`

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
phantumcode: Updated to use `try?`

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
phantumcode: That's a very good point.  I've added static getter methods that will fallback to Userdefault lookup if keychain is empty.   I've left the implementation in `EndpointClient` since it's only a couple static functions.  I've also made similar updates to the AWS Pinpoint framework in aws-ios-sdk, please also review the [PR 4223](https://github.com/aws-amplify/aws-sdk-ios/pull/4223) with the latest updates.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
phantumcode: Updated constructor and also specify the service.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
```diff
@@ -285,6 +285,42 @@ actor EndpointClient: EndpointClientBehaviour {
...
+        fallbackTo defaultSource: UserDefaultsBehaviour
+    ) -> Data? {
+        if let data = try? keychain.getData(key) {
+            return data
+        } else {
+            return defaultSource.data(forKey: key)
+        }
+    }
+    
+    private static func getStoredAttributes(
```
ruisebas, line 301: nit: `getStoredAttributes` and `getStoredMetrics` are pretty much the same thing, just with different keys and return types. 

You can use a generic version instead:

```
private static func get<T: Decodable>(
    key: String,
    as: T.Type,
    from keychain: KeychainStoreBehavior,
    fallbackTo defaultSource: UserDefaultsBehaviour,
    using archiver: AmplifyArchiverBehaviour
) -> T? {
    guard let data = try? keychain.getData(key) else {
        return defaultSource.object(forKey: key) as? T
    }

    return try? archiver.decode(T.self, from: data)
}
```
---
Also, if you don't wanna be passing around 3 storage arguments each time on these retrieval functions, you can follow what we did on `PinpointContext` [and have a `struct` that groups all of them](https://github.com/aws-amplify/amplify-ios/blob/dev-preview/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift#L111-L116), so you can just [store that as a property](https://github.com/aws-amplify/amplify-ios/blob/dev-preview/AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift#L134-L137) and pass it around like 
```
let token = Self.getData(forKey: Constants.deviceTokenKey, from: storage)
``` 

But these are suggestions, up to you :)

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
```diff
@@ -148,10 +146,12 @@ actor EndpointClient: EndpointClientBehaviour {
         }
 
         // 2. Look for a valid PinpointEndpointProfile object stored in the Keychain. It needs to match the current applicationId, otherwise we'll discard it.
-        if let endpointProfileData = try? keychain.getData(Constants.endpointProfileKey),
-           let decodedEndpointProfile = try? archiver.decode(PinpointEndpointProfile.self, from: endpointProfileData),
+        if let endpointProfileData = Self.getStoredData(from: keychain, forKey: Constants.endpointProfileKey, fallbackTo: userDefaults),
+            let decodedEndpointProfile = try? archiver.decode(PinpointEndpointProfile.self, from: endpointProfileData),
```
ruisebas, line 150: nit: the indentation went wrong here

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
```diff
@@ -148,10 +146,12 @@ actor EndpointClient: EndpointClientBehaviour {
...
            decodedEndpointProfile.applicationId == configuration.appId {
             return await configure(endpointProfile: decodedEndpointProfile)
+        } else {
```
ruisebas, line 153: **very** nit*ty*: Having the removal inside an `else` clause it not really needed, as the `return` on the happy path will break the execution.

ruisebas (changes requested):

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPluginTests/Pinpoint/EndpointClientTests.swift
ruisebas: Oh got it! 

If you face a similar situation in the future, there is an `async` version of the `setUp` function as well for tests, so you can just override that one if you need to await for other methods :)

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
phantumcode: removed the `else` clause

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
phantumcode: Should be fixed!

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Endpoint/EndpointClient.swift
phantumcode: I combined the two functions to be generic and leaving the arguments as separate for now to keep to keep the argument names clear instead of combining them into a re-usable struct.

AmplifyPlugins/Analytics/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/PinpointContext.swift
```diff
@@ -150,7 +150,8 @@ class PinpointContext {
                                                              isOptOut: configuration.isApplicationLevelOptOut),
                                         pinpointClient: targetingPinpointClient,
                                         currentDevice: currentDevice,
-                                        userDefaults: userDefaults)
+                                        userDefaults: userDefaults,
+                                        keychain: KeychainStore(service: Constants.Keychain.service))
```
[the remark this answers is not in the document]
phantumcode: This should be fixed.

ruisebas (approved):

ruisebas (approved):

PR #2046, opened 2022-07-27
feat(datastore): Remove synchronous load from List

lawmicha:
*Issue #, if available:*

*Description of changes:*
This removes the behavior of implicitly lazy loading elements from the list when accessed, and requires explicit `fetch` call to be made access the elements.

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/API/Tests/AWSAPIPluginTests/Core/AppSyncListProviderTests.swift
+++ b/AmplifyPlugins/API/Tests/AWSAPIPluginTests/Core/AppSyncListProviderTests.swift
@@ -105,18 +105,22 @@ class AppSyncListProviderTests: XCTestCase {
         XCTAssertEqual(associatedField, "post")
     }
 
-    func testLoadedStateSynchronousLoadSuccess() {
+    func testLoadedStateLoadSuccess() {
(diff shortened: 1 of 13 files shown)


AmplifyPlugins/API/Tests/AWSAPIPluginTests/Core/AppSyncListProviderTests.swift
```diff
@@ -105,18 +105,22 @@ class AppSyncListProviderTests: XCTestCase {
...
+        listProvider.load { result in
+            guard case .success(let posts) = result else {
+                XCTFail("Should be .success")
+                return
+            }
+            XCTAssertEqual(posts.count, 2)
+            loadCompleted.fulfill()
         }
-        XCTAssertEqual(posts.count, 2)
+        wait(for: [loadCompleted], timeout: 1)
```
ameter, line 120: This is fine in these tests because they are synchronous, but note that in async tests, you need to use `await waitForExpectations(...)` as it has been updated to work with Swift Concurrency, whereas `wait(...)` will block the thread.

ameter (approved):

PR #2067, opened 2022-08-02
fix(datastore): String conform to ModelIdentifierProtocol

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/pull/1752#discussion_r917123859

*Description of changes:*
When compiling Amplify DataStore's category APIs using Xcode 14 Beta 4, for an API like
```swift
func query<M: Model>(_ modelType: M.Type,
                         byIdentifier id: String,
                         completion: DataStoreCallback<M?>) where M: ModelIdentifiable,
                                                                  M.IdentifierFormat == ModelIdentifierFormat.Default
```

the error shown is
image

There are two `Identifier`, where `M: Model` and `Model` contains `Identifier`:
image
and where `M: ModelIdentifiable`, and `ModelIdentifiable` contains `Identifer`:
image

## Solution 

This PR adds a default conformance of String to ModelIdentifierProtocol. By conforming to `ModelIdentifierProtocol`, unfortunately makes all String values be able to return "fields" as `(name: "id", value: self)`. However, Model.Identifier is deprecated and can be removed in the future.

This change was tested on Xcode 14 beta 4 and back on Xcode 13.4.1 for any regressions

## Alternative 1 
Changing Model's Identifier does not work because that would be a breaking change to the public API of Model.

## Alternative 2

Changing ModelIdentifiable's associated type `Identifier` to something like `IdentifierProtocol` 
```swift
/// Defines requirements for a model to be identifiable with a unique identifier
/// that can be either a single field or a combination of fields
public protocol ModelIdentifiable {
    associatedtype IdentifierFormat: AnyModelIdentifierFormat
    // associatedtype Identifier: ModelIdentifierProtocol
   associatedtype IdentifierProtocol: ModelIdentifierProtocol
}
```
will also not be ideal since that will require codegen changes, and impacts the DX of public APis
```swift
extension PostWithCompositeKey: ModelIdentifiable {
  public typealias IdentifierFormat = ModelIdentifierFormat.Custom
  //public typealias Identifier = ModelIdentifier<Self, ModelIdentifierFormat.Custom> // previously 
   public typealias IdentifierProtocol = ModelIdentifier<Self, ModelIdentifierFormat.Custom> // new
}

//extension PostWithCompositeKey.Identifier { // previously
extension PostWithCompositeKey.IdentifierProtocol { // alternative
  public static func identifier(id: String,
      title: String) -> Self {
    .make(fields: [(name: "id", value: id), (name: "title", value: title)])
  }
}
```

Call site in alternative 2 (no ideal DX):
```
// Previously
PostWithCompositeKey.Identifier.identifier(id: "123", title: "title")

// New
PostWithCompositeKey.IdentifierProtocol.identifier(id: "123", title: "title")
```

- [ ] All unit tests pass

### DIFF
--- a/Amplify/Categories/DataStore/Model/ModelIdentifiable.swift
+++ b/Amplify/Categories/DataStore/Model/ModelIdentifiable.swift
@@ -53,6 +53,15 @@ public protocol ModelIdentifierProtocol {
     var predicate: QueryPredicate { get }
 }
 
+extension Model.Identifier: ModelIdentifierProtocol {
+   public var fields: Fields {
+       [(name: ModelIdentifierFormat.Default.name, value: self)]
+   }
+   public var stringValue: String {
+       self
+   }
+}
+
 public extension ModelIdentifierProtocol {
     var stringValue: String {
         if fields.count == 1, let field = fields.first {


drochetti:
Regarding the call-site example:

```swift
// Previously
PostWithCompositeKey.Identifier.identifier(id: "123", title: "title")

// New
PostWithCompositeKey.IdentifierProtocol.identifier(id: "123", title: "title")
```

I believe the suggested DX is to always use the shortcut call, like `.identifier(id: "123", title: "title")`, isn't it?

drochetti (approved):

lawmicha:
> Regarding the call-site example:
> 
> ```swift
> // Previously
> PostWithCompositeKey.Identifier.identifier(id: "123", title: "title")
> 
> // New
> PostWithCompositeKey.IdentifierProtocol.identifier(id: "123", title: "title")
> ```
> 
> I believe the suggested DX is to always use the shortcut call, like `.identifier(id: "123", title: "title")`, isn't it?

I think that is true for most cases when using the query/delete APIs that take in the strongly type argument, it can infer the type to be `PostWithCompositeKey.Identifier` / `PostWithCompositeKey.IdentifierProtocol`. I believe there are times when the identifer is used as the value of the `where: QueryPredicate` where it is not so easy to infer the type and requires the developer to type out the entire thing

PR #2084, opened 2022-08-04
Implement Task.sleep with seconds

brennanMKE:
*Issue #, if available:*

*Description of changes:*

Task has a sleep function which uses nanoseconds. A double value for seconds is easier to use.

- [ ] All unit tests pass

### DIFF
--- a/Amplify/Core/Support/Task+Seconds.swift
+++ b/Amplify/Core/Support/Task+Seconds.swift
@@ -1,0 +1,15 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+extension Task where Success == Never, Failure == Never {
+    static func sleep(seconds: Double) async throws {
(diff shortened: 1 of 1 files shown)


Amplify/Core/Support/Task+Seconds.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+extension Task where Success == Never, Failure == Never {
+    static func sleep(seconds: Double) async throws {
+        let nanoseconds = UInt64(seconds * Double(NSEC_PER_SEC))
+        try await Task.sleep(nanoseconds: nanoseconds)
```
ameter, line 13:
```suggestion
        try await Task.sleep(nanoseconds: NSEC_PER_SEC * seconds)
```
Can be simplified. :-)

ameter (approved):
Approved with comment.

Amplify/Core/Support/Task+Seconds.swift
brennanMKE: Discussed off GH. The `seconds` are `Double` so `UInt64` is required and nesting the work in that parameter quickly becomes less readable.

PR #2086, opened 2022-08-04
adds AsyncChannel and AsyncThrowingChannel with tests

brennanMKE:
*Issue #, if available:*

*Description of changes:*

adds AsyncChannel and AsyncThrowingChannel with tests

- [ ] All unit tests pass

### DIFF
--- a/Amplify/Core/Support/AsyncChannel.swift
+++ b/Amplify/Core/Support/AsyncChannel.swift
@@ -7,23 +7,23 @@
 
 import Foundation
 
-public actor AsyncChannel<Element: Sendable>: AsyncSequence {
-    public struct Iterator: AsyncIteratorProtocol, Sendable {
+actor AsyncChannel<Element: Sendable>: AsyncSequence {
+    struct Iterator: AsyncIteratorProtocol, Sendable {
         private let channel: AsyncChannel<Element>
 
-        public init(_ channel: AsyncChannel<Element>) {
+        init(_ channel: AsyncChannel<Element>) {
             self.channel = channel
         }
 
-        public mutating func next() async -> Element? {
+        mutating func next() async -> Element? {
             await channel.next()
         }
     }
 
-    public enum InternalFailure: Error {
+    enum InternalFailure: Error {
         case cannotSendAfterTerminated
     }
-    public typealias ChannelContinuation = CheckedContinuation<Element?, Never>
+    typealias ChannelContinuation = CheckedContinuation<Element?, Never>
 
     private var continuations: [ChannelContinuation] = []
     private var elements: [Element] = []
@@ -37,21 +37,21 @@ public actor AsyncChannel<Element: Sendable>: AsyncSequence {
         terminated && elements.isEmpty && !continuations.isEmpty
     }
 
-    public init() {
+    init() {
     }
 
-    public nonisolated func makeAsyncIterator() -> Iterator {
+    nonisolated func makeAsyncIterator() -> Iterator {
         Iterator(self)
     }
 
-    public func next() async -> Element? {
+    func next() async -> Element? {
         await withCheckedContinuation { (continuation: ChannelContinuation) in
             continuations.append(continuation)
             processNext()
         }
     }
 
-    public func send(_ element: Element) throws {
+    func send(_ element: Element) throws {
         guard !terminated else {
             throw InternalFailure.cannotSendAfterTerminated
         }
@@ -59,7 +59,7 @@ public actor AsyncChannel<Element: Sendable>: AsyncSequence {
         processNext()
     }
 
-    public func finish() {
+    func finish() {
         terminated = true
         processNext()
     }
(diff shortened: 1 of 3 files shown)


Amplify/Core/Support/AsyncChannel.swift
```diff
@@ -0,0 +1,88 @@
...
+actor AsyncChannel<Element: Sendable>: AsyncSequence {
+    struct Iterator: AsyncIteratorProtocol, Sendable {
+        private let channel: AsyncChannel<Element>
+
+        init(_ channel: AsyncChannel<Element>) {
+            self.channel = channel
+        }
+
+        mutating func next() async -> Element? {
+            await channel.next()
```
royjit, line 19: Check for cancellation? https://developer.apple.com/documentation/swift/asynciteratorprotocol

Amplify/Core/Support/AsyncChannel.swift
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private var canTerminate: Bool {
+        terminated && elements.isEmpty && !continuations.isEmpty
+    }
+
+    init() {
+    }
+
+    nonisolated func makeAsyncIterator() -> Iterator {
```
royjit, line 43: Is this okay? Will the channel work if we iterate in two different task?

Amplify/Core/Support/AsyncChannel.swift
```diff
@@ -0,0 +1,88 @@
...
+    private var continuations: [ChannelContinuation] = []
+    private var elements: [Element] = []
+    private var terminated: Bool = false
+
+    private var hasNext: Bool {
+        !continuations.isEmpty && !elements.isEmpty
+    }
+
+    private var canTerminate: Bool {
+        terminated && elements.isEmpty && !continuations.isEmpty
```
royjit, line 37: Should these be OR cases?

Amplify/Core/Support/AsyncChannel.swift
```diff
@@ -0,0 +1,88 @@
...
+        guard !terminated else {
+            throw InternalFailure.cannotSendAfterTerminated
+        }
+        elements.append(element)
+        processNext()
+    }
+
+    func finish() {
+        terminated = true
+        processNext()
```
royjit, line 64: why are we calling processNext after finishing?

Amplify/Core/Support/AsyncChannel.swift
```diff
@@ -0,0 +1,88 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+actor AsyncChannel<Element: Sendable>: AsyncSequence {
```
royjit, line 10: nit: Call it `AmplifyAsyncChannel` ?

Amplify/Core/Support/AsyncChannel.swift
```diff
@@ -0,0 +1,88 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+public actor AsyncChannel<Element: Sendable>: AsyncSequence {
```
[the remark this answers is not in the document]
brennanMKE: done, though the `AsyncSequence` is a protocol with associated type. There is no way to represent it in Swift for a return value without the implementation

Amplify/Core/Support/AsyncChannel.swift
brennanMKE: this logic is correct

Amplify/Core/Support/AsyncChannel.swift
brennanMKE: What can be canceled here?

Amplify/Core/Support/AsyncChannel.swift
brennanMKE: this function cannot be isolated in the actor. it also does not mutate any state. It creates the iterator which is used by the sequence and all those functions are isolated within the actor.

Amplify/Core/Support/AsyncChannel.swift
brennanMKE: see the `processNext` function

Amplify/Core/Support/AsyncChannel.swift
brennanMKE: there is reason to specialize this name

royjit: I think if developer uses both Amplify and AsyncAlgorithms this names can conflict. We cannot use module name to distinguish between them because of the known issue in SPM where it cannot figure out the type `Amplify` vs the module `Amplify`. https://github.com/apple/swift/issues/43510

royjit (approved):
LGTM. This api can change based on api review.

PR #2089, opened 2022-08-05
fix(datastore): Remove DispatchSemaphore - RemoteSyncEngine.performInitialSync

lawmicha:
*Issue #, if available:*

*Description of changes:*
Previously the code would use a semaphore to block on returning from `performInitialSync()` but it doesn't matter if it returns since the call to `self.stateMachine.notify(action: .performedInitialSync)` is what transitioned the state to perform the next action. Since we are no longer blocking, `performInitialSync` will return, and the `stateMachine.notify` will cause it to move onto `activateCloudSubscriptions()`, which now sets `self.initialSyncOrchestrator = nil`.

### Alternatives discussion

The current PR appears to be a feasible solution but couples `activateCloudSubscriptions` with the previous action, it knows that it's the step after performInitialSync and sets `initialSyncOrchestrator` to nil. Was the semaphore there because the following would have no effect? 

```swift
initialSyncOrchestrator.sync { [weak self] result in
            defer {
                // semaphore.signal()
                self?.initialSyncOrchestrator = nil // would this work?
            }

            guard let self = self else {
                return
            }
            if case .failure(let dataStoreError) = result {
                self.log.error(dataStoreError.errorDescription)
                self.log.error(dataStoreError.recoverySuggestion)
                if let underlyingError = dataStoreError.underlyingError {
                    self.log.error("\(underlyingError)")
                }
                self.stateMachine.notify(action: .errored(dataStoreError))
            } else {
                self.log.info("Successfully finished sync")
                self.remoteSyncTopicPublisher.send(.performedInitialSync)
                self.stateMachine.notify(action: .performedInitialSync)
            }
        }
```

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/RemoteSyncEngine.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/RemoteSyncEngine.swift
@@ -342,12 +342,12 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
                 self.remoteSyncTopicPublisher.send(.performedInitialSync)
                 self.stateMachine.notify(action: .performedInitialSync)
             }
+            self.initialSyncOrchestrator = nil
         }
     }
 
     private func activateCloudSubscriptions() {
         log.debug(#function)
-        self.initialSyncOrchestrator = nil
         guard let reconciliationQueue = reconciliationQueue else {
             let error = DataStoreError.internalOperation("reconciliationQueue is unexpectedly `nil`", "", nil)
             stateMachine.notify(action: .errored(error))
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -350,13 +343,11 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
                 self.stateMachine.notify(action: .performedInitialSync)
             }
         }
```
ameter, line 345:
```suggestion

            self.initialSyncOrchestrator = nil
        }
```

ameter (commented):
Recommend setting self.initialSyncOrchestrator to nil at the end of the closure to allow it to be released.  See suggestion.

ameter (approved):

PR #2090, opened 2022-08-05
fix(datastore): Remove DispatchGroup - RemoteSyncEngine.isUserLoggedIn() async -> Bool

lawmicha:
*Issue #, if available:*

*Description of changes:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/RemoteSyncEngine+AuthModeStrategyDelegate.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/RemoteSyncEngine+AuthModeStrategyDelegate.swift
@@ -33,19 +33,6 @@ extension RemoteSyncEngine: AuthModeStrategyDelegate {
             return false
         }
 
-        // Note: blocking is not recommended
-        let group = DispatchGroup()
-        var isSignedIn = false
-
-        group.enter()
-        auth.getCurrentUser { result in
-            if case .success(let user) = result {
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/RemoteSyncEngine+AuthModeStrategyDelegate.swift
```diff
@@ -33,19 +33,6 @@ extension RemoteSyncEngine: AuthModeStrategyDelegate {
             return false
         }
 
-        // Note: blocking is not recommended
-        let group = DispatchGroup()
```
lawmicha, line 37: Could this have been a DispatchSemaphore instead of a DispatchGroup? It looks like we are only calling group.enter() once

ameter: Yes, it could have been.  With that said, I actually think a `DispatchGroup` was better here (and would have been better most places a semaphore was used in our codebase).  The `DispatchGroup` is basically for waiting for one or more threads to be done.  The `DispatchSemaphore` can be used for that as well (and frequently was within the Amplify codebase), but is more general and can also be used to allow a limited number of threads to continue to execute simultaneously.  Given that the intended behavior was the former, I would have used a `DispatchGroup`, but again, either would work.

ameter (approved):

PR #2113, opened 2022-08-08
chore(auth): Update credentials to store signedIn info

royjit:
*Issue #, if available:*

*Description of changes:*

- [ ] All unit tests pass


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/HostedUIProviderInfo.swift
```diff
@@ -28,8 +28,8 @@ extension HostedUIProviderInfo: Codable {
 
     init(from decoder: Decoder) throws {
         let values = try decoder.container(keyedBy: CodingKeys.self)
-        idpIdentifier = try values.decode(String.self, forKey: .idpIdentifier)
-        federationProviderName = try values.decode(String.self, forKey: .federationProviderName)
+        idpIdentifier = try? values.decode(String.self, forKey: .idpIdentifier)
+        federationProviderName = try? values.decode(String.self, forKey: .federationProviderName)
```
harsh62, line 32: what is the reason of this change?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/SignedInData.swift
```diff
@@ -15,16 +15,14 @@ struct SignedInData {
...
+    init(signedInDate: Date,
+         signInMethod: SignInMethod,
+         deviceMetadata: DeviceMetadata = .noData,
+         cognitoUserPoolTokens: AWSCognitoUserPoolTokens
     ) {
-        self.userId = userId
-        self.userName = userName
+        let user = try? TokenParserHelper.getAuthUser(accessToken: cognitoUserPoolTokens.accessToken)
+        self.userId = user?.userId ?? "unknown"
+        self.userName = user?.username ?? "unknown"
```
harsh62, line 25: This will eventually be passed back to the customer.. Are we okay passing `unknown` value back?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/HostedUIProviderInfo.swift
royjit: these values are optional, but decoding fails if we are expecting non-optional value. So to avoid it changed the try to optional.

harsh62: We should use `decodeIfPresent` in that case rather than `decode`.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/SignedInData.swift
royjit: As discussed this will be always passed back from Cognito. "Unknown" is just a fallback so that we do not need to crash the app, in case parsing fails.

harsh62 (approved):

PR #2128, opened 2022-08-11
Defines Async API primarily with AmplifyTask and Task Adapter for Operations

brennanMKE:
*Issue #, if available:*

*Description of changes:*

Amplify V2 will have an async API and encapsulate Operations as an internal implementation. `AmplifyTask` will define the public API for Tasks. Operations are given to `ChildTask` to provide async properties `value` and `inProcess`.

* adds AmplifyTask
* adds ChildTask
* revises AsyncChannel and AsyncThrowingChannel to use AsyncStream
* adds BufferingSequence to support AsyncStreams
* implements Task.sleep(seconds:)
* revises Combine support in Operations to work with AmplifyTask
* adds RequestIdentifier
* integrates RequestIdentifier with AmplifyTask
* adds AsyncPassthroughSubject to work with AmplifyTask
* adds AsyncExpectation to support testing

- [ ] All unit tests pass


AmplifyTests/CoreTests/Operations/FastOperation.swift
```diff
@@ -0,0 +1,95 @@
...
+            self = .unknown(errorDescription, error)
+        }
+    }
+}
+
+public typealias FastOperationResult = Result<FastOperationSuccess, FastOperationError>
+public typealias FastOperationResultListener = (FastOperationResult) -> Void
+
+public class FastOperation: AmplifyOperation<FastOperationRequest, FastOperationSuccess, FastOperationError> {
+    public typealias TaskAdapter = AmplifyOperation<Request, Success, Failure>
```
lawmicha, line 70: should this be typealias for `AmplifyOperationTaskAdapter` ?

brennanMKE: This is the mechanism which bridges to `AmplifyTask` by passing along the generic types of the Operation. When the instance of the adapter is created it passes in the actual `Operation`. 

```swift
    init(operation: AmplifyOperation<Request, Success, Failure>) {
        self.operation = operation
        self.childTask = ChildTask(parent: operation)
        resultToken = operation.subscribe(resultListener: resultListener)
    }
```

Then all of the stored properties which reference these types are defined for that instance.

```swift
class AmplifyOperationTaskAdapter<Request: AmplifyOperationRequest, Success, Failure: AmplifyError>: AmplifyTask {
    let operation: AmplifyOperation<Request, Success, Failure>
    let childTask: ChildTask<Void, Success, Failure>
```

Then `AmplifyTask` has the associated types defined for the protocol.

```swift
public protocol AmplifyTask {
    associatedtype Request
    associatedtype Success
    associatedtype Failure: AmplifyError
```

This typealias allows the one outside this type to inherit those type definitions so they do not have to be repeated. Below you can see it maps `FastOperation` to `FastTask` using the nested `TaskAdapter` definition.

```swift
public typealias FastTask = FastOperation.TaskAdapter
```

Without the nested typealias it would be necessary to repeat the list of generic types which is redundant and potentially easy to cause a mismatch. Nesting prevents these concerns.

Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift
```diff
@@ -0,0 +1,109 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+class AmplifyOperationTaskAdapter<Request: AmplifyOperationRequest, Success, Failure: AmplifyError>: AmplifyTask {
```
lawmicha, line 10:
```suggestion
public class AmplifyOperationTaskAdapter<Request: AmplifyOperationRequest, Success, Failure: AmplifyError>: AmplifyTask {
```

Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift
```diff
@@ -0,0 +1,109 @@
...
+    let operation: AmplifyOperation<Request, Success, Failure>
+    let childTask: ChildTask<Void, Success, Failure>
+    var resultToken: UnsubscribeToken? = nil
+
+    init(operation: AmplifyOperation<Request, Success, Failure>) {
```
lawmicha, line 15:
```suggestion
    public init(operation: AmplifyOperation<Request, Success, Failure>) {
```

Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+    var result: Success {
+        get async throws {
+            try await childTask.result
+        }
+    }
+
+    var progress: AsyncChannel<InProcess> {
```
lawmicha, line 80: discussed offline, to change to `inProcess`

lawmicha: define the "subscription" for async channels related to GraphQLSubscriptionTask, extend AmplifyInProcessReportingOperationTaskAdapter where it is a GraphQLSubscriptionTask

will remove this from APIPlugin:
```
public extension GraphQLSubscriptionTask {
    var subscription: AsyncChannel<InProcess> {
        get async {
            await progress
        }
    }
}
```

Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private func resultListener(_ result: Result<Success, Failure>) {
+        Task {
+            await childTask.finish(result)
+        }
+    }
+}
+
+class AmplifyInProcessReportingOperationTaskAdapter<Request: AmplifyOperationRequest, InProcess, Success, Failure: AmplifyError>: AmplifyTask, AmplifyInProcessReportingTask {
```
lawmicha, line 52:
```suggestion
public class AmplifyInProcessReportingOperationTaskAdapter<Request: AmplifyOperationRequest, InProcess, Success, Failure: AmplifyError>: AmplifyTask, AmplifyInProcessReportingTask {
```

AmplifyTests/CoreTests/Operations/FastOperation.swift
lawmicha:
```suggestion
    public typealias TaskAdapter = AmplifyOperationTaskAdapter <Request, Success, Failure>
```

Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift
lawmicha: the constructor needs to be public as well for the plugins to use

Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+public class AmplifyInProcessReportingOperationTaskAdapter<Request: AmplifyOperationRequest, InProcess, Success, Failure: AmplifyError>: AmplifyTask, AmplifyInProcessReportingTask {
+    let operation: AmplifyInProcessReportingOperation<Request, InProcess, Success, Failure>
+    let childTask: ChildTask<InProcess, Success, Failure>
+    var resultToken: UnsubscribeToken? = nil
+    var inProcessToken: UnsubscribeToken? = nil
+
+    init(operation: AmplifyInProcessReportingOperation<Request, InProcess, Success, Failure>) {
```
lawmicha, line 58: the constructor needs to be public as well for the plugins to use

lawmicha:
```suggestion
    public init(operation: AmplifyInProcessReportingOperation<Request, InProcess, Success, Failure>) {
```

Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift
lawmicha: i'll add this to Amplify when I add the APIPlugin APIs:
```
public extension MyTask {
    func subscription<R : Decodable>() async -> AsyncStream<InProcess>
        where InProcess == SubscriptionOperation<R>.InProcess {
            return await inProcess
    }
}

File: Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift (hunk 3 above)
Comment (brennanMKE) [reply Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift:15]: The instance is only created internally so the init should not have to be public.

Comment (lawmicha) [reply Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift:15]: In the plugin code, for example, https://github.com/aws-amplify/amplify-ios/pull/2140/files#diff-fa2f4789a9a71f094ae985b1d3b57e4c8b52caef6e7cf6fea0c6234652e803e9

```
func query<R: Decodable>(request: GraphQLRequest<R>) async throws -> GraphQLTask<R>.Success {
        let operation = AWSGraphQLOperation(request: request.toOperationRequest(operationType: .query),
                                            session: session,
                                            mapper: mapper,
                                            pluginConfig: pluginConfig,
                                            resultListener: nil)
        let task = AmplifyOperationTaskAdapter(operation: operation)
        queue.addOperation(operation)
        return try await task.result
    }
```

how should the plugins be creating the task after creating the operation?

File: Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift [hunk 7]
```diff
@@ -0,0 +1,109 @@
...
+        resultToken = operation.subscribe(resultListener: resultListener)
+    }
+
+    deinit {
+        if let resultToken = resultToken {
+            Amplify.Hub.removeListener(resultToken)
+        }
+    }
+
+    public var result: Success {
```
lawmicha, line 27: `value`

Amplify/Core/Support/AmplifyOperation.swift
```diff
@@ -154,6 +160,10 @@ open class AmplifyOperation<Request: AmplifyOperationRequest, Success, Failure:
         )
         publish(result: .failure(cancellationError))
 #endif
+        if let token = resultListenerUnsubscribeToken {
+            unsubscribe(token: token)
+            resultListenerUnsubscribeToken = nil
```
lawmicha, line 165: redundant `resultListenerUnsubscribeToken = nil` here, resultListenerUnsubscribeToken is being set to `nil` in `subscribe(token: token)` call

brennanMKE: This is the `unsubscribe` function so this will is intentional. It is called when an Operation is cancelled which could be due to the caused by the `ChildTask` being cancelled. The `subscribe` function which only will remove the listener after the result is received, but it does not set the stored property for `resultListenerUnsubscribeToken` to nil. The cancel function can be called before a result is received and should prevent the listener from being sent a value.  The nil state of `resultListenerUnsubscribeToken` is also used to indicate the listener was removed from the Hub for the `unsubscribe` and `cancel` functions.

AmplifyTestCommon/Helpers/Sequence+sum.swift
```diff
@@ -0,0 +1,21 @@
...
+    @inlinable func sum() -> Element {
+        reduce(0, +)
+    }
+}
+
+public extension Sequence where Element: UnsignedInteger {
+    @inlinable func sum() -> Element {
+        reduce(0, +)
+    }
+}
```
atierian, line 20:
```suggestion
public extension Sequence where Element: AdditiveArithmetic {
    func sum() -> Element {
        reduce(.zero, +)
    }
}
```

If this is really only for testing purposes, the `@inlinable` is unnecessary but it also doesn't hurt.

lawmicha:
> unit tests failing in the github actions workflow
> 
> ```
> ❌  /Users/runner/work/amplify-ios/amplify-ios/Amplify/Core/Support/ChildTask.swift:15:27: type 'InProcess' does not conform to protocol 'Sendable'
> 
>     var inProcessChannel: AsyncChannel<InProcess>? = nil
> ```

https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#xcode Xcode 13.2.1

lawmicha (dismissed):

Amplify/Core/Support/AmplifyOperation.swift
lawmicha: sorry i meant `unsubscribe(token)` calls `resultListenerUnsubscribeToken = nil` as well as this `cancel()` method calls `unsubscribe(token)` and `resultListenerUnsubscribeToken = nil`

AmplifyTestCommon/AsyncExpectation.swift
```diff
@@ -0,0 +1,130 @@
...
+
+        await withThrowingTaskGroup(of: Void.self) { group in
+            for exp in expectations {
+                group.addTask {
+                    try await exp.wait()
+                }
+            }
+        }
+
+        timeout.cancel()
```
ameter, line 92:
```suggestion
        let timeoutTask = Task {
            try await Task.sleep(seconds: timeout)
            for exp in expectations {
                await exp.timeOut(file: file, line: line)
            }
        }

        await withThrowingTaskGroup(of: Void.self) { group in
            for exp in expectations {
                group.addTask {
                    try await exp.wait()
                }
            }
        }

        timeoutTask.cancel()
```

Amplify/Core/Support/Task+Seconds.swift
```diff
@@ -7,7 +7,7 @@
 
 import Foundation
 
-extension Task where Success == Never, Failure == Never {
+public extension Task where Success == Never, Failure == Never {
```
ameter, line 10: should this be public given that we don't own this type?

ameter (approved):

lawmicha (approved):

Amplify/Core/Support/Task+Seconds.swift
lawmicha: https://github.com/aws-amplify/amplify-ios/issues/2154

lawmicha:
> > unit tests failing in the github actions workflow
> > ```
> > ❌  /Users/runner/work/amplify-ios/amplify-ios/Amplify/Core/Support/ChildTask.swift:15:27: type 'InProcess' does not conform to protocol 'Sendable'
> > 
> >     var inProcessChannel: AsyncChannel<InProcess>? = nil
> > ```
> 
> https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#xcode Xcode 13.2.1

https://github.com/aws-amplify/amplify-ios/issues/2155

PR #2149, opened 2022-08-15
feat: DataStore API subscribe observe observeQuery APIs

lawmicha:
*Issue #, if available:*

*Description of changes:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/API/Sources/AWSAPIPlugin/AWSAPIPlugin+GraphQLBehavior.swift
+++ b/AmplifyPlugins/API/Sources/AWSAPIPlugin/AWSAPIPlugin+GraphQLBehavior.swift
@@ -70,7 +70,7 @@ public extension AWSAPIPlugin {
             return operation
     }
     
-    func subscribe<R>(request: GraphQLRequest<R>) async throws -> GraphQLSubscriptionTask<R> {
+    func subscribe<R>(request: GraphQLRequest<R>) async -> AmplifyAsyncThrowingSequence<GraphQLResponse<R>> {
         let operation = AWSGraphQLSubscriptionOperation(
             request: request.toOperationRequest(operationType: .subscription),
             pluginConfig: pluginConfig,
@@ -81,6 +81,7 @@ public extension AWSAPIPlugin {
             resultListener: nil)
         let task = AmplifyInProcessReportingOperationTaskAdapter(operation: operation)
         queue.addOperation(operation)
-        return task
+        return Fatal.preconditionFailure("Need to return AmplifyAsyncThrowingSequence")
+        // return task
     }
 }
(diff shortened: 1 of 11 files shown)


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/AWSDataStorePlugin+DataStoreSubscribeBehavior.swift
```diff
@@ -63,4 +61,25 @@ extension AWSDataStorePlugin: DataStoreSubscribeBehavior {
...
+                                                          predicate: predicate,
+                                                          sortInput: sortInput?.asSortDescriptors(),
+                                                          storageEngine: storageEngine,
+                                                          dataStorePublisher: dataStorePublisher,
+                                                          dataStoreConfiguration: dataStoreConfiguration,
+                                                          dispatchedModelSyncedEvent: dispatchedModelSyncedEvent)
+        operationQueue.addOperation(operation)
+        let task = AmplifyInProcessReportingOperationTaskAdapter(operation: operation)
+        return Fatal.preconditionFailure("Need to return AmplifyAsyncThrowingSequence")
+        // return task.inProcess
```
lawmicha, line 83: Here `AWSDataStoreObserveQueryOperation` is created and passed to the current task adapter, which returns `AmplifyAsyncSequence`. It doesn't support returning an `AmplifyAsyncThrowingSequence`

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/AWSDataStorePlugin+DataStoreSubscribeBehavior.swift
```diff
@@ -25,6 +23,11 @@ extension AWSDataStorePlugin: DataStoreSubscribeBehavior {
         return publisher.filter { $0.modelName == modelName }.eraseToAnyPublisher()
     }
 
+    public func observe<M: Model>(for modelType: M.Type) async -> AmplifyAsyncThrowingSequence<MutationEvent> {
+        // There's `AWSDataStoreObserveOpertion` operation that can be removed
+        return Fatal.preconditionFailure("Need to return AmplifyAsyncThrowingSequence")
+    }
```
lawmicha, line 29: Here, we used to only return the Combine publisher from `publisher(modelType:)` API. In this case, we don't need to create a new operation, but thought it would work with the task adapter and returning `task.inProcess` from there, but that is not the case right now. So we can actually just delete `AWSDataStoreObserveOpertion` if it is not useful and make a provider out of the publisher, and pass the provider to the AmplifyAsyncThrowingSequence to be returned?

AmplifyPlugins/API/Sources/AWSAPIPlugin/AWSAPIPlugin+GraphQLBehavior.swift
```diff
@@ -70,7 +70,7 @@ public extension AWSAPIPlugin {
             return operation
     }
     
-    func subscribe<R>(request: GraphQLRequest<R>) async throws -> GraphQLSubscriptionTask<R> {
+    func subscribe<R>(request: GraphQLRequest<R>) async -> AmplifyAsyncThrowingSequence<GraphQLResponse<R>> {
         let operation = AWSGraphQLSubscriptionOperation(
```
lawmicha, line 74: For API.subscribe, it is also backed by the operation `AWSGraphQLSubscriptionOperation`. This is the same issue with `observeQuery` where the `taskAdapter(operation).inProcess` isn't the AmplifyAsyncThrowingSequence. The solution for DataStore.observeQuery and API.subscribe should be identical

lawmicha:
this PR is now outdated, latest PRs change these to AmplifyAsyncThrowingSequence

PR #2159, opened 2022-08-17
chore(auth): Changes the internal statemachine to structure code

royjit:
*Issue #, if available:*

*Description of changes:*

- [x] All unit tests pass

### DIFF
--- a/.swiftpm/xcode/xcshareddata/xcschemes/Amplify-Package.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/Amplify-Package.xcscheme
@@ -314,6 +314,20 @@
                ReferencedContainer = "container:">
             </BuildableReference>
          </BuildActionEntry>
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "YES"
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "HSM"
+               BuildableName = "HSM"
+               BlueprintName = "HSM"
+               ReferencedContainer = "container:">
+            </BuildableReference>
+         </BuildActionEntry>
       </BuildActionEntries>
    </BuildAction>
    <TestAction
@@ -424,6 +438,16 @@
                BlueprintIdentifier = "AWSPinpointAnalyticsPluginUnitTests"
                BuildableName = "AWSPinpointAnalyticsPluginUnitTests"
                BlueprintName = "AWSPinpointAnalyticsPluginUnitTests"
+               ReferencedContainer = "container:">
+            </BuildableReference>
+         </TestableReference>
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "HSMUniTests"
+               BuildableName = "HSMUniTests"
+               BlueprintName = "HSMUniTests"
                ReferencedContainer = "container:">
             </BuildableReference>
          </TestableReference>
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateAsyncSequence.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateAsyncSequence.swift
@@ -34,4 +34,7 @@ class StateAsyncSequence<Element: Sendable>: AsyncSequence {
         continuation.finish()
     }
 
+    deinit {
+        print("deinit!!!")
+    }
 }
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateAsyncSequence.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateAsyncSequence.swift
@@ -1,0 +1,37 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+class StateAsyncSequence<Element: Sendable>: AsyncSequence {
+
+    typealias Iterator = AsyncStream<Element>.Iterator
+    private var continuation: AsyncStream<Element>.Continuation! = nil
+
+    private var asyncStream: AsyncStream<Element>! = nil
+
+    init(bufferingPolicy: AsyncStream<Element>.Continuation.BufferingPolicy = .unbounded) {
+        asyncStream = AsyncStream<Element>(
+            Element.self,
+            bufferingPolicy: bufferingPolicy) { continuation in
+                self.continuation = continuation
+            }
+    }
+
+    func makeAsyncIterator() -> Iterator {
+        asyncStream.makeAsyncIterator()
+    }
+
+    func send(_ element: Element) {
+        continuation.yield(element)
+    }
+
+    func cancel() {
+        continuation.finish()
+    }
+
+}
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
@@ -8,40 +8,10 @@
 import Foundation
 
 /// Captures a weak reference to the value
-class WeakWrapper<T>: Hashable where T: AnyObject, T: Hashable {
+class WeakWrapper<T> where T: AnyObject {
     private(set) weak var value: T?
     init(_ value: T) {
         self.value = value
-    }
-
-    static func == (lhs: WeakWrapper<T>, rhs: WeakWrapper<T>) -> Bool {
-        lhs === rhs
-    }
-
-    func hash(into hasher: inout Hasher) {
-        value?.hash(into: &hasher)
-    }
-}
-
-/// Wrap any type in a reference
-class Box<BoxedValue>: Hashable where BoxedValue: Hashable {
-    let value: BoxedValue
-    init(_ value: BoxedValue) {
-        self.value = value
-    }
-
-    static func == (lhs: Box<BoxedValue>, rhs: Box<BoxedValue>) -> Bool {
-        lhs === rhs
-    }
-
-    func hash(into hasher: inout Hasher) {
-        value.hash(into: &hasher)
-    }
-}
-
-extension Box where BoxedValue == UUID {
-    func store(in tokens: inout Set<Box<BoxedValue>>) {
-        tokens.insert(self)
     }
 }
 
@@ -60,19 +30,16 @@ actor StateMachine<
     StateType,
     EnvironmentType: Environment
 > where StateType: State {
-    /// Listeners are invoked a minimum of one time: Each listener receives the current
-    /// state as soon as it subscribes, and will be invoked on each subsequent state change.
-    typealias StateChangedListener = (StateType) -> Void
-    typealias StateChangeListenerToken = Box<UUID>
-    typealias OnSubscribedCallback = () -> Void
+
+    /// AsyncSequences are invoked a minimum of one time: Each sequence receives the current
+    /// state as soon as it `listen()` was invoked, and will be invoked on each subsequent state change.
+    typealias StateChangeSequence = StateAsyncSequence<StateType>
 
     private let environment: EnvironmentType
     private let resolver: AnyResolver<StateType>
 
     private(set) var currentState: StateType
-
-    private var subscribers: [WeakWrapper<StateChangeListenerToken>: StateChangedListener]
-    private let pendingCancellations: AtomicValue<Set<StateChangeListenerToken>>
+    private var subscribers: [WeakWrapper<StateAsyncSequence<StateType>>]
 
     init<ResolverType>(
         resolver: ResolverType,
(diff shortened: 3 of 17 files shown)


Package.swift
```diff
@@ -175,7 +182,15 @@ let authTargets: [Target] = [
             "AWSCognitoAuthPlugin",
             "AWSPluginsTestCommon"
         ],
-        path: "AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests"
+        path: "AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests",
```
royjit, line 185: These changes are temporary added to aid in testing and generating small PRs. Will be removed before merging to main dev-preview branch

.swiftpm/xcode/xcshareddata/xcschemes/Amplify-Package.xcscheme
```diff
@@ -314,6 +314,20 @@
                ReferencedContainer = "container:">
             </BuildableReference>
          </BuildActionEntry>
+         <BuildActionEntry
```
royjit, line 317: These changes are temporary added to aid in testing and generating small PRs. Will be removed before merging to main dev-preview branch

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/hierarchical-state-machine-swiftTests/StateMachineTests.swift
```diff
@@ -156,8 +140,9 @@ class StateMachineTests: XCTestCase {
             eventType: .incrementAndDoActions([action1])
         )
 
-        testMachine.send(event)
-        wait(for: [action1WasExecuted, action2WasExecuted], timeout: 0.1, enforceOrder: true)
```
royjit, line 160: Note: I could not figure out a way to assert the order of expectation in async test context.

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/hierarchical-state-machine-swiftTests/StateMachineListenerTests.swift
```diff
@@ -87,27 +87,27 @@ class StateMachineListenerTests: XCTestCase {
...
 
-    func testDoesNotNotifyIfCancelledImmediately() {
-        stateMachine.send(Counter.Event(id: "test", eventType: .increment))
+    func testDoesNotNotifyIfCancelledImmediately() async {
+        await stateMachine.send(Counter.Event(id: "test", eventType: .increment))
 
         let notified = expectation(description: "notified")
-        notified.isInverted = true
-        let token = stateMachine.listen({ _ in
+        notified.expectedFulfillmentCount = 1
```
royjit, line 100: Note: I think this logic is the right one because, immediately after listening you should get back the current state. Only then cancel should take effect.

palpatim: I agree. Let's get this documented and agreed to across all platforms

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/hierarchical-state-machine-swiftTests/StateMachineTests.swift
```diff
@@ -6,32 +6,22 @@
...
-        testMachine.send(increment)
-        let notified = expectation(description: "notified")
-        testMachine.getCurrentState {
-            XCTAssertEqual($0.value, 1)
-            notified.fulfill()
-        }
-
-        waitForExpectations(timeout: 0.1)
+        await testMachine.send(increment)
+        let state = await testMachine.getCurrentState
```
palpatim, line 23: Something that falls naturally out of the async `send` and the fact that we run event processing to completion is that we could now return the new state from the `send` method: `let state = await testMachine.send(increment)`. This would be a change to the API contract, but would reduce boilerplate by removing the `send/currentState` dance. Thoughts?

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/hierarchical-state-machine-swiftTests/StateMachineTests.swift
```diff
@@ -76,23 +62,18 @@ class StateMachineTests: XCTestCase {
...
-            testMachine.send(increment)
-            let notified = expectation(description: "notified")
-            allExpectations.append(notified)
-            testMachine.getCurrentState {
-                XCTAssertEqual($0.value, iteration)
-                notified.fulfill()
-            }
+            await testMachine.send(increment)
+
+            let state = await testMachine.getCurrentState
```
palpatim, line 73: If we do adjust the `send` behavior to return the current state, we should retain this test: `machine.currentState` should still be actor-isolated and therefore async

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/hierarchical-state-machine-swiftTests/StateMachineTests.swift
palpatim: We do need to test this. I'm assuming there's something about the async context that is causing the `wait(for:,timeout:,enforceOrder:)` method to not work, so maybe we have assertions built into task 2 that ensures that task 1 has already been invoked? something like:

```swift
let action1WasExecuted = AtomicValue(initialValue: false)
let action2WasExecuted = AtomicValue(initialValue: false)

let action1 = BasicAction(identifier: "basic") { dispatcher, _ in
    action1WasExecuted.toggle()
    let action2 = BasicAction(identifier: "basic") { _, _ in
        XCTAssert(action1WasExecuted.value)
        action2WasExecuted.toggle()
    }
    ...
}

await testMachine.send(event)
XCTAssert(action1WasExecuted.value)
XCTAssert(action2WasExecuted.value)
```

palpatim (changes requested):
Throughout: Update doc comments--there are lots of stale references to invoking on the operationQueue

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
```diff
@@ -195,7 +152,7 @@ class StateMachine<
```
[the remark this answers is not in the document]
royjit: PendingCancel is still required because cancel is nonisolated

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/hierarchical-state-machine-swiftTests/StateMachineTests.swift
royjit: I am not sure whether it is really helpful in real case scenarios where the immediate state change is not relevant. Most of the use cases, we need to listen to state changes via listener and act on it.

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/hierarchical-state-machine-swiftTests/StateMachineTests.swift
royjit: Thank you, added a variation of this.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
```diff
@@ -73,70 +67,41 @@ class StateMachine<
...
     private let resolver: AnyResolver<StateType>
 
-    /// Backing queue for both consistencyQueue and notificationQueue
-    private let concurrentQueue: DispatchQueue
-
-    /// Manages consistency of internal state machine state
-    let operationQueue: OperationQueue
-
-    private var currentState: StateType
+    var currentState: StateType
```
palpatim, line 72: Should make this a `private(set) var` right? We don't want anything except the state machine resolver logic to update the state.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
```diff
@@ -220,13 +163,11 @@ extension StateMachine: EventDispatcher {
     /// machine's state and invokes listeners with the new state. Regardless of whether
     /// the state is new or not, the state machine will execute any effects from the
     /// event resolution process.
-    func send(_ event: StateMachineEvent) {
-        operationQueue.addOperation {
-            self.process(event: event)
-        }
+    func send(_ event: StateMachineEvent) async {
+        process(event: event)
```
palpatim, line 167: New behavior note: `send` will now always invoke state machine change listeners before it returns. Callers should be ready to handle state machine change updates before invoking `send`. Previously, the order of "return from method" vs. "invoke change listener" was undefined, but tended to be that it returned first.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
```diff
@@ -146,43 +111,23 @@ class StateMachine<
...
-                    continuation.resume(returning: self.currentState)
-                }
-            }
-        }
     }
 
     // MARK: - Isolated methods
 
-    /// Must be invoked on operationQueue
+    ///  Add the token in a weakly retained map
```
palpatim, line 124: suggestion: 
```suggestion
    ///  Weakly retain the listener and synchronously invoke it with the current state
```

See note below about sync invocation

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
```diff
@@ -193,14 +138,12 @@ class StateMachine<
 
         subscribers[wrappedToken] = listener
 
-        onSubscribe?()
-
-        concurrentQueue.async {
+        Task {
             listener(currentState)
```
palpatim, line 142: By invoking listener async, I'm just now realizing that we are not guaranteeing the order of delivery: it would be possible for the Task to be deferred until after a new state machine transition has been processed. We should invoke `listener` synchronously from this method, right? (This was a bug in my original implementation, too, so hurray for desk checking!)

We should make sure we have tests that assert the order of listener invocation. I know we've got one below that asserts that we get the initial state, but we need to gain some confidence that we will always invoke the listener prior to processing the next actor message.

I'll need to update the sequence diagrams as well; they currently don't show that the SM invokes with current state when the subscription message is processed.

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/hierarchical-state-machine-swiftTests/StateMachineListenerTests.swift
```diff
@@ -17,97 +17,75 @@ class StateMachineListenerTests: XCTestCase {
         stateMachine = CounterStateMachine.logging()
     }
 
-    func testNotifiesOnListen() {
-        stateMachine.send(Counter.Event(id: "test", eventType: .increment))
-        let subscribed = expectation(description: "subscribed")
+    func testNotifiesOnListen() async {
```
palpatim, line 20: As noted above, we need another test that asserts the order of execution:

* **Given** a state machine in state S<sub>0</sub> at T<sub>0</sub>
* **When** 
    * I subscribe to the state machine at time T<sub>1</sub>, and
    * At time T<sub>2</sub>, another process sends an event that triggers a transition to state S<sub>1</sub>
* **Then**
    * My listener is invoked at time T<sub>i</sub> > T<sub>1</sub> with the state S<sub>0</sub>; and subsequently
    * My listener is invoked at time T<sub>j</sub> > T<sub>i</sub> with the state S<sub>1</sub>

Note that there is no assertion about the truth of _T<sub>2</sub> <= T<sub>i</sub>_

Since this is async, we may need to test this in a loop to gain confidence that we don't have a hard-to-reproduce race condition. What we want to simulate is basically that Thread1 starts a subscription at almost exactly the same time as Thread2 sends the event. The State Machine must serialize the processing of messages such that the subscription listener is invoked prior to processing the state transition. In the current code, that assertion may fail since we invoke the initial listener asynchronously.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateAsyncSequence.swift
```diff
@@ -0,0 +1,37 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+class StateAsyncSequence<Element: Sendable>: AsyncSequence {
```
ameter, line 10: Would it make sense to reuse [AmplifyAsyncSequence](https://github.com/aws-amplify/amplify-ios/blob/dev-preview/Amplify/Core/Support/AmplifyAsyncSequence.swift) here?

royjit: Few things for not using AmplifyAsyncSequence:
- Do not want to rely on a public type here and depend on core, statemachine logic might be moved to a separate place in future
- We do not need most of the functionalities inside AmplifyAsyncSequence
- I kept the sequence as a `class` to help in identifying if it was dereferenced

palpatim (approved):
Approved w/comments

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
```diff
@@ -220,13 +74,14 @@ extension StateMachine: EventDispatcher {
     /// machine's state and invokes listeners with the new state. Regardless of whether
     /// the state is new or not, the state machine will execute any effects from the
     /// event resolution process.
-    func send(_ event: StateMachineEvent) {
-        operationQueue.addOperation {
-            self.process(event: event)
+    func send(_ event: StateMachineEvent) async {
+        if Task.isCancelled {
```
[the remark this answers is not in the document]
royjit: Yes, that is right

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/hierarchical-state-machine-swift/StateMachine.swift
```diff
@@ -250,31 +103,25 @@ extension StateMachine: EventDispatcher {
     ///   - newState: The new state to be sent
```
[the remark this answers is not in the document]
royjit: Updated

PR #2183, opened 2022-08-20
feat: add support for AsyncSequence Amplify.Publisher

ameter:
*Description of changes:*
add support for AsyncSequence Amplify.Publisher

- [x] All unit tests pass
- ~~[ ] Documentation update for the change if required~~

### DIFF
--- a/Amplify/Core/Support/Amplify+Publisher.swift
+++ b/Amplify/Core/Support/Amplify+Publisher.swift
@@ -21,7 +21,7 @@ public extension Amplify {
         /// - Parameter operation: The Task for which to create the Publisher.
         /// - Returns: The Publisher for the given Task.
         public static func create<Success>(
-            operation: @escaping @Sendable () async throws -> Success
+            _ operation: @escaping @Sendable () async throws -> Success
         ) -> AnyPublisher<Success, Error> {
             let task = Task(operation: operation)
             return Future() { promise in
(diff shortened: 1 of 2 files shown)


Amplify/Core/Support/Amplify+Publisher.swift
```diff
@@ -27,6 +50,52 @@ extension Amplify {
...
+        ///
+        /// - Parameter sequence: The AsyncSequence for which to create the Publisher.
+        /// - Returns: The Publisher for the given AsyncSequence.
+        public static func create<Sequence: AsyncSequence>(
+            _ sequence: Sequence
+        ) -> AnyPublisher<Sequence.Element, Error> {
+            let subject = PassthroughSubject<Sequence.Element, Error>()
+            let task = Task {
+                do {
+                    for try await value in sequence {
```
brennanMKE, line 78: this `await` could be waiting a long time before it is able to continue to the next iteration. it should use the `withTaskCancellationHandler` function so that the onCancel handler can be called immediately when the Task is cancelled.

ameter: Done.  This allows it to immediately stop waiting for values from the sequence in the case of a slow iterator.  I kept an internal check for cancellation inside the loop as well, which prevents it from continuing to process values after cancellation in the case of a fast iterator.

brennanMKE (approved):
👍

PR #2217, opened 2022-08-29
adds async function for getURL in Storage Plugin

brennanMKE:
*Issue #, if available:*

*Description of changes:*

- [ ] All unit tests pass

### DIFF
--- a/Amplify/Categories/Storage/StorageCategory+ClientBehavior.swift
+++ b/Amplify/Categories/Storage/StorageCategory+ClientBehavior.swift
@@ -81,4 +81,12 @@ extension StorageCategory: StorageCategoryBehavior {
                      resultListener: StorageListOperation.ResultListener?) -> StorageListOperation {
         return plugin.list(options: options, resultListener: resultListener)
     }
+
+    // MARK: - Async API -
+
+    @discardableResult
+    public func getURL(key: String,
+                options: StorageGetURLOperation.Request.Options?) async throws -> URL {
+        try await plugin.getURL(key: key, options: options)
+    }
 }
--- a/Amplify/Categories/Storage/StorageCategoryBehavior.swift
+++ b/Amplify/Categories/Storage/StorageCategoryBehavior.swift
@@ -104,4 +104,17 @@ public protocol StorageCategoryBehavior {
     @discardableResult
     func list(options: StorageListOperation.Request.Options?,
               resultListener: StorageListOperation.ResultListener?) -> StorageListOperation
+
+    // MARK: - Async API -
+
+    /// Retrieve the remote URL for the object from storage.
+    ///
+    /// - Parameters:
+    ///   - key: The unique identifier for the object in storage.
+    ///   - options: Parameters to specific plugin behavior
+    /// - Returns: requested Get URL
+    @discardableResult
+    func getURL(key: String,
+                options: StorageGetURLOperation.Request.Options?) async throws -> URL
+
 }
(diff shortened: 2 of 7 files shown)


Amplify/Categories/Storage/StorageCategory+ClientBehavior.swift
```diff
@@ -81,4 +81,12 @@ extension StorageCategory: StorageCategoryBehavior {
                      resultListener: StorageListOperation.ResultListener?) -> StorageListOperation {
         return plugin.list(options: options, resultListener: resultListener)
     }
+
+    // MARK: - Async API -
+
+    @discardableResult
+    public func getURL(key: String,
```
royjit, line 88: 1. Are we keeping the call back apis?
2. Let make the Options take a default nil  `options: StorageGetURLOperation.Request.Options? = nil`

brennanMKE: the listener API is unchanged for this PR. once the async functions are merged to dev-preview the listener API can be removed. this will make for a more focused PR for each change.

Amplify/Categories/Storage/StorageCategory+ClientBehavior.swift
```diff
@@ -81,4 +81,12 @@ extension StorageCategory: StorageCategoryBehavior {
                      resultListener: StorageListOperation.ResultListener?) -> StorageListOperation {
         return plugin.list(options: options, resultListener: resultListener)
     }
+
+    // MARK: - Async API -
+
+    @discardableResult
```
lawmicha, line 87: the reason why the callback apis had the `@discardableResult` was because the result was returned on the callback listener, and most use cases do not care about the operation. Do we want to have `@discardableResult` for these async APIs as well? The fast operation return type is the URL that the developer cares about. I think it make sense to warn them if they are not doing anything with the result, forcing them to explicitly putting down `_ = Amplify.Storage.getURL...` if they really don't care about the return value

Amplify/Categories/Storage/StorageCategory+ClientBehavior.swift
```diff
@@ -81,4 +81,12 @@ extension StorageCategory: StorageCategoryBehavior {
                      resultListener: StorageListOperation.ResultListener?) -> StorageListOperation {
         return plugin.list(options: options, resultListener: resultListener)
     }
+
+    // MARK: - Async API -
```
lawmicha, line 85: [suggestion] since I've done this for API and DataStore and List APIs.. 
- It didn't seem too much work to all all async flavours in one single PR, see API https://github.com/aws-amplify/amplify-ios/pull/2140/files and DataStore https://github.com/aws-amplify/amplify-ios/pull/2126 Up to you to add ontop of this PR or do another one later
- I added them undernearth each corresponding previous API, instead of creating a new section below them. just pointing it out, up to your personal preference. Might be easier to spot the API differences and easier to remove previous API later and use the same API documentation (if it exists)

lawmicha (approved):
LGTM, one comment about `@discardableResult`

Amplify/Categories/Storage/StorageCategory+ClientBehavior.swift
lawmicha: looks like this PR has more unit testing and better documentation, will retract my suggestion above

Amplify/Categories/Storage/StorageCategory+ClientBehavior.swift
brennanMKE: We still support the Hub so that may be used instead of the return value so I think it will be alright to leave this attribute.

lawmicha: okay sounds good

PR #2218, opened 2022-08-29
chore(api): fix remaining GraphQlConnection(1,2) to async api

manaswi223:
*Issue #, if available:*

*Description of changes:*
fixed remaining tests of the GraphQLConnectionScrenario1Tests and GraphQLConnectionScrenario2Tests, in continuation with [https://github.com/aws-amplify/amplify-ios/pull/2178](url)
Screen Shot 2022-08-29 at 1 06 26 PM

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/GraphQLConnectionScenario1Tests.swift
+++ b/AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/GraphQLConnectionScenario1Tests.swift
@@ -137,33 +137,22 @@ class GraphQLConnectionScenario1Tests: XCTestCase {
         }
     }
 
-    func testPaginatedListProjects() {
-        guard let team = createTeam(name: "name"),
-              let projecta = createProject(team: team),
-              let projectb = createProject(team: team) else {
+    func testPaginatedListProjects() async throws {
+        guard let team = try await createTeam(name: "name"),
+              let projecta = try await createProject(teamID: team.id, team: team),
+              let projectb = try await createProject(teamID: team.id, team: team) else {
             XCTFail("Could not create team and two projects")
             return
         }
-
-        let listProjectByTeamIDCompleted = expectation(description: "list projects completed")
         var results: List<Project1>?
         let predicate = Project1.keys.id == projecta.id || Project1.keys.id == projectb.id
         let request: GraphQLRequest<List<Project1>> = GraphQLRequest<Project1>.list(Project1.self, where: predicate)
-        Amplify.API.query(request: request) { result in
-            switch result {
-            case .success(let result):
-                switch result {
-                case .success(let projects):
-                    results = projects
-                    listProjectByTeamIDCompleted.fulfill()
-                case .failure(let response):
-                    XCTFail("Failed with: \(response)")
-                }
-            case .failure(let error):
-                XCTFail("\(error)")
-            }
+        let result = try await Amplify.API.query(request: request)
+        guard case .success(let projects) = result else {
+            XCTFail("Missing Successful response")
+            return
         }
-        wait(for: [listProjectByTeamIDCompleted], timeout: TestCommonConstants.networkTimeout)
+        results = projects
         guard var subsequentResults = results else {
             XCTFail("Could not get first results")
             return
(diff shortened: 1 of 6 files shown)


AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/GraphQLConnectionScenario1Tests.swift
```diff
@@ -136,34 +136,23 @@ class GraphQLConnectionScenario1Tests: XCTestCase {
...
-
-        let listProjectByTeamIDCompleted = expectation(description: "list projects completed")
+    
+    func testPaginatedListProjects() async throws {
+        let team = Team1(name: "name")
+        _ = try await Amplify.API.mutate(request: .create(team))
+        let projecta = Project1(team: team)
+        _ = try await Amplify.API.mutate(request: .create(projecta))
+        let projectb = Project1(team: team)
+        _ = try await Amplify.API.mutate(request: .create(projectb))
```
lawmicha, line 146: we need to check if the result is a success case as well (due to the nested DX)

AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/GraphQLConnectionScenario1Tests.swift
```diff
@@ -190,49 +179,32 @@ class GraphQLConnectionScenario1Tests: XCTestCase {
...
-        wait(for: [requestInvokedSuccessfully], timeout: TestCommonConstants.networkTimeout)
         return result
     }
 
     func createProject(id: String = UUID().uuidString,
-                       name: String? = nil,
-                       team: Team1? = nil) -> Project1? {
+                        name: String? = nil,
+                        teamID: String,
+                        team: Team1? = nil) async throws -> Project1? {
```
lawmicha, line 198:
```suggestion
                        team: Team1? = nil) async throws -> Project1 {
```
Return `Project1` not optional since that's the return type of the inner switch statement. The goal is to reduce the nested result types by throwing the result's failure case. the method will have this structure:
```
func methodThatThrowsTwoDifferentErrors() async throws -> Project1 {
  let result = try await methodThatThrowsAndReturnsResult()
  switch result {
   case .success(let project): return project
   case .failure(let error): throw error
  }
}
```

AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/GraphQLConnectionScenario1Tests.swift
```diff
@@ -190,49 +179,32 @@ class GraphQLConnectionScenario1Tests: XCTestCase {
...
         let project = Project1(id: id, name: name, team: team)
         var result: Project1?
-        let requestInvokedSuccessfully = expectation(description: "request completed")
-        Amplify.API.mutate(request: .create(project)) { event in
+        let event = try await Amplify.API.mutate(request: .create(project))
             switch event {
```
lawmicha, line 202:
```suggestion
        let event = try await Amplify.API.mutate(request: .create(project))
        switch event {
```
no need to indent on `switch event {`

AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/GraphQLConnectionScenario1Tests.swift
```diff
@@ -190,49 +179,32 @@ class GraphQLConnectionScenario1Tests: XCTestCase {
...
     }
 
     func createProject(id: String = UUID().uuidString,
-                       name: String? = nil,
-                       team: Team1? = nil) -> Project1? {
+                        name: String? = nil,
+                        teamID: String,
+                        team: Team1? = nil) async throws -> Project1? {
         let project = Project1(id: id, name: name, team: team)
         var result: Project1?
```
lawmicha, line 200:
```suggestion
```
we don't need `result` to hold it, the `await` call does all the work for us in suspending and resuming the code execution, making it like procedural code

```swift
// let result: Project1?
let event = try await Amplify.API.mutate(request: .create(project)) // code suspends here, then resumes on the next line when the call returns
switch event {
case .success(let project):
   return project 
...
```

AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/GraphQLConnectionScenario1Tests.swift
```diff
@@ -190,49 +179,32 @@ class GraphQLConnectionScenario1Tests: XCTestCase {
...
             switch event {
-            case .success(let data):
-                switch data {
-                case .success(let project):
+            case .success(let project):
                     result = project
-                default:
-                    XCTFail("Could not get data back")
+            case .failure(let graphQLResponseError):
+                    XCTFail("Failed with error: \(graphQLResponseError)")
```
lawmicha, line 206: Failing here works as well and has some intention behaviors for our testing since a call to `createProject` which receives a GraphQLResponseError will then hit this line, `XCTFail`, to fail the test. Failing the test is not always the case since tests can be configured to continue on failure. I believe if you throw `graphQLResponseError` then the code will happily compile since all code paths will complete the method. Having this `XCTFail` is probably the reason why we are holding `var result: Project1?` and returning `nil`, which is required since XCTFail does not complete the method.

lawmicha:
```suggestion
   throw graphQLResponseError
```

AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/GraphQLConnectionScenario2Tests.swift
```diff
@@ -140,61 +140,42 @@ class GraphQLConnectionScenario2Tests: XCTestCase {
...
                 case .success(let projects):
                     print(projects)
-                    listProjectByTeamIDCompleted.fulfill()
-                case .failure(let response):
-                    XCTFail("Failed with: \(response)")
-                }
-            case .failure(let error):
-                XCTFail("\(error)")
+                case .failure(let graphQLResponse):
+                    XCTFail("Failed with: \(graphQLResponse)")
```
manaswi223, line 158: throw graphQLResponseError

AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/GraphQLConnectionScenario1Tests.swift
lawmicha: my mistake, it the return type should still be optional `Project1` if the `case .success(let project)` provides `project` as an optional type

manaswi223: done! got you resolving the con

PR #2225, opened 2022-08-30
adds async function for downloadFile in Storage Plugin

brennanMKE:
*Issue #, if available:*

*Description of changes:*

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin+AsyncClientBehavior.swift
+++ b/AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin+AsyncClientBehavior.swift
@@ -50,6 +50,22 @@ extension AWSS3StoragePlugin {
     }
 
     @discardableResult
+    public func downloadFile(key: String,
+                      local: URL,
+                      options: StorageDownloadFileOperation.Request.Options?) async throws -> StorageDownloadFileTask {
+        let options = options ?? StorageDownloadFileRequest.Options()
+        let request = StorageDownloadFileRequest(key: key, local: local, options: options)
+        let operation = AWSS3StorageDownloadFileOperation(request,
+                                                          storageConfiguration: storageConfiguration,
+                                                          storageService: storageService,
+                                                          authService: authService)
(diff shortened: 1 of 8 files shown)


AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin+AsyncClientBehavior.swift
```diff
@@ -49,6 +49,22 @@ extension AWSS3StoragePlugin {
...
+                      options: StorageDownloadFileOperation.Request.Options?) async throws -> StorageDownloadFileTask {
+        let options = options ?? StorageDownloadFileRequest.Options()
+        let request = StorageDownloadFileRequest(key: key, local: local, options: options)
+        let operation = AWSS3StorageDownloadFileOperation(request,
+                                                          storageConfiguration: storageConfiguration,
+                                                          storageService: storageService,
+                                                          authService: authService)
+
+        queue.addOperation(operation)
+        let taskAdapter = AmplifyInProcessReportingOperationTaskAdapter(operation: operation)
```
lawmicha, line 64: i think these two lines should be swapped because we do not control when the operation's `main()` method is executed, thus adding it to the operation queue is the intention to start the execution and may cause it to prematurely finish before being hooked up in the taskAdapter. At runtime, this may not actually be a problem since creating the task adapter will probably always be faster than the queue dequeuing an operation and performing it. Logically reading this though, I believe it should be 
1. Create the taskAdapter with the operation to hook everything up, the progress and result listeners
2. Then add it to the queue, as our intention for the operation to run
3. Then return the taskAdapter

lawmicha: Adding Todo to an existing Storage issue here https://github.com/aws-amplify/amplify-ios/issues/2223#issuecomment-1231906147

lawmicha (approved):
LGTM, tracking issue in https://github.com/aws-amplify/amplify-ios/pull/2225#discussion_r958705031

PR #2239, opened 2022-08-31
adds new target for AmplifyAsyncTesting to centralize this support for more targets

brennanMKE:
*Issue #, if available:*

*Description of changes:*

- [ ] All unit tests pass


ruisebas (dismissed):

AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/Base/AsyncExpectation.swift
```diff
@@ -1,130 +0,0 @@
...
-// All Rights Reserved.
-//
-// SPDX-License-Identifier: Apache-2.0
-//
-
-import Foundation
-import XCTest
-import Amplify
-
-public actor AsyncExpectation {
```
lawmicha, line 12: by removing this, how does APIHostApp and the rest of the integration test targets pull in AsyncExpectation?

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginIntegrationTests/TestSupport/AsyncExpectation.swift
```diff
@@ -1,196 +0,0 @@
...
-// All Rights Reserved.
-//
-// SPDX-License-Identifier: Apache-2.0
-//
-
-import Foundation
-import XCTest
-import Amplify
-
-public actor AsyncExpectation {
```
lawmicha, line 12: DataStoreHostApp/APIHostApp, which has integration test targets, pulls in package.swift which has the following products available:
```
products: [
        .library(
            name: "Amplify",
            targets: ["Amplify"]
        ),
        .library(
            name: "AWSAPIPlugin",
            targets: ["AWSAPIPlugin"]
        ),
        .library(
            name: "AWSCognitoAuthPlugin",
            targets: ["AWSCognitoAuthPlugin"]
        ),
        .library(
            name: "AWSDataStorePlugin",
            targets: ["AWSDataStorePlugin"]
        ),
        .library(
            name: "AWSS3StoragePlugin",
            targets: ["AWSS3StoragePlugin"]
        ),
        .library(
            name: "AWSLocationGeoPlugin",
            targets: ["AWSLocationGeoPlugin"]
        ),
        .library(
            name: "AWSPinpointAnalyticsPlugin",
            targets: ["AWSPinpointAnalyticsPlugin"]
        )
    ],
```

We do not publiclly expose AmplifyTestCommon, which is the reason why this file was duplicated. By removing it here, the integration test targets are broken since they still don't have access to AmplifyTestCommon, and if i understand correctly, this PR only moves the AsyncExpectation and related under AmplifyAsyncTesting, which is still under AmplifyTestCommon

lawmicha (changes requested):

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginIntegrationTests/TestSupport/AsyncExpectation.swift
lawmicha: All integration tests are opened via their host app's xcodeproj file, for example DataStoreHostApp and APIHostApp. switching to an test target, uses the host app as the host application. Building fails because the file is removed:
image

ruisebas: Yeah you're right, we need to add `AmplifyAsyncTesting`  here if we want to use them in the integration tests host apps.

lawmicha: integration testing is done via the host apps, ie. DataStoreHostApp and APIHostApp. Each test target runs on the host app, which has package.swift added to the project, the hostapp adds the products needed like the plugin targets. the hostapp and test targets do not have access to AsyncExpectation because it doesn't have access to AmplifyTestCommon/AmplifyAsyncTesting

image

Package.swift
```diff
@@ -306,7 +322,10 @@ let package = Package(
         .library(
             name: "AWSPinpointAnalyticsPlugin",
             targets: ["AWSPinpointAnalyticsPlugin"]
-        )
+        ),
+        .library(
+            name: "AmplifyAsyncTesting",
+            targets: ["AmplifyAsyncTesting"]),
```
lawmicha, line 328: With this change and removing the AsyncExpectation from the HostApps will break builds for the integration tests. The follow up to this is to add AmplifyAsyncTesting to the HostApp and also XCTest.framework with "Do not Embed" (fails to compile without this).
image

lawmicha: Just adding AmplifyAsyncTesting shows this error
image
Adding XCTest.framework shows duplicate error
Set to "Do not Embed" compiles file

lawmicha (approved):

Package.swift
brennanMKE: yes, normally XCTest is not added to an app project. this could also be added to linking but this works too.

PR #2276, opened 2022-09-07
fix(Storage): Various bug fixes

ruisebas:
*Description of changes:*
Fixing the following Storage issues:

**Download**
- Attempting to download to a file always fails with "file already exists". Caused by a missing `!` in a validation.
- `.httpStatusError` is returned instead of `.keyNotFound` and `accessDenied` when applicable. Fixed by implementing a check for these status codes in `StorageTransferResponse.responseError`.

**List**
- Setting a `path` value in its options results in said value being removed from each `item.key` that is returned. 
  E.g. if you provide the full path for a given item, the item will have an empty key. If you provide the folder, each `item.key` will not include the folder.
  Fixed by adding a `path` attribute to the `AWSS3ListObjectsV2Request` instead of concatenating everything into its `prefix`.

**GetURL**
- Setting an `expires` value in its options does nothing, the URL does not expire. Fixed by propagating the `expires` argument instead of `nil`.

**Remove**
- Doesn't work on protected nor private buckets. This was caused due to the Swift SDK setting `useDoubleURIEncode: false` when creating the `SigV4Config` struct. 
   Fixed with a local extension that creates the right struct.

- [X] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Dependency/AWSS3Adapter.swift
+++ b/AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Dependency/AWSS3Adapter.swift
@@ -34,7 +34,7 @@ class AWSS3Adapter: AWSS3Behavior {
         Task {
             let input = DeleteObjectInput(bucket: request.bucket, key: request.key)
             do {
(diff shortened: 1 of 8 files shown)


royjit (approved):
LGTM

PR #2282, opened 2022-09-08
chore(datastore): Fix deprecated MutationSyncMetadata API

thisisabhash:
*Issue #, if available:* None

*Description of changes:* Fix deprecated `MutationSyncMetadata` API

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
@@ -103,8 +103,8 @@ final class SQLiteMutationSyncMetadataMigrationDelegate: MutationSyncMetadataMig
             sql += "SELECT id, \'\(modelName)\' as tableName FROM \(modelName)"
         }
         sql = "INSERT INTO \(MutationSyncMetadataMigration.MutationSyncMetadataCopy.modelName) (id,deleted,lastChangedAt,version) " +
-        "select models.tableName || '|' || mm.id, mm.deleted, mm.lastChangedAt, mm.version " +
-        "from MutationSyncMetadata mm INNER JOIN (" + sql + ") as models on mm.id=models.id"
+        "select mm.id, mm.deleted, mm.lastChangedAt, mm.version " +
+        "from MutationSyncMetadata mm INNER JOIN (" + sql + ") as models on mm.id=models.tableName || '|' || models.id"
         try connection.execute(sql)
         return sql
     }
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
@@ -103,8 +103,8 @@ final class SQLiteMutationSyncMetadataMigrationDelegate: MutationSyncMetadataMig
             sql += "SELECT id, \'\(modelName)\' as tableName FROM \(modelName)"
         }
         sql = "INSERT INTO \(MutationSyncMetadataMigration.MutationSyncMetadataCopy.modelName) (id,deleted,lastChangedAt,version) " +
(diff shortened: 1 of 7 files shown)


AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Migration/SQLiteMutationSyncMetadataMigrationDelegateTests.swift
```diff
@@ -75,14 +75,14 @@ class SQLiteMutationSyncMetadataMigrationDelegateTests: MutationSyncMetadataMigr
         try setUpAllModels()
         let restaurant = Restaurant(restaurantName: "name")
         save(restaurant)
-        let metadata = MutationSyncMetadata(id: restaurant.id, deleted: false, lastChangedAt: 1, version: 1)
+        let metadata = MutationSyncMetadata(modelId: restaurant.id, modelName: Restaurant.modelName, deleted: false, lastChangedAt: 1, version: 1)
```
lawmicha, line 78: i don't think we should change this test, despite it's marked as deprecated, the old init is allowing us to set up the migration script testing with old day and verifying that the migration works

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Migration/MutationSyncMetadataMigrationDelegate+SQLite.swift
```diff
@@ -103,8 +103,8 @@ final class SQLiteMutationSyncMetadataMigrationDelegate: MutationSyncMetadataMig
             sql += "SELECT id, \'\(modelName)\' as tableName FROM \(modelName)"
         }
         sql = "INSERT INTO \(MutationSyncMetadataMigration.MutationSyncMetadataCopy.modelName) (id,deleted,lastChangedAt,version) " +
-        "select models.tableName || '|' || mm.id, mm.deleted, mm.lastChangedAt, mm.version " +
-        "from MutationSyncMetadata mm INNER JOIN (" + sql + ") as models on mm.id=models.id"
+        "select mm.id, mm.deleted, mm.lastChangedAt, mm.version " +
+        "from MutationSyncMetadata mm INNER JOIN (" + sql + ") as models on mm.id=models.tableName || '|' || models.id"
```
lawmicha, line 107: what backfill does is if the existing data on the app is not in the format of "<modelName>|<modelId>" then backfill it to this format. I don't think we can change this logic, if we change it it's like we might as well remove the migration script all together. doing so will be a one-way door because some existing apps will break if the developer upgrades to 2.0.0 and ships the app.

lawmicha (changes requested):
Thanks for looking into this, I don't think we can ship this since it changes the migration script logic. However we've uncovered the tech debt regarding MutationSyncMetadata having the incorrect schema. We should scope this out further, such as 

- introduce a new table name like “ModelMetadata” or something (perhaps aligned across other platforms) that has the composite key `modelName` and `modelId` (or just id)
- then we would also need to write another migration script that migrates MutationSyncMetadata to ModelMetadata
- the sync engine portion of the code should create `ModelMetadata` rows when reconciling mutation responses.

AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Sync/MutationQueue/OutgoingMutationQueueNetworkTests.swift
```diff
@@ -265,7 +265,8 @@ class OutgoingMutationQueueNetworkTests: SyncEngineTestBase {
             let mockResponse = MutationSync(
                 model: model,
                 syncMetadata: MutationSyncMetadata(
-                    id: model.id,
+                    modelId: model.id,
+                    modelName: model.modelName,
```
lawmicha, line 269: this also looks like a reasonable change to do

lawmicha (approved):

lawmicha:
rerunning unrelated test failure in API 
```
AWSAPIPluginTests.GraphQLMutateCombineTests
  testMutateHandlesResponseError, Asynchronous wait failed: Exceeded timeout of 0.05 seconds, with unfulfilled expectations: "Received response error", "Received finished".
  /Users/runner/work/amplify-ios/amplify-ios/AmplifyPlugins/API/Tests/AWSAPIPluginTests/Operation/GraphQLMutateCombineTests.swift:85
  ```

        waitForExpectations(timeout: 0.05)
        sink.cancel()
  ```
```

PR #2301, opened 2022-09-13
fix(Storage): Adding support for pausing, resuming and cancelling multipart uploads

ruisebas:
*Description of changes:*
Tweaking the code in order to support pausing, resuming and cancelling multipart uploads.

Basically the main `StorageTransferTask` that was used for multipart uploads had no way of interacting with each of the subtasks corresponding to each part. So i've added a `subTasks` array only for multipart uploads, which is then called respectively on each corresponding method.

- [X] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Support/Internal/StorageMultipartUploadSession.swift
+++ b/AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Support/Internal/StorageMultipartUploadSession.swift
@@ -181,6 +181,7 @@ class StorageMultipartUploadSession {
 
     func startUpload() {
         do {
+            transferTask.notify(progress: Progress(totalUnitCount: 0))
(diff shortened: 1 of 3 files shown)


AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Support/Internal/StorageMultipartUploadSession.swift
```diff
@@ -181,6 +181,7 @@ class StorageMultipartUploadSession {
 
     func startUpload() {
         do {
+            transferTask.notify(progress: Progress(totalUnitCount: 0))
```
ruisebas, line 184: Given that the main `transferTask` lacks a `sessionTask`, this is the only way to set its `status` to `.inProgress`.

As mentioned above, that is necessary for the `pause` method to work

PR #2338, opened 2022-09-22
fix(datastore-v2): InitializeSubscription race condition disconnected event

lawmicha:
*Issue #, if available:*

*Description of changes:*
This PR is the V2 version of https://github.com/aws-amplify/amplify-ios/pull/2319

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
@@ -207,7 +207,30 @@ extension AWSInitialSyncOrchestrator {
               else {
             return false
         }
+        
+        // The following check is to catgorize the error as an unauthorized error when the process fails to retrieve
+        // the autorization token. This is taken directly from `AuthTokenURLRequestInterceptor`'s error handling path
+        // that returns an APIError.operationError with an underlying AuthError.
+        //
+        // A signed out user, or a signed in user's session that has expired, will result the `getToken()` to
+        // return an error. The request is never sent over the network to the service to apply its auth check and is
+        // returned immediately to this calling code.
+        //
+        // The check itself is not the most ideal contract for checking when there is an authorization error, however it
+        // does have some stability since `operationError` is a local implementation detail. The underlying error check
+        // for AuthError means that the AuthError could be any case. For example, if the AuthError changes, it will
+        // still be categorized as unauthorized by this code. If a specific type like `AuthError.unauthorized` is
+        // introduced in the future, this code will still return true, which is crucial to prevent the sync
+        // orchestrator from failing the entire sync process.
+        if case let .api(amplifyError, _) = datastoreError,
+           let apiError = amplifyError as? APIError,
+           case .operationError(_, _, let underlyingError) = apiError,
+           (underlyingError as? AuthError) != nil {
+            return true
+        }
 
+        // If token was retrieved, and request was sent, but service returned an error response,
+        // check that it is an Unauthorized error from the GraphQL response payload.
         if case let .api(apiError, _) = datastoreError,
            let responseError = apiError as? GraphQLResponseError<ResponseType>,
            let graphQLError = graphqlErrors(from: responseError)?.first,
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
@@ -208,8 +208,8 @@ extension AWSInitialSyncOrchestrator {
             return false
         }
         
-        // The following check is to catgorize the error as an unauthorized error when the process fails to retrieve
-        // the autorization token. This is taken directly from `AuthTokenURLRequestInterceptor`'s error handling path
+        // The following check is to categorize the error as an unauthorized error when the process fails to retrieve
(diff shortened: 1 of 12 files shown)


lawmicha:
Running through the unit / integration tests and these are the following that need to be fixed:
- OutgoingMutationQueueNetworkTests/testLastMutationSentWhenNoNetwork() - has been hitting us a couple of times now

lawmicha:
The macOS tests that are failing should not be blocking this PR since they were failing before and recently enabled in the pipeline.

image

Opening an issue to track this

drochetti (dismissed):

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -207,7 +207,30 @@ extension AWSInitialSyncOrchestrator {
...
+        // The check itself is not the most ideal contract for checking when there is an authorization error, however it
+        // does have some stability since `operationError` is a local implementation detail. The underlying error check
+        // for AuthError means that the AuthError could be any case. For example, if the AuthError changes, it will
+        // still be categorized as unauthorized by this code. If a specific type like `AuthError.unauthorized` is
+        // introduced in the future, this code will still return true, which is crucial to prevent the sync
+        // orchestrator from failing the entire sync process.
+        if case let .api(amplifyError, _) = datastoreError,
+           let apiError = amplifyError as? APIError,
+           case .operationError(_, _, let underlyingError) = apiError,
+           (underlyingError as? AuthError) != nil {
```
thisisabhash, line 228: Is it intended to classify all Auth errors as unauthorized?

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
```diff
@@ -262,7 +264,10 @@ extension AWSModelReconciliationQueue {
         return errorTypeValue
     }
 
-    private func isUnauthorizedError(_ error: Error?) -> Bool {
+    private func isUnauthorizedError(errorDescription: String, _ error: Error?) -> Bool {
+        if errorDescription.range(of: "Unauthorized", options: .caseInsensitive) != nil {
```
thisisabhash, line 268: Can we use `Error` type instead of using finding presence of string "Unauthorized" in error description?

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -207,7 +207,30 @@ extension AWSInitialSyncOrchestrator {
               else {
             return false
         }
+        
+        // The following check is to catgorize the error as an unauthorized error when the process fails to retrieve
```
thisisabhash, line 211: nit : categorize

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -207,7 +207,30 @@ extension AWSInitialSyncOrchestrator {
...
+        // the autorization token. This is taken directly from `AuthTokenURLRequestInterceptor`'s error handling path
```
thisisabhash, line 212: nit: authorization

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/AWSModelReconciliationQueue.swift
lawmicha: The `Error` type here is the underlying error coming from `APIError.operationError` for subscription messages that contain a valid GraphQL response with the errors in the response payload, containing the unauthorized error. This check against the message is for another scenario at the top level websocket connection that can receive an unauthorized message. 

Some additional details around this.. The AppSync RT client returns the error back to APIPlugin, which then sets the message with operationError that represents "Unauthorized". This was done in this PR https://github.com/aws-amplify/amplify-ios/pull/1994 In that PR we also exposed APIError.isUnauthorized to make it easy for the developer using APIPlugin directly to used. We couldn't use that extension method here in DataStore since APIPlugin is injected at runtime, it doesn't have a dependency on AWSAPIPlugin, making this check the check that we need to do for the websocket connection failures. Alternatively, instead of checking against the string, i think the ideal solution is to check that the underlying error contains AuthError. This just means that APIPlugin needs to fabricate the return object to be an APIError with an underlying AuthError, which isn't hard to do.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
lawmicha: Yes, but that's not all that what we're doing. It's in addition to checking if it's "APIError.operationError" and the underlying error is an AuthError. It's the error that the library created when the interceptor code in AuthTokenURLRequestInterceptor could not retrieve the token for the request, so the request was not actually performed, and the service never got a chance to validate the request and return unauthorized. From DataStore's perspective, the request is unauthorized because the user is either not signed in or the token has expired

thisisabhash (approved):

PR #2358, opened 2022-09-27
fix(datastore): Setting nil values for empty associated models

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-ios/issues/2336

*Description of changes:*
Verified that there was a regression from the CPK changes at this file: https://github.com/aws-amplify/amplify-ios/pull/1752/files#diff-2fa30b26d50da3e22bdbe4d8e4eb16bb22011672699f804c0926fa34629c5cc2 CPK was released in 1.28.0, and verified that the failing test `DataStoreConnectionOptionalAssociations.testRemovePostFromCommentAndBlogFromPost` is passing in 1.27.1 (last 1.27.x version).

This PR updates the logic for translating a model instance over to the GraphQL request variables "input" object. 

The following example uses Comment and Post, where a post has many comments and the comment belongs to a post.

Previously, before CPK changes, when a model with an associated field with its value populated will be translated to 
```
"input": {
   "id": "commentId",
   "postId": "associatedPostId"
}
```
When the model with associated field does not have the value populated, it will be translated to 
```
"input": {
   "id": "commentId",
   "postId": nil
}
```
With the CPK changes, the model is now being translated to 
```
"input": {
   "id": "commentId"
}
```
The problem is that it is missing the associated field and `nil` value in the input. For create mutations, this is a valid request, and the comment gets created without associating to any post. The comment can be updated later to be associated with a post. For update mutations, where the comment was created with a post and then removed from the comment instance, then the update mutation request is created, is what's causing issues. The removal of the post is not sent as part of the update mutation for the deassociation. 

By always setting the associated fields to `nil` when the value does not exist will fix the problem. The create mutation will also be affected since uses the same translation code, which simply reverts the behavior back to what it was previously. We may consider passing in the mutationType to better control when this logic happens.

Testing  (status: done)
- DataStoreIntegrationTests (includes V2 and CPK)
- DataStoreAuthTests (includes MultiAuth, Cognito Auth, IAM Auth)
- CPK sample app testing

PR for V2 --> https://github.com/aws-amplify/amplify-ios/pull/2361

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
+++ b/AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
@@ -41,12 +41,28 @@ extension Model {
             let name = modelField.name
 
             guard let value = modelFieldValue else {
-                // don't invalidate fields of type .model
-                // as we'll take care of this later on (see line 61)
+                // Special case for associated models when the value is `nil`, by setting all of the associated
+                // model's primary key fields (targetNames) to `nil`.
                 if case .model = modelField.type {
-                    continue
+                    let fieldNames = getFieldNameForAssociatedModels(modelField: modelField)
+                    for fieldName in fieldNames {
+                        // Only set to `nil` if it has not been set already. For hasOne relationships, where the
+                        // target name of the associated model is explicitly on this model as a field property, we
+                        // cannot guardteed which field is processed first, thus if there is a value for the explicit
+                        // field and was already set, don't overwrite it.
+                        if input[fieldName] == nil {
+                            // Always setting the value to `nil` is not necessary for create mutations, since leaving it
+                            // out will also reflect that there's no associated model. However, we always set it to `nil`
+                            // to account for the update mutation use cases where the caller may be de-associating the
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -41,12 +41,28 @@ extension Model {
...
-                // as we'll take care of this later on (see line 61)
+                // Special case for associated models when the value is `nil`, by setting all of the associated
+                // model's primary key fields (targetNames) to `nil`.
                 if case .model = modelField.type {
-                    continue
+                    let fieldNames = getFieldNameForAssociatedModels(modelField: modelField)
+                    for fieldName in fieldNames {
+                        // Only set to `nil` if it has not been set already. For hasOne relationships, where the
+                        // target name of the associated model is explicitly on this model as a field property, we
+                        // cannot guardteed which field is processed first, thus if there is a value for the explicit
```
thisisabhash, line 51: nit: guarantee

thisisabhash (dismissed):

thisisabhash (approved):

PR #2379, opened 2022-09-30
chore: Making `AmplifyAsyncTesting` package internal 

ruisebas:
*Description of changes:*

I've made `AmplifyAsyncTesting` an internal target instead, so it can continue to be used by all unit tests.

For integration tests, the approach I've taken is to just add a reference to the `AmplifyAsyncTesting/Sources/AsyncTesting` folder within the HostApp project, so it can then just use those classes without needing to duplicate code.

Something to call out about this approach is that it's explicitly adding the existing 3 files from that folder:
- AsyncExpectation.swift
- AsyncTesting.swift
- XCTestCase+AsyncTesting.swift

So if in the future we add more files in that folder, we'd need to manually create references for it in each Host app project. Bummer

- [ ] All unit tests pass


royjit (approved):
LGTM

PR #2397, opened 2022-10-05
fix: pin sdk transitive dependencies

atierian:
*Issue #, if available:*
N/A

*Description of changes:*
We pin our dependency on the AWS SDK for Swift to an exact version to ensure their new releases don't introduce breaking changes to our customers. However, the AWS SDK for Swift introduces two transitive dependencies `smithy-swift` and `aws-crt-swift`, which use the version requirement rule `from` rather than `exact`.  [[example](https://github.com/awslabs/aws-sdk-swift/blob/0.2.6/Package.swift#L323-L324)] 
New releases of `smithy-swift` and `aws-crt-swift` results in customer applications pulling in those versions, which can cause breaking behavioral changes. One such [change](https://github.com/awslabs/smithy-swift/pull/442) was discovered during development of this branch, leading to this PR.

Thankfully SPM has some sane rules around version requirement resolution. If we add direct dependencies on `smithy-swift` and `aws-crt-swift` in our own Package manifest, SPM will honor the more specific requirement. Thus adding `exact` will pin the versions of those dependencies.

- [ ] ~All unit tests pass~

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -7,7 +7,10 @@ let dependencies: [Package.Dependency] = [
     .package(url: "https://github.com/awslabs/aws-sdk-swift.git", exact: "0.2.6"),
     .package(url: "https://github.com/aws-amplify/aws-appsync-realtime-client-ios.git", from: "2.1.1"),
     .package(url: "https://github.com/stephencelis/SQLite.swift.git", exact: "0.12.2"),
-    .package(url: "https://github.com/mattgallagher/CwlPreconditionTesting.git", from: "2.1.0")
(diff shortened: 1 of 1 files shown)


atierian:
putting this in draft because although it works as expected when added locally, pinning isn't honored when pulling in Amplify remotely.

atierian:
Opened PRs in `AWS SDK for Swift` to pin dependency versions. 
Linking here for reference.
https://github.com/awslabs/aws-sdk-swift/pull/635
https://github.com/awslabs/aws-sdk-swift/pull/636

atierian:
Closing this PR as the underlying issue can't be resolved by changes to this repo.

PR #2414, opened 2022-10-10
chore: add non-throwing create publisher overload

atierian:
*Issue #, if available:*
N/A

*Description of changes:*

### Current State
`Amplify.Publisher.create` allows for converting an async throwing function (e.g. `func foo() async throws -> Int`) into a `AnyPublisher<Int, Error>`. This publisher can then be leveraged like so:
```swift
Amplify.Publisher.create { 
  try await foo()
}.sink(
  receiveCompletion: { competion in 
    // handle completion
  },
  receiveValue: { value in 
    // handle value
  }
)
```
The `receiveValue` block will be invoked upon `foo()` successfully returning an `Int` and the `receiveCompletion` block will be invoked if `foo()` decides to throw. 

This works perfectly when the function throws, but is less ergonomic and borderline confusing when an async non-throwing function is passed into `create(_:)`. Let's look at an example: 

```swift
func bar() async -> String { "Hello, world!" }

Amplify.Publisher.create { 
  await bar()
}.sink(
  receiveCompletion: { competion in 
    // when is this invoked?
  },
  receiveValue: { value in 
    // handle value
  }
)
```

`create(_:)` returns an `AnyPublisher<Success, Error>` regardless of whether the argument received throws or not. Resulting in the caller being required to include a `receiveCompletion` block, despite it being unnecessary. 

### Change
Nothing changes on the existing`create(_:)` API. This change adds an overload that accepts a non-throwing argument and returns `AnyPublisher<Success, Never>`

This allows for:
```swift
func bar() async -> String { "Hello, world!" }

Amplify.Publisher.create { 
  await bar()
}
.sink(receiveValue: { value in 
  // handle value
})
```

- [x] All unit tests pass

### DIFF
--- a/Amplify/Core/Support/Amplify+Publisher.swift
+++ b/Amplify/Core/Support/Amplify+Publisher.swift
@@ -50,6 +50,34 @@ public extension Amplify {
             .handleEvents(receiveCancel: { task.cancel() } )
             .eraseToAnyPublisher()
         }
+
+        /// Create a Combine Publisher for a given non-throwing Task.
+        ///
+        /// Example Usage
+        /// ```
+        /// let sink = Amplify.Publisher.create {
+        ///     try await Amplify.Auth.signOut()
+        /// }
+        ///     .sink(receiveValue: { value in
+        ///         // handle value
+        ///     })
+        /// ```
+        ///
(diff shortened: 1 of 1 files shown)


ameter (approved):
LGTM

PR #2421, opened 2022-10-11
fix(datastore): fix observe query unit test

atierian:
*Issue #, if available:*
N/A

*Description of changes:*
Moves initial saves to after initial query for `testObserveQuery`.

- [x] All unit tests pass


lawmicha:
ObserveQuery InitialQuery in parallel with saves may result in missed items from the saves happening in parallel.

The implementation change we'd like to do is subscribe to changes before intiial query with a change in how we process the changes.
- Scenario 1. Initial query has not finished, aggregate items to be applied to the initial query results. this is the new subscribe we’d like to add
- Scenario 2 (already exists). Initial query has finished && ! model synced, existing implementation does not change
- Scenario 3 (already exists). Initial query has finished && model Synced, existing implementation does not change

Once the initial query finishes, we take the changes observed while initial query has not finished and apply them to the snapshot. This is to allow observeQuery to be kicked off in parallel with the saves.

PR #2448, opened 2022-10-14
test(datastore): Lazy loading integration tests

lawmicha:
*Issue #, if available:*

*Table of Contents*:
1.  #2174
2. #2448 (You are here)
3. #2449

*Description of changes:*
This PR consists of the integration tests against a live backend using the expected codegenerated models for a variety of schemas being tested.

DataStore.save API will persist the data using the SQL adapter, then query for the data. The queried data as part of the save API call will continue to be eager loaded as the model is needed for syncing to the cloud. If the data is lazy loaded, then there's a problem model only containing the SQL identifier information `@@primaryKey` column value. We might be able to allow lazy loading on the save data flow with some more changes to the implementation.

DataStore.query API will allow lazy loading of SQL data

- [ ] All unit tests pass


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -135,7 +135,10 @@ extension Statement: StatementModelConvertible {
             // For example, when the value is the `id` of the Blog, then the field.isPrimaryKey is satisfied.
             // Every association of the Blog, such as the has-many Post is populated with the List with
             // associatedId == blog's id. This way, the list of post can be lazily loaded later using the associated id.
-            if let id = modelValue as? String, field.isPrimaryKey {
+            if let id = modelValue as? String,
+                (field.name == ModelIdentifierFormat.Custom.sqlColumnName || // this is the `@@primaryKey` (CPK)
+                 (schema.primaryKey.fields.count == 1 // or there's only one primary key (not composite key)
+                  && schema.primaryKey.indexOfField(named: field.name) != nil)) { // and this field is the primary key
```
lawmicha, line 141: bug where lazy List was not being created due to `field.primaryKey` of the generated code was not returning true for Post4V2. this change can be applied directly on `main` and cherry picked over to `v1`.

lawmicha: https://github.com/aws-amplify/amplify-swift/issues/2482

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -59,6 +59,8 @@ extension Model {
                             input.updateValue(nil, forKey: fieldName)
                         }
                     }
+                } else if case .collection = modelField.type { // skip all "has-many"
+                    continue
```
lawmicha, line 63: bug where collections's are being set to `nil` in the GraphQL input, so for example, when creating a Post and saving it through DataStore. the sync process will translate the post to the GraphQL input variables containing
```
{
  "id": "postid123",
   "comments": null
}
```

The problem is this fails as there's no input type "comments". We should skip has-many relationships like in this code snippet. This can be done in `main` and `v1`

lawmicha: https://github.com/aws-amplify/amplify-swift/issues/2483

PR #2452, opened 2022-10-14
fix(Storage): Addressing issues with race conditions

ruisebas:
*Description of changes:*
- Making `AtomicDictionary` a class again with an `NSLock` mechanism, as the `actor` approach created issues when the result was immediately dispatched without having time to register the listener (due to being an asynchronous call made inside the synchronous `subscribe` method)
- Wrapping operation listeners in closures that don't retain the adapter, so they are not kept around after the operations complete.
- Fixing `StorageBackgroundEventsRegistryTests.testRegisteringAndUnregister` which was attempting to resume a continuation before it finished being stored in the registry.

- [ ] All unit tests pass

### DIFF
--- a/Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift
+++ b/Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift
@@ -69,11 +69,11 @@ public class AmplifyInProcessReportingOperationTaskAdapter<Request: AmplifyOpera
     public init(operation: AmplifyInProcessReportingOperation<Request, InProcess, Success, Failure>) {
         self.operation = operation
         self.childTask = ChildTask(parent: operation)
-        resultToken = operation.subscribe(resultListener: {[weak self] result in
+        resultToken = operation.subscribe(resultListener: { [weak self] result in
             guard let self = self else { return }
             self.resultListener(result)
         })
-        inProcessToken = operation.subscribe(inProcessListener: {[weak self] inProcess in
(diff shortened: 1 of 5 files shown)


royjit (approved):
LGTM

Amplify/Core/Support/AmplifyTask+OperationTaskAdapters.swift
```diff
@@ -69,8 +69,14 @@ public class AmplifyInProcessReportingOperationTaskAdapter<Request: AmplifyOpera
...
-        resultToken = operation.subscribe(resultListener: resultListener)
-        inProcessToken = operation.subscribe(inProcessListener: inProcessListener)
+        resultToken = operation.subscribe(resultListener: {[weak self] result in
+            guard let self = self else { return }
+            self.resultListener(result)
+        })
+        inProcessToken = operation.subscribe(inProcessListener: {[weak self] inProcess in
+            guard let self = self else { return }
+            self.inProcessListener(inProcess)
+        })
```
atierian, line 79: nit
```suggestion
        resultToken = operation.subscribe(resultListener: { [weak self] result in
            guard let self = self else { return }
            self.resultListener(result)
        })
        inProcessToken = operation.subscribe(inProcessListener: { [weak self] inProcess in
            guard let self = self else { return }
            self.inProcessListener(inProcess)
        })
```

atierian (dismissed):

atierian (approved):

royjit (approved):

PR #2481, opened 2022-10-19
implements new Storage API for handleEventsForBackgroundURLSession

brennanMKE:
*Issue #, if available:*

N/A

*Description of changes:*

- [ ] All unit tests pass


AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Support/Internal/StorageBackgroundEventsRegistry.swift
```diff
@@ -7,30 +7,44 @@
 
 import Foundation
 
+extension Notification.Name {
```
royjit, line 10: This notification was setup just for unit test? Looks like this will be fired in the production app as well. We should think about a better way to make this class testable?

AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Support/Internal/StorageBackgroundEventsRegistry.swift
```diff
@@ -7,30 +7,44 @@
...
+    static let StorageBackgroundEventsRegistryWaiting = Notification.Name("StorageBackgroundEventsRegistryWaiting")
+}
+
 /// Background events registry.
 ///
 /// Discussion:
 ///    Multiple URLSession instances could be running background events with their own unique identifier. Those can be run
 ///    independently of the Amplify Storage plugin and this function will indiciate if it will handle the given identifier.
-class StorageBackgroundEventsRegistry {
```
royjit, line 15: what is the point in making this an actor? I might be wrong, but does making this actor help in making the static variables protected against multiple access?

AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Support/Internal/StorageBackgroundEventsRegistry.swift
brennanMKE: What other options do you suggest?

royjit: The test worked without the notification expectation:

```swift
    func testRegisteringAndUnregister() async throws {
        let identifier = UUID().uuidString
        let otherIdentifier = UUID().uuidString
        await StorageBackgroundEventsRegistry.shared.register(identifier: identifier)

        let done = asyncExpectation(description: "done")
        Task {
            let handled = await StorageBackgroundEventsRegistry.shared.handleEventsForBackgroundURLSession(identifier: identifier)
            await done.fulfill()
            XCTAssertTrue(handled)
        }

        let didContinue = await handleEvents(for: identifier)
        XCTAssertTrue(didContinue)
        await waitForExpectations([done])

        let otherDone = asyncExpectation(description: "other done")

        Task {
            let otherHandled = await StorageBackgroundEventsRegistry.shared.handleEventsForBackgroundURLSession(identifier: otherIdentifier)
            await otherDone.fulfill()
            XCTAssertFalse(otherHandled)
        }

        let didNotContinue = await handleEvents(for: otherIdentifier)
        XCTAssertFalse(didNotContinue)
        await waitForExpectations([otherDone])
    }
```

AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Support/Internal/StorageBackgroundEventsRegistry.swift
royjit: I see that the static variables are converted to instance variables, resolving this comment.

AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Support/Internal/StorageBackgroundEventsRegistry.swift
brennanMKE: You have to run it on macOS many times repeatedly to get it to fail. Without the notification it is a flaky test which will fail based on timing.

royjit: This issue can then happen in the real app as well right? What if the URLSession call back reach before we setup the continuation?

royjit: Another option could be to create an overloaded function just for test:

```
    func handleEventsForBackgroundURLSession(identifier: String) async -> Bool {
        return await handleEventsForBackgroundURLSession(identifier: identifier, notifyTest: nil)
    }

    func handleEventsForBackgroundURLSession(identifier: String, notifyTest: ((String) -> Void)? ) async -> Bool {
        guard self.identifier == identifier else { return false }
        return await withCheckedContinuation { (continuation: CheckedContinuation<Bool, Never>) in
            self.continuation = continuation
            if let notifyTest = notifyTest {
                notifyTest(identifier)
            }
        }
    }
```

akshaaatt:
When can we expect this to released?

PR #2488, opened 2022-10-20
fix(Auth): Making improvements to federation flow

harsh62:
*Issue #, if available:*
* Adding the ability to clear credentials when in an error state
* Fixing fetchAuthSession and federation workflows when in an error state

*Description of changes:*

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/Federation/ClearFederationToIdentityPool.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/Federation/ClearFederationToIdentityPool.swift
@@ -18,8 +18,18 @@ struct ClearFederationToIdentityPool: Action {
 
         let event: StateMachineEvent
         do {
-            try await credentialStoreClient?.deleteData(type: .amplifyCredentials)
-            event = AuthenticationEvent.init(eventType: .clearedFederationToIdentityPool)
+            // Check if we have credentials are of type federation, otherwise throw an error
+            if case .amplifyCredentials(let credentials) = try await credentialStoreClient?.fetchData(
+                type: .amplifyCredentials),
+               case .identityPoolWithFederation = credentials {
+
+                try await credentialStoreClient?.deleteData(type: .amplifyCredentials)
+                event = AuthenticationEvent.init(eventType: .clearedFederationToIdentityPool)
+            } else {
+                let error = AuthenticationError.service(message: "Unable to find credentials to clear for federation.")
+                event = AuthenticationEvent.init(eventType: .error(error))
+                logError("\(#fileID) Sending event \(event.type) with error \(error)", environment: environment)
+            }
 
         } catch {
             let error = AuthenticationError.unknown(message: "Unable to clear credential store: \(error)")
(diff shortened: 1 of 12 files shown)


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/Federation/ClearFederationToIdentityPool.swift
```diff
@@ -18,8 +18,18 @@ struct ClearFederationToIdentityPool: Action {
 
         let event: StateMachineEvent
         do {
-            try await credentialStoreClient?.deleteData(type: .amplifyCredentials)
-            event = AuthenticationEvent.init(eventType: .clearedFederationToIdentityPool)
+            // Check if we have credentials are of type federation, otherwise throw an error
+            if case .amplifyCredentials(let credentials) = try await credentialStoreClient?.fetchData(
```
royjit, line 22: This action should not be called if there is not federated credentials right?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/States/AuthorizationState.swift
```diff
@@ -19,8 +19,10 @@ enum AuthorizationState: State {
 
     case clearingFederation
 
-    case federatingToIdentityPool(FetchAuthSessionState,
-                                  FederatedToken)
+    case federatingToIdentityPool(
+        FetchAuthSessionState,
+        FederatedToken,
+        existingCredentials: AmplifyCredentials)
```
royjit, line 25: Why was this `existingCredentials` added? Not sure where this is used

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/Federation/ClearFederationToIdentityPool.swift
royjit: I think this is to clear an error state?

royjit (approved):
LGTM

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/States/AuthorizationState.swift
harsh62: Had a discussion offline..

> So this was added because when session error happens during federation, it expects to have a credential passed to it, so that fetchAuthSession can determine the type of retry that could happen.. case sessionError(FetchSessionError, AmplifyCredentials)

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/Federation/ClearFederationToIdentityPool.swift
harsh62: Had a discussion offline. 

>  I can avoid using the action to determine clearFederation is needed or not, by utilizing the credentials associated with the session error... But there could be a situation, where we don't get a sessionError and the user won't be able to clear..

Issue #2471, opened 2022-10-18
Amplify drops event attributes, quota in AWS iOS SDK is different from PinPoint documentation


frankiesimon:
### Describe the bug

We started noticing that our events (sent from iOS client via Amplify to AWS PinPoint and forwarded with Kinesis to a lambda function) are missing random attribute fields.

After talking to both AWS support and `ruisebas` on the Amplify discord this seems to be the issue.

AWS iOS SDK drops attributes from an event if the combined number of metrics and attributes is 50 or above:
https://github.com/aws-amplify/aws-sdk-ios/blob/main/AWSPinpoint/AWSPinpointEvent.m#L114-L126

Meanwhile AWS PinPoint has a documented quota of 40 attributes/metrics which is apparently checked separately:
https://docs.aws.amazon.com/pinpoint/latest/developerguide/quotas.html#quotas-events

While the code removes "excess" attributes to fulfill "attributes+metrics<50", the actual payload can still fail if the number of *attributes* is *40* or above.

Here's debug output from such a case

(requires)
```
import AWSCore
...
AWSDDLog.sharedInstance.logLevel = .verbose
AWSDDLog.add(AWSDDTTYLogger.sharedInstance)
```

```
Response body:
{"Results":{"95B0C5F7-584B-44A7-B801-F77A6D736836":{"EndpointItemResponse":{"StatusCode":202,"Message":"Accepted"},"EventsItemResponse":{"525FD708-4105-4C9D-BF2E-7CF83AE386A6":{"StatusCode":400,"Message":"Number of attributes per event submitted should be less than 40."}}}}}

PutEventsResponse received: [<AWSPinpointTargetingPutEventsResponse: 0x2817c5c30> {
[...]

Issue #2401, opened 2022-10-06
PinPoint events are not cached while offline and are not sent later


frankiesimon:
### Describe the bug

When my app is offline events that are recorded are not cached anywhere and are not sent when the app goes back online or in a subsequent app session (restarting it).

When debugging I got to `AWSPinpointAnalyticsPlugin+ClientBehavior.swift -> flushEvents()`.

```
public func flushEvents() {
        if !isEnabled {
            log.warn("Cannot flushEvents. Analytics is disabled. Call Amplify.Analytics.enable() to enable")
            return
        }

        pinpoint.submitEvents().continueWith { (task) -> Any? in
            guard task.error == nil else {
                // TODO: some error mapping
                let error = task.error! as NSError
                Amplify.Hub.dispatchFlushEvents(AnalyticsErrorHelper.getDefaultError(error))
                return nil
            }

            if let pinpointEvents = task.result as? [AWSPinpointEvent] {
                // TODO: revist this, this is exposing internal implementation
                Amplify.Hub.dispatchFlushEvents(pinpointEvents)
            }

            return nil
        }
    }
```
If internet is down the task has an error and there are no `pinpointEvents` parsed from the `task.result`, it flushes an "error" event but ignores the rest.
As far as I can tell this is what loses the events. If there is some caching behavior I didn't see it.

I'm having trouble understanding if this is a bug or if this was never implemented.

We are porting our code from `AWSMobileAnalytics` and this behavior was definitely supported there. Documented in the FAQ here:

https://aws.amazon.com/mobileanalytics/faqs/

> Q: Is data cached when a user’s device is offline?
Yes, when using the AWS Mobile SDK, data is cached on the user’s device and is uploaded when a network connection is next established.

It's a big deal for us and I couldn't understand if this is a known regression in Amplify iOS.

By the way, it seems that Amplify Android SDK does have support for "retryable errors" (though we also found a bug in the implementation there preventing offline events from being cached).

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Setup Amplify with pinpoint
2. Start app without internet and record events
3. Go online and record more events
4. Offline events were not sent.
```

### Expected behavior

Expected offline events to be cached and sent later on.

### Amplify Framework Version

1.28.1

### Amplify Categories

Analytics, Auth

### Dependency manager

Swift PM

### Swift version

5.6.1

### CLI version

10.0.0

### Xcode version

13.4.1 13F100

### Is this a regression?

Yes

[...]

PR #2497, opened 2022-10-25
fix(Analytics): Fixing Analytics issues

ruisebas:
*Description of changes:*

This PR addresses four issues with the current Analytics implementation:

----

#### If the `PutEvents` operation fails, the events in the database are updated but never deleted.

If the `pinpoint.putEvents(input:)` call inside the [`submit(pinpointEvents:endpointProfile)` method](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Analytics/EventRecorder.swift#L104-L105) throws an error, we handle it [inside a `catch` block](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Analytics/EventRecorder.swift#L155) that updates the events accordingly (i.e. either increases the counter or marks it as dirty, depending on the type of error thrown) and we [rethrow the error](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Analytics/EventRecorder.swift#L166).

But because the enclosing [`processBatch(_:endpointProfile)` method](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Analytics/EventRecorder.swift#L95) also throws, the [`try storage.removeFailedEvents()` statement](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/Dependency/Pinpoint/Analytics/EventRecorder.swift#L97) is never executed, as the error gets propagated.

This causes each failed event to get property updated, but since we fail to act on it we will keep trying to submit dirty events and/or events that exceeded their retry count until we get a successful response from Pinpoint.

So i've added a new `catch` block to attempt to clean up the events in the case we get an error, and then rethrow it.

#### If the `PutEvents` operation succeeds but **all** events are rejected, we don't recognize it as a failure.

When Pinpoint rejects an event, it does not cause the `pinpointClient.putEvents(input:)` to throw an error, but to state which event was rejected and its reason inside a `EventItemResponse` object that is part of the response.

I'd argue that if _all_ events in a batch are rejected, then it's likely there's an issue going on and we should consider the attempt a failure, instead of carrying on with the next batch. 
So I've added a small validation that will throw an error if 0 events were successful, which should interrupt the submission and allow the consumer to act.

####  [If the device lacks internet, all events that are submitted fail and are discarded](https://github.com/aws-amplify/amplify-swift/issues/2401)
If we attempt to call `pinpointClient.putEvents(input:)` and the device does not have internet, the request fails and it's considered a retryable error and the events retry counter is updated.

If the device stays offline during the next 3 attempts, the events will be discarded since they will exceed their retry-limit. This is not desired, as connectivity-related errors should be retryable indefinitely (and Android already does this).
So I've implemented a new check that first validates if the error was caused due to connectivity issues, and it so just immediately rethrow the error without doing updates to the database.

#### [Pinpoint event quota limits are incorrect](https://github.com/aws-amplify/amplify-swift/issues/2471)
We're currently validating that an event cannot exceed 50 metrics and attributes combined, and that each attribute value can be up to 1000 in length. These values are not aligned with [the Pinpoint documentation](https://docs.aws.amazon.com/pinpoint/latest/developerguide/quotas.html#quotas-events), which states that each event can have up to 40 metrics and 40 attributes, and that the maximum value length is 200.

I'm also exposing these values in a `AWSPinpointAnalytics.Constants.Event` struct, so that consumers can use them in their own validations if they want to.

----
#### Additional misc changes

I added extra `log.verbose` logs to match what we had on V1/legacy SDK, which provides additional debugging information, as well as other relevant information (e.g. if an event has exceeded its retry count)

I also created a `AmplifyStringConvertible` protocol to easily "prettify" the SDK objects when they are printed, and defined a custom `description` for the `PinpointEvent`s as well. 

----

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/AWSPinpointAnalyticsPlugin+ClientBehavior.swift
+++ b/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/AWSPinpointAnalyticsPlugin+ClientBehavior.swift
@@ -100,7 +100,7 @@ extension AWSPinpointAnalyticsPlugin {
         }
 
         // Do not attempt to submit events if we detect the device is offline, as it's gonna fail anyway
-        guard networkMonitor.currentPath.status == .satisfied else {
+        guard networkMonitor.isOnline else {
             let errorMessage = "Cannot flushEvents. \(AnalyticsPluginErrorConstant.deviceOffline.errorDescription)"
             log.error(errorMessage)
             Amplify.Hub.dispatchFlushEvents(.unknown(errorMessage))
--- a/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/AWSPinpointAnalyticsPlugin+ClientBehavior.swift
+++ b/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/AWSPinpointAnalyticsPlugin+ClientBehavior.swift
@@ -46,7 +46,7 @@ extension AWSPinpointAnalyticsPlugin {
         Task {
             do {
                 try await pinpoint.record(pinpointEvent)
-                Amplify.Hub.dispatchRecord(event)
+                Amplify.Hub.dispatchRecord(pinpointEvent.asAnalyticsEvent())
             } catch {
                 Amplify.Hub.dispatchRecord(AnalyticsErrorHelper.getDefaultError(error))
             }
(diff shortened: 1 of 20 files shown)


phantumcode (approved):
LGTM

Issue #2167, opened 2022-08-17
Error happened when I return back from offline mode and turn on the wifi, Also a crash happened if I restarted the data store again


medhatIbsais-Harri:
### Describe the bug

When using DataStore, after configuring the Amplify and start syncing, try to turn off the wifi, then turn it on again, an error will be displayed in XCode logs, if you then forced syncing your DataStore by stopping the DataStore and started it again a crash will happen.

### Steps To Reproduce

```swift
1. Open your app normally
2. wait until the DataStore and Amplify to be configured and synced
3. turn off the wifi
4. turn it on again
5. An error will be logged that there is an error caused by connection
6. force syncing the DataStore (stop DataStore, then start DataStore)
7. A crash will happen in Storage Engine

        Amplify.DataStore.stop { _ in
            Amplify.DataStore.start { _ in

            }
        }
```

### Expected behavior

To sync data normally.

### Amplify Framework Version

1.28.0

### Amplify Categories

DataStore

### Dependency manager

Cocoapods

### Swift version

5.0

### CLI version

Not Installed

### Xcode version

13.3.1

### Relevant log output

```shell
Error Log:

User receives a network connection status: true
************************************** HubPayload(eventName: "DataStore.networkStatus", context: nil, data: Optional(AmplifyPlugins.NetworkStatusEvent(active: true)))
2022-08-17 20:30:01.038161+0300 TeamHub[9288:290017] [RealtimeConnectionProvider] Realtime connection is stale, disconnecting.
2022-08-17 20:30:01.039196+0300 TeamHub[9288:290017] ConnectionProviderError.connection
2022-08-17 20:30:01.039787+0300 TeamHub[9288:290017] ConnectionProviderError.connection
2022-08-17 20:30:01.042814+0300 TeamHub[9288:290017] ConnectionProviderError.connection
2022-08-17 20:30:01.055788+0300 TeamHub[9288:290070] [AWSModelReconciliationQueue] receiveCompletion: error: DataStoreError: Subscription item event failed with error
Caused by:
APIError: Subscription item event failed with error
Caused by:
connection
2022-08-17 20:30:01.060776+0300 TeamHub[9288:290017] [AWSDataStorePlugin] StorageEngine completed with error: DataStoreError: Subscription item event failed with error
Caused by:
DataStoreError: Subscription item event failed with error
Caused by:
APIError: Subscription item event failed with error
Caused by:
connection

------------------------------------------------------------------------------------------------

[...]

PR #2517, opened 2022-10-28
fix(datastore): fix stop then start API call pattern

manaswi223:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-swift/issues/2167
*Description of changes:*

This PR fixes a problem with the following code snippets:
```swift
Amplify.DataStore.stop { _ in 
    Amplify.DataStore.start { _ in       
    }
}
```
`stop` enters the queue.sync block and calls its completion handler which calls `start`. `start` attempts to enter the queue and waits for `stop` to finish, which will never happen. This PR prevents a crash by calling `completion` from a different thread.

`main` PR https://github.com/aws-amplify/amplify-swift/pull/2529

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
@@ -281,15 +281,15 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
                 dispatchedModelSynced.set(false)
             }
             if storageEngine == nil {
-                workQueue.async {
+                queue.async {
                     completion(.successfulVoid)
                 }
                 return
             }
 
             storageEngine.stopSync { result in
-                self.workQueue.async {
-                    self.storageEngine = nil
+                self.storageEngine = nil
+                self.queue.async {
                     completion(result)
                 }
             }
@@ -312,14 +312,14 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
                 dispatchedModelSynced.set(false)
             }
             if storageEngine == nil {
-                workQueue.async {
+                queue.async {
                     completion(.successfulVoid)
                 }
                 return
             }
             storageEngine.clear { result in
-                self.workQueue.async {
-                    self.storageEngine = nil
+                self.storageEngine = nil
+                self.queue.async {
                     completion(result)
                 }
             }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
@@ -281,14 +281,17 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
                 dispatchedModelSynced.set(false)
             }
             if storageEngine == nil {
-
-                completion(.successfulVoid)
+                workQueue.async {
+                    completion(.successfulVoid)
+                }
                 return
             }
 
             storageEngine.stopSync { result in
-                self.storageEngine = nil
-                completion(result)
+                self.workQueue.async {
+                    self.storageEngine = nil
+                    completion(result)
+                }
             }
         }
     }
@@ -309,12 +312,16 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
                 dispatchedModelSynced.set(false)
             }
             if storageEngine == nil {
-                completion(.successfulVoid)
+                workQueue.async {
+                    completion(.successfulVoid)
+                }
                 return
             }
             storageEngine.clear { result in
-                self.storageEngine = nil
-                completion(result)
+                self.workQueue.async {
+                    self.storageEngine = nil
+                    completion(result)
+                }
             }
         }
     }
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -46,6 +46,8 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
 
     var storageEngine: StorageEngineBehavior!
     var storageEngineInitQueue = DispatchQueue(label: "AWSDataStorePlugin.storageEngineInitQueue")
+    let workQueue = DispatchQueue(label: "AWSDataStorePlugin.storageEngineInitQueue",
+                                  target: DispatchQueue.global())
     var storageEngineBehaviorFactory: StorageEngineBehaviorFactory
 
     var iStorageEngineSink: Any?
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
@@ -46,8 +46,7 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
 
     var storageEngine: StorageEngineBehavior!
     var storageEngineInitQueue = DispatchQueue(label: "AWSDataStorePlugin.storageEngineInitQueue")
-    let workQueue = DispatchQueue(label: "AWSDataStorePlugin.storageEngineInitQueue",
-                                  target: DispatchQueue.global())
+    let queue = DispatchQueue(label: "AWSDataStorePlugin.queue", target: DispatchQueue.global())
     var storageEngineBehaviorFactory: StorageEngineBehaviorFactory
 
     var iStorageEngineSink: Any?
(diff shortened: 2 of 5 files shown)


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin.swift
```diff
@@ -46,6 +46,8 @@ final public class AWSDataStorePlugin: DataStoreCategoryPlugin {
 
     var storageEngine: StorageEngineBehavior!
     var storageEngineInitQueue = DispatchQueue(label: "AWSDataStorePlugin.storageEngineInitQueue")
+    let workQueue = DispatchQueue(label: "AWSDataStorePlugin.storageEngineInitQueue",
+                                  target: DispatchQueue.global())
```
lawmicha, line 50: how about just "queue"?  and update `label` to differentiate from the one above
```
let queue = DispatchQueue(label: "AWSDataStorePlugin.queue", target: DispatchQueue.global())` 
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
```diff
@@ -15,6 +15,7 @@ import Combine
 @testable import AmplifyTestCommon
 @testable import AWSDataStoreCategoryPlugin
 
+// swiftlint:disable all
```
lawmicha, line 18: should disable only what you need instead of disabling everything, for example if the file is too long then you can either create another file and extend the class `DataStoreEndToEndTests` 
```
// File name can be something like `DataStoreEndToEndTests+Restart.swift`
extension DataStoreEndToEndTests {
  func testStopStart() 
  func testClearStart()
 etc..
```

or be more specific abotu what we are disabling, ie 
```
// swiftlint:disable type_body_length
// swiftlint:disable file_length
// swiftlint:disable line_length
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
```diff
@@ -366,6 +367,175 @@ class DataStoreEndToEndTests: SyncEngineIntegrationTestBase {
...
+    ///  a DispatchQueue
+    ///
+    /// - Given:  DataStore has just configured, but not started
+    /// - When:
+    ///    - DataStore.stop
+    ///    - Followed by DataStore.start in the completion of the stop
+    /// - Then:
+    ///    - Datastore should be started successfully
+    ///
+    func testConfigureAmplifyThenStopStartonDispatch() throws {
```
lawmicha, line 444: what are we testing here by running the code on the background thread? should we just remove this test?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -281,14 +281,17 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
...
+                    completion(.successfulVoid)
+                }
                 return
             }
 
             storageEngine.stopSync { result in
-                self.storageEngine = nil
-                completion(result)
+                self.workQueue.async {
+                    self.storageEngine = nil
```
lawmicha, line 292: we should set this to `nil` outside of the async block, because the sync block performs work serially, so that access to `storageEngine` is serial, thus putting it inside the async block will defeat that purpose

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSDataStorePluginTests.swift
```diff
@@ -61,6 +61,92 @@ class AWSDataStorePluginTests: XCTestCase {
...
+                case .finished:
+                    finishNotReceived.fulfill()
+                case .failure(let error):
+                    XCTFail("Error \(error)")
+                }
+            } receiveValue: { _ in }
+
+            storageEngine.responders[.startSync] = StartSyncResponder { _ in
+                currCount = self.expect(startExpectation, currCount, 2)
+            }
```
lawmicha, line 102: does this need to be here, and not with the other storageEngine responder set up at line 70?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSDataStorePluginTests.swift
```diff
@@ -61,6 +61,92 @@ class AWSDataStorePluginTests: XCTestCase {
...
+        }
+        let dataStorePublisher = DataStorePublisher()
+        let plugin = AWSDataStorePlugin(modelRegistration: TestModelRegistration(),
+                                        storageEngineBehaviorFactory: storageEngineBehaviorFactory,
+                                        dataStorePublisher: dataStorePublisher,
+                                        validAPIPluginKey: "MockAPICategoryPlugin",
+                                        validAuthPluginKey: "MockAuthCategoryPlugin")
+        let finishNotReceived = expectation(
+            description: "publisher should not receive .finished completion event after stop() is called")
+        finishNotReceived.isInverted = true
```
lawmicha, line 85: can this be moved above, with the rest of the expectations?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
lawmicha: the workaround for the customer without this PR where the DispatchQueue.async is performed before the second call, not before both calls, like this:
```
DataStore.stop {   
  DispatchQueue.async {
      DataStore.start { 
       }
   }
}
```

We can provide this info to the customer

manaswi223: I feel this test is to make sure that running consecutive start-stop on a single thread (DispatchQueue) 🧵 is still successful. 
I can remove this if we don't require!

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginTests/Core/AWSDataStorePluginTests.swift
manaswi223: not required, moved it up!

lawmicha:
On `main`, this problem is avoided because of the wrapper to async await APIs. I'm guessing the following code snippet works as intended:
```swift
try await Amplify.DataStore.stop()
try await Amplify.DataStore.start()
```
Since there is no `completion` callback being executed within the `stop` call, the issue described in this PR is not possible in `main`. 
However, the implementation of the async await APIs are simply a wrapper ontop of the existing callback APIs, this means we should be able to reproduce it in `main` is using the plugin directly (as shown in the unit test of this PR). 

I think we should cherry-pick this change over to `main` so we're not second guessing whether the issue exists or not in `main`, have the unit test for the same scenario, and have the integration test against the async await APIs

AmplifyPlugins/DataStore/AWSDataStoreCategoryPluginIntegrationTests/DataStoreEndToEndTests.swift
lawmicha: i don't think this test is required because whether we decide to run it on a new queue or run it in current executing thread of the test, it's still just a thread we're running on. I test that may be useful to have is to have the workaround as a test (see second comment of this comment thread) to show that the implementation changes in this PR does not cause the workaround to no longer work

lawmicha:
can you confirm that these existing integration tests are all passing?
- [ ] Ran AWSDataStorePluginIntegrationTests (includes V2 and CPK)
- [ ] Ran AWSDataStorePluginAuthIntegrationTests (includes MultiAuth, Cognito Auth, IAM Auth)

lawmicha (approved):
LGTM, please don't merge until we run the integration tests in this PR and make sure other tests are all still passing. We should also cherry-pick this over to `main` to keep the codebase aligned as much as possible

manaswi223:
Integration Tests are failing with this PR:
1. AWSDataStoreCategoryPluginAuthIntegrationTests ❌:

    a. AWSDataStoreCategoryPluginIAMAuthIntegrationTests 
    b. AWSDataStoreMultiAuthCombinationTests 
    c. AWSDataStoreMultiAuthSingleRuleTests 
    d. AWSDataStoreMultiAuthThreeRulesTests 
    e. AWSDataStoreMultiAuthTwoRulesTests 

2. AWSDataStoreCategoryPluginIntegrationTests

    a. AWSDataStorePrimaryKeyIntegrationTests ❌
    
    b. V2 Connection ❌:
b1. DataStoreConnectionScenario3aV2Tests
b2. DataStoreConnectionScenario3V2Tests
b3. DataStoreConnectionScenario4V2Tests
b4. DataStoreConnectionScenario5V2Tests
b5. DataStoreConnectionScenario7V2Tests
b6. DataStoreConnectionScenario8V2Tests

lawmicha:
> Integration Tests that are failing with this PR:
> 
> 1. AWSDataStoreCategoryPluginAuthIntegrationTests ❌:
>    a. AWSDataStoreMultiAuthCombinationTests
>    b. AWSDataStoreMultiAuthSingleRuleTests
>    c. AWSDataStoreMultiAuthThreeRulesTests
>    d. AWSDataStoreMultiAuthTwoRulesTests
> 2. AWSDataStoreCategoryPluginIntegrationTests
>    a. AWSDataStorePrimaryKeyIntegrationTests - Flaky

So i'm coming back to this PR to run the integ tests and these appear to run fine for me
image

if the entire test class is failing, i think is an indiciation that something is misaligned with the backend. One issue you mentioned earlier was when you didn't know what to enter for the OIDC configuration so we never actually ran `amplify push`. Is this comment referring to before or after that problem we solved? 

Let's at a minimum update the README.md to instructions that make it clearer to provision the backend

lawmicha (approved):

PR #2526, opened 2022-11-01
feat(api): adds selection set customization

drochetti:
## Custom selection set for GraphQL queries in the API category

This feature will enable customers to define which associations will be included in selection set when querying data using the static typed models. Previously developers had no control over that and had to fallback to untyped/string-based GraphQL queries when they needed more control. This feature enables fine control with full type information of which associations should be included in a query.

A quick example of the added capability is represented in the following code snippet:

```swift
let postsWithComments = try await Amplify.API.query(request: .list(Post.self, includes: { [$0.comments] }))
```
Developers can go as deep as needed and/or add multiple associations. Next I'll go into more details on how this was accomplished to provide folks reviewing this PR more information about the implementation details and decisions.

---

- [ ] All unit tests pass

---


Amplify/Core/Error/DataError.swift
```diff
@@ -0,0 +1,78 @@
...
+    /// Indicates that expected data could not be loaded due to the
+    /// lack of integrity.
+    case dataIntegrity(underlyingError: Error? = nil)
+
+    /// Indicates an unexpected scenario happened and the specific
+    /// error could not be resolved.
+    case unknown(ErrorDescription, RecoverySuggestion, Error)
+}
+
+extension DataError: AmplifyError {
```
lawmicha, line 24: we currently have `CoreError` which is returned on List operations. I think we should merge the two into one, and name it either CoreError or DataError as you have it here

Amplify/Categories/DataStore/Model/PropertyPath.swift
```diff
@@ -0,0 +1,126 @@
...
+        return ModelType.self
+    }
+
+    public func isRoot() -> Bool {
+        return metadata.parent == nil
+    }
+
+    public func id(_ name: String = "id") -> FieldPath<String> {
+        FieldPath(name: name, parent: self)
+    }
```
lawmicha, line 105: Is there a special use case for this or just visually appealing to have it codegenerated as "id"? When hand-writing the schema with composite keys, both keys were not "id" so they are both `string(<fieldName>)`, which means in this case that the primary key of a model with composite key doesn't show which of the fields are its primary keys in the ModelPath extension. Should we just remove this completely? Currently, I haven't added the check for "is primaryKey and field value is 'id'" and simply generated "string(<fieldName>)" for all string fields including properties called "id"

Amplify/Categories/DataStore/Model/PropertyPath.swift
```diff
@@ -0,0 +1,126 @@
...
+
+    public func string(_ name: String) -> FieldPath<String> {
```
lawmicha, line 107: missing FieldPath<Double>

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/SelectionSet.swift
```diff
@@ -44,14 +52,14 @@ extension SelectionSet {
                 let associatedModelName = field.associatedModelName,
                 let schema = ModelRegistry.modelSchema(from: associatedModelName) {
                 let child = SelectionSet(value: .init(name: field.name, fieldType: .model))
-                child.withModelFields(schema.graphQLFields)
+                child.withModelFields(schema.primaryKey.fields)
```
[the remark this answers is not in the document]
drochetti: Sounds good! I reviewed your approach and will keep this PR like this since it'll be fixed once you apply yours on top of it.

AmplifyPlugins/Core/AWSPluginsCore/Model/GraphQLRequest/GraphQLRequest+Model.swift
```diff
@@ -233,11 +257,17 @@ extension GraphQLRequest: ModelGraphQLRequestFactory {
...
+                                      where predicate: QueryPredicate? = nil,
+                                      includes: IncludedAssociations<M> = { _ in [] },
+                                      limit: Int? = nil) -> GraphQLRequest<List<M>> {
         var documentBuilder = ModelBasedGraphQLDocumentBuilder(modelSchema: modelType.schema, operationType: .query)
         documentBuilder.add(decorator: DirectiveNameDecorator(type: .list))
 
+        if let modelPath = modelType.rootPath as? ModelPath<M> {
+            let associations = includes(modelPath)
+            documentBuilder.add(decorator: IncludeAssociationDecorator(associations))
+        }
```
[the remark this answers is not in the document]
drochetti: That makes sense. So should I leave it like this here since your PR will change it?

Amplify/Categories/DataStore/Model/PropertyPath.swift
drochetti: That's a good callout. It made sense before when we had ids based on the naming convention. That not being the case anymore calls for a different approach, I think we can simply use the `string()` function.

lawmicha:
This PR has been merged with https://github.com/aws-amplify/amplify-swift/pull/2583 to `data-dev-preview`

Issue #2523, opened 2022-10-31
Flutter Todo sync during network off fails and is dropped


lawmicha:
### Describe the bug

I have followed the below tutorial to replicate:
[+] https://docs.amplify.aws/start/getting-started/setup/q/integration/flutter/ 

Version 0.6.8 and version 0.6.9

### Steps To Reproduce

```swift
1. Added todo1 -> synced
2. Disconnected wifi
3. Added offline-todo2, offline-todo3, offline-todo4, offline-todo5
4. reconnected
5. offline-todo3 is synced.
6. Added reconnected-todo7.
7. all the previously added todos except offline-todo2 are synced to datastore.

```

### Expected behavior

Data is not dropped

### Amplify Framework Version

v1

### Amplify Categories

DataStore

### Dependency manager

Cocoapods

### Swift version

N/A

### CLI version

N/A

### Xcode version

N/A

### Relevant log output

```shell
Successfully saved model: FlutterSerializedModel(values: ["updatedAt": Amplify.JSONValue.null, "__typename": Amplify.JSONValue.string("Todo"), "isComplete": Amplify.JSONValue.boolean(false), "description": Amplify.JSONValue.string("when connected!"), "createdAt": Amplify.JSONValue.null, "name": Amplify.JSONValue.string("todo1"), "id": Amplify.JSONValue.string("933efeb7-0099-44c1-a902-124268c12dc9")], _modelName: "Todo")
Encoder Foundation.(unknown context at $7fff59c1aab8).__JSONEncoder
Failed to parse the event DataStoreError: No decoder found for model named MutationEvent
Recovery suggestion: There is no decoder registered for the model named MutationEvent. Register models with `ModelRegistry.register(modelName:)` at startup.
Encoder Foundation.(unknown context at $7fff59c1aab8).__JSONEncoder
Encoder Foundation.(unknown context at $7fff59c1aab8).__JSONEncoder
Successfully saved model: FlutterSerializedModel(values: ["isComplete": Amplify.JSONValue.boolean(false), "updatedAt": Amplify.JSONValue.null, "id": Amplify.JSONValue.string("d477a2eb-6dc7-4e97-ad11-c52aeefa4ef9"), "__typename": Amplify.JSONValue.string("Todo"), "description": Amplify.JSONValue.string("when offline"), "createdAt": Amplify.JSONValue.null, "name": Amplify.JSONValue.string("offline-todo2")], _modelName: "Todo")
[...]

PR #2532, opened 2022-11-02
fix(datastore): improve sync event error handling - cannotParseResponse

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-swift/issues/2523

1. **`v1` PR https://github.com/aws-amplify/amplify-swift/pull/2532**  (You are here)
2. `main` PR https://github.com/aws-amplify/amplify-swift/pull/2536

*Description of changes:*
In the scenario of toggling Wifi off and then saving a model, a mutation event will be produced and sent to AppSync. The API call can fail undeterministically with different URLError's. Most of the time it's the error "The Internet connection appears to be offline." corresponding to `.notConnectedToInternet` and the retry mechanism kicks in.

Sometimes it's "cannot parse response" which gets processed by `ProcessMutationErrorFromCloudOperation` as no-op, returns success to the OutgoingMutationQueue, which will then delete mutation event, dropping the event from being processed again, without notifying the ErrorHandler.

This PR adds `.cannotParseResponse` to the list for the retry strategy to consider as retryable. We maintain a list of URL errors that are considered for retry: 

```swift
case .notConnectedToInternet,
             .dnsLookupFailed,
             .cannotConnectToHost,
             .cannotFindHost,
             .timedOut,
             .dataNotAllowed,
             .cannotParseResponse: // This PR adds this one
```

Secondly, "without notifying the ErrorHandler" is a bug because DataStore should not drop a mutation event silently. Developers should be notified of sync failures. This PR ensures each level of error handling branch calls the ErrorHandler to pass back the mutationEvent before continue on to deleting the mutation event from the local database. In the future, the errorHandler payload should be more descriptive such as passing back a SyncError (similar to JS's current implementation).

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
+++ b/AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
@@ -55,25 +55,42 @@ class ProcessMutationErrorFromCloudOperation: AsynchronousOperation {
             return
         }
 
-        if let apiError = apiError, isAuthSignedOutError(apiError: apiError) {
+        if let apiError = apiError {
+            if isAuthSignedOutError(apiError: apiError) {
+                log.verbose("User is signed out, passing error back to the error handler, and removing mutation event.")
+            } else if let underlyingError = apiError.underlyingError {
+                log.debug("Received APIError: \(apiError.localizedDescription) with underlying error: \(underlyingError.localizedDescription)")
+            } else {
+                log.debug("Received APIError: \(apiError.localizedDescription)")
+            }
             dataStoreConfiguration.errorHandler(DataStoreError.api(apiError, mutationEvent))
             finish(result: .success(nil))
             return
         }
 
-        guard let graphQLResponseError = graphQLResponseError,
-            case let .error(graphQLErrors) = graphQLResponseError else {
-                finish(result: .success(nil))
-                return
+        guard let graphQLResponseError = graphQLResponseError else {
+            dataStoreConfiguration.errorHandler(
+                DataStoreError.api(APIError.unknown("This is unexpected. Missing APIError and GraphQLError.", ""),
+                                   mutationEvent))
+            finish(result: .success(nil))
+            return
+        }
+
+        guard case let .error(graphQLErrors) = graphQLResponseError else {
+            dataStoreConfiguration.errorHandler(DataStoreError.api(graphQLResponseError, mutationEvent))
+            finish(result: .success(nil))
+            return
         }
 
         guard graphQLErrors.count == 1 else {
             log.error("Received more than one error response: \(String(describing: graphQLResponseError))")
+            dataStoreConfiguration.errorHandler(DataStoreError.api(graphQLResponseError, mutationEvent))
             finish(result: .success(nil))
             return
         }
 
         guard let graphQLError = graphQLErrors.first else {
+            dataStoreConfiguration.errorHandler(DataStoreError.api(graphQLResponseError, mutationEvent))
             finish(result: .success(nil))
             return
         }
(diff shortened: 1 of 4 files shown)


lawmicha:
Something's going on with this unit test 
```
com.amplifyframework.datastore.appsync.ModelWithMetadataAdapterTest > adapterCanDeserializeJsonOfSerializedModelIntoMwm FAILED
    java.lang.NullPointerException at ModelWithMetadataAdapterTest.java:141
```
but hard to debug due to not being able to run local unit tests.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/MutationSync/OutgoingMutationQueue/ProcessMutationErrorFromCloudOperation.swift
```diff
@@ -85,22 +102,26 @@ class ProcessMutationErrorFromCloudOperation: AsynchronousOperation {
                 let payload = HubPayload(eventName: HubPayload.EventName.DataStore.conditionalSaveFailed,
                                          data: mutationEvent)
                 Amplify.Hub.dispatch(to: .dataStore, payload: payload)
+                dataStoreConfiguration.errorHandler(DataStoreError.api(graphQLResponseError, mutationEvent))
```
drochetti, line 105: If the error is always of the same case with the cause/info part of the GraphQLResponseError, I wonder how the developer is supposed to handle then? Case matching won't be available, so how to differentiate between an invalid input or an unauthorized scenario?

lawmicha: I wanted to the scope the changes in this PR to simply pass back the info back to the developer and let them log whatever is called in the callback, and the only reasonable DataStoreError case to do that with is `DataStoreError.api` since it has the associated types of AmplifyError and MutationEvent. We were already doing this in some other places in this code, so i'm extending it to the missed cases. The long term plan is to have a similar error type as what JS called `SyncError`, which on iOS will conform to AmplifyError. The JS SyncError looks like this https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/types.ts#L933 
```typescript
export type SyncError<T extends PersistentModel> = {
	message: string;
	errorType: ErrorType;
	errorInfo?: string;
	recoverySuggestion?: string;
	model?: string;
	localModel: T;
	remoteModel: T;
	process: ProcessName;
	operation: string;
	cause?: Error;
};

export type ErrorType =
	| 'ConfigError'
	| 'BadModel'
	| 'BadRecord'
	| 'Unauthorized'
	| 'Transient'
	| 'Unknown';

export enum ProcessName {
	'sync' = 'sync',
	'mutate' = 'mutate',
	'subscribe' = 'subscribe',
}
```

With this PR, I don't think a developer will reasonably be successful in handling the errorHandler callback (if i understand, this is what you're asking) becuase they need to check that it's a DataStoreError, with case `.api`, then cast the first parameter to APIError or GraphQLResponseError, and further check for `.errors` case, and process the GraphQLError's. I think what this PR enables the developer is the ability to get visibility on sync errors, such they can log what's coming back from the errorHandler, which helps them and us debug their situation further. Later on, we can take the error processing logic here and map that to the SyncError. So for example, in this code, the developer receiving a callback on the errorHandler can cast the AmplifyError to SyncError, the processName should be "mutate". For unauthorized, it will be the errorType "Unauthorized", and for conditionalSaveFailed, it will most likely be "BadModel" (not 100% sure on this one). The mapping exercise we will have to do and synchronize with the JS's mapping logic.

drochetti: That makes sense Michael, thanks for the thoughtful response and I think the scope of the changes make sense.

Next we should try to document the error handler usage scenarios, I believe that will help us define an API that will help customer take action in the handler when needed.

drochetti (approved):

lawmicha:
this was released in https://github.com/aws-amplify/amplify-swift/releases/tag/v2.1.0
and https://github.com/aws-amplify/amplify-swift/releases/tag/v1.28.3

PR #2540, opened 2022-11-07
chore(datastore): test run V2 Integ tests on main

manaswi223:
*Issue #, if available:*

*Description of changes:*

- [ ] All unit tests pass

### DIFF
--- a/.github/workflows/integ_test_datastore_v2.yml
+++ b/.github/workflows/integ_test_datastore_v2.yml
@@ -1,0 +1,59 @@
+name: DataStore Integration V2 Tests
+on:
+  push:
+    branches: [chore/test-run-V2]
+
+permissions:
+    id-token: write
+    contents: read 
+
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  prepare-for-test:
+    runs-on: macos-12
+    environment: IntegrationTest
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        with:
+          persist-credentials: false
+
+      - name: Verify copy resources
+        uses: ./.github/composite_actions/download_test_configuration
+        with:
+          resource_subfolder: NA
+          aws_role_to_assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          aws_region: ${{ secrets.AWS_REGION }}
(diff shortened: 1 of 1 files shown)


.github/workflows/integ_test_datastore_v2.yml
```diff
@@ -0,0 +1,59 @@
+name: DataStore Integration V2 Tests
```
lawmicha, line 1: I think referencing TransformerV2 in the name, which shows up as the name of the workflow under Actions tab, will make more clear that we're talking about TransformerV2 and not some other version like the library V2 version
```suggestion
name: DataStore TransformerV2 Tests
```

.github/workflows/integ_test_datastore_v2.yml
```diff
@@ -0,0 +1,59 @@
...
+on:
+  push:
+    branches: [chore/test-run-V2]
```
lawmicha, line 4: a reminder to update this back to `main` before merging it in

lawmicha:
```suggestion
    branches: [main]
```

.github/workflows/integ_test_datastore_v2.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
   datastore-integration-v2-test:
     continue-on-error: true
-    timeout-minutes: 20
+    timeout-minutes: 17
```
lawmicha, line 33: I can see this successful test run here https://github.com/aws-amplify/amplify-swift/actions/runs/3413375249/jobs/5680024350 that ran for 14 mins, and a very short "prepare-for-test" step that took 20 seconds.. I would say 20 or 25 seems like a reasonable amount of time to build a buffer for when it runs long. And less than 30 minutes so when something hangs, it times out and fails the test run so it's not hanging forever. regarding the test run, 14 minutes still seems awfully high though.. keeping the timeout higher will help us capture more data points on a couple of runs. If the test doesn't have a syncExpression to filter out data created before the current test and it's waiting for sync, then that might be contributing to the overall time. If this is what's happening (i haven't looked into these tests to say definitively) but let's say it takes 30 seconds to fully sync the backend data and there are 20 tests, each test clears and resyncs is going to be at least 10 mins

lawmicha:
PR title suggestion: `chore(datastore): enable TransformerV2 tests`

PR #2553, opened 2022-11-09
chore(datastore): add logging to cascadeDelete

manaswi223:
*Issue #, if available:*

*Description of changes:*
This PR adds logging to the cascadeDelete operation for simplified debugging into the cascadeDelete operation & provide better CX.
In any Amplify configured App with verbose logging enabled, searching `[cascadeDelete` will print out the state of cascadeDelete Operation like:
```
[cascadeDelete.1] Start cascade logic for Blog with id `blog123`
[cascadeDelete.2] Query for Blog associated models, found Post 1 Comment 1
[cascadeDelete.3] Local cascade delete of Blog successful!
[cascadeDelete.4] sending 3 delete mutations
```

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
@@ -31,6 +31,8 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
 
     private let serialQueueSyncDeletions: DispatchQueue
 
+    private var mutationEvent: String
+
     init(storageAdapter: StorageEngineAdapter,
          syncEngine: RemoteSyncEngineBehavior?,
          modelType: M.Type,
@@ -51,6 +53,7 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
         self.completionForWithId = completion
         self.completionForWithFilter = nil
         self.serialQueueSyncDeletions = DispatchQueue(label: "com.amazoncom.Storage.CascadeDeleteOperation.concurrency")
+        self.mutationEvent = self.modelSchema.name
         super.init()
     }
 
@@ -68,6 +71,7 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
         self.completionForWithId = nil
         self.completionForWithFilter = completion
         self.serialQueueSyncDeletions = DispatchQueue(label: "com.amazoncom.Storage.CascadeDeleteOperation.concurrency")
+        self.mutationEvent = self.modelSchema.name
         super.init()
     }
 
@@ -144,11 +148,24 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
         
         associatedModels = await self.recurseQueryAssociatedModels(modelSchema: self.modelSchema, ids: modelIds)
         
+        if mutationEvent != "MutationEvent" {
+            self.log.debug("[CascadeDelete.1] Start cascade logic for \(mutationEvent) with id \(modelIds)")
+        }
+        
+        let associatedModelNames = associatedModels.compactMap { $1.modelName }
+        
+        if mutationEvent != "MutationEvent" {
+            self.log.debug("[CascadeDelete.2] Query for \(mutationEvent) associated models, found \(associatedModelNames)")
+        }
+
         deletedResult = await withCheckedContinuation { continuation in
             self.storageAdapter.delete(self.modelType,
                                        modelSchema: self.modelSchema,
                                        filter: self.deleteInput.predicate) { result in
                 continuation.resume(returning: result)
+            }
+            if mutationEvent != "MutationEvent" {
+                self.log.debug("[CascadeDelete.3] Local cascade delete of \(mutationEvent) successful!")
             }
         }
         return collapseResults(queryResult: queriedResult,
@@ -295,6 +312,8 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
             // mutation?
             switch deleteInput {
             case .withIdentifier, .withIdentifierAndCondition:
+                let syncDeletionsCount = (queryAndDeleteResult.associatedModels.count + queryAndDeleteResult.deletedModels.count)
+                self.log.debug("[CascadeDelete.4] sending \(syncDeletionsCount) delete mutations")
                 syncDeletions(withModels: queryAndDeleteResult.deletedModels,
                               associatedModels: queryAndDeleteResult.associatedModels,
                               syncEngine: syncEngine) {
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/StorageEngine.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/StorageEngine.swift
@@ -343,7 +343,7 @@ final class StorageEngine: StorageEngineBehavior {
             case .failure(let dataStoreError):
                 completion(.failure(dataStoreError))
             case .success(let mutationEvent):
-                self.log.verbose("\(#function) successfully submitted to sync engine \(mutationEvent)")
+                self.log.verbose("\(#function) successfully submitted \(mutationEvent.modelName) to sync engine \(mutationEvent)")
                 completion(.success(savedModel))
             }
         }


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -144,12 +148,25 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
...
+
         deletedResult = await withCheckedContinuation { continuation in
             self.storageAdapter.delete(self.modelType,
                                        modelSchema: self.modelSchema,
                                        filter: self.deleteInput.predicate) { result in
                 continuation.resume(returning: result)
             }
+            if mutationEvent != "MutationEvent" {
+                self.log.debug("[CascadeDelete.3] Local cascade delete of \(mutationEvent) successful!")
+            }
```
lawmicha, line 169: is `storageAdapter.delete` completion closure an escaping closure? if it is, then this code is running before the delete is completed and would be inaccurate. If it's not an escaping closure, then `continuation.resume` is called to resume the suspended task and then the caller resumes while we call this code in parallel to that, which seems error prone. When leveraging async await's `withCheckedContinuation` to convert the completion handler to async should be a simple wrapper that doesn't involve other logic. Also this doesn't check if the `result` is successful or not so it should be based on the `result` being a `.success` case. `deleteResult` (which is the same as `result`, passed back through `continuation.resume(returning: result)`) is passed to `collapseResults` which i think is where this log line can live

lawmicha: In `collapseResults`, before returning success, it can log that the local delete was successful

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -295,6 +312,8 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
             // mutation?
             switch deleteInput {
             case .withIdentifier, .withIdentifierAndCondition:
+                let syncDeletionsCount = (queryAndDeleteResult.associatedModels.count + queryAndDeleteResult.deletedModels.count)
+                self.log.debug("[CascadeDelete.4] sending \(syncDeletionsCount) delete mutations")
                 syncDeletions(withModels: queryAndDeleteResult.deletedModels,
```
lawmicha, line 317: `syncDeletions` is called in at least two places, here and in the `case .withFilter(let filter)` just below this. If we move this log line into `syncDeletions` then it will we'll automatically gain this log for both code paths

lawmicha: Adding the two numbers together seems like we're losing more information than we originally started with, if we add this to `syncDeletions(withModels:)` call, and don't add the numbers together, the log line will look like this

```
Begin syncing  \(models.count) \(modelName) models for deletion. 
```

For associated models, I can see this is lacking some information for us since it contains a count of all associated models, and we don't know the subset count for each model. I think this is okay for now since the logging context contains a log above this that says how many of each associated model was queried.
```
Begin syncing \(associatedModels.count) associated models for deletion. 
```

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -144,12 +148,25 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
         
         associatedModels = await self.recurseQueryAssociatedModels(modelSchema: self.modelSchema, ids: modelIds)
         
+        if mutationEvent != "MutationEvent" {
```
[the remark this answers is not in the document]
manaswi223: w.r.t cascadeDelete Operation, logging the actual models(like Post, Blog etc) and skipping MutationEvent(here a form of model) is both cx and developer friendly to only track the state of cascadeDelete operation of the required actual model. 
If I don't skip logging on MutationEvents, my logs look like:
```
[CascadeDelete.1] Start cascade logic for MutationEvent with id ["456"]
[CascadeDeleteOperation<MutationEvent>] [CascadeDelete.2] Query for MutationEvent associated models, found []
[CascadeDeleteOperation<MutationEvent>] [CascadeDelete.3] Local cascade delete of MutationEvent successful!
[CascadeDeleteOperation<MutationEvent>] [CascadeDelete.1] Start cascade logic for MutationEvent with id ["567"]
[CascadeDeleteOperation<MutationEvent>] [CascadeDelete.2] Query for MutationEvent associated models, found []
[CascadeDeleteOperation<MutationEvent>] [CascadeDelete.3] Local cascade delete of MutationEvent successful!
[CascadeDeleteOperation<MutationEvent>] [CascadeDelete.1] Start cascade logic for MutationEvent with id ["678"]
[CascadeDeleteOperation<MutationEvent>] [CascadeDelete.2] Query for MutationEvent associated models, found []
[CascadeDeleteOperation<MutationEvent>] [CascadeDelete.3] Local cascade delete of MutationEvent successful!
[CascadeDeleteOperation<Blog6V2>] [CascadeDelete.1] Start cascade logic for Blog with id ["123"]
[CascadeDeleteOperation<Blog6V2>] [CascadeDelete.2] Query for Blog associated models, found ["Post", "Comment"]
[CascadeDeleteOperation<Blog6V2>] [CascadeDelete.3] Local cascade delete of Blog6V2 successful!
[CascadeDeleteOperation<Blog6V2>] [CascadeDelete.4] sending 3 delete mutations
```
The filtering of MutationEvents(here a form of model) is to refine the logs of required models & show their state. for example, with filtering the MutationEvents, the logs look clean as in:
```
[CascadeDeleteOperation<Blog6V2>] [CascadeDelete.1] Start cascade logic for Blog with id ["123"]
[CascadeDeleteOperation<Blog6V2>] [CascadeDelete.2] Query for Blog associated models, found ["Post", "Comment"]
[CascadeDeleteOperation<Blog6V2>] [CascadeDelete.3] Local cascade delete of Blog6V2 successful!
[CascadeDeleteOperation<Blog6V2>] [CascadeDelete.4] sending 3 delete mutations
```

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
manaswi223: sure, added that! I thought the combined numbers(associatedModel.count + deletedModel.count) will give clear understanding of the total no.of models that are sent for delete Mutations. Along with your suggestions, now the logs look like:
```
[CascadeDelete.4] sending a total of 3 delete mutations
[CascadeDeleteOperation<Blog>] [CascadeDelete.4] Begin syncing 1 Blog models for deletion
[CascadeDeleteOperation<Blog>] [CascadeDelete.4] Begin syncing 2 associated models for deletion.
[CascadeDeleteOperation<Blog>] syncDeletions(withModels:predicate:associatedModels:syncEngine:completion:) successfully submitted Comment to sync engine MutationEvent(...)
syncDeletions(withModels:predicate:associatedModels:syncEngine:completion:) successfully submitted Post to sync engine MutationEvent(...)
syncDeletions(withModels:predicate:associatedModels:syncEngine:completion:) successfully submitted Blog to sync engine MutationEvent(...)
```

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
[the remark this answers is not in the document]
lawmicha: filtering out the MutationEvent deletions seems fair, let's do this for the remaining system model scehams then. Let's create a computed property called `isDeveloperDefinedModel` in this class that checks for the model names of the system models

```
ModelSyncMetadata.modelName
MutationEvent.modelName
MutationSyncMetadata.modelName
```
 so that we can reuse wherever you want to create the developer friendly logging. We also should not be hardcoding the string "MutationEvent" and use the model's properties directly. Hardcoding a string multiple times is prone to typos and if the name changes, this string needs to change as well, but is not enforced by the compiler.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -142,8 +146,19 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
...
         let modelIds = queriedModels.map { $0.identifier(schema: self.modelSchema).stringValue }
         
-        associatedModels = await self.recurseQueryAssociatedModels(modelSchema: self.modelSchema, ids: modelIds)
+        if modelName != "MutationEvent" {
+            self.log.debug("[CascadeDelete.1] Deleting \(modelName) with identifiers: \(modelIds)")
+        }
         
+        associatedModels = await self.recurseQueryAssociatedModels(modelSchema: self.modelSchema, ids: modelIds)
+        if self.log.logLevel >= .debug {
+            if modelName != "MutationEvent" {
```
lawmicha, line 155: two if statements can be combined into one if statement with && operator

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -221,6 +236,9 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
                 if let deleteResult = deleteResult {
                     switch deleteResult {
                     case .success:
+                        if modelName != "MutationEvent" {
+                            self.log.debug("[CascadeDelete.3] [Delete from Local Store] Local cascade delete of \(modelName) successful!")
```
lawmicha, line 240: nit: why the brackets on "Delete From Local Store"? would suggest just "Local delete for \(modelName) successful". Not all models are cascade deleted either, only models with associations, so I would suggest dropping the "cascade delete" unless we have also determined that it is a cascade delete for this model

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -295,6 +313,8 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
             // mutation?
             switch deleteInput {
             case .withIdentifier, .withIdentifierAndCondition:
+                let syncDeletionsCount = (queryAndDeleteResult.associatedModels.count + queryAndDeleteResult.deletedModels.count)
```
lawmicha, line 316: this looks like code duplication, in this case, i believe you can just do this one above the `switch deleteInput` line, on line 301, or even higher up

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -346,6 +368,7 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
                                associatedModels: [(ModelName, Model)],
                                syncEngine: RemoteSyncEngineBehavior,
                                completion: @escaping DataStoreCallback<Void>) {
+        self.log.debug("[CascadeDelete.4] Begin syncing  \(models.count) \(modelName) model for deletion")
```
lawmicha, line 371: this looks out of place, i think when `associatedModels.isEmpty` (two lines below) then we call a private `syncDeletions` method that has no `asssociatedModels` parameter. This log line should go in there

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -142,8 +146,19 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
...
         
-        associatedModels = await self.recurseQueryAssociatedModels(modelSchema: self.modelSchema, ids: modelIds)
+        if modelName != "MutationEvent" {
+            self.log.debug("[CascadeDelete.1] Deleting \(modelName) with identifiers: \(modelIds)")
+        }
         
+        associatedModels = await self.recurseQueryAssociatedModels(modelSchema: self.modelSchema, ids: modelIds)
+        if self.log.logLevel >= .debug {
+            if modelName != "MutationEvent" {
+                self.log.debug("[CascadeDelete.2] Queried for \(modelName) associated models, found \(associatedModels.compactMap { $1.modelName })")
```
lawmicha, line 156: doing this outside of the method recurseQueryAssociatedModels seems inefficient because we doing additional processing just for logging. I would look at recurseQueryAssociatedModels again to see if we have all the information available there to log. let's review recurseQueryAssociatedModels to see if we have all the information there

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -142,8 +146,19 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
...
+                
+                self.log.debug("[CascadeDelete.2] [Query ids of associated Models] Queried for \(associatedModels.compactMap({$1.modelName})), retrieved ids for deletion: \(associatedModels.map { $0.1.identifier(schema: modelSchema).stringValue })")
```
lawmicha, line 158: see above comment about taking a look at `recurseQueryAssociatedModels` to see if we can do this logging during the recursive query logic

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -144,12 +148,25 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
         
         associatedModels = await self.recurseQueryAssociatedModels(modelSchema: self.modelSchema, ids: modelIds)
         
+        if mutationEvent != "MutationEvent" {
+            self.log.debug("[CascadeDelete.1] Start cascade logic for \(mutationEvent) with id \(modelIds)")
+        }
+        
+        let associatedModelNames = associatedModels.compactMap { $1.modelName }
```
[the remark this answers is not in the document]
lawmicha: resolving this comment, i've added a new comment based on the latest changes

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
lawmicha: i see that this was moved, so will resolve this comment

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
lawmicha: sounds good, i see the latest code has been moved to `syncDeletion`, i've added new comments w.r.t those places

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
manaswi223: tried doing this in `recurseQueryAssociatedModels` in couple of ways but looked more efficient and apt in the queryResult success case of `collapseResults` function.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -142,8 +146,12 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
         
         let modelIds = queriedModels.map { $0.identifier(schema: self.modelSchema).stringValue }
         
-        associatedModels = await self.recurseQueryAssociatedModels(modelSchema: self.modelSchema, ids: modelIds)
+        if self.log.logLevel >= .debug && modelName != MutationEvent.modelName {
```
lawmicha, line 149: we are checking MutationEvent.modelName here and then later in the code, we're checking `isDeveloperDefined.MutationEvent.rawValue`, what's the point of creating `isDeveloperDefined` enum if the all the values are not used? A computed property called `isDeveloperDefined` should check if the instance property `modelName` is one of the three system model types, like
```
var isDeveloperDefinedModel Bool {
   return !systemModelNames.contains(modelSchema.name)
}
```

Then use it wherever you want to control the logging, ie.
```
if self.log.logLevel >= .debug && isDeveloperDefinedModel {
 // only log when log level is debug level or higher and is a developer defined model
}
```

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -486,6 +508,12 @@ extension CascadeDeleteOperation {
             }
         }
     }
+    
+    enum isDeveloperDefined: ModelName {
+        case ModelSyncMetadata = "ModelSyncMetadata"
+        case MutationEvent = "MutationEvent"
+        case MutationSyncMetadata = "MutationSyncMetadata"
+    }
```
lawmicha, line 516: should use the system Model type's themselves instead of hardcoding the system model names, what if they change? this code will continue to compile. Using the types give you access to the model's name, for example, the MutationEvent type should have a `modelName` or `name` property that should be used for model name string checking

lawmicha: based on my comment about creating a computed property that returns whether it is a developer defined model or not should render this enum unused. We can probably use a Set or an Array to hold the system model names and use that to check when checking if `modelSchema.name` is a developer defined model or not.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -349,13 +369,15 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
         var savedDataStoreError: DataStoreError?
 
         guard !associatedModels.isEmpty else {
+            self.log.debug("[CascadeDelete.4] Begin syncing  \(models.count) \(modelName) model for deletion")
```
lawmicha, line 372: If `associatedModels.isEmpty` is true then this log statement is performed. I still think this log line should be within the `syncDeletions(withModels:predicate:syncEngine:dataStoreError:completion:)` method because that way it will be logged in all code paths that call this method is called. The `syncDeleteions(withModels:predicate:associatedModels:)` method will sync the associated models first before syncing the parent models (as logged by this log line). If there are associated models, then this guard statement body is not executed, thus going further below to submit the associated models mutation events. Once `mutationEventsSubmitCompleted == associatedModels.count` on line 408 is true, then it will sync the parent models. With this log line being here, for the scenario where there are associated models, we are not logging the parent model's "begin syncing". If we put it inside `syncDeleteions(withModels:predicate:associatedModels:)` then any code paths calling this will log the number of parent models syncing count

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
lawmicha: can you expand on your comment? i would like see the different ways you mentioned to better understand why this option is better

lawmicha:
```
2022-11-14 14:07:38.206589-0500 DataStoreHostApp[32649:334143] [CascadeDeleteOperation<Blog6V2>] [CascadeDelete.1] Deleting Blog6V2 with identifiers: ["981C9D8D-E6ED-4ABF-A05C-D62AF493FE6F"]
2022-11-14 14:07:38.206665-0500 DataStoreHostApp[32649:334143] [CascadeDeleteOperation<Blog6V2>] [CascadeDelete.2] Querying for Blog6V2's associated model Post6V2.
2022-11-14 14:07:38.216816-0500 DataStoreHostApp[32649:334143] [CascadeDeleteOperation<Blog6V2>] [CascadeDelete.2] Queried for Post6V2, retrieved ids for deletion: ["D2036A03-446C-47C5-87C6-F176FD73927F"]
2022-11-14 14:07:38.216899-0500 DataStoreHostApp[32649:334143] [CascadeDeleteOperation<Blog6V2>] [CascadeDelete.2] Querying for Blog6V2's associated model Comment6V2.
2022-11-14 14:07:38.218082-0500 DataStoreHostApp[32649:334143] [CascadeDeleteOperation<Blog6V2>] [CascadeDelete.2] Queried for Comment6V2, retrieved ids for deletion: ["0924A192-C592-449E-A450-B95B16C9A274"]
2022-11-14 14:07:38.219467-0500 DataStoreHostApp[32649:334143] [CascadeDeleteOperation<Blog6V2>] [CascadeDelete.3] Local cascade delete of Blog6V2 successful!
2022-11-14 14:07:38.219556-0500 DataStoreHostApp[32649:334143] [CascadeDeleteOperation<Blog6V2>] [CascadeDelete.4] sending a total of 3 delete mutations
2022-11-14 14:07:38.219610-0500 DataStoreHostApp[32649:334143] [CascadeDeleteOperation<Blog6V2>] [CascadeDelete.4] Begin syncing 2 associated models for deletion.
2022-11-14 14:07:38.223275-0500 DataStoreHostApp[32649:334150] [CascadeDeleteOperation<Blog6V2>] syncDeletions(withModels:predicate:associatedModels:syncEngine:completion:) successfully submitted Comment6V2 to sync engine MutationEvent(id: "DD368C23-D675-4ED2-8B39-E11A2E2F416A", modelId: "0924A192-C592-449E-A450-B95B16C9A274", modelName: "Comment6V2", json: "{\"post\":{\"createdAt\":\"2022-11-14T19:07:37.858Z\",\"id\":\"D2036A03-446C-47C5-87C6-F176FD73927F\",\"title\":\"title\",\"updatedAt\":\"2022-11-14T19:07:37.858Z\",\"blog\":{\"posts\":[],\"id\":\"981C9D8D-E6ED-4ABF-A05C-D62AF493FE6F\",\"updatedAt\":\"2022-11-14T19:07:37.645Z\",\"name\":\"name\",\"createdAt\":\"2022-11-14T19:07:37.645Z\"},\"comments\":[]},\"id\":\"0924A192-C592-449E-A450-B95B16C9A274\",\"content\":\"content\",\"updatedAt\":\"2022-11-14T19:07:38.067Z\",\"createdAt\":\"2022-11-14T19:07:38.067Z\"}", mutationType: "delete", createdAt: Amplify.Temporal.DateTime(foundationDate: 2022-11-14 19:07:38 +0000), version: Optional(1), inProcess: true, graphQLFilterJSON: nil)
2022-11-14 14:07:38.225180-0500 DataStoreHostApp[32649:334150] [CascadeDeleteOperation<Blog6V2>] syncDeletions(withModels:predicate:associatedModels:syncEngine:completion:) successfully submitted Post6V2 to sync engine MutationEvent(id: "9354FE59-94F3-447F-8100-43CE54812E77", modelId: "D2036A03-446C-47C5-87C6-F176FD73927F", modelName: "Post6V2", json: "{\"createdAt\":\"2022-11-14T19:07:37.858Z\",\"id\":\"D2036A03-446C-47C5-87C6-F176FD73927F\",\"title\":\"title\",\"updatedAt\":\"2022-11-14T19:07:37.858Z\",\"blog\":{\"posts\":[],\"id\":\"981C9D8D-E6ED-4ABF-A05C-D62AF493FE6F\",\"updatedAt\":\"2022-11-14T19:07:37.645Z\",\"name\":\"name\",\"createdAt\":\"2022-11-14T19:07:37.645Z\"},\"comments\":[]}", mutationType: "delete", createdAt: Amplify.Temporal.DateTime(foundationDate: 2022-11-14 19:07:38 +0000), version: Optional(1), inProcess: false, graphQLFilterJSON: nil)
2022-11-14 14:07:38.225230-0500 DataStoreHostApp[32649:334150] [CascadeDeleteOperation<Blog6V2>] [CascadeDelete.4] Begin syncing 1 Blog6V2 model for deletion
2022-11-14 14:07:38.228602-0500 DataStoreHostApp[32649:334144] [CascadeDeleteOperation<Blog6V2>] syncDeletions(withModels:predicate:syncEngine:dataStoreError:completion:) successfully submitted to sync engine MutationEvent(id: "C6B35587-1767-4042-A5C8-4F7A1B9144C3", modelId: "981C9D8D-E6ED-4ABF-A05C-D62AF493FE6F", modelName: "Blog6V2", json: "{\"posts\":[],\"id\":\"981C9D8D-E6ED-4ABF-A05C-D62AF493FE6F\",\"updatedAt\":\"2022-11-14T19:07:37.645Z\",\"name\":\"name\",\"createdAt\":\"2022-11-14T19:07:37.645Z\"}", mutationType: "delete", createdAt: Amplify.Temporal.DateTime(foundationDate: 2022-11-14 19:07:38 +0000), version: nil, inProcess: false, graphQLFilterJSON: nil)

```

lawmicha (approved):

royjit (dismissed):

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -31,6 +31,16 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
 
     private let serialQueueSyncDeletions: DispatchQueue
 
+    private var modelName: ModelName
+
+    private let systemModelNames: Set = [MutationEvent.modelName,
+                                         ModelSyncMetadata.modelName,
+                                         MutationSyncMetadata.modelName]
+    
+    private var shouldLogDeveloperDefinedModel: Bool {
```
atierian, line 40: Do we need to evaluate this on the fly every time it's accessed? 

It looks like `systemModelNames.contains(modelName)` can be evaluated once on initialization or lazily because `systemModelNames` and `modelName` are both immutable.

Do we need to support respecting a dynamically changed `log.logLevel` after the `CascadeDeleteOperation` is initialized? The cost of doing this is a blocking `nsLock.execute { }` on each `logLevel` `get` ([source](https://github.com/aws-amplify/amplify-swift/blob/e753770efddacd63125375a84fc8dc4c56ebc0ca/Amplify/Categories/Logging/LoggingCategory.swift#L45-L49)) , which seems less than ideal. Can we also evaluate that once with a lazy var?

i.e. 
```swift
private lazy var shouldLogDeveloperDefinedModel: Bool = { 
  log.logLevel >= .debug && Set([
    MutationEvent.modelName,
    ModelSyncMetadata.modelName,
    MutationSyncMetadata.modelName
  ]).contains(modelName)
}()
```

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -143,6 +155,10 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
         
         let modelIds = queriedModels.map { $0.identifier(schema: self.modelSchema).stringValue }
 
+        if shouldLogDeveloperDefinedModel {
```
atierian, line 158: I recommend creating a local function to replace these five `if` branches scattered throughout the code.

```swift
private func _log(
  _ message: @escaping @autoclosure () -> String,
  level log: (() -> String) -> Void
) { 
  guard shouldLogDeveloperDefinedModel else { return }
  log(message)
}
```

Then you can replace all of the
```swift
if shouldLogDeveloperDefinedModel { 
  self.log.<level>(...)
}
```

with
```swift
_log(..., level: log.<level>)
```

e.g.
```swift
// this
if shouldLogDeveloperDefinedModel {
  self.log.debug("[CascadeDelete.1] Deleting \(modelName) with identifiers: \(modelIds)")
}

// becomes this
_log("[CascadeDelete.1] Deleting \(modelName) with identifiers: \(modelIds)", level: log.debug)
```

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
atierian: On second thought, we can drop the `log.logLevel >= .debug` check here all together, can't we?

We're calling the applicable log level methods (e.g. `log.debug(...)`), so the logger internals will take care of that check.

atierian: Assuming that's the case, we don't need `self` to do the evaluation anymore, so we can drop the `lazy` recommended above and define a local constant that's assigned in the `init`. Afaict, this also let's us drop the `modelName` property.

```swift
private let isDeveloperDefinedModel: Bool

init(...) {
  // ...

  self.isDeveloperDefinedModel = !Set([
    MutationEvent.modelName,
    ModelSyncMetadata.modelName,
    MutationSyncMetadata.modelName
  ]).contains(modelSchema.name)
}
```

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -465,10 +505,23 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
                                     completion: @escaping DataStoreCallback<MutationEvent>) {
         syncEngine.submit(mutationEvent, completion: completion)
     }
+
+    private func _log(
```
5d, line 509: the `_` prefix on function name looks like a `C++` program style.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -137,12 +157,13 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
...
             return collapseResults(queryResult: queriedResult,
-                                                    deleteResult: deletedResult,
-                                                    associatedModels: associatedModels)
+                                   deleteResult: deletedResult,
+                                   associatedModels: associatedModels)
         }
         
         let modelIds = queriedModels.map { $0.identifier(schema: self.modelSchema).stringValue }
-
+        _log("[CascadeDelete.1] Deleting \(modelSchema.name) with identifiers: \(modelIds)")
```
5d, line 165: is the number inside `[CascadeDelete.1]` define the order should be followed?

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
```diff
@@ -29,57 +29,77 @@ public class CascadeDeleteOperation<M: Model>: AsynchronousOperation {
...
         self.modelType = modelType
         self.modelSchema = modelSchema
-        self.deleteInput = .withFilter(filter)
-        self.completionForWithId = nil
-        self.completionForWithFilter = completion
-        self.serialQueueSyncDeletions = DispatchQueue(label: "com.amazoncom.Storage.CascadeDeleteOperation.concurrency")
+        self.deleteInput = deleteInput
+        self.completionForWithId = completionForWithId
+        self.completionForWithFilter = completionForWithFilter
+        self.isDeveloperDefinedModel = !Set([
```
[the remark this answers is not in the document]
royjit: Used the same from StorageEngine here, I didnot move to a common location because StoageEnginer. systemModelSchemas is used in a lot of places.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
royjit: Yes, looks like it.  Though there are repeated numbers, but they represent the same task and follows the sequence.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
royjit: Agreed, renamed to `logMessage`

5d (approved):

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/CascadeDeleteOperation.swift
royjit: Updated to use jus the system models and a common log function

Issue #2152, opened 2022-08-16
DataStore does not restart after toggle airplane mode off and on


dpilch:
### Describe the bug

The issue was originally reported in [Discord](https://discord.com/channels/705853757799399426/707328996995760179/1009011237348184105). Customer states that this issue started appearing on iOS 15.6. 

When using DataStore and airplane mode is switched to off and then back on DataStore will stop but not restart.

Appears to be related to https://github.com/aws-amplify/amplify-ios/pull/1901.

### Steps To Reproduce

1. Create basic app with DataStore and the default Todo schema
1. In main app setup DataStore and verbose logging.

```swift
import SwiftUI
import Amplify
import AWSDataStorePlugin
import AWSAPIPlugin

class AppDelegate: NSObject, UIApplicationDelegate {
    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey : Any]? = nil) -> Bool {
        do {
            Amplify.Logging.logLevel = .verbose
            let apiPlugin = AWSAPIPlugin(modelRegistration: AmplifyModels());
            try Amplify.add(plugin: apiPlugin)
            let dataStorePlugin = AWSDataStorePlugin(modelRegistration: AmplifyModels())
            try Amplify.add(plugin: dataStorePlugin)
            try Amplify.configure()
            print("Amplify configured with DataStore plugin")
            Amplify.DataStore.start { result in print("DataStore started") }
        } catch {
            print("An error occurred setting up Amplify: \(error)")
        }
        
        return true
    }
}
@main
struct MyIosAppApp: App {
    @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate

        var body: some Scene {
            WindowGroup {
                ContentView()
            }
        }
}
```
1. Run app on physical device.
1. Toggle airplane mode off then on. (Customer states this must be done when app is not in foreground, however I was able to reproduce the issue while the app is in foreground.
1. Observe logs
1. Create Todo on another client and observe that `websocketDidReceiveMessage` message is not present in logs

### Expected behavior

DataStore should restart and reconnect.

### Amplify Framework Version

1.28.0

### Amplify Categories

DataStore

### Dependency manager

Swift PM

### Swift version

5.6.1

### CLI version

9.2.1

### Xcode version

13.4.1

### Relevant log output

<details>
<summary>Log Messages</summary>

```
[...]

PR #2571, opened 2022-11-15
fix(datastore): retry on subscription connection error

lawmicha:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-swift/issues/2152

- `main` PR https://github.com/aws-amplify/amplify-swift/pull/2571 (This PR)
- `v1` PR https://github.com/aws-amplify/amplify-swift/pull/2581 

*Description of changes:*
Once DataStore is in an active state, then the network goes down, then recovers, the subscription websocket will send an error on network recovery. This error propagated back to API/DataStore which handles the error by moving DataStore to a stopped state. It is not processed as retryable. This delays the recoverability of DataStore until the next DataStore operation is called, since it remains in a stopped state. The scenario is reproducible by running an iOS app on a device or running a macOS app on the macbook. Running on an iOS simulator does not have the same effect and the websockets do not disconnect, rather the subscription event is received with a delay (~5-10 seconds) after network recovers.

The AppSyncRealTime websocket client, used by Amplify.API (API plugin) does have its own retry logic on classification of errors coming from the underlying websocket, but in certain scenarios, whether it's an unclassifed error or retry has been exhausted, will eventually propagate a `ConnectionProviderError.connection` error back to the caller and close the websocket connection. In this case, the API Plugin receives this terminating error and sends back `API.operationError` to DataStore. 

DataStore’s sync engine used to be much more aggressive in retrying the sync process. This caused retry storms on AppSync and was fixed in https://github.com/aws-amplify/amplify-swift/pull/1901 It was a code bug that retried on all errors, regardless of the underlying error. By fixing this, it has created another issue where in this scenario where retry would have fixed the problem, was not initiated until the next explicit DataStore operation (start/save/query/delete/etc..) was called, because it has transitioned to the stopped state. When the network is turned back on, DataStore gets the `API.operationError` and stops. We were able to observe this on multiple attempts, https://github.com/aws-amplify/amplify-swift/issues/2152#issuecomment-1217173993 and https://github.com/aws-amplify/amplify-swift/issues/2152#issuecomment-1289616392.

1. API plugin should classify `ConnectionProvider.connection` errors from AppSyncRealTimeClient as an `APIError.networkError`. Convert `ConnectionProvider.connection` to some Foundation’s URLError case that makes sense like `URLError.networkConnectionLost`
2. DataStore should not handle anything related to AppSyncRealTimeClient errors and should only act on `APIError` and extract out the URLError when it is a `APIError.networkError` as an indication to retry the sync process. It passes the URLError to RequestRetryablePolicy, which now classifies `.networkConnectionLost` as retryable.

With this change, when the wifi is turned back on from off, DataStore receives the websocket error and restarts the sync process, and transitions to the active state successfully.

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSGraphQLSubscriptionTaskRunner.swift
+++ b/AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSGraphQLSubscriptionTaskRunner.swift
@@ -174,6 +174,11 @@ public class AWSGraphQLSubscriptionTaskRunner<R: Decodable>: InternalTaskRunner,
             return
         } else if case ConnectionProviderError.unauthorized = error {
             errorDescription += ": \(APIError.UnauthorizedMessageString)"
+        } else if case ConnectionProviderError.connection = error {
+            errorDescription += ": connection"
+            let error = URLError(.networkConnectionLost)
+            fail(APIError.networkError(errorDescription, nil, error))
+            return
         }
 
         fail(APIError.operationError(errorDescription, "", error))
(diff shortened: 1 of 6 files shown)


atierian (approved):
LGTM

atierian (approved):

PR #2639, opened 2022-12-22
chore(datastore): Add stress tests for Datastore category

thisisabhash:
*Issue #, if available:* None 

*Description of changes:*

This PR adds end to end integration tests to Datastore category. The rationale behind adding these tests is to improve pre-release confidence and also increase/maintain high code quality. This is a foundational work with more tests to be added in future as we get more data (from customer bugs and scenarios on all platforms). This is completely targeting the parallelism of Amplify library APIs instead of targeting AWS Services. 
The concurrency limit is chosen as an arbitrary value currently in accordance with other platform(Android). This can be updated in future (maxing out service limits/ data from customer scenarios).
To test for parallelism, `DispatchQueue.concurrentPerform()` is being used to start multiple tasks in parallel. This is the same as starting multiple tasks in a loop. `TaskGroup` can also be used to achieve concurrency when a job has many tasks which can be run at the same time but are related in some way (for e.g. populating and storing results from all these tasks in a specific way) It is dependent on what the test is for. Both constructs can be used accordingly.

- [x] All unit tests pass


AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/DataStoreStressTests.swift
```diff
@@ -0,0 +1,432 @@
...
+
+        let capturedPosts = posts
+
+        DispatchQueue.concurrentPerform(iterations: concurrencyLimit) { index in
+            Task {
+                _ = try await Amplify.DataStore.save(capturedPosts[index])
+            }
+        }
+        
+        await waitForExpectations([postsSyncedToCloud], timeout: networkTimeout)
```
lawmicha, line 157: this code makes sense to be the body of `testMultipleSave`, but now it's duplicated in `testMultipleQueryByID` as the "set up" portion of the test. we can refactor this to something like `let savedAndSyncedPosts = try await saveAndSyncPosts(concurrenyLimit: concurrenyLimit)` for example

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/DataStoreStressTests.swift
```diff
@@ -0,0 +1,432 @@
...
```
lawmicha, line 228: same for this i think, looks like code duplication from `testMultipleQueryByID`

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/DataStoreStressTests.swift
```diff
@@ -0,0 +1,432 @@
...
+        
+        let localQueryForPosts = asyncExpectation(description: "Query for the post is successful")
+        
+        // query by predicate where title of the post is "title 25", should be successful for
+        // only one query
+        let predicate = Post.keys.id.eq(capturedPosts[25].id).and(Post.keys.title.eq(capturedPosts[25].title))
```
lawmicha, line 234: why `25` here? shouldn't the predicate change for each `index` into the `capturedPosts` array

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/DataStoreStressTests.swift
```diff
@@ -0,0 +1,432 @@
...
+        }
+
+        let capturedPosts = posts
+
+        DispatchQueue.concurrentPerform(iterations: concurrencyLimit) { index in
+            Task {
+                _ = try await Amplify.DataStore.save(capturedPosts[index])
+            }
+        }
+        await waitForExpectations([postsSyncedToCloud], timeout: networkTimeout)
```
lawmicha, line 303: code duplication for this as well

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/DataStoreStressTests.swift
```diff
@@ -0,0 +1,432 @@
...
+                try await Amplify.DataStore.delete(capturedPosts[index])
+            }
+        }
+        
+        await waitForExpectations([postsDeletedLocally, postsDeletedFromCloud], timeout: networkTimeout)
+    }
+
+    // MARK: - Helpers
+
+    func validateSavePost() async throws {
```
lawmicha, line 345: I didn't see this used, can it be removed or is this useful for future stress testing use cases?

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/DataStoreStressTests.swift
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    func setUp(withModels models: AmplifyModelRegistration, logLevel: LogLevel = .error) async {
+        continueAfterFailure = false
+        await Amplify.reset()
+        Amplify.Logging.logLevel = logLevel
+
+        do {
+            try Amplify.add(plugin: AWSAPIPlugin(modelRegistration: models))
+            try Amplify.add(plugin: AWSDataStorePlugin(modelRegistration: models))
```
lawmicha, line 41: After reviewing the tests being added, I don't see any tests that stress the performance of the initial DataStore sync, so it will be better for the latency of a test by setting the max sync count to 100. Default is 10000 which will be the normal amount synced down after a few CI test runs against this backend.

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+type CustomerOrder @model
+   @key(fields: ["orderId","id"]) {
+   id: ID!
+   orderId: String!
+   email: String!
+}
+
+```
+3. If you are using the latest CLI, update cli.json to include `"useExperimentalPipelinedTransformer": false` to ensure that it will use the v1 transformer.
```
lawmicha, line 239: we should be using the latest defaults for new tests we are writing, ie. TransformerV2

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/README.md
```diff
@@ -0,0 +1,250 @@
...
+? Enter a description for the API key:
+? After how many days from now the API key should expire (1-365): `365`
+? Do you want to configure advanced settings for the GraphQL API `Yes, I want to make some additional changes.`
+? Configure additional auth types? `No`
+? Configure conflict detection? `Yes`
+? Select the default resolution strategy `Auto Merge`
+? Choose a schema template: `Blank Schema`
+? Do you want to edit the schema now? Y
+```
+When asked to provide the schema, create the `schema.graphql` file
```
lawmicha, line 25: is this the same schema as Android stress tests? If not, there will be more work in the future to align the tests cases.

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/README.md
```diff
@@ -0,0 +1,250 @@
...
+    id: ID!
+    content: String!
+    createdAt: AWSDateTime!
+    post: Post @connection(name: "PostComment")
+}
+
+## These are examples from https://docs.amplify.aws/cli/graphql-transformer/connection
+
+# 1 - Project has a single optional Team
+type Project1 @model {
```
lawmicha, line 55: Since these models are in the schema but not used in the test, is there a future plan to add different types of stress tests against them? Can you comment a bit on what future tests and their correlation to the remaining model graphs? Having them here now will make it easier to build the new tests in the future without having to run `amplify push` with the updated schema. We can just write new tests and reuse the same backend. If there are no future plans, i would prefer to remove unused models

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/README.md
lawmicha: which means we should also use V2 schema for testing, migrate `@connection` usage to `@hasMany`/`@belongsTo` etc

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/README.md
thisisabhash: Currently this schema is different than android.

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/DataStoreStressBaseTest.swift
```diff
@@ -0,0 +1,75 @@
...
+import AWSAPIPlugin
+
+@testable import Amplify
+@testable import AWSDataStorePlugin
+@testable import DataStoreHostApp
+
+class DataStoreStressBaseTest: XCTestCase {
+ 
+    static let amplifyConfigurationFile = "testconfiguration/AWSDataStoreStressTests-amplifyconfiguration"
+    let concurrencyLimit = 1
```
lawmicha, line 21: how do we reason about what `concurrencyLimit` to set?

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/README.md
lawmicha: looks like the schema has been greatly reduced to just one model, which means it's probably not that necessary to reuse the same schema, as Android stress tests could simply do the same paralleism on any model regardless of the name.

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreStressTests/DataStoreStressBaseTest.swift
thisisabhash: sorry, that was a typo meant for testing - updated it to previous value.
thanks for catching that!

lawmicha: how do we decide on the initial number and what data do we use to adjust this number? could it be done dynamically base on the system its running on? I feel like the rationale should live in the description of the PR to give the reader an overview or inline comments

lawmicha (approved):
LGTM, from just reading the code this looks like a good start for having tests that did not exist before. I would suggest capturing some of the answers from the comments and move it over to the description of the PR 
- why we are doing this at all, what's in scope (expected success in parallel API testing) and not in scope (ie. performance testing)
- rationale for the currency limit, and how this can be adjusted in the future
- which construct to use and why it achieves the goal of stress testing (ie. DispatchQueue.concurrentPerform vs TaskGroup)
- anything else you can think of for the reader coming back to this PR in the future

PR #2647, opened 2022-12-27
feat(api): Update url cache for storage and api to nil

royjit:
*Issue #, if available:*

*Description of changes:* Updates the cache control of Storage and API category to disable caching by default. This can be overridden by setting the Cache-Control request header for API rest category.

- [x] All unit tests pass

### DIFF
--- a/Amplify/Categories/API/Request/RESTRequest.swift
+++ b/Amplify/Categories/API/Request/RESTRequest.swift
@@ -32,9 +32,13 @@ public class RESTRequest {
                 headers: [String: String]? = nil,
                 queryParameters: [String: String]? = nil,
                 body: Data? = nil) {
+        if headers?["Cache-Control"] == nil {
+            self.headers = ["Cache-Control": "no-store"]
+        } else {
+            self.headers = headers
+        }
         self.apiName = apiName
         self.path = path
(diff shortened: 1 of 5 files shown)


Amplify/Categories/API/Request/RESTRequest.swift
```diff
@@ -32,9 +32,13 @@ public class RESTRequest {
                 headers: [String: String]? = nil,
                 queryParameters: [String: String]? = nil,
                 body: Data? = nil) {
+        if headers?["Cache-Control"] == nil {
+            self.headers = ["Cache-Control": "no-store"]
```
ruisebas, line 36: 🤔  wouldn't this be a problem if we pass a `header` parameter that does not have  a `"Cache-Control` value, but it does have others values?

For example, if we just provide `["content-type": "application/json"]` to the init, we'd end up with only `["Cache-Control": "no-store"]`, right?

ruisebas (changes requested):

Amplify/Categories/API/Request/RESTRequest.swift
royjit: Fixed

ruisebas (approved):

PR #2683, opened 2023-01-19
fix(api): decorate URLRequest with interceptors for subscription API request

5d:
## Description
- decorate URLRequest with interceptors for subscription API request

## General Checklist

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
+++ b/AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
@@ -64,85 +64,125 @@ final public class AWSGraphQLSubscriptionOperation<R: Decodable>: GraphQLSubscri
         }
 
         // Validate the request
+        if case let .failure(error) = validate(request: request) {
+            dispatch(result: .failure(error))
+            finish()
+            return
+        }
+
+        let endpointConfig = getEndpointConfig(from: pluginConfig, apiName: request.apiName)
+        let authType = (request.options.pluginOptions as? AWSPluginOptions).flatMap { $0.authType }
+        let urlRequestInterceptors = endpointConfig.flatMap {
+            getURLRequestInterceptors(
+                pluginConfig: pluginConfig,
+                endpointConfig: $0,
+                authType: authType
+            )
+        }
+
+        let urlRequest = endpointConfig.map { makeURLRequest(with: $0) }.flatMap { request in
+            urlRequestInterceptors.flatMap { interceptors in
+                decorateURLRequest(urlRequestInterceptors: interceptors, urlRequest: request)
+            }
+        }
+
+        switch (endpointConfig, urlRequest) {
+        case let (.failure(error), _), let (_, .failure(error)):
+            dispatch(result: .failure(error))
+            finish()
+        case let (.success(endpointConfig), .success(urlRequest)):
+            // Retrieve the subscription connection
+            subscriptionQueue.sync {
+                do {
+                    subscriptionConnection = try subscriptionConnectionFactory
+                        .getOrCreateConnection(for: endpointConfig,
+                                               urlRequest: urlRequest,
+                                               authService: authService,
+                                               authType: authType,
+                                               apiAuthProviderFactory: apiAuthProviderFactory)
+                } catch {
+                    let error = APIError.operationError("Unable to get connection for api \(endpointConfig.name)", "", error)
+                    dispatch(result: .failure(error))
+                    finish()
+                    return
+                }
+
+                // Create subscription
+
+                subscriptionItem = subscriptionConnection?.subscribe(requestString: request.document,
+                                                                     variables: request.variables,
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
```diff
@@ -90,6 +98,24 @@ final public class AWSGraphQLSubscriptionOperation<R: Decodable>: GraphQLSubscri
...
         }
 
+        var urlRequest = URLRequest(url: endpointConfig.baseURL)
+        do {
+            // set default user-agent value
+            urlRequest.setValue(AmplifyAWSServiceConfiguration.baseUserAgent(),
+                                forHTTPHeaderField: URLRequestConstants.Header.userAgent)
+            for inspector in reqeustInspectors {
+                urlRequest = try inspector.intercept(urlRequest)
+            }
```
lawmicha, line 108: The first change, setting the User-Agent is a miss that we definitely need to get in. The second change, decorating the request with the interceptors, have some more considerations.

Customers are able to provide the library with a custom interceptor, which is used for mutations and queries. Here, it is an array of interceptors because based on the auth type, we provide a default interceptor, and chain it with the customer provided interceptor. For example, if APIKey is the auth mode and the customer provided us an interceptor, there will be two interceptors in the chain. We will be unnecessarily intercepting the subscription request with the APIKey interceptor. Currently as an array, we don't know which is the APIKey interceptor and which is the customer's interceptor, unless we explicitly check if the interceptor is of a type that we know of, ie. APIKeyInterceptor/IAMInterceptor/CognitoUserPoolInterceptor/etc.  To my understanding, AppSync subscription connect requests's underlying HTTP request's headers are ignored, so decorating with all the interceptors seems uncessary. 

It seems like a bug that the customer's interceptor isn't used for subscription requests, but at the same time this isn't done for Android as well. (cc @manueliglesias), so I think if we agree that the subscription request should be decorated with the customer's interceptor, then we should do it across platforms, Android should have the same parity (not necessarily need to ship the changes at the same time, but perhaps a github issue on Android repo to track the changes). And we should figure out how to get the customer's interceptor and only use that, instead of the array of interceptors. 

Is there a risk of breaking existing customer's subscription requests? We don't know what the customer's interceptors are doing, and applying that to the subscription request in this change is a behavior change, which could impact some existing app's ability to establish the subscription properly. One way to avoid this is to have separate interceptors. At the call site, currently, developers add an interceptor through the plugin
```swift
try apiPlugin.add(interceptor: CustomInterceptor(), for: "yourApiName")
```
There are some ways to make this change opt-in for customers looking to intercept the subscrption requests going forward, for example
```
try apiPlugin.add(interceptor: CustomInterceptor(), for: "yourApiName", applyOnSubscriptions: true)
```

```
try apiPlugin.add(subscriptionInterceptor: CustomInterceptor(), for: "yourApiName")
```

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
```diff
@@ -64,58 +64,125 @@ final public class AWSGraphQLSubscriptionOperation<R: Decodable>: GraphQLSubscri
...
-                dispatch(result: .failure(error))
-                finish()
-                return
-            }
+    private func makeURLRequest(
+        with endpointConfig: AWSAPICategoryPluginConfiguration.EndpointConfig
+    ) -> URLRequest {
+        var urlRequest = URLRequest(url: endpointConfig.baseURL)
+        urlRequest.setValue(AmplifyAWSServiceConfiguration.baseUserAgent(),
+                            forHTTPHeaderField: URLRequestConstants.Header.userAgent)
```
lawmicha, line 150: we should cherry-pick this change and release it as is, without enabling subscription interceptors. This PR can continue to be the proof of concept for enabling that use case, but requires some cross platform involvement to make sure we have a consolidated developer experience

AmplifyPlugins/API/AWSAPICategoryPlugin/Operation/AWSGraphQLSubscriptionOperation.swift
manueliglesias: > if we agree that the subscription request should be decorated with the customer's interceptor, then we should do it across platforms

Yes, I agree on that.

> we should figure out how to get the customer's interceptor and only use that, instead of the array of interceptors

On iOS, is the same http client used for both queries/mutations and the WebSockets upgrade request? (On Android they are separate http clients)

> Is there a risk of breaking existing customer's subscription requests?

Yes, as you identified

> One way to avoid this is to have separate interceptors.

Right, I am aligned on this

On Android, consumers get [access to the whole http client builder, not just a way to add interceptors](https://docs.amplify.aws/lib/graphqlapi/authz/q/platform/android/#none). These preexisting deviations between platforms on the API will dictate what we can do next, but the overall theme of configuring http and websockets separately will prevail.

PR #2684, opened 2023-01-20
fix(datastore): observe API mutation event decode to model successfully

lawmicha:
## Description

This PR addresses a bug related to DataStore's observe API. When observing data through this API, a MutationEvent is received on the async sequence. The MutationEvent contains data about the mutation that occurred, version of the model, and the model encoded. To retrieve the model back, the developer can decode the mutation event's `json` property to a model type:
```
let receivedPost = try? mutationEvent.decodeModel(as: Post.self)

let receivedComment = try? mutationEvent.decodeModel(as: Comment.self)
```

The decoded model should have the lazy loading functionality available, for example. `receivedPost` should be able to lazy load its comments. `receivedComment` should be able to lazy load the posts. The resulting decoded model should have the same functionality as if a query occurred. This is essentially what's happening when the model gets reconciled, in the last step before emitting the MutationEvent, in ReconcileAndLocalSaveOperation. A query is performed to retrieve the model and is emitted to the Observe API subscribers as an encoded model inside a MutationEvent object.

This PR fixes the problem with encoding the queried model, for the MutationEvent, so that when the caller decodes the model from the MutationEvent, the lazy loading functionality will work.

The bug is that the encoding logic for LazyReference and List did not delegate it to the modelProvider, which contains the state of the object. Delegating the encoding functionality to the underlying modelProvider allows the modelProvider to control the shape of the encoded object, thus when the custom decoding logic runs, it will decode successfully back into a modelProvider though the ModelListDecoder/ModelProviderDecoders (DataStoreListDecoder/AppSyncListDecoder/DataStoreModelDecoder/AppSyncModelDecoder).

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/StorageEngineAdapter+UntypedModel.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/StorageEngineAdapter+UntypedModel.swift
@@ -10,7 +10,9 @@ import SQLite
 
 extension SQLiteStorageEngineAdapter {
 
-    func save(untypedModel: Model, completion: DataStoreCallback<Model>) {
+    func save(untypedModel: Model,
+              eagerLoad: Bool = true,
+              completion: DataStoreCallback<Model>) {
         guard let connection = connection else {
             completion(.failure(.nilSQLiteConnection()))
             return
@@ -33,8 +35,6 @@ extension SQLiteStorageEngineAdapter {
             let shouldUpdate = try exists(modelSchema,
                                           withIdentifier: untypedModel.identifier(schema: modelSchema))
 
-            // TODO serialize result and create a new instance of the model
-            // (some columns might be auto-generated after DB insert/update)
             if shouldUpdate {
                 let statement = UpdateStatement(model: untypedModel, modelSchema: modelSchema)
                 _ = try connection.prepare(statement.stringValue).run(statement.variables)
@@ -42,8 +42,23 @@ extension SQLiteStorageEngineAdapter {
                 let statement = InsertStatement(model: untypedModel, modelSchema: modelSchema)
                 _ = try connection.prepare(statement.stringValue).run(statement.variables)
             }
-
-            completion(.success(untypedModel))
+            
+            query(modelSchema: modelSchema,
+                  predicate: untypedModel.identifier(schema: modelSchema).predicate,
+                  eagerLoad: eagerLoad) {
+                switch $0 {
+                case .success(let result):
+                    if let saved = result.first {
+                        completion(.success(saved))
+                    } else {
+                        completion(.failure(.nonUniqueResult(model: modelSchema.name,
+                                                             count: result.count)))
+                    }
+                case .failure(let error):
+                    completion(.failure(error))
+                }
+            }
+            
         } catch {
             completion(.failure(causedBy: error))
         }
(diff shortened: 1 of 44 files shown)


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
```diff
@@ -54,6 +56,13 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
         self.mutationEventPublisher = PassthroughSubject<ReconcileAndLocalSaveOperationEvent, DataStoreError>()
 
         self.cancellables = Set<AnyCancellable>()
+        
+        // `isEagerLoad` is true by default, unless the models contain the rootPath
+        // which is indication that codegenerated model types support for lazy loading.
+        if isEagerLoad && ModelRegistry.modelType(from: modelSchema.name)?.rootPath != nil {
+            self.isEagerLoad = false
+        }
```
lawmicha, line 64: We need to control the eagerload or lazyloading, used by the queries that result in models encoded in MutationEvents. If the Model types do not support lazy loading (LazyReference encapsulation) then we need to continue eager loading on queries, much like the `DataStore.query` API path.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/StorageEngineAdapter+UntypedModel.swift
```diff
@@ -33,33 +35,51 @@ extension SQLiteStorageEngineAdapter {
...
             } else {
                 let statement = InsertStatement(model: untypedModel, modelSchema: modelSchema)
                 _ = try connection.prepare(statement.stringValue).run(statement.variables)
             }
-
-            completion(.success(untypedModel))
+            
+            query(modelSchema: modelSchema,
+                  predicate: untypedModel.identifier(schema: modelSchema).predicate,
+                  eagerLoad: eagerLoad) {
```
lawmicha, line 48: This method `save` used to return a model from the input, which is now inaccurate in the lazy loading cases. The input model `untypedModel` may be a model that was from the response of an Amplify.API mutation call. Imagine returning this model back to the caller in a MutationEvent on the DataStore Observe API. This will mean that the lazy loading functionality will delegate to the AppSyncModelProvider and AppSyncListProvider, calling into API directly. To avoid this, we do a query to get the latest data from the local store and return that data to the caller (model -> MutationEvent -> emit to Observe API)

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/SubscriptionSync/AWSIncomingEventReconciliationQueue.swift
```diff
@@ -209,7 +209,11 @@ extension AWSIncomingEventReconciliationQueue: Resettable {
 
         log.verbose("Resetting reconcileAndSaveQueue")
         reconcileAndSaveQueue.cancelAllOperations()
-        reconcileAndSaveQueue.waitUntilOperationsAreFinished()
+        // Reset is used in internal testing only. Some operations get kicked off at this point and do not finish
+        // We're sometimes hitting a deadlock when waiting for them to finish. Commenting this out and letting
+        // the tests continue onto the next works pretty well, but ideally ReconcileAndLocalSaveOperation's should 
+        // always finish. We can uncomment this to explore a better fix that will still gives us test stability.
+        //reconcileAndSaveQueue.waitUntilOperationsAreFinished()
```
lawmicha, line 216: See comment above, this seemed to help with test stability in the LazyLoading tests. A deadlock happens when we call `DataStore.save`, wait for first mutation event, and then finish by calling reset. Reset calls `reconcileAndSaveQueue.waitUntilOperationsAreFinished()` while another subscription event comes in which causes a ReconcileAndLocalSaveOperation to start, but never finishes, while this thread is waiting for all operations to finish. I think this must have been happening in the DataStoreIntegrationTests as well.

lawmicha:
**Workflow dispatch w/ this branch**

[DataStore Integration test](https://github.com/aws-amplify/amplify-swift/actions/runs/3969440515)

[DataStore LazyLoading integ tests](https://github.com/aws-amplify/amplify-swift/actions/runs/3969441901)

PR #2687, opened 2023-01-20
fix(auth): Unblock fetchAuthSession call during a signIn flow

royjit:
## Issue \#
https://github.com/aws-amplify/amplify-swift/issues/2608

## Description
Current implementation of the auth state machine blocks any fetch Auth session call to be invoked if there is a signIn is in progress. This causes issue if we want to call fetchAuthSession when the library is waiting for user input for confirm signIn. This PR unblocks the fetchAuthSession api during signIn invocation. 

Note that this does not interfere in a signIn flow because only one api will be executing at a time in the auth plugin.

## General Checklist

- [x] All unit tests pass


jcjimenez (approved):
LGTM

PR #2689, opened 2023-01-24
fix: Add missing PropertyPath types double and int

lawmicha:
## Description

Add missing `int` and `double` to the PropertyPath. This PR also adds two sets of models (LL14 and LL15) to the lazy loading schema and have been used to update the shared backend provisioning. The two schemas, each has a use case that utilizes the missing property path method `int` and `double`.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/Amplify/Categories/DataStore/Model/PropertyPath.swift
+++ b/Amplify/Categories/DataStore/Model/PropertyPath.swift
@@ -123,4 +123,8 @@ open class ModelPath<ModelType: Model> : PropertyContainerPath {
     public func time(_ name: String) -> FieldPath<Temporal.Time> {
         FieldPath(name: name, parent: self)
     }
+    
+    public func int(_ name: String) -> FieldPath<Int> {
+        FieldPath(name: name, parent: self)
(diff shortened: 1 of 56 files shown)


Amplify/Categories/DataStore/Model/PropertyPath.swift
```diff
@@ -123,4 +123,12 @@ open class ModelPath<ModelType: Model> : PropertyContainerPath {
...
         FieldPath(name: name, parent: self)
     }
+    
+    public func int(_ name: String) -> FieldPath<Int> {
+        FieldPath(name: name, parent: self)
+    }
+    
+    public func double(_ name: String) -> FieldPath<Double> {
+        FieldPath(name: name, parent: self)
+    }
```
lawmicha, line 133: This is the only functional change in this PR which is why it is a "fix", the newly added schemas have fields of type `int` and `double` which uses these methods. They are added to the integraton test target to ensure they compile, provisioned to the shared backend to anticipate more testing

PR #2694, opened 2023-01-25
fix(auth): Remove static deviceMetaData in auth flow

royjit:
## Issue \#
NA

## Description
DeviceMetaData is stored in keychain which are mapped to `username`. But the `username` value changes during the signIn flow, because in the username defined by Cognito is different from the one entered by user. For example in case of user alias if user entered a phone number as `username`, on `initiateAuth` will return a UUID as `username`. 

We should use this username to see if we have a mapping of deviceMetaData on the next step of authflow and should not keep a static deviceMetadata.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/SignIn/DeviceSRPAuth/VerifyDevicePasswordSRP.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/SignIn/DeviceSRPAuth/VerifyDevicePasswordSRP.swift
@@ -33,12 +33,15 @@ struct VerifyDevicePasswordSRP: Action {
             let secretBlock = try secretBlock(secretBlockString)
             let serverPublicB = try serverPublic(parameters)
 
-            guard case .metadata(let deviceData) = stateData.deviceMetadata else {
+            let deviceMetadata = await DeviceMetadataHelper.getDeviceMetadata(
+                for: username,
+                with: environment)
+            guard
+                case .metadata(let deviceData) = deviceMetadata
+            else {
                 let authError = SignInError.service(error: SRPError.calculation)
                 logVerbose("\(#fileID) DevciceSRPSignInError \(authError)", environment: environment)
-                let event = SignInEvent(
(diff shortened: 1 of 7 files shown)


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/SignIn/DeviceSRPAuth/VerifyDevicePasswordSRP.swift
```diff
@@ -58,6 +61,7 @@ struct VerifyDevicePasswordSRP: Action {
                 session: authResponse.session,
                 secretBlock: secretBlockString,
                 signature: signature,
+                deviceMetadata: deviceMetadata,
```
royjit, line 64: Check if we need to send this on respondToAuthChallenge, in a separate PR.

palpatim (approved):
Approved to merge into a feature branch. Don't ship this until we understand how Cognito uses device metadata during the initial `InitiateAuth` call.

Specifically, what does Cognito do in each of the following circumstances:

* Pass device metadata key `123` during `InitiateAuth`, but pass device metadata key `456` during `RespondToAuthChallenge: PASSWORD_VERIFIER`
* Do not pass device metadata during `InitiateAuth`, but DO pass `123` during `RespondToAuthChallenge: PASSWORD_VERIFIER`
* Do not pass device metadata during both `InitiateAuth` and `RespondToAuthChallenge: PASSWORD_VERIFIER`, but DO pass `123` during `RespondToAuthChallenge: DEVICE_SRP_AUTH` and `RespondToAuthChallenge: DEVICE_PASSWORD_VERIFIER`
* Pass `123` during both `InitiateAuth` and `RespondToAuthChallenge: PASSWORD_VERIFIER`, but pass `456` during `RespondToAuthChallenge: DEVICE_SRP_AUTH` and `RespondToAuthChallenge: DEVICE_PASSWORD_VERIFIER`
* Pass `123` during `RespondToAuthChallenge: DEVICE_SRP_AUTH` but pass `456` `RespondToAuthChallenge: DEVICE_PASSWORD_VERIFIER`

PR #2704, opened 2023-01-26
chore(auth): Integration tests fail fast on failed assertions

jcjimenez:
## Issue

https://github.com/aws-amplify/amplify-swift/issues/2705

## Description

Ensuring integration tests fail fast on assertions such as those for a successful sign-in. This helps reduce noise during investigations. Think of this as the unit test equivalent of an [andon cord](https://en.wikipedia.org/wiki/Andon_(manufacturing)).

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/AuthDeleteUserTests/AuthDeleteUserTests.swift
+++ b/AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/AuthDeleteUserTests/AuthDeleteUserTests.swift
@@ -15,10 +15,12 @@ import AmplifyTestCommon
 class AuthDeleteUserTests: AWSAuthBaseTest {
 
     override func setUpWithError() throws {
+        continueAfterFailure = false
         try initializeAmplify()
     }
 
     override func tearDownWithError() throws {
+        _ = Amplify.Auth.signOut()
         Amplify.reset()
         sleep(2)
     }
@@ -50,10 +52,10 @@ class AuthDeleteUserTests: AWSAuthBaseTest {
             defer {
                 authSignedInSessionExpectation.fulfill()
             }
-            do {
-                let session = try result.get()
+            switch result {
+            case .success(let session):
                 XCTAssertTrue(session.isSignedIn, "Auth session should be signedIn")
-            } catch {
+            case .failure(let error):
                 XCTFail("Fetch auth session failed with error - \(error)")
             }
         }
@@ -66,7 +68,7 @@ class AuthDeleteUserTests: AWSAuthBaseTest {
             }
             switch result {
             case .success:
-                print("Success deleteUser")
+                break
             case .failure(let error):
                 XCTFail("deleteUser should not fail - \(error)")
             }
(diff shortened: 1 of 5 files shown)


AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/AuthDeleteUserTests/AuthDeleteUserTests.swift
```diff
@@ -39,34 +40,33 @@ class AuthDeleteUserTests: AWSAuthBaseTest {
         AuthSignInHelper.registerAndSignInUser(username: username,
                                                password: password,
                                                email: email) { didSucceed, error in
-            registerAndSignInExpectation.fulfill()
+            if didSucceed {
+                registerAndSignInExpectation.fulfill()
+            }
             XCTAssertTrue(didSucceed, "SignIn operation failed - \(String(describing: error))")
```
ameter, line 46: An alternative to checking and then asserting for success would be to defer the expectation fulfillment.  The end result would be the same in this case, but it would only require one check.  The deferral pattern would have added benefit in cases where more assertions are added because it would ensure that the expectation is only fulfilled after all assertions have been met.

ameter (approved):
approved with comment

AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/AuthDeleteUserTests/AuthDeleteUserTests.swift
jcjimenez: Yeah, there are some cases where we do follow this pattern (some of which I left intact). However, the failed assertions didn't cause the unit test to halt (unless `continueAfterFailure` is set to false) and those things would proceed to cause chaos after the `wait(for:timeout:)`. I may be misunderstanding - you meant something like the example below, right?

```
func testSomething() throws {
    let e = expectation(description: "foo")
    performSomethingAsync { (didSucceed, error) in
        defer {
            e.fulfill()
        }
        XCTAssertTrue(didSucceed)
    }
    wait(for: [e], timeout: 1.0)
    // If the operation above succeeded, perform the following:
   ...
}
```

Another way we could do this is save a value representing the success (e.g. `didSucceed` or `session`), then assert after waiting for the expectation (and setting `continueAfterFailure` to `false`). What do you think?

ameter: Right, the above requires `continueAfterFailure` to be set to false if you want to end the test after a failure.  I don't have a strong opinion about which pattern to use, just pointing out alternatives.

ameter: for the purposes of this PR, I think it's good as written

AmplifyPlugins/Auth/AWSCognitoAuthPluginIntegrationTests/AuthDeleteUserTests/AuthDeleteUserTests.swift
```diff
@@ -15,10 +15,12 @@ import AmplifyTestCommon
 class AuthDeleteUserTests: AWSAuthBaseTest {
 
     override func setUpWithError() throws {
+        continueAfterFailure = false
```
jcjimenez, line 18: Considered putting this in `AWSAuthBaseTest.init` but didn't want to touch other tests as part of this.

ameter (approved):
LGTM

PR #2710, opened 2023-01-27
fix(test): data race on reseting in tear down

5d:
## Description

- according to the `HubChannelDispatcher.destroy` [description](https://github.com/aws-amplify/amplify-swift/blob/7f094a9e88612b5b8fda1bce01ce71e7f1915053/Amplify/DefaultPlugins/AWSHubPlugin/Internal/HubChannelDispatcher.swift#L60-L64), wait 1 second after `Amplify.reset` for all operations to be finished in dispatcher queue.

## General Checklist

- [ ] All unit tests pass


AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/ClientBehaviorAPITests/AWSCognitoAuthDeviceBehaviorTests.swift
```diff
@@ -22,10 +25,6 @@ class AWSCognitoAuthDeviceBehaviorTests: XCTestCase {
                          hubEventHandler: MockAuthHubEventBehavior())
     }
 
-    override func tearDown() {
-        Amplify.reset()
-    }
```
lawmicha, line 27: It seems more logically to do the reset in the `tearDown` method, as a "clean up" step of all tests. Is there a reason to move this to the "setUp"? I would think we can still call `wait(for: 1)` in the tearDown after `Amplify.reset()`

AmplifyPlugins/API/AWSAPICategoryPluginTests/AWSAPICategoryPluginTestBase.swift
```diff
@@ -71,7 +72,12 @@ class AWSAPICategoryPluginTestBase: XCTestCase {
         }
     }
 
-    override func tearDown() {
-        Amplify.reset()
```
lawmicha, line 75: Will it fix the problem with the data race if we kept the `Amplify.reset()` in `tearDown`, added the `wait(for: 1)` here and remove `Amplify.reset()` from the `setUp()`?

lawmicha: i think this means some tests are not calling `Amplify.reset()` on tearDown, which could happen if not all tests are using this test base. In that case, then calling reset in the setUp of this test makes sense since we can't guaranteed  that other tests are resetting properly

lawmicha (approved):

PR #2714, opened 2023-01-27
chore(auth): Cleanup on FetchAuthSessionOperationhelper

royjit:
## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/CredentialStorage/AmplifyCredentials+CognitoSession.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/CredentialStorage/AmplifyCredentials+CognitoSession.swift
@@ -9,9 +9,7 @@ import Foundation
 import Amplify
 
 extension AmplifyCredentials {
-
     static let expiryBufferInSeconds = TimeInterval.seconds(2 * 60)
-    
     var cognitoSession: AWSAuthCognitoSession {
 
         switch self {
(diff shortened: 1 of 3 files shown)


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/CredentialStorage/AmplifyCredentials+CognitoSession.swift
```diff
@@ -48,4 +50,37 @@ extension AmplifyCredentials {
             return AuthCognitoSignedOutSessionHelper.makeSessionWithNoGuestAccess()
         }
     }
+
+    func areValid() -> Bool {
```
atierian, line 54: Does the name `areValid()` convey the right meaning here? It looks like this function defines whether credentials "are valid" in three different ways:
1. no credentials
2. credentials that expire within predefined buffer
3. credentials that are already expired

**Question:** Does it make sense to extract those out in private helper functions to keep the complexity down? 
**Recommendation:** Use a function body local to drop the `if` statements nested in the `switch`. Something like:
```swift
let valid: Bool
// ...
switch self {
case ...:
  valid = !data.cognitoUserPoolTokens.doesExpire(in: ...)
}
return valid
```
**Request:** Regardless of how you want to handle ^, I'd ask that we add a comment in the `identityPoolWithFederation` case explaining why it doesn't need/use `expiryBufferInSeconds`

royjit: Thanks for the suggestions.
Function split to better understand and added the recommendation. I added the same buffer time for federation with identity pool

atierian (approved):

PR #2720, opened 2023-01-31
fix: add granular read ops enum

AaronZyLee:
## Issue \#

https://github.com/aws-amplify/amplify-flutter/issues/2526
## Description

Add granular read operation enums in model auth operation rule
Refer https://docs.amplify.aws/cli/graphql/authorization-rules/#how-it-works
## General Checklist

- [ ] All unit tests pass

### DIFF
--- a/Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
+++ b/Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
@@ -22,6 +22,12 @@ public enum ModelOperation {
     case update
     case delete
     case read
+    // granular read access
+    case `get`
+    case list 
+    case sync 
+    case listen 
+    case search
 }
 
 /// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly


phani-srikar (commented):
Are there any existing tests associated with these additional operations?

royjit (approved):

5d:
> Are there any existing tests associated with these additional operations?

@phani-srikar ; No, we don't have test cases cover these additional operations.

@AaronZyLee ; Have you performed manual testing to ensure that these additional operation types are functioning as expected?

Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
```diff
@@ -22,6 +22,12 @@ public enum ModelOperation {
     case update
     case delete
     case read
+    // granular read access
+    case `get`
+    case list 
+    case sync 
+    case listen 
+    case search
 }
```
lawmicha, line 31: Can we start off by attaching the generated Model swift types that utilizes these new granular read access cases? By doing so, we can update the source schema that we use to reference the generated types. 

Is there an sort of relevant API calls that we can write, either in the unit or integration tests, utilizing these models, and asserting the expected behavior?

lawmicha:
In DataStore, we get the authtypes based on `.read`. Since developers can use this schema with DataStore as long as they allow listen and sync. then this code probably needs to be updated to getting auth types for `.read`, `sync` or `listen`. ?
```
var authTypes = await authModeStrategy.authTypesFor(schema: modelSchema, operation: .read)
```

Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
```diff
@@ -22,6 +22,12 @@ public enum ModelOperation {
     case update
     case delete
     case read
+    // granular read access
+    case `get`
+    case list 
+    case sync 
+    case listen 
+    case search
```
lawmicha, line 30: I bet a lot of logic is determined by checking if the operation is `.read` and does something. Now that we allow compiling the more granular "read" types, that means all logic based on "read" should now be updated to

1. If the previous check for read expected all granular read access then update the code to "if operation == read || operations.containsAll(list, sync, listen, search)"  
2. If the previous check for read expected just the sync API (for example DataStore's sync operation), then code can be updated to "if operation == read || operation == sync)"
3. If previous check for read expected that the subscription API is used then --> "operation == read or operation ==  "listen"

Amplify/Categories/DataStore/Model/Internal/Schema/AuthRule.swift
AaronZyLee: The example schema using it is 
```
type Todo @model
@auth(rules:[
  { allow: public, operations: [get, list, listen, sync, search] },
  { allow: owner }
])
{
  id: ID!
  name: String
}
```
The model files can be generated by the current version of modelgen, which has the following line generated
```
    model.authRules = [
      rule(allow: .public, operations: [.get, .list, .listen, .sync, .search]),
      rule(allow: .owner, ownerField: "owner", identityClaim: "cognito:username", provider: .userPools, operations: [.create, .update, .delete, .read])
    ]
```

AaronZyLee:
> In DataStore, we get the authtypes based on `.read`. Since developers can use this schema with DataStore as long as they allow listen and sync. then this code probably needs to be updated to getting auth types for `.read`, `sync` or `listen`. ?
> 
> ```
> var authTypes = await authModeStrategy.authTypesFor(schema: modelSchema, operation: .read)
> ```

The datastore usage is out of scope of this PR as mentioned in the description. This is for the API users to get the modelgen files compiled

lawmicha:
> > In DataStore, we get the authtypes based on `.read`. Since developers can use this schema with DataStore as long as they allow listen and sync. then this code probably needs to be updated to getting auth types for `.read`, `sync` or `listen`. ?
> > ```
> > var authTypes = await authModeStrategy.authTypesFor(schema: modelSchema, operation: .read)
> > ```
> 
> The datastore usage is out of scope of this PR as mentioned in the description. This is for the API users to get the modelgen files compiled

Customers can use DataStore, and if they have more granular access including listen and sync, then their models may not work as expected because DataStore's sync code checks for "read" access, not "read" or "sync" access. By allowing them to compile the App with the generated models, will cause other unintended behaviors when using DataStore at runtime

lawmicha:
Next steps, this comment https://github.com/aws-amplify/amplify-swift/pull/2720#discussion_r1313212958 captures some the concern over introducing more granular operations that was "sub-operations" of `read`. Logic that checks for "read" should consider checking for "read" OR one or more of the granular operations

marcoboerner:
Is there any update on this? I'm using Amplify Gen2 as well and don't get any errors when adding `get` to my schema, the backend deploys just fine, but then the code generation fails / produces unknown enum cases.

PR #2701, opened 2023-01-25
fix: do not set nil values for create mutation graphQL input translation

lawmicha:
## Description

Create mutations do not need to set `nil` values for fields that are not populated on the model instance.

Add additional integration tests such as 
- start of some DataStore HasOneParent/HasOneChild tests
- API list query tests
- DataStore observeQuery tests
- API nested selection set tests

## General Checklist

- [ ] Added new tests to cover change, if needed
- [ ] Build succeeds with all target using Swift Package Manager
- [ ] All unit tests pass
- [ ] All integration tests pass
- [ ] Security oriented best practices and standards are followed (e.g. using input sanitization, principle of least privilege, etc)
- [ ] Documentation update for the change if required
- [ ] PR title conforms to conventional commit style
- [ ] New or updated tests include `Given When Then` inline code documentation and are named accordingly `testThing_condition_expectation()`
- [ ] If breaking change, documentation/changelog update with migration instructions

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2721, opened 2023-01-31
refactor(datastore): convert with schema fields

5d:
## Description

This is a functional change in `Statement+Model.swift` during the phase where SQL results are converted to a dictionary for decoding to the model types. The convert method has been refactored to iterate over the model fields rather than the SQL results themselves. This allows us cleaner control over what is being populated for each key of the dictionary. 

## General Checklist

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/Statement+Model.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/Statement+Model.swift
@@ -90,6 +90,12 @@ extension Statement: StatementModelConvertible {
         eagerLoad: Bool = true,
         path: [String] = []
     ) throws -> ModelValues? {
+        guard let maxDepth = columnNames.map({ $0.split(separator: ".").count }).max(),
+              path.count < maxDepth
+        else {
+            return nil
+        }
+
         let modelValues = try modelSchema.fields.mapValues {
             try convert(field: $0, schema: modelSchema, using: statement, from: row, eagerLoad: eagerLoad, path: path)
         }
@@ -124,9 +130,10 @@ extension Statement: StatementModelConvertible {
             return DataStoreModelDecoder.lazyInit(identifier: targetBinding)
 
         case let (.model(modelName), true):
-            guard !path.contains(field.name),
-                  let modelSchema = getModelSchema(for: modelName, with: statement)
-            else { return nil }
+            guard let modelSchema = getModelSchema(for: modelName, with: statement)
+            else {
+                return nil
+            }
 
             return try convert(
                 row: element,


lawmicha (commented):
great refactor and fixing the problem with uni-directional has-one!

lawmicha:
Testing notes
- LazyLoading integration tests are all passing! Ideally we get https://github.com/aws-amplify/amplify-swift/pull/2689 and https://github.com/aws-amplify/amplify-swift/pull/2701 in and rebase this PR to get the additional tests that were added from those two PRs
- DataStore Integration tests are also passing!
- V2 integration tests have some failures in the OptionalAssociations tests
 
image

V2 integration tests refer to the V2 Transformer used for those schemas (see readme for provisioning steps). We do need to fix the implementation so that it works with the previously codegenerated types that do not have the LazyReference and ModelPaths. This is to allow backwards compatiable in the library w/o making a breaking change version. Looks like the only test failures are in  `DataStoreConnectionOptionalAssociations` from the V2 test target.

Some of the unit test failures also appear to be related to optional associations for the previous generated models, ie. `StorageEngineTestsOptionalAssociation`

lawmicha:
From the first commit, running the tests in `StorageEngineTestsOptionalAssociation`

```
### values ["elements": 
    Optional([
        ["id": Optional("F8F2B220-34EC-48A8-A707-7B666BA10D84"), 
         "post": Optional([
            "updatedAt": nil, 
            "__typename": Optional("Post8"), 
            "blog": Optional([
                "__typename": Optional("Blog8"), 
                "updatedAt": nil, 
                "name": nil, 
                "customs": nil, 
                "notes": nil, 
                "createdAt": nil, 
                "posts": nil, 
                "id": nil]), 
            "name": nil, 
            "comments": nil, 
            "id": nil, 
            "randomId": nil, 
            "createdAt": nil]), 
        "content": Optional("content"), 
        "createdAt": nil, 
        "updatedAt": nil, 
        "__typename": Optional("Comment8")]])]
```

this payload fails to decode to the Comment8 model due to 

```
valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "elements", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "post", intValue: nil), CodingKeys(stringValue: "id", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
```

For previously generated models, we always have to keep eager loading to true. In the test scenario where Comment optionally belongs to a Post, we're initially saving a comment without associating a post. The storage engine adapter does an InsertStatement, followed by a query to return the saved data. The query is done with eager load true, and converts the data to `values` above, before decoding to `Comment8`. As seen in the data, the post has all of its fields with `nil` values. This fails to decode to the post field on the Comment with "Expected String value but found null instead". A payload capable of decoding would look something like this:

```
### values ["elements": 
    Optional([
        ["id": Optional("F8F2B220-34EC-48A8-A707-7B666BA10D84"), 
         "post": nil, 
        "content": Optional("content"), 
        "createdAt": nil, 
        "updatedAt": nil, 
        "__typename": Optional("Comment8")]])]
```

For some historical context, I previously added an implementation that did some post processing within the `convert` method here https://github.com/aws-amplify/amplify-swift/pull/1849 . I doubt we need this specific piece of code in this PR since this logic has been changed, but just a reference for what was done to make this work before

lawmicha:
Testing notes on latest commits

- LazyLoading test ✅
- V2  ✅
- MultiAuth - I had some issues running locally, and think it's related to my backend API key expired. Kicked off a workflow here: https://github.com/aws-amplify/amplify-swift/actions/runs/4066664590
- IAM ✅
- Cognito ✅
- CPK ✅

lawmicha (approved):

lawmicha (approved):

PR #2737, opened 2023-02-03
fix(datastore): load hasOne and belongsTo lazy reference with composite key

5d:
## Description

- build `LazyReference` field for `hasOne` and `belongsTo` based on foreign table's primary key and query result columns.
- fix broken test cases.

## General Checklist

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Core/DataStoreModelDecoder.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Core/DataStoreModelDecoder.swift
@@ -24,15 +24,7 @@ public struct DataStoreModelDecoder: ModelProviderDecoder {
         }
 
         func toJsonObject() -> Any? {
-            return [
-                "identifiers": identifiers.map({
-                    [
-                        "name": $0.name,
-                        "value": $0.value
-                    ]
-                }),
-                "source": source
-            ]
+            try? JSONSerialization.jsonObject(with: JSONEncoder().encode(self))
         }
     }
     


Amplify/Categories/DataStore/Model/Model.swift
```diff
@@ -46,6 +46,35 @@ extension Model {
...
+    }
+
+    public static func identifier(
+        of lazyReferenceField: ModelField,
+        from assoicatedModelName: String,
+        associatedFieldsData: [String: Persistable?]
+    ) -> ModelIdentifierProtocol? {
+        // according to codegen logic:
+        // https://github.com/aws-amplify/amplify-codegen/blob/main/packages/appsync-modelgen-plugin/src/utils/process-connections.ts#L41
+        // associated field name has format ${type}${field}${primaryKey} in camel case.
```
lawmicha, line 60: can this move over to the datastore plugin package? At first glance, the reference to codegen/transformer is an implementation detail.

Looks only used in plugin's `Statement+Model` 
```
return DataStoreModelDecoder.lazyInit(
                    identifier: associatedFieldsData.flatMap {
                        modelType.identifier(of: field, from: modelName, associatedFieldsData: $0)
                    }
                )
```

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -68,7 +68,7 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
         storageEngine.save(model,
                            modelSchema: modelSchema,
                            condition: condition,
-                           eagerLoad: true,
+                           eagerLoad: isEagerLoad,
```
lawmicha, line 71: Thinking about this some more, `isEagerLoad` will be set to false for latest codegenerated types with LazyReference/ModelPaths. When this is false, it will lazy load on save's (create/update), the queried model used to sync to AppSync will contain a LazyReference in a not loaded state. 

A DataStore-sourced LazyReference's not loaded state will have identifiers in the format of (fieldName: `@@primaryKey`, value: `idValue123#titlesortKey123`) - when model is CPK. The logic to extract out the value will be in `Model+GraphQL.swift`. is this correct so far?

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -162,24 +162,30 @@ extension Model {
...
-                preconditionFailure(
-                """
-                Associated model target names and values for field \(field.name) of model \(modelName) mismatch.
-                There is a possibility that is an issue with the generated models.
-                """
-                )
-            } else if mutationType == .update {
-                // otherwise, pad the values with `nil` to account for removals of associations on updates.
-                values = [Persistable?](repeating: nil, count: fieldNames.count)
+            let defaultValues = [Persistable?](repeating: nil, count: fieldNames.count)
```
lawmicha, line 173: Do we still need the check on the `mutationType` that was removed in this PR? 

Some context on this: create mutations don't have to send `defaultValues`, only on update mutations so that it acts as a removal of the association. No-op if there is no association originally. Let's take Comment, Post with CPK, example

if it is a create mutation, the comment object translated to the graphQL input will fail when `null` values are sent for the post foreign key fields
```
// this will fail on AppSync, trying to set foreign key to null
{
   "id": "commentId123"
   "postId": null,
   "postTitle": null
}
```

For create mutations, we don't pad the values with null so that only the fields that are being persisted are sent:
```
{
   "id": "commentId123"
}
```
There was also an upper level check done in [here](https://github.com/aws-amplify/amplify-swift/pull/2701/files#r1093600308) so you may want to rebase with latest `data-dev-preview` to get the latest changes, so I wonder if this check is still required.

For update mutations, we do want to pad it with `nil` values:
```
// this will fail on AppSync, trying to set foreign key to null
{
   "id": "commentId123"
   "postId": null,
   "postTitle": null
}
```
When the comment is not associated with a post, then `null` updates are okay since no change in data. When the comment is deassociated with a post, then the `null` values will update the existing values in DynamoDB to `null`

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -162,24 +162,30 @@ extension Model {
...
+            if field.isAssociationOwner && fieldNames.count > 1 && values.count == 1 {
+                let discreteKeys = values.compactMap { persistable in
+                    persistable.map({ String(describing: $0) })
+                        .map({ $0.split(separator: ModelIdentifierFormat.Custom.separator.first!) })
+                        .map({ $0.map({
+                            $0.trimmingCharacters(in: CharacterSet(charactersIn: "\""))}) as? Persistable
+                        })
+                }
+                values = discreteKeys.count == values.count ? discreteKeys : defaultValues
```
lawmicha, line 182: This code path is used by API sourced objects as well. For example
```
var comment = try await Amplify.API.query(.get(Comment.self, byId: "commentId"))
comment.post = newPost
var updatedComment = try await Amplify.mutate(.update(comment))
```
The queried comment `comment` will contain a LazyReference post in a not loaded state with fields: ["postId", "postTitle"] and identifiers: ["postId123", "postTitle123"]. Does the translation still work as expected for API sourced objects? Let's make sure the API integration tests still pass, especially the ones in AWSAPIPluginLazyLoadTests. I see some failures 

image

which look related to
- saving a comment without a post
- saving project and updating by removing the team associated to it. 

This may be related to the `mutationType` check on `update` mutations.

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginLazyLoadTests/LL5/Team1+Schema.swift
```diff
@@ -28,7 +28,7 @@ extension Team1 {
     model.fields(
       .field(team1.teamId, is: .required, ofType: .string),
       .field(team1.name, is: .required, ofType: .string),
-      .belongsTo(team1.project, is: .optional, ofType: Project1.self, targetNames: ["team1ProjectProjectId", "team1ProjectName"]),
+      .belongsTo(team1.project, is: .optional, ofType: Project1.self, associatedFields: [Project1.keys.projectId, Project1.keys.name], targetNames: ["team1ProjectProjectId", "team1ProjectName"]),
```
lawmicha, line 31: These files, which are code generated from `amplify codegen models` process, need to be updated to generate `associatedFields`. We can definitely explore doing that in this file: https://github.com/aws-amplify/amplify-codegen/pull/541/files#diff-eaba881c8c73aea316d5f63356d638f20a3bf275c7a4b66d8e97e1070c2de177 similar to how we've added the `associatedFields` for has-many and has-one model fields which replaces "associatedWith", but I'm evaluating whether we really need to, or if we already have all the information. 

## Codegen

image

In the Codegen code, the swift visitor doesn't populate `connectionInfo` and is processed at the platform agnostic level. There will be a much heavier lift to achieve this by populating connectionInfo with associatedFields and then codegenerating the info into the swift schema model files.

## has-many

has-many "associatedFields" is required because we're storing the parent's foreign key in the lazy child list. We don't know which FK on the child is the parent for uni-directional has-many unless we pass that information to the child lazy object from the parent. We were already generating "associatedWith" but should have been generating `associatedFields` for CPK uni-directional has-many use case. Lazy loading a list of child models uses the parent's FK fields on the child and the predicate values are the parent's identifiers, ie. `query(Child.self, where: Child.keys.parentFK == "parentId123")` 

## has-one

Evaluating whether the library needs to use the information of the has-one "associatedFields" / "associatedWith" for our use cases. It uses targetNames to extract out the value, ie. `element[targetname]`. Then the primary key values should be passed to the LazyReference.

LazyReference child only needs its own primary key (the child primary key values). DataStore can lazy load itself by using `ModelType.schema.primaryKey.sqlName` as the key since it will resolve to "id" for non-CPK and `@@primaryKey` for CPK use cases. If `@@primaryKey` then concat the identifiers using "#" to get the final query to be `query(Child.self, where: "@@primaryKey" == "id123#title123".  The child's schema already knows which fields are the primary keys, we only need the value (or values). 

Syncing a parent only needs its the child primary key values as well. API Model helpers can translate the parent object to the GraphQL input by pulling out the child primary key values from the not loaded Lazy Reference and then use the targetName as the key:
```
 { 
"projectid": "project123",
...
"hasOneParentChildId": "childId123" // "childId123" pulled from the LazyReference, targetName "hasOneParentChildId" pulled the parent's hasOne modelField's targetNames
}
```

It doesn't seems like we need `associatedWith`/`associatedFields` at all. The parent element and the targetNames gives us the child primary key values, the child schema gives us the primary key field names.

## belongs-to

Evaluating whether the library needs to use the belongs-to's "associatedFields" for our use cases. Applying the same logic from the has-one case to see if it checks out:

Use the targetNames to extract out the value, ie. `ChildElement[targetname]`. Then the primary key values should be passed to the LazyReference parent.

LazyReference parent only needs its own primary key (the parent's primary key values). DataStore can lazy load itself by using `ModelType.schema.primaryKey.sqlName` as the key since it will resolve to "id" for non-CPK and `@@primaryKey` for CPK use cases. If `@@primaryKey` then concat the identifiers using "#" to get the final query to be `query(Parent.self, where: "@@primaryKey" == "id123#title123".  The child's schema already knows which fields are the primary keys, we only need the value (or values). 

Syncing a child only needs its the parent primary key values as well. API Model helpers can translate the child object to the GraphQL input by pulling out the parent primary key values from the not loaded Lazy Reference and then use the targetName as the key:
```
 { 
"id": "parent123",
...
"childParentId": "parentId123" // "parent123" pulled from the LazyReference, targetName "chidlParentId" pulled the child's belongsTo modelField's targetNames
}
```

Amplify/Categories/DataStore/Model/Internal/Schema/ModelField+Association.swift
```diff
@@ -88,28 +88,48 @@ import Foundation
 ///   directly by host applications. The behavior of this may change without warning.
 public enum ModelAssociation {
     case hasMany(associatedFieldName: String?, associatedFieldNames: [String] = [])
-    case hasOne(associatedFieldName: String?, targetNames: [String])
-    case belongsTo(associatedFieldName: String?, targetNames: [String])
+    case hasOne(associatedFieldName: String?, associatedFieldNames: [String] = [], targetNames: [String])
+    case belongsTo(associatedFieldName: String?, associatedFieldNames: [String] = [], targetNames: [String])
```
lawmicha, line 92: It appears we have defined `belongsTo` to hold the associatedFieldName but the codegen output never generates this information (see https://github.com/aws-amplify/amplify-codegen/pull/541/files#diff-eaba881c8c73aea316d5f63356d638f20a3bf275c7a4b66d8e97e1070c2de177 line 634)

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
lawmicha: Continuation of the functional change to enable `isEagerLoaded` computed property for DataStore.save's. 

A DataStore-sourced LazyReference's not loaded state identifiers format will have to change (as you've done in this PR). By syncing models with lazy loading, the identifiers of a non loaded LazyReference needs to be extracted out of the LazyReference objects and the targetNames used as keys

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -121,13 +121,24 @@ extension Statement: StatementModelConvertible {
...
+            .map { String(describing: $0) }
+
+            if associatedFieldValues.isEmpty {
+                associatedFieldValues = associatedValues(
+                    from: path + [field.foreignKeySqlName(withAssociationTargets: targetNames)],
+                    element: element
+                )
+            }
+            guard associatedFieldValues.count == field.associatedFieldNames.count else {
+                return nil
```
lawmicha, line 137: I took a stab at the problem we're solving by remove the required codegen changes from this PR, over in PR https://github.com/aws-amplify/amplify-swift/pull/2749/
- removed the public API changes that depended on codegen to generate the new `associatedFields` for hasOne and belongsTo. 
- Removed retrieving the associatedFieldNames at this line of code, basically just padded it with empty strings (see [code](https://github.com/aws-amplify/amplify-swift/pull/2749/files#r1098937871))

Amplify/Categories/DataStore/Model/Internal/Schema/ModelField+Association.swift
```diff
@@ -91,13 +91,13 @@ public enum ModelAssociation {
     case hasOne(associatedFieldName: String?, targetNames: [String])
     case belongsTo(associatedFieldName: String?, targetNames: [String])
 
-    public static let belongsTo: ModelAssociation = .belongsTo(associatedFieldName: nil, targetNames: [])
+    public static let belongsTo: ModelAssociation = .belongsTo()
```
lawmicha, line 94: let's revert this change as well since `case belongsTo(associatedFieldName: String?, targetNames: [String])` doesn't have default values for `associatedFieldName` and `targetNames`, we should keep the static method `.belongsTo` with the default values

AmplifyPlugins/Core/AWSPluginsCore/Model/Support/Model+GraphQL.swift
```diff
@@ -166,24 +166,35 @@ extension Model {
...
+            if field.isAssociationOwner && fieldNames.count > 1 && values.count == 1 {
+                let discreteKeys = values.compactMap { persistable in
+                    persistable.map({ String(describing: $0) })
+                        .map({ $0.split(separator: ModelIdentifierFormat.Custom.separator.first!) })
+                }.flatMap({ $0 })
+                .map({ $0.trimmingCharacters(in: CharacterSet(charactersIn: "\"")) })
+                values = discreteKeys.count == fieldNames.count ? discreteKeys : defaultValues
+            } else {
+                values = defaultValues
             }
```
lawmicha, line 187: is this logic needed if the `var values = getModelIdentifierValues(from: value, modelSchema: associateModelSchema)` come from the LazyReference's not loaded `[LazyReferenceIdentifier]` or loaded `model.identifier.values`? It should have already been processed to their individual values, so it's always going to be an array of the identifier values, not the concatenated version we store in SQL

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginLazyLoadTests/LL12/CompositePK/CompositePKChild+Schema.swift
```diff
@@ -66,4 +66,4 @@ extension ModelPath where ModelType == CompositePKChild {
   public var updatedAt: FieldPath<Temporal.DateTime>   {
       datetime("updatedAt") 
     }
-}
```
lawmicha, line 69: I think this PR is missing some changes after the rebase to `AWSDataStoreLazyLoadCompositePKTests`, such as changing 
```
assertLazyReference(queriedCompositePKChild._parent,
                            state: .notLoaded(identifiers: [.init(name: "@@primaryKey", value: savedParent.identifier)]))
```
to
```
assertLazyReference(queriedCompositePKChild._parent,
                            state: .notLoaded(identifiers: [.init(name: "CompositePKParent.keys.customId", value: savedParent.customId),
.init(name: "CompositePKParent.keys.content", value: savedParent.content]))
```

etc..

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginLazyLoadTests/LL1/Comment4V2+Schema.swift
```diff
@@ -32,7 +32,7 @@ extension Comment4V2 {
     model.fields(
       .field(comment4V2.id, is: .required, ofType: .string),
       .field(comment4V2.content, is: .required, ofType: .string),
-      .belongsTo(comment4V2.post, is: .optional, ofType: Post4V2.self, targetNames: ["postID"]),
+      .belongsTo(comment4V2.post, is: .optional, ofType: Post4V2.self, associatedWith: Post4V2.keys.id, targetNames: ["postID"]),
```
lawmicha, line 35: let's revert this change as well, since codegen is not currently generating this with `associatedWith`

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginLazyLoadTests/LL12/HasOneParentChild/AWSDataStoreLazyLoadHasOneTests.swift
```diff
@@ -76,6 +76,29 @@ class AWSDataStoreLazyLoadHasOneTests: AWSDataStoreLazyLoadBaseTest {
...
+        queriedParent.hasOneParentChildId = savedNewChild.id
+        try await updateAndWaitForSync(queriedParent)
+        print("New Child \(savedNewChild)")
+        print("Parent: \(savedParent)")
+
+        let queriedParentV2 = try await query(for: parent)
+        print("ParentV2: \(queriedParentV2)")
+        XCTAssertEqual(queriedParentV2.hasOneParentChildId, savedNewChild.id)
+    }
+
```
lawmicha, line 101: can we remove print and replace with asserts, like
```swift
func testUpdateParentWithNewChild() async throws {
        await setup(withModels: HasOneModels())
        let child = HasOneChild()
        let savedChild = try await saveAndWaitForSync(child)
        let parent = HasOneParent(child: savedChild, hasOneParentChildId: savedChild.id)
        let savedParent = try await saveAndWaitForSync(parent)

        var queriedParent = try await query(for: savedParent)
        XCTAssertEqual(queriedParent.hasOneParentChildId, savedChild.id)

        // The child can be lazy loaded.
        assertLazyReference(queriedParent._child,
                            state: .notLoaded(identifiers: [.init(name: "", value: child.id)]))

        let newChild = HasOneChild()
        let savedNewChild = try await saveAndWaitForSync(newChild)

        // Update parent to new child
        queriedParent.setChild(savedNewChild)
        queriedParent.hasOneParentChildId = savedNewChild.id
        try await updateAndWaitForSync(queriedParent)

        let queriedParentWithNewChild = try await query(for: parent)
        // The child can be lazy loaded.
        assertLazyReference(queriedParentWithNewChild._child, state: .notLoaded(identifiers: [.init(name: "", value: newChild.id)]))
        XCTAssertEqual(queriedParentWithNewChild.hasOneParentChildId, savedNewChild.id)
    }
```

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginLazyLoadTests/LL13/PhoneCall+Schema.swift
```diff
@@ -30,8 +30,8 @@ extension PhoneCall {
     
     model.fields(
       .field(phoneCall.id, is: .required, ofType: .string),
-      .belongsTo(phoneCall.caller, is: .required, ofType: Person.self, targetNames: ["callerId"]),
-      .belongsTo(phoneCall.callee, is: .required, ofType: Person.self, targetNames: ["calleeId"]),
+      .belongsTo(phoneCall.caller, is: .required, ofType: Person.self, associatedWith: Person.keys.id, targetNames: ["callerId"]),
+      .belongsTo(phoneCall.callee, is: .required, ofType: Person.self, associatedWith: Person.keys.id, targetNames: ["calleeId"]),
```
lawmicha, line 34: revert this since codegen is not doing to generate this, and in other places as well

AmplifyTestCommon/Models/TransformerV2/4aV2/Project4aV2+Schema.swift
```diff
@@ -31,7 +31,7 @@ extension Project4aV2 {
     model.fields(
       .id(),
       .field(project4aV2.name, is: .optional, ofType: .string),
-      .hasOne(project4aV2.team, is: .optional, ofType: Team4aV2.self, associatedWith: Team4aV2.keys.project, targetName: "project4aV2TeamId"),
+      .hasOne(project4aV2.team, is: .optional, ofType: Team4aV2.self, associatedWith: Team4aV2.keys.id, targetName: "project4aV2TeamId"),
```
lawmicha, line 34: AmplifyTestCommon files are used by unit tests. is this change expected? If so, would prefer to make this change outside this PR with the details on the issue

lawmicha (approved):

PR #2746, opened 2023-02-06
fix: DefaultModelProvider return nil from not loaded state

lawmicha:
## Description

When there is no metadata to create a ModelProvider, the lazy reference will be instantiated with `nil` identifiers with a DefaultModelProvider. When this happens, `try await` on the computed property should simply return `nil` on `get()` and throw on `require()`. 

## General Checklist

- [ ] All unit tests pass

### DIFF
--- a/Amplify/Categories/DataStore/Model/Lazy/DefaultModelProvider.swift
+++ b/Amplify/Categories/DataStore/Model/Lazy/DefaultModelProvider.swift
@@ -24,7 +24,7 @@ public struct DefaultModelProvider<Element: Model>: ModelProvider {
     public func load() async throws -> Element? {
         switch loadedState {
         case .notLoaded:
-            throw CoreError.clientValidation("DefaultModelProvider does not provide loading capabilities", "")
+            return nil
         case .loaded(let model):
             return model
         }
--- a/AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginLazyLoadTests/LL12/CompositePK/GraphQLLazyLoadCompositePKChildTests.swift
+++ b/AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginLazyLoadTests/LL12/CompositePK/GraphQLLazyLoadCompositePKChildTests.swift
@@ -1,0 +1,124 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import Combine
+import XCTest
+
+@testable import Amplify
+import AWSPluginsCore
+
+extension GraphQLLazyLoadCompositePKTests {
+
+    // MARK: - CompositePKParent / CompositePKChild
+    
+    func initChild(with parent: CompositePKParent? = nil) -> CompositePKChild {
+        CompositePKChild(childId: UUID().uuidString, content: "content", parent: parent)
+    }
+    
+    func testSaveCompositePKChild() async throws {
+        await setup(withModels: CompositePKModels())
+        
+        let parent = initParent()
+        let savedParent = try await mutate(.create(parent))
+        let child = initChild(with: savedParent)
+        try await mutate(.create(child))
+    }
+    
+    func testUpdateCompositePKChild() async throws {
+        await setup(withModels: CompositePKModels())
+        let parent = initParent()
+        let savedParent = try await mutate(.create(parent))
+        let child = initChild(with: parent)
+        var savedChild = try await mutate(.create(child))
+        let loadedParent = try await savedChild.parent
+        XCTAssertEqual(loadedParent?.identifier, savedParent.identifier)
+        
+        // update the child to a new parent
+        let newParent = initParent()
+        let savedNewParent = try await mutate(.create(newParent))
+        savedChild.setParent(savedNewParent)
+        let updatedChild = try await mutate(.update(savedChild))
+        let loadedNewParent = try await updatedChild.parent
+        XCTAssertEqual(loadedNewParent?.identifier, savedNewParent.identifier)
+    }
+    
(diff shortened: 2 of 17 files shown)


Amplify/Categories/DataStore/Model/Lazy/DefaultModelProvider.swift
```diff
@@ -24,7 +24,7 @@ public struct DefaultModelProvider<Element: Model>: ModelProvider {
     public func load() async throws -> Element? {
         switch loadedState {
         case .notLoaded:
-            throw CoreError.clientValidation("DefaultModelProvider does not provide loading capabilities", "")
+            return nil
```
lawmicha, line 27: This is the main functional change. If the default model provider is called from a not loaded state, it should simply return `nil`. The caller (LazyReference) is responsible for returning `nil` on `get()` or throwing if `require()` is called.

AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginLazyLoadTests/LL12/CompositePK/GraphQLLazyLoadCompositePKChildTests.swift
```diff
@@ -0,0 +1,124 @@
...
+    
+    func testUpdateFromNoParentCompositePKChild() async throws {
+        await setup(withModels: CompositePKModels())
+        let parent = initParent()
+        let savedParent = try await mutate(.create(parent))
+        
+        let childWithoutParent = initChild()
+        var savedChild = try await mutate(.create(childWithoutParent))
+        let nilParent = try await savedChild.parent
+        XCTAssertNil(nilParent)
```
lawmicha, line 58: This test was what triggered the implementation change in the PR, the child without a parent should be query-able and can be traversed to the parent, even if the parent is nil. Previously, this call was throwing due to the default model provider throwing. Since the parent is an optional association, LazyReference gets `nil` back from the default model provider, and returns `nil` on `get()`.

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginLazyLoadTests/LL12/CompositePK/CompositePKParent+Schema.swift
```diff
@@ -34,7 +34,7 @@ extension CompositePKParent {
       .hasMany(compositePKParent.children, is: .optional, ofType: CompositePKChild.self, associatedWith: CompositePKChild.keys.parent),
       .hasMany(compositePKParent.implicitChildren, is: .optional, ofType: ImplicitChild.self, associatedWith: ImplicitChild.keys.parent),
       .hasMany(compositePKParent.strangeChildren, is: .optional, ofType: StrangeExplicitChild.self, associatedWith: StrangeExplicitChild.keys.parent),
-      .hasMany(compositePKParent.childrenSansBelongsTo, is: .optional, ofType: ChildSansBelongsTo.self, associatedWith: ChildSansBelongsTo.keys.compositePKParentChildrenSansBelongsToCustomId),
+      .hasMany(compositePKParent.childrenSansBelongsTo, is: .optional, ofType: ChildSansBelongsTo.self, associatedFields: [ChildSansBelongsTo.keys.compositePKParentChildrenSansBelongsToCustomId, ChildSansBelongsTo.keys.compositePKParentChildrenSansBelongsToContent]),
```
lawmicha, line 37: this use case is the uni-directional has-many changes. `childrenSansBelongsTo ` doesn't have a belongs to reference to the parent, and the parent (this schema) has a has-many field to the `childrenSansBelongsTo` child. With the CPK feature, it should have `associatedFields` with the CPK of the parent foreign key fields.

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginLazyLoadTests/LL12/CompositePK/CompositePKChild.swift
```diff
@@ -34,7 +34,7 @@ public struct CompositePKChild: Model {
       self.createdAt = createdAt
       self.updatedAt = updatedAt
   }
-  public mutating func setParent(parent: CompositePKParent? = nil) {
+  public mutating func setParent(_ parent: CompositePKParent? = nil) {
```
lawmicha, line 37: This is what should be codegenerated (see https://github.com/aws-amplify/amplify-codegen/pull/504/files#r1098000791) and is updated here manually.

AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginLazyLoadTests/LL12/CompositePK/AWSDataStoreLazyLoadCompositePKChildTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        let queriedParent = try await query(for: savedParent)
+        assertList(queriedParent.children!, state: .isNotLoaded(associatedIds: [queriedParent.identifier],
+                                                                associatedFields: ["parent"]))
+        try await queriedParent.children?.fetch()
+        assertList(queriedParent.children!, state: .isLoaded(count: 1))
+        
+        // query child and load the parent - CompositePKChild
+        let queriedCompositePKChild = try await query(for: savedCompositePKChild)
+        assertLazyReference(queriedCompositePKChild._parent,
+                            state: .notLoaded(identifiers: [.init(name: "@@primaryKey", value: savedParent.identifier)]))
```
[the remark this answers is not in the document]
lawmicha: Sounds good, i'll rebase 2737 with `data-dev-preview` and re-run/fix the integration tests

PR #2749, opened 2023-02-07
fix(datastore): lazy load on save with sync on identifiers

lawmicha:
## Description

This PR is a cherry-pick of https://github.com/aws-amplify/amplify-swift/pull/2737 to minimize the necessary changes needed. We remove the updated codegen changes and instead of storing "associatedFields" / "associatedkey" for lazy references (has-one  and belongs-to use cases), we're just storing the empty string "". Storing LazyReferenceIdentifier is an implementation detail, which is exposed via the `_state` property for internal plugins logic to operate against, such as Model+GraphQL.

Datastore.save has been updated to perform query with lazy load to return a model with associated models that are not yet loaded. The same model returned to the caller is synced to AppSync. Previously this model was eager loaded. Model+GraphQL translates the model to the GraphQL Input by extracting the identifiers out of associated models. Since the associated model is not yet loaded, the identifers are extracted out of the lazy reference `.notLoaded(identifiers)`.

DataStoreModelProvider, responsible for lazy loading a single model, used to hold the a single identifier value, which may be the value or a value concatenated with "#" for CPK. On`load()`, it will use the schema to find the primary key or use the  `@@primaryKey` column name.

Model+GraphQL will use the targetNames as for the GraphQL input keys and the identifiers array from  `.notLoaded(identifers)` of the LazyReference. This was previously not possible since the identifiers was already the concatenated version of the primary key.

## General Checklist

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Core/DataStoreModelDecoder.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Core/DataStoreModelDecoder.swift
@@ -15,18 +15,33 @@ public struct DataStoreModelDecoder: ModelProviderDecoder {
     
     /// Metadata that contains the foreign key value of a parent model, which is the primary key of the model to be loaded.
     struct Metadata: Codable {
-        let identifier: String?
+        let identifiers: [LazyReferenceIdentifier]
         let source: String
         
-        init(identifier: String?, source: String = DataStoreSource) {
-            self.identifier = identifier
+        init(identifiers: [LazyReferenceIdentifier], source: String = DataStoreSource) {
+            self.identifiers = identifiers
             self.source = source
+        }
+        
+        func toJsonObject() -> Any? {
+            return [
+                "identifiers": identifiers.map({
+                    [
+                        "name": $0.name,
+                        "value": $0.value
+                    ]
+                }),
+                "source": source
+            ]
         }
     }
     
     /// Create a SQLite payload that is capable of initializting a LazyReference, by decoding to `DataStoreModelDecoder.Metadata`.
-    static func lazyInit(identifier: Binding?) -> [String: Binding?] {
-        return ["identifier": identifier, "source": DataStoreSource]
+    static func lazyInit(identifiers: [LazyReferenceIdentifier]) -> Metadata? {
+        if identifiers.isEmpty {
+            return nil
+        }
+        return Metadata(identifiers: identifiers)
     }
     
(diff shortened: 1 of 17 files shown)


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/Statement+Model.swift
```diff
@@ -121,14 +121,23 @@ extension Statement: StatementModelConvertible {
...
+                    from: path + [field.foreignKeySqlName(withAssociationTargets: targetNames)],
+                    element: element
+                )
             }
-            return DataStoreModelDecoder.lazyInit(identifier: targetBinding)
-
+            let emptyFieldNames = [String](repeating: "", count: associatedFieldValues.count)
+            return DataStoreModelDecoder.lazyInit(identifiers: zip(emptyFieldNames,
+                                                                   associatedFieldValues).map {
+                LazyReferenceIdentifier(name: $0.0, value: $0.1)
```
lawmicha, line 139: padding with empty strings

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Core/DataStoreModelDecoder.swift
```diff
@@ -15,18 +15,33 @@ public struct DataStoreModelDecoder: ModelProviderDecoder {
     
     /// Metadata that contains the foreign key value of a parent model, which is the primary key of the model to be loaded.
     struct Metadata: Codable {
-        let identifier: String?
+        let identifiers: [LazyReferenceIdentifier]
```
lawmicha, line 18: maybe this can just be an array of strings, `[String]`, then we don't have to pad it from Statement+Model. When DataStoreModelProvider.init(metadata) is called, it can create the LazyReferenceIdentifier's with the following outputs:

```
// regular
LazyReferenceIdentifier(name: "id", value:"id123")

// CPK
[LazyReferenceIdentifier(name: "@@primaryKey", value:"id123"),
LazyReferenceIdentifier(name: "", value:"title123")] 
```

Then we can change the `load()` slightly to always use the first name ("id" or "@@primaryKey") and concat the values.

lawmicha:
This PR is no longer needed and was an attempt at investigating the PR https://github.com/aws-amplify/amplify-swift/pull/2737 We have incorporated the changes in https://github.com/aws-amplify/amplify-swift/pull/2737 instead

PR #2757, opened 2023-02-08
fix(datastore): full sync when sync predicate changes

lawmicha:
## Issue \#

https://github.com/aws-amplify/amplify-swift/issues/3383

## Description

Perform a full sync when the sync expression for the model changes. To do this, we have to persist and retrieve it from the sync metadata table for comparison with the incoming sync predicate.

The sync expression can be made up of mulitple predicates, each correponding to a specific model. InitialSyncOperation extracts out the first sync predicate from the sync expression for the corresponding model. This `QueryPredicate` is then used to compare with the existing persisted sync predicate from the last sync run, stored in the ModelSyncMetadata table.

We extend the ModelSyncMetadata table to add a new column, make QueryPredicate Encodable so we can encode the sync predicate and store it as a String. The encoder has the output formatting set to `.sortedKeys` to make sure the same string is generated every time. We do a string comparision of the incoming sync predicate with the persisted one to see if it was changed.

We do a full sync on these scenarios:
- no previous sync predicate and now there is a sync predicate
- there is a previous sync predicate and no current sync predicate
- both exist but they do not match

When there are matching sync predicates or no previous and current sync predicates, then the sync type falls back to the previous logic to determine full or delta (based on the sync interval).

### Existing app upgrade scenario

Extending the ModelSyncMetadata table means that existing apps that have already created the table will not recreate it for the new column. The most optimal way, without deleting the local data, is to add a new column if we can detect the column doesn't exist. This is in the ModelSyncMetadataMigration logic.

### Verification

```
Detected ModelSyncMetadata table exists without syncPredicate column.
ALTER TABLE "ModelSyncMetadata" ADD COLUMN "syncPredicate" "text";
ModelSyncMetadata table altered to add syncPredicate column.
```

## General Checklist

- [x] All unit tests pass
- [ ] All integration tests pass https://github.com/aws-amplify/amplify-swift/actions/runs/7062531087
- [ ] ~~If breaking change, documentation/changelog update with migration instructions~~

### DIFF
--- a/Amplify/Categories/DataStore/Model/Internal/Persistable.swift
+++ b/Amplify/Categories/DataStore/Model/Internal/Persistable.swift
@@ -20,7 +20,7 @@ import Foundation
 /// - `Temporal.Time`
 /// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly
 ///   by host applications. The behavior of this may change without warning.
-public protocol Persistable {}
+public protocol Persistable: Encodable {}
 
 extension Bool: Persistable {}
 extension Double: Persistable {}
--- a/Amplify/Categories/DataStore/Query/QueryOperator.swift
+++ b/Amplify/Categories/DataStore/Query/QueryOperator.swift
@@ -7,7 +7,7 @@
 
 import Foundation
 
-public enum QueryOperator {
+public enum QueryOperator: Encodable {
     case notEqual(_ value: Persistable?)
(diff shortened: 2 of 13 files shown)


MahmoudAlsamman:
@lawmicha  What is the status of this fix? is it going to be merged soon?

5d (approved):

Issue #2508, opened 2022-10-27
Enable Keychain Sharing for App Extensions


bdhazman:
### Is your feature request related to a problem? Please describe.

Amplify manages user authentication and credentials but while Amplify functionality (including authentication) can be used in App Extensions such as Share Extensions, authentication sessions are not shared between apps and their app extensions. This is a problem for developers who want to use the Amplify library for managing iOS authentication to their services but also build software that extends the functionality of their application with app extensions. (examples of feature request: [1](https://stackoverflow.com/questions/70033427/how-can-i-share-an-ios-amplify-projects-login-credentials-between-main-app-and), [2](https://github.com/aws-amplify/aws-sdk-ios/issues/3013))

### Describe the solution you'd like

An ideal solution is allowing for cross-app authentication by taking advantage of Apple's ['shared keychain' through keychain access groups](https://developer.apple.com/documentation/security/keychain_services/keychain_items/sharing_access_to_keychain_items_among_a_collection_of_apps) in a manner similar to how [Firebase does](https://firebase.google.com/docs/auth/ios/single-sign-on). Allowing developers to choose the keychain group their application uses would allow them to support cross-app authentication including App Extensions.

Amplify already [relies on iOS's keychain](https://docs.amplify.aws/sdk/auth/how-it-works/q/platform/ios/#offline-support) so implementing this change would be a matter of making changes to how Amplify accesses and writes credentials to the system's keychain. 

Making this change would give developers the opportunity to build suites of apps that require only one log in and extend functionality of their apps through deeper iOS integration with extensions such as Share Extensions and widgets.

### Describe alternatives you've considered

Alternatives to this feature can result in poor user experiences or less secure workarounds.

- Developers can require users to re-authenticate in app extensions or apps that could share authentication sessions but this can create more friction for users resulting in a poor user experience. 
- A workaround developers could take is to insecurely store auth tokens or credentials and pass them between apps (example: [1](https://github.com/aws-amplify/aws-sdk-ios/pull/1155#issuecomment-618123577), [2](https://github.com/aws-amplify/aws-sdk-ios/issues/1106#issuecomment-684413547)) and app extensions or app groups but this isn't as secure as relying solely on the keychain in all apps/extensions.

I am open to hearing any other suggested alternatives for accessing existing authentication sessions in extensions of apps.

### Is the feature request related to any of the existing Amplify categories?

Auth

### Additional context

[...]

PR #2770, opened 2023-02-17
Support Keychain Access Groups

ekurutepe:
## Issue \#
#2508 

## Description
This change enables secure sharing of Cognito session between the main app and app extensions

## General Checklist
- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
@@ -32,7 +32,10 @@ extension AWSCognitoAuthPlugin {
         let authConfiguration = try ConfigurationHelper.authConfiguration(jsonValueConfiguration)
 
         let credentialStoreResolver = CredentialStoreState.Resolver().eraseToAnyResolver()
-        let credentialEnvironment = credentialStoreEnvironment(authConfiguration: authConfiguration)
+        let credentialEnvironment = credentialStoreEnvironment(
+            authConfiguration: authConfiguration,
+            accessGroup: jsonValueConfiguration["KeychainAccessGroup"]?.stringValue
+        )
         let credentialStoreMachine = StateMachine(resolver: credentialStoreResolver,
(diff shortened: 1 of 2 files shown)


harsh62:
Thank you for creating PR to support passing Keychain Groups. Since the configuration file is generated by using @aws-amplify/amplify-cli , adding `KeychainAccessGroup` would first need to be implemented with CLI.
Manually editing the configuration file is discouraged as could be overwritten during a change done through the CLI. 
So we have two options here: 
* Propagate the right changes with @aws-amplify/amplify-cli as well to support this feature.
* Allow customers to dynamically pass the keychain access group to the Auth Plugin.

harsh62:
Digging deeper into the PR and the existing implementation, I found a scenario where we also need to consider how existing items that are there in the keychain can be updated to use access group that was passed in.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
```diff
@@ -217,12 +220,16 @@ extension AWSCognitoAuthPlugin {
...
             authConfiguration: authConfiguration,
             credentialStoreEnvironment: BasicCredentialStoreEnvironment(
-                amplifyCredentialStoreFactory: makeCredentialStore,
-                legacyKeychainStoreFactory: makeLegacyKeychainStore(service:)
+                amplifyCredentialStoreFactory: {
+                    self.makeCredentialStore(accessGroup: accessGroup)
+                },
+                legacyKeychainStoreFactory:  { service in
+                    self.makeLegacyKeychainStore(service: service, accessGroup: accessGroup)
+                }
```
harsh62, line 232: We should not be adding `accessGroup`to the legacy keychain. This would cause migration to fail from V1 to V2.

ekurutepe:
@harsh62 thanks for looking into this PR. We have been using this branch in production for a few weeks now. I understand that the bar to be included in a popular public SDK is very high and this probably needs to address various different use cases that I am not familiar with and did not consider. Please feel free to take over and take it to the finish line. I'd kindly suggest to prioritize this internally since this feature is required to be able to safely share cognito credentials between the main app and app extensions.

Issue #2059, opened 2022-07-29
Deleting a parent in datastore doesn't delete the children


depthzadmin:
### Describe the bug

In iOS, when you delete a parent object with a one to many relationship, the children are not removed from the DataStore.

Note that the expected behavior of deleting the children does work outside the iOS context. As an example, when I perform the same delete action in the Amplify Studio Content tab, the children are removed.

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Setup a schema with a 1:M relationship

extension User {
  // MARK: - CodingKeys 
   public enum CodingKeys: String, ModelKey {
    case id
    case username
    case email
    case Tests
    case createdAt
    case updatedAt
  }
  
  public static let keys = CodingKeys.self
  //  MARK: - ModelSchema 
  
  public static let schema = defineSchema { model in
    let user = User.keys
    
    model.authRules = [
      rule(allow: .private, operations: [.create, .update, .delete, .read])
    ]
    
    model.pluralName = "Users"
    
    model.fields(
      .id(),
      .field(user.username, is: .required, ofType: .string),
      .field(user.email, is: .optional, ofType: .string),
      .hasMany(user.Tests, is: .optional, ofType: Test.self, associatedWith: Test.keys.userID),
      .field(user.createdAt, is: .optional, isReadOnly: true, ofType: .dateTime),
      .field(user.updatedAt, is: .optional, isReadOnly: true, ofType: .dateTime)
    )
    }
}

extension Test {
  // MARK: - CodingKeys 
   public enum CodingKeys: String, ModelKey {
    case id
    case test
    case userID
    case createdAt
    case updatedAt
  }
  
  public static let keys = CodingKeys.self
  //  MARK: - ModelSchema 
  
  public static let schema = defineSchema { model in
    let test = Test.keys
    
    model.authRules = [
      rule(allow: .private, operations: [.create, .update, .delete, .read])
    ]
    
    model.pluralName = "Tests"
    
    model.attributes(
      .index(fields: ["userID"], name: "byUser")
    )
    
    model.fields(
      .id(),
      .field(test.test, is: .optional, ofType: .string),
      .field(test.userID, is: .required, ofType: .string),
      .field(test.createdAt, is: .optional, isReadOnly: true, ofType: .dateTime),
      .field(test.updatedAt, is: .optional, isReadOnly: true, ofType: .dateTime)
    )
    }
}


2. In iOS, use the method:

[...]

PR #2781, opened 2023-02-24
fix(datastore): mutation event got ignored while executing parallel saving

5d:
## Issue \#

https://github.com/aws-amplify/amplify-swift/issues/2059

## Related PR

v1 fix - https://github.com/aws-amplify/amplify-swift/pull/2782

## Description

In the `MutationEvent` [saving](https://github.com/aws-amplify/amplify-swift/blob/ac2b6a0/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/AWSMutationDatabaseAdapter/AWSMutationDatabaseAdapter%2BMutationEventIngester.swift#L218-L242) implementation, we check `nextEventPromise` at the beginning to detect wether put the `MutationEvent` to `inProgress` status. After it's written to database, we change the `nextEventPromise` to `nil`. The operation between `non-nil` check at the beginning and set value `nil` for `nextEventPromise` doesn't lock the access of `nextEventPromise`. It may took over hundred milliseconds for writing to the database. In this duration, if there is any `MutationEvent` get saved, it will have `inProcess = true`. But only one of them will be handled back to the `MutationEventPublisher`.

This change simply `getAndSet` the `nextEventPromise` to `nil` to ensure only one thread has the access to stored `nextEventPromise` in Atomic container. Thus other parallel calls will have the MutationEvent with `inProgress = false`.

## General Checklist

- [X] All unit tests pass

### DIFF
--- a/Amplify/Core/Support/Optional+Extension.swift
+++ b/Amplify/Core/Support/Optional+Extension.swift
@@ -1,0 +1,26 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+		
+
+import Foundation
+
+extension Optional {
+    ///
+    /// Performing side effect function when data is exist
+    /// - parameters:
+    ///     - f: a function may perform side effects on wrapped data
+    /// - returns:
+    /// The original Optional object without touched.
+    @discardableResult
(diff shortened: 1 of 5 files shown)


royjit (dismissed):
LGTM

royjit (approved):

PR #2788, opened 2023-03-02
test(datastore): fix flaky integration test cases

5d:
## Description

For integration test cases of `DataStoreConsecutiveUpdatesTests`, there is chance that some MutationEvents are not processed or put into inProcess state when we've received previous operation response from server side. According to the [current reconciliation algorithm](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/AWSMutationDatabaseAdapter/AWSMutationDatabaseAdapter%2BMutationEventIngester.swift#L198-L214), these incoming event will be ignored and will not be published as hub event. 
Therefore, this fix patch is updating the assert logic focus on confirming the data is eventually consistent.

## General Checklist

- [X] All unit tests pass


royjit (approved):
LGTM

Issue #1049, opened 2021-02-01
QueryField should support notContains operator


richardmcclellan:
[This documentation](https://docs.amplify.aws/lib/datastore/data-access/q/platform/ios#predicates) indicates that DataStore supports the `notContains` operator, but it [does not appear to be implemented](https://github.com/aws-amplify/amplify-ios/blob/e2f7f64e2a501f3b2807481ca921e71710c82b22/Amplify/Categories/DataStore/Query/QueryField.swift#L34-L44) in the library.   

On amplify-js, It appears to be [supported](https://github.com/aws-amplify/amplify-js/blob/6e28eaf37525ce231d7793bf82a960046fc7f8f4/packages/datastore/src/types.ts#L295-L299).

On Android, it is not supported, currently.  A customer reported this missing functionality, [here](https://github.com/aws-amplify/amplify-android/issues/1142).  Support is being added for Android in https://github.com/aws-amplify/amplify-android/pull/1145/

lawmicha:
Hi @richardmcclellan, thanks for opening the issue

What does the SQL look like? I couldn't decipher exactly from the Android PR. 
Is it "NOT LIKE"? ` select * from table_name where table_column_name NOT LIKE '%1%'`
[SO ref](https://stackoverflow.com/questions/19896831/sqlite-select-where-column-does-not-contain-certain-data)

What does the AppSync operator look like? this one looks simply as `notContains`

richardmcclellan:
> What does the SQL look like?

The SQL for `contains` is: `select * from table_name where instr(table_column_name,?) > 0`

I believe `LIKE` could be used instead of `instr`?   The performance of the two [appears to be about the same](https://stackoverflow.com/questions/2451608/which-is-faster-instr-or-like).   But, given that we were already using `instr` for contains, I chose to add support for `notContains` with a similar SQL command:

`select * from table_name where instr(table_column_name,?) = 0`

> What does the AppSync operator look like? this one looks simply as `notContains`

Correct, it is `notContains`.

atierian:
Working implementation for strings, but still early exploratory phase.
https://github.com/aws-amplify/amplify-swift/pull/2789

PR #2789, opened 2023-03-04
feat(datastore): add notContains query operator

atierian:
## Issue \#

https://github.com/aws-amplify/amplify-swift/issues/1049

## Description

Related Android PR: https://github.com/aws-amplify/amplify-android/pull/1145

This change adds support for the `notContains` predicate for DataStore `query` and `delete` operations. This is a documented feature [[docs/swift/datastore/data-access/predicates](https://docs.amplify.aws/lib/datastore/data-access/q/platform/ios/#predicates)], however it is currently not supported in Amplify Swift. This addition reaches feature parity with regards to `notContains` with Android.

Functionally, it exposes a new static method `notContains(_ value: String) -> QueryPredicateOperation`, which is mapped to a `NOT LIKE '%value%'` operator.

Add tests cover multiple scenarios included `query` and `delete`, along with test cases for chained predicates (`&&` and `||`) for each top level API.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/Amplify/Categories/DataStore/Query/ModelKey.swift
+++ b/Amplify/Categories/DataStore/Query/ModelKey.swift
@@ -56,6 +56,11 @@ extension CodingKey where Self: ModelKey {
         return key.contains(value)
     }
 
+    // MARK: - not contains
+    public func notContains(_ value: String) -> QueryPredicateOperation {
+        return field(stringValue).notContains(value)
+    }
+
     // MARK: - eq
 
     public func eq(_ value: Persistable?) -> QueryPredicateOperation {
--- a/Amplify/Categories/DataStore/Query/QueryOperator.swift
+++ b/Amplify/Categories/DataStore/Query/QueryOperator.swift
@@ -15,6 +15,7 @@ public enum QueryOperator {
     case greaterOrEqual(_ value: Persistable)
     case greaterThan(_ value: Persistable)
     case contains(_ value: String)
+    case notContains(_ value: String)
     case between(start: Persistable, end: Persistable)
     case beginsWith(_ value: String)
 
@@ -37,6 +38,10 @@ public enum QueryOperator {
                 return targetString.contains(predicateString)
             }
             return false
+        case .notContains(let predicateString):
+            if let targetString = target as? String {
+                return !targetString.contains(predicateString)
+            }
         case .between(let start, let end):
             return PersistableHelper.isBetween(start, end, target)
         case .beginsWith(let predicateValue):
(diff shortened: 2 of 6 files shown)


Amplify/Categories/DataStore/Query/ModelKey.swift
```diff
@@ -56,6 +56,11 @@ extension CodingKey where Self: ModelKey {
         return key.contains(value)
     }
 
+    // MARK: - not contains
+    public func notContains(_ value: String) -> QueryPredicateOperation {
```
jcjimenez, line 60: This represents a substring check, right? In other words, it's not meant to also be used for an equivalent to SQL `WHERE NOT IN (<some list or set>)`?

Amplify/Categories/DataStore/Query/QueryOperator.swift
```diff
@@ -37,6 +38,10 @@ public enum QueryOperator {
                 return targetString.contains(predicateString)
             }
             return false
+        case .notContains(let predicateString):
+            if let targetString = target as? String {
+                return !targetString.contains(predicateString)
+            }
```
jcjimenez, line 44: Should we return `false` like for `.contains` above when the target is not a string?

jcjimenez (approved):

Amplify/Categories/DataStore/Query/QueryOperator.swift
atierian: That would happen as is due to the default [`return false` underneath the switch statement](https://github.com/aws-amplify/amplify-swift/blob/68b61be6487a40669eb943f208f8874850c9a886/Amplify/Categories/DataStore/Query/QueryOperator.swift#L52). I don't think an explicit `return false` in this `case` block is necessary, but happy to add it if you think it provides value.

Amplify/Categories/DataStore/Query/ModelKey.swift
atierian: That's correct. It only support `String` checking by doing a `NOT LIKE %value%` ([not like](https://github.com/aws-amplify/amplify-swift/pull/2789/files/68b61be6487a40669eb943f208f8874850c9a886#diff-c2beb770af448b950a26b1bde597b3840e5bcdb6c7b3b263bdff2ea18717bfc2R33)) ([%value%](https://github.com/aws-amplify/amplify-swift/pull/2789/files/68b61be6487a40669eb943f208f8874850c9a886#diff-c2beb770af448b950a26b1bde597b3840e5bcdb6c7b3b263bdff2ea18717bfc2R53)).

Amplify/Categories/DataStore/Query/QueryOperator.swift
jcjimenez: Oh, I missed that, thanks for clarifying!

PR #2852, opened 2023-04-05
chore: Adding mechanism to provide plugin extensions

ruisebas:
## Description
This PR adds:
- A `@_spi` annotated `jsonConfiguration` property to `AWSCognitoAuthPlugin`, that we can use to retrieve the final configuration that's been used.
- An `@_spi` annotated protocol called `AWSPluginExtension ` that allows us to provide additional functionality to a plugin
- An `@_spi` annotated `UserAgentSuffixAppender` class that conforms to `AWSPluginExtension` and `HttpClientEngine`, which can "intercept" requests and add the provided suffix to the User-Agent header.
- An `@_spi` annotated `addPluginExtension` method to `AWSCognitoAuthPlugin`
- An **internal** `CustomHttpEngine` protocol.  This is done in order to prevent anybody but us from being able to submit custom `HttpClientEngine`s to the plugin. `UserAgentSuffixAppender` also conforms to this.

## General Checklist

- [X] All unit tests pass


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
```diff
@@ -89,6 +91,15 @@ extension AWSCognitoAuthPlugin {
                 frameworkMetadata: AmplifyAWSServiceConfiguration.frameworkMetaData(),
                 region: userPoolConfig.region
             )
+
+            if let provider = self as? UserAgentSuffixProvider, !provider.userAgentSuffix.isEmpty {
```
jcjimenez, line 95: So how does the receiver (`self`) sometimes conform and not to the `UserAgentSuffixProvider` protocol? by subclassing then extending `UserAgentSuffixProvider`? (apologies if I'm missing something here).

ruisebas: Oh, you just make the plugin conform to `UserAgentSuffixProvider` :) 

```swift
extension AWSCognitoAuthPlugin: UserAgentSuffixProvider {
    var userAgentSuffix: String {
        "yourCustomSuffix"
    }
}
```

jcjimenez: Sorry, I was referring to two cases with the same plugin:

1. Used from the context where you **do** want to implement `UserAgentSuffixProvider` (like what you have now)
2. Used from a context where you **do not** want the user agent suffix.

I assumed we wanted to support both cases from the same plugin, but it sounds like we don't. If that's the case, this looks good.

jcjimenez: I was thinking about this a bit more, and maybe another approach to consider would be:

Making `AWSCognitoAuthPlugin` itself conform to `HttpClientEngine`.

Pros
* No need for an additional semi-public SPI
* Gives us the flexibility to decide at **run-time** to append a suffix or not.

Cons
* Requires `AWSCognitoAuthPlugin` to hold to yet one more reference (in this case, the `HttpClientEngine` which it proxies).
* ?

ruisebas: >Used from the context where you **do** want to implement UserAgentSuffixProvider (like what you have now)
>Used from a context where you **do not** want the user agent suffix.

Yes, this approach is either all or nothing, which is OK for what we need right now. The protocol conformance does allow us to avoid having the Proxy set up for **all** requests though. So for a normal use case (i.e. the plugin not conforming to the protocol), nothing would change.

----

>No need for an additional semi-public SPI

We would still need an new external `@_spi` API to let the plugin know what suffix to include in the User-Agent, right?
Or am I missing something?

> Gives us the flexibility to decide at run-time to append a suffix or not.
🤔  Isn't this what we are doing already? I.e. at runtime we check if the plugin conforms to the protocol (either `HttpClientEngine` or `UserAgentSuffixProvider ` and if so we create the proxy.

If you meant making the plugin **always** conform to `HttpClientEngine`, then we would end up creating the proxy for all users (even those who don't need extra User-Agent stuff) and we wanted to avoid that.

jcjimenez: From what you're saying, and what is missing from this PR, I think I understand your proposal (please correct me if I'm wrong): Some other library or app (not amplify-swift) will contribute this extension in their own library (hoping somebody else isn't doing it also):

```
extension AWSCognitoAuthPlugin: UserAgentSuffixProvider {
...
```

By doing so, that other target decides what suffix to provide. Is this correct?

> We would still need an new external @_spi API to let the plugin know what suffix to include in the User-Agent, right? Or am I missing something?

No, we could do away with `UserAgentSuffixProvider` altogether using the same approach you have here with `HttpClientEngine`.

> 🤔 Isn't this what we are doing already? I.e. at runtime we check if the plugin conforms to the protocol (either HttpClientEngine or UserAgentSuffixProvider  and if so we create the proxy.

Assuming that what you are proposing is what I asked above, the runtime check aspect of it, yes. However, this still leads us with a build-time "hard-coded" implementation of `UserAgentSuffixProvider` (weather done from the `AWSCognitoAuthPlugin` target or elsewhere). What I'm trying to see is if it's worth doing a little extra work so that the decision on wether to contribute a suffix can be done dynamically.

<hr>

For the record, I'm not a big fan of using category extensions to append logic (possibly non-deterministically) to a logical unit (e.g. AWSCognitoAuthPlugin) from outside its own target, e.g. leading to `Redundant conformance of protocol` errors where it may not be clear where the original one is or not being able to do anything about it if someone beat you to it.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+PluginExtension.swift
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+@_spi(InternalAmplifyPluginExtension) import AWSPluginsCore
+import Foundation
+import ClientRuntime
+
+extension AWSCognitoAuthPlugin {
+    @_spi(InternalAmplifyPluginExtension)
+    public func addPluginExtension(_ pluginExtension: AWSPluginExtension) {
```
jcjimenez, line 14: Excellent! Selector suggestion to consider: `add(pluginExtension:)`

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Support/Utils/CustomHttpEngine.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+@_spi(InternalCustomHttpEngine) @_spi(InternalAmplifyPluginExtension) import AWSPluginsCore
+import ClientRuntime
+import Foundation
+
+protocol CustomHttpEngine: HttpClientEngine {
+    func setHttpClientEngine(_ httpClientEngine: HttpClientEngine)
```
jcjimenez, line 13: This setter makes me think that this make the notion of a proxy a bit more solid. What do you think about this?

```
protocol HttpClientEngineProxy: HttpClientEngine {
    var target: HttpClientEngine?
}
```

jcjimenez (approved):

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin.swift
```diff
@@ -29,6 +30,11 @@ public final class AWSCognitoAuthPlugin: AWSCognitoAuthPluginBehavior {
 
     var taskQueue: TaskQueue<Any>!
 
+    var customHttpEngine: CustomHttpEngine?
+
+    @_spi(InternalAmplifyConfiguration)
+    internal(set) public var jsonConfiguration: JSONValue?
```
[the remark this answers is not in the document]
ruisebas: It is updated from the same target, but the idea is to be able to grab it from the Authenticator, which is an external target.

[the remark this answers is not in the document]
jcjimenez: Oh, I see.

jcjimenez (approved):

PR #2863, opened 2023-04-13
fix(datastore-v1): stop datastore only stop sync engine

5d:
## Description

- The proposed PR offers an alternative solution that minimizes the changes required to fix the issue.
- When calling DataStore.stop, we only stop the SyncEngine and set it to nil, while keeping the StorageEngine alive to maintain local storage functionality.
- This approach may make the logic more difficult to understand but results in minimal changes.

## General Checklist

- [x] All unit tests pass


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -105,27 +109,28 @@ final class StorageEngine: StorageEngineBehavior {
 
         try storageAdapter.setUp(modelSchemas: StorageEngine.systemModelSchemas)
         if #available(iOS 13.0, *) {
-            let syncEngine = isSyncEnabled ? try? RemoteSyncEngine(storageAdapter: storageAdapter,
```
royjit, line 108: Dont we need syncEngine when we initialize the StorageEngine for the first time?

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
```diff
@@ -232,6 +232,10 @@ class RemoteSyncEngine: RemoteSyncEngineBehavior {
         stateMachine.notify(action: .finished)
     }
 
+    func isSyncing() -> Bool {
+        stateMachine.state.displayName != State.notStarted.displayName
```
royjit, line 236: Can we directly check the enum cases instead of checking the property `displayName`?


```
    func isSyncing() -> Bool {
        if case .notStarted = stateMachine.state {
            return false
        }
        return true
    }
```

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -272,7 +272,7 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
 
     public func start(completion: @escaping DataStoreCallback<Void>) {
         initStorageEngineAndStartSync { result in
-            completion(result)
+            self.queue.async { completion(result) }
```
royjit, line 275: As discussed this is required to avoid deadlock

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -355,6 +354,11 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
         guard #available(iOS 13.0, *) else {
             return
         }
+
+        guard let storageEngine = storageEngine else {
+            return
```
royjit, line 359: Lets add a verbose log here so that we can debug if something goes wrong.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
5d: We don't need the syncEngine when we initialize the StorageEngine. I seperated the initialization of StorageEngine and SyncEngine. We can stop the syncEngine and reinitalized it again later.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Sync/RemoteSyncEngine.swift
5d: will change to that.

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
5d: will add the log

AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/AWSDataStorePlugin+DataStoreBaseBehavior.swift
```diff
@@ -355,6 +353,15 @@ extension AWSDataStorePlugin: DataStoreBaseBehavior {
...
             return
         }
+
+        guard let storageEngine = storageEngine else {
+            log.info(
+                "StorageEngine is nil; " +
+                "Skip publishing the mutaitonEvent for \(mutationType) - \(modelSchema.name)"
+            )
+            return
+        }
```
royjit, line 363:
```suggestion
        guard let storageEngine = storageEngine else {
            log.info(
                """
                StorageEngine is nil; Skip publishing the mutaitonEvent for
                \(mutationType) - \(modelSchema.name)
                """
            )
            return
        }
```

5d: updated!

royjit (approved):
LGTM

PR #2888, opened 2023-04-23
fix(datastore): fix sync with lazy references

sapir-muallem:
## Description
- The decoding logic of `LazyReference` relied on `ModelProviderRegistry.decoders` to handle the decoding process.
- In case the `ModelProviderRegistry.decoders` failed to decode, the logic attempted to decode it into a model instance.
- However, if required fields were missing from the JSON, the decode failed and returned nil.
- Handling this scenario by returning the unloaded model state instead of nil was missing.
- Fixed this issue by adding a logic that handles not fully loaded models.

## General Checklist
- [X] All unit tests pass

### DIFF
--- a/AmplifyTests/CategoryTests/DataStore/Model/LazyReferenceTests.swift
+++ b/AmplifyTests/CategoryTests/DataStore/Model/LazyReferenceTests.swift
@@ -84,4 +84,23 @@ final class LazyReferenceTests: XCTestCase {
         let json = try comment.toJSON()
         XCTAssertEqual(json, "{\"post\":[{\"name\":\"id\",\"value\":\"postId\"}],\"id\":\"commentId\",\"content\":\"content\",\"updatedAt\":null,\"createdAt\":null}")
     }
+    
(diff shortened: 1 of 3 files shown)


AmplifyTests/CategoryTests/DataStore/Model/LazyReferenceTests.swift
```diff
@@ -84,4 +84,23 @@ final class LazyReferenceTests: XCTestCase {
         let json = try comment.toJSON()
         XCTAssertEqual(json, "{\"post\":[{\"name\":\"id\",\"value\":\"postId\"}],\"id\":\"commentId\",\"content\":\"content\",\"updatedAt\":null,\"createdAt\":null}")
     }
+    
+    func testDecodePrimaryKeysOnly() async throws {
```
MorCohenAres, line 88: Conventional test name should include the expectation:
`testDecodePrimaryKeysOnlyNotLoaded`

MorCohenAres (approved):

PR #2939, opened 2023-05-05
fix(datastore-v1): stop storageEngine do not publish completion event

5d:
## Description

Since we have made changes to the `storageEngine`, stopping it now only stops the `syncEngine`. 
Therefore, the completion event from the `syncEngine` should not be propagated by the `storageEngine` publisher. If this event is passed, the subscription will be terminated, and any future `syncEngine` events will not be propagated if the `syncEngine` restarts.

## General Checklist

- [X] All unit tests pass


AmplifyPlugins/DataStore/AWSDataStoreCategoryPlugin/Storage/StorageEngine.swift
```diff
@@ -133,9 +133,9 @@ final class StorageEngine: StorageEngineBehavior {
     func onReceiveCompletion(receiveCompletion: Subscribers.Completion<DataStoreError>) {
         switch receiveCompletion {
         case .failure(let dataStoreError):
-            storageEnginePublisher.send(completion: .failure(dataStoreError))
+            log.error("RemoteSyncEngine completed with error \(dataStoreError)")
```
royjit, line 136: Do we know if there is any other scenario where we need to handle this?

5d: Good question. The behavior remains unchanged even we've changed this line. 

We have updated [the way we stop the datastore plugin](https://github.com/aws-amplify/amplify-swift/pull/2863) so that it now only stops the syncEngine when the storageEngine is stopped. It is essential to maintain the subscription between the datastore plugin and the storageEngine.

royjit (approved):

PR #2942, opened 2023-05-09
fix(datastore): stop datastore plugin only stops syncEngine

5d:
## Issue \#

This is the ported fix for v1 https://github.com/aws-amplify/amplify-swift/pull/2863
## Description

In the current implementation, if the customer triggers DataStore.stop, it's possible that the local DB connection may not be closed immediately. If the customer attempts to perform another DataStore operation while this connection is still open, a second connection to the database may be initiated from a different process. This could cause conflicts with the [SQLite locking system](https://www.sqlite.org/lockingv3.html) and potentially result in a failed write operation.

This PR is to address the issue with minimal changes. Specifically, when the `DataStore.stop` API is called, only the `SyncEngine` is stopped and set to nil, while the `StorageEngine` remains active to preserve the connection to local database.

## General Checklist

- [X] All unit tests pass


royjit (approved):
LGTM

PR #2979, opened 2023-05-19
fix(api): wrap URLSession with validation state

5d:
## Issue \#

When performing `Amplify.reset`, th `URLSession` of the API plugin is also invalidated and reset. This means that any ongoing operations that utilize this `URLSession` to create new URL tasks may encounter an exception with the message `"Terminating app due to uncaught exception 'NSGenericException', reason: 'Task created in a session that has been invalidated'"` Unfortunately, `URLSession` does not provide a method to check if the current session has been invalidated.

## Description

- create a new class called `AmplifyURLSession` that acts as a wrapper for `URLSession` and also conforms to the `URLSessionBehavior` protocol.
- in the `AmplifyURLSession` class, include a variable that tracks the state of the `URLSession`.
- when attempting to generate a dataTask, check the state of the `URLSession` in `AmplifyURLSession` to ensure it is not already invalidated. If the session is invalidated, prevent the generation of the dataTask.

we can effectively manage the `URLSession` state and avoid generating data tasks when the session is already invalidated.

## General Checklist

- [X] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreHostApp/AmplifyURLSession.swift
+++ b/AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreHostApp/AmplifyURLSession.swift
@@ -1,0 +1,77 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+
+import Foundation
+import Amplify
+import AWSAPIPlugin
+
+enum AmplifyURLSessionState {
+    case active
+    case inactive
+}
+
+class AmplifyURLSessionNoOperationDataTask: URLSessionDataTaskBehavior {
+    var taskBehaviorIdentifier: Int
+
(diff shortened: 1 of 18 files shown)


AmplifyPlugins/DataStore/Tests/DataStoreHostApp/DataStoreHostApp/AmplifyURLSession.swift
```diff
@@ -0,0 +1,77 @@
...
+
+    init(taskBehaviorIdentifier: Int, taskBehaviorResponse: URLResponse? = nil) {
+        self.taskBehaviorIdentifier = taskBehaviorIdentifier
+        self.taskBehaviorResponse = taskBehaviorResponse
+    }
+
+    static let shared: AmplifyURLSessionNoOperationDataTask =
+        AmplifyURLSessionNoOperationDataTask(taskBehaviorIdentifier: -1)
+
+    func cancel() { }
```
royjit, line 31: nit: Add some logs so that we know when this methods get called. Might help in debugging if the test fails for some reason.

PR #2988, opened 2023-05-24
fix(datastore): fix failed test cases for PR#2979

5d:
## Issue \#

https://github.com/aws-amplify/amplify-swift/pull/2979

## Description

After importing `DataStoreHostApp`, there are redundant declarations of extesnsion for String to confirm `Error` protocol. The build target is also downgrade to iOS13 with adding `availablility` check for `StateObject`.

## General Checklist

- [X] All unit tests pass

### DIFF
--- a/AmplifyPlugins/API/Tests/APIHostApp/APIHostApp/SubscriptionView.swift
+++ b/AmplifyPlugins/API/Tests/APIHostApp/APIHostApp/SubscriptionView.swift
@@ -55,6 +55,7 @@ class SubscriptionViewModel: ObservableObject {
         self.todos.append(todo)
     }
 }
+@available(iOS 14.0, *)
 struct SubscriptionView: View {
     @StateObject var vm = SubscriptionViewModel()
     
@@ -73,6 +74,7 @@ struct SubscriptionView: View {
     
 }
 
+@available(iOS 14.0, *)
 struct SubscriptionView_Previews: PreviewProvider {
     static var previews: some View {
         SubscriptionView()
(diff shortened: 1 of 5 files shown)


AmplifyPlugins/API/Tests/APIHostApp/APIHostApp/SubscriptionView.swift
```diff
@@ -55,6 +55,7 @@ class SubscriptionViewModel: ObservableObject {
         self.todos.append(todo)
     }
 }
+@available(iOS 14.0, *)
```
atierian, line 58: I'm assuming the answer is "no", but just to make sure. The `SubscriptionView` isn't necessary for any of the tests that are running on iOS 13?

atierian: It looks like we don't need it based on the tests. If we do at some point, based on what `SubscriptionView` is doing, it looks like we can inject an `@ObservableObject` instead.

atierian (approved):

AmplifyPlugins/API/Tests/APIHostApp/APIHostApp/SubscriptionView.swift
5d: I think it won't work with iOS13. But we are using [the latest OS version](https://github.com/aws-amplify/amplify-swift/blob/main/.github/composite_actions/run_xcodebuild_test/action.yml#L17) of [Github macos-12 Xcode](https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md), which should be iOS 16.2

PR #3032, opened 2023-06-21
feat: Adding support for tvOS and watchOS

ruisebas:
## Description
This PR includes all the _development_ changes needed to the code in order to support tvOS and watchOS.
These mainly consist of compiler check flags to either exclude or tweak some functionality in order to accommodate to the target platform.

Differences:
 - **DevMenu** remains an iOS-only feature
 - **HostedUI** is not available in tvOS/watchOS
 - `Amplify.API.reachabilityPublisher(for:)` is not available in watchOS.
 - `Amplify.Notifications.Push.recordNotificationOpened(_:)` is not available in tvOS, as alert notification are not supported

## General Checklist
- [ ] Added new tests to cover change, if needed (Coming in a separate PR)
- [X] Build succeeds with all target using Swift Package Manager
- [] All unit tests pass (Coming in a separate PR)
- [ ] All integration tests pass (Coming in a separate PR)
- [X] Security oriented best practices and standards are followed (e.g. using input sanitization, principle of least privilege, etc)
- [X] Documentation update for the change if required
- [X] PR title conforms to conventional commit style
- [ ] New or updated tests include `Given When Then` inline code documentation and are named accordingly `testThing_condition_expectation()`
- [X] If breaking change, documentation/changelog update with migration instructions

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3036, opened 2023-06-22
test: Adding unit and integration tests for tvOS and watchOS

ruisebas:
## Description
This PR includes all the _testing_ changes needed to the code in order to support tvOS and watchOS.
These mainly consist on **some** compilation checks, **many** project/target new configurations, and finally adding the new platforms on the CircleCI and GitHub Workflow files.

### ⚠️ CircleCI checks will probably fail
This is expected as this PR only includes code changes in tests, but not their corresponding `development`. See #3032 for those changes.

#### What's with all the new `*Watch` schemes/targets?
For watchOS integration tests, we cannot reuse the existing host apps (we can for tvOS). Instead, we need dedicated Watch apps for each, and have the tests run using those apps.
While there's totally a way of sharing the test schemes and dynamically select the host app based on the platform is being built for, unfortunately I was unable to make it work 🤷‍♂️. This could be caused by some manual build configurations we have in our apps, but in the sake of time I just decided to duplicate the schemes. The actual files are the same though.

## General Checklist
- [X] All unit tests pass


royjit (approved):
lgtm

PR #3078, opened 2023-07-10
feat(logging): integrate AWS CloudWatch Plugin 

phantumcode:
## Description
Integrating all components of AWS CloudWatch plugins

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/Amplify/Core/Configuration/AmplifyConfiguration.swift
+++ b/Amplify/Core/Configuration/AmplifyConfiguration.swift
@@ -135,9 +135,11 @@ extension Amplify {
             }
         }
 
-        // Always configure Logging, Hub and Auth first, so they are available to other categories.
-        // Auth is a special case for other plugins which depend on using Auth when being configured themselves.
+        // Always configure logging first since Auth dependings on logging
         try configure(CategoryType.logging.category, using: resolvedConfiguration)
+
+        // Always configure Hub and Auth next, so they are available to other categories.
+        // Auth is a special case for other plugins which depend on using Auth when being configured themselves.
         let manuallyConfiguredCategories = [CategoryType.hub, .auth]
         for categoryType in manuallyConfiguredCategories {
             try configure(categoryType.category, using: resolvedConfiguration)
--- a/Amplify/Core/Configuration/AmplifyConfiguration.swift
+++ b/Amplify/Core/Configuration/AmplifyConfiguration.swift
@@ -137,7 +137,8 @@ extension Amplify {
 
         // Always configure Logging, Hub and Auth first, so they are available to other categories.
         // Auth is a special case for other plugins which depend on using Auth when being configured themselves.
-        let manuallyConfiguredCategories = [CategoryType.logging, .hub, .auth]
+        try configure(CategoryType.logging.category, using: resolvedConfiguration)
+        let manuallyConfiguredCategories = [CategoryType.hub, .auth]
         for categoryType in manuallyConfiguredCategories {
             try configure(categoryType.category, using: resolvedConfiguration)
         }
--- a/Amplify/DefaultPlugins/AWSUnifiedLoggingPlugin/Internal/ConsoleLoggingConfiguration.swift
+++ b/Amplify/DefaultPlugins/AWSUnifiedLoggingPlugin/Internal/ConsoleLoggingConfiguration.swift
@@ -1,0 +1,81 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+public struct ConsoleLoggingConfiguration: Codable {
+    public init(enable: Bool = true) {
+         self.enable = enable
+    }
+
+   public let enable: Bool
+}
+
+extension ConsoleLoggingConfiguration {
+    init?(bundle: Bundle) {
+        guard let path = bundle.path(forResource: "amplifyconfiguration_logging", ofType: "json") else {
+            return nil
+        }
+
+        let url = URL(fileURLWithPath: path)
+
+        if let config = try? ConsoleLoggingConfiguration.loadConfiguration(from: url) {
+            self = config
+        } else {
+            return nil
+        }
+    }
+
+    static func loadConfiguration(from url: URL) throws -> ConsoleLoggingConfiguration? {
+        let fileData: Data
+        do {
+            fileData = try Data(contentsOf: url)
+        } catch {
+            throw LoggingError.configuration(
+                """
+                Could not extract UTF-8 data from `\(url.path)`
+                """,
+
+                """
+                Could not load data from the file at `\(url.path)`. Inspect the file to ensure it is present.
+                The system reported the following error:
+                \(error.localizedDescription)
+                """,
+                error
+            )
+        }
+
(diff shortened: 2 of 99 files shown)


Amplify/Core/Configuration/AmplifyConfiguration.swift
```diff
@@ -137,7 +137,8 @@ extension Amplify {
 
         // Always configure Logging, Hub and Auth first, so they are available to other categories.
         // Auth is a special case for other plugins which depend on using Auth when being configured themselves.
-        let manuallyConfiguredCategories = [CategoryType.logging, .hub, .auth]
+        try configure(CategoryType.logging.category, using: resolvedConfiguration)
```
ruisebas, line 140: Isn't this the same as before? 🤔

Amplify/DefaultPlugins/AWSUnifiedLoggingPlugin/Internal/ConsoleLoggingConfiguration.swift
```diff
@@ -0,0 +1,81 @@
...
+//
+
+import Foundation
+
+public struct ConsoleLoggingConfiguration: Codable {
+    public init(enable: Bool = true) {
+         self.enable = enable
+    }
+
+   public let enable: Bool
```
ruisebas, line 15:
```suggestion
    public let enable: Bool
```

AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/AWSCloudWatchLoggingCategoryClient.swift
```diff
@@ -0,0 +1,195 @@
...
+import Network
+
+/// Concrete implementation of
+/// [LoggingCategoryClientBehavior](x-source-tag://LoggingCategoryClientBehavior)
+/// ensures the right log files and cloud watch streams are used according to
+/// the application's authentication state.
+///
+/// - Tag: CloudWatchLoggingCategoryClient
+final class AWSCloudWatchLoggingCategoryClient {
+    private var _enable: Bool = true
```
ruisebas, line 23: 🤔 why the underscore?

AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/AWSCloudWatchLoggingSessionController.swift
```diff
@@ -0,0 +1,193 @@
...
+    
+    func disable() {
+        self.batchSubscription = nil
+        self.authSubscription = nil
+        self.session = nil
+        self.consumer = nil
+    }
+
+    private func updateConsumer() {
+        do {
```
ruisebas, line 92: You can just do

```swift
consumer = try? createConsumer()
```

AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/AWSCloudWatchLoggingSessionController.swift
```diff
@@ -0,0 +1,193 @@
...
+    
+    var client: CloudWatchLogsClientProtocol?
+    private let localStoreMaxSizeInMB: Int
+    private let credentialsProvider: CredentialsProvider
+    private let authentication: AuthCategoryUserBehavior
+    private let category: String
+    private let namespace: String?
+    private var session: AWSCloudWatchLoggingSession?
+    private var consumer: LogBatchConsumer?
+    private var batchSubscription: AnyCancellable? { willSet { batchSubscription?.cancel() } }
```
ruisebas, line 39: nit: properly format the `willSet` observations to match `didSet`

AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/AWSCloudWatchLoggingCategoryClient.swift
```diff
@@ -0,0 +1,195 @@
...
+        self.logger(forCategory: namespace)
+    }
+    
+    func logger(forCategory category: String, forNamespace namespace: String) -> Logger {
+        let defaultLogLevel = logFilter.getDefaultLogLevel(forCategory: category, userIdentifier: self.userIdentifier)
+        return self.logger(forCategory: category, namespace: namespace, logLevel: defaultLogLevel)
+    }
+    
+    func getInternalClient() -> CloudWatchLogsClientProtocol {
+        loggersByKey.first!.value.client!
```
ruisebas, line 178: Force unwrapping is always risky, but we're doing it two times here. 
How confident are we that it's not gonna result in a crash? Perhaps we could use a `preconditionFailure` instead?

Amplify/Core/Configuration/AmplifyConfiguration.swift
phantumcode: Logging needs to be the first, order wasn't guaranteed in the previous code and initialization would randomly fail..  I've added additional code comments to clarify.

Amplify/DefaultPlugins/AWSUnifiedLoggingPlugin/Internal/ConsoleLoggingConfiguration.swift
phantumcode: Spacing updated.

AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/AWSCloudWatchLoggingCategoryClient.swift
phantumcode: This was to avoid naming collision with the public func `enable()`.  I've updated the internal variable to `enabled` in the latest rev.

AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/AWSCloudWatchLoggingSessionController.swift
phantumcode: Good catch, code has been updated to the simplified recommendaiton.

AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/AWSCloudWatchLoggingCategoryClient.swift
phantumcode: There should not be a case where the client is nil if the plugin is configured.  I've updated the logic with a `guard let` check with a `preconditionFailure`

Amplify/Core/Configuration/AmplifyConfiguration.swift
ruisebas: > order wasn't guaranteed

in arrays!? TIL 🤯

ruisebas (approved):

Amplify/Core/Configuration/AmplifyConfiguration.swift
ruisebas: Could it be that there's something else going on behind the scenes that's causing the fail? Arrays should keep their order, it's like one of their features.

It could totally be that Swift's for loop is messing up, I'm just being extra cautious

PR #3080, opened 2023-07-11
feat: Add default reset impl for Plugins; change all Resettable conformances to async functions

palpatim:
## Issue \#
N/A

## Description
This change adds a default async `reset` implementation for Plugins, to remove the need for plugin authors to add a no-op `reset` method. It also removes the now-redundant no-op `reset` method on the `AWSCognitoAuthPlugin`.

By doing so, this change makes the default async `reset` method a higher precedence than a sync method on a concrete plugin type. That means that in normal invocations of `await Amplfiy.reset()`, sync plugin `reset()` methods would not be invoked, because Swift would dispatch calls to the default implementation instead. Therefore, this change also:

1. Adds `async` keyword to all current `Resettable` conformances
2. Invokes `reset()` from the protocol existential in tests, rather than the concrete implementing type
3. Replaces `wait(for:timeout:)` with `await fulfillment(of:timeout:)` in several async tests (this was an opportunistic change to clean up some test warnings, not strictly related to the resettable default implementation)

**NOTE** This change introduces one new warning: Because this change adds `async` to  `AWSInitialSyncOrchestrator/reset()`, it now emits a warning:

> Instance method 'waitUntilAllOperationsAreFinished' is unavailable from asynchronous contexts; Use addBarrierBlock or a dependent Operations instead; this is an error in Swift 6

This will need to be fixed before supporting Swift 6.

## General Checklist

- [ ] ~~Added new tests to cover change, if needed~~
- [X] All unit tests pass
- [ ] ~~Security oriented best practices and standards are followed (e.g. using input sanitization, principle of least privilege, etc)~~
- [ ] ~~Documentation update for the change if required~~
- [ ] ~~New or updated tests include `Given When Then` inline code documentation and are named accordingly `testThing_condition_expectation()`~~
- [ ] ~~If breaking change, documentation/changelog update with migration instructions~~


thisisabhash (approved):
LGTM

Issue #3107, opened 2023-07-31
AuthError: iOS Amplify V2 Token returned by service could not be parsed


bevanchristian:
### Describe the bug

So, I have an application that uses Amplify v1 and has been running well for a year, and now I want to upgrade to Amplify v2.

in app delegate i configure my amplify using custom configuration

```swift
    private func initializeAmplify() {
        do {
            let authConf = AuthCategoryConfiguration(plugins: AuthConfig().buildConfig())
            let config = AmplifyConfiguration(
                        analytics: nil,
                        api: nil,
                        auth: authConf,
                        dataStore: nil,
                        hub: nil,
                        logging: nil,
                        predictions: nil,
                        storage: nil
                    )
            Amplify.Logging.logLevel = .verbose
            try Amplify.add(plugin: AWSCognitoAuthPlugin())
            try Amplify.configure(config)
        } catch {
            print("Failed to initialize Amplify with \(error)")
        }
    }

```
this is my function for build config

```swift
    func buildConfig() -> [String: JSONValue] {
            return [
                   "awsCognitoAuthPlugin": [
                       "IdentityManager": [
                           "Default": [:]
                       ],
                       "Auth": [
                           "Default": [
                                "authenticationFlowType": "USER_SRP_AUTH",
                                "OAuth": [
                                    "WebDomain": "xxxx",
                                    "AppClientId": "xxxxxx",
                                    "SignInRedirectURI": "myapp://profile/me/",
                                    "SignOutRedirectURI": "myapp://home/",
                                    "Scopes": [
                                        "aws.cognito.signin.user.admin",
                                        "email",
                                        "openid",
                                        "phone",
                                        "profile"
                                    ]
                                ]
                           ]
                       ],
                       "CognitoUserPool": [
                            "Default": [
                                "PoolId": "xxxxx",
                                "AppClientId": "xxxxxxx",
                                "Region": "ap-southeast-1"
                            ]
                       ],
                       "CredentialsProvider": [
                           "CognitoIdentity": [
                                "Default": [
                                    "PoolId": "xxxxxxxx",
                                    "Region": "ap-southeast-1"
                                ]
                           ]
                       ]
                   ]
               ]
    }
```
and i call this code for user to sign in

```swift
    let defaultPresentationAnchor = UIApplication.shared.windows.filter
        {$0.isKeyWindow}.first ?? UIWindow()
        
        do {
[...]

PR #3114, opened 2023-08-01
fix(Auth): Refactor the expires in logic of the authentication result

harsh62:
## Issue \#
#3107 

## Description
Using the min expiration time of the idToken and accessToken to detemine Authentication result expiry in the abscence of expiry details from the service. 

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
@@ -41,10 +41,10 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
         self.expiration = expiration
     }
 
-    public init(idToken: String,
-                accessToken: String,
-                refreshToken: String,
-                expiresIn: Int? = nil) {
+    init(idToken: String,
+         accessToken: String,
+         refreshToken: String,
+         expiresIn: Int? = nil) {
 
         self.idToken = idToken
         self.accessToken = accessToken
@@ -71,7 +71,7 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
             self.expiration = Date().addingTimeInterval(TimeInterval((expirationDoubleValue ?? 0)))
         }
     }
-
+    
 }
 
 extension AWSCognitoUserPoolTokens: Equatable { }
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
@@ -16,6 +16,7 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
 
     public let refreshToken: String
 
+    @available(*, deprecated, message: "Use of `expiration` is deprecated, use `exp` claim in the `idToken` or `accessToken` for expiries")
     public let expiration: Date
 
     public init(idToken: String,
@@ -49,18 +50,24 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
                       refreshToken: refreshToken,
                       expiresIn: expiresIn)
         } else {
-            // If for some reason, failed to extract "exp" value, use the value as zero and don't block the user
-            var expirationDoubleValue: Double = 0
-            if let idTokenClaims = try? AWSAuthService().getTokenClaims(tokenString: idToken).get(),
-               let accessTokenClaims = try? AWSAuthService().getTokenClaims(tokenString: accessToken).get(),
-               let idTokenExpiration = idTokenClaims["exp"]?.doubleValue,
-               let accessTokenExpiration = accessTokenClaims["exp"]?.doubleValue {
-                expirationDoubleValue = min(idTokenExpiration, accessTokenExpiration)
+
+            var expirationDoubleValue: Double? = nil
+            if let idTokenExpiration = try? AWSAuthService().getTokenClaims(tokenString: idToken).get()["exp"]?.doubleValue {
+                expirationDoubleValue = idTokenExpiration
             }
+            if let accessTokenExpiration = try? AWSAuthService().getTokenClaims(tokenString: accessToken).get()["exp"]?.doubleValue {
+                if let unwrappedExpirationDoubleValue = expirationDoubleValue {
+                    expirationDoubleValue = min(unwrappedExpirationDoubleValue, accessTokenExpiration)
+                } else {
+                    expirationDoubleValue = accessTokenExpiration
+                }
+            }
+
             self.init(idToken: idToken,
                       accessToken: accessToken,
                       refreshToken: refreshToken,
-                      expiresIn: Int(expirationDoubleValue.rounded(.down)))
+                      // If for some reason, failed to extract "exp" value, use the value as zero and don't block the user
+                      expiresIn: Int((expirationDoubleValue ?? 0).rounded(.down)))
         }
     }
 
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
@@ -19,6 +19,7 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
     @available(*, deprecated, message: "Use of `expiration` is deprecated, use `exp` claim in the `idToken` or `accessToken` for expiries")
     public let expiration: Date
 
+    @available(*, deprecated, message: "Use of `init(idToken,accessToken,refreshToken:expiresIn)` is deprecated, use `exp` claim in the `idToken` or `accessToken` instead")
     public init(idToken: String,
                 accessToken: String,
                 refreshToken: String,
@@ -29,6 +30,7 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
         self.expiration = Date().addingTimeInterval(TimeInterval(expiresIn))
     }
 
+    @available(*, deprecated, message: "Use of `init(idToken,accessToken,refreshToken:expiration)` is deprecated, use `exp` claim in the `idToken` or `accessToken` instead")
     public init(idToken: String,
                 accessToken: String,
                 refreshToken: String,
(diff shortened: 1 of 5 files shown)


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
```diff
@@ -38,6 +38,32 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
...
+                expiresIn: Int? = nil) {
+
+        if let expiresIn =  expiresIn {
+            self.init(idToken: idToken,
+                      accessToken: accessToken,
+                      refreshToken: refreshToken,
+                      expiresIn: expiresIn)
+        } else {
+            // If for some reason, failed to extract "exp" value, use the value as zero and don't block the user
+            var expirationDoubleValue: Double = 0
```
lawmicha, line 53: if we can't extract the "exp" values, how does expiresIn in 0 get used later on? what's the impact of setting it to 0?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
```diff
@@ -38,6 +38,32 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
...
+            if let idTokenClaims = try? AWSAuthService().getTokenClaims(tokenString: idToken).get(),
+               let accessTokenClaims = try? AWSAuthService().getTokenClaims(tokenString: accessToken).get(),
```
lawmicha, line 55: what if you can get one "exp" from one token, idToken or accessToken, but not the other? Would that be such a case? If so, maybe we can split it out as 0 by default, idToken's exp if accessToken doesn't exist, accessToken's exp if idToken doesn't exist, and then the min of the two if both exists. Just a thought, not sure if this there is such a scenario of token claims data.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
harsh62: There is no impact. As we are not expecting customers to use it.. 
So I have also marked this `expiry` property as deprecated.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
harsh62: Nice.. Ideally this won't happen and it would never reach here. But updated the logic..

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
```diff
@@ -38,6 +39,38 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
...
+            if let idTokenExpiration = try? AWSAuthService().getTokenClaims(tokenString: idToken).get()["exp"]?.doubleValue {
+                expirationDoubleValue = idTokenExpiration
+            }
+            if let accessTokenExpiration = try? AWSAuthService().getTokenClaims(tokenString: accessToken).get()["exp"]?.doubleValue {
+                if let unwrappedExpirationDoubleValue = expirationDoubleValue {
+                    expirationDoubleValue = min(unwrappedExpirationDoubleValue, accessTokenExpiration)
+                } else {
+                    expirationDoubleValue = accessTokenExpiration
+                }
+            }
```
lawmicha, line 64: I think a switch statement might be easier to read, for example, i gave this a try and looks like this:
```swift
struct TwoOptionals {
    
    func returnDoubleOrNil() throws -> Double? {
        return nil
    }
    
    func example() -> Int {
        let finalValue: Double
        switch (try? returnDoubleOrNil(), try? returnDoubleOrNil()) {
        case (.some(let value1), .none):
            finalValue = value1
        case (.none, .some(let value2)):
            finalValue = value2
        case (.some(let value1), .some(let value2)):
            finalValue = min(value1, value2)
        case (.none, .none):
            finalValue = 0
        }
        return Int(finalValue.rounded(.down))
        
    }
}

```

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
```diff
@@ -38,6 +39,38 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
         self.expiration = expiration
```
lawmicha, line 39: sorry i may be missing some context here, if `expiration` is deprecated, then i think the init `init(idToken:accessToken:refreshToken:expiration)` should also be deprecated. Then the new init that takes in `expiresIn` should hold/calculate the value from the claims "exp" in a new property called `expiresIn`? Introducing a new property will allow us to remove the deprecated one in a major version bump. Or perhaps `expiration` shouldn't be deprecated but only the `init(expiration)` should be, ie.
```
@available(*, deprecated, message: "Use of `init(idToken,accessToken,refreshToken:expiration)` is deprecated, use `exp` claim in the `idToken` or `accessToken` instead")
public init(idToken: String,
                accessToken: String,
                refreshToken: String,
                expiresIn: Int)
```

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
```diff
@@ -38,6 +39,38 @@ public struct AWSCognitoUserPoolTokens: AuthCognitoTokens {
...
     }
 
+    public init(idToken: String,
+                accessToken: String,
+                refreshToken: String,
+                expiresIn: Int? = nil) {
```
lawmicha, line 45: Do developers use this class init directly in their code? I think we should just expose `init(idToken, accessToken, refreshToken)` judging from the deprecation message

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
lawmicha: i meant, the expiration property is set to 0, and then how does the plugin use this value later on in the implementation? i can see that customers wouldn't need to read the value or do anything with it

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
lawmicha: let me know if this is off, my current understanding is
- I think `expiration` property shouldn't be deprecated because we are still setting the value, calculated from the claims. 
- the `expiration` property can still be read by the customer if they wanted to, but the init available to customer should be `init(idToken:accessToken:refreshToken)` without `expiresIn` property. this init does the caluclation and sets `self.expiration = ` directly

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
harsh62: Even the plugin doesn't use it in any way.. Its just something Cognito has and we passed it back to the customer.. Hoping to remove it in the next major version.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
harsh62: I am gonna mark the methods as deprecated but not in favor of adding a new property because we are not relying on it any longer. Will be removed in the next major version.. This is a workaround for the customer to unblock them.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
harsh62: Mmmm.. They don't use it.. But I can not be certain about it. Hence I kept it as a optional param and marked it as deprecated. 
Hopefully if some customers are using it could complain about the deprecation with a new issue, but that is highly unlikely.

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/TaskTests/HostedUITests/AWSAuthHostedUISignInTests.swift
```diff
@@ -88,6 +88,24 @@ class AWSAuthHostedUISignInTests: XCTestCase {
         XCTAssertTrue(result.isSignedIn)
     }
 
+    @MainActor
+    func testSuccessfulSignIn_missingExpiresIn() async throws {
```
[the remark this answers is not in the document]
harsh62: Added tests.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
lawmicha: I think keeping `expiresIn` as an optional parameter means we're essentailly not deprecating it (the logic of the deprecated init just got moved over to the new init). Imagine a major version bump where we remove the deprecated init, but developers do not have a breaking change and continue to use the new init if they were passing in `expiresIn`. I may not have the full picture of how AWSCognitoUserPoolTokens is created and used, so just my thought on `expiresIn` being an optional parameter and the fact that this init supports both use cases appears to be a pretty safe, non-breaking change.

lawmicha (dismissed):

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoUserPoolTokens.swift
harsh62: Thought more about it.. Made the new init non-public.

lawmicha (approved):

PR #3179, opened 2023-08-25
feat(auth): add userID property to AuthSignUpResult

atierian:
## Description

Adds a `userID` property to `AuthSignUpResult`. 

Currently, this property is accessible when the `AuthSignUpResult().nextStep` is `AuthSignUpStep.confirmUser` through the case's associated values. 

The typical flow is `signUp() --> .confirmUser --> confirmSignUp() --> .done`.  

`confirmSignUp` ([ConfirmSignUp](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ConfirmSignUp.html)) doesn't return a `userID` (`UserSub`). However, `signUp() --> .done` is also a valid flow with the use of a custom verification flow. In this case, we receive the `userID` but we're preventing the caller from accessing it. This change provides the `userID` in the `AuthSignUpResult` if it's returned by the service, regardless of the `nextStep`.

### Platform Parity
This change brings Amplify Swift in parity with Android, Flutter, and JS.
- [Android - AuthSignUpResult.getUserId()](https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/auth/result/AuthSignUpResult.java#L67-L74)
- [Flutter - SignUpResult.userId](https://github.com/aws-amplify/amplify-flutter/blob/main/packages/amplify_core/lib/src/types/auth/sign_up/sign_up_result.dart#L28)
- [JS - AuthSignUpResult.userId](https://github.com/aws-amplify/amplify-js/blob/next/packages/auth/src/types/results.ts#L32)

**NOTE**:
This change can be confusing because `AuthSignUpResult` and `AuthSignUpStep.confirmUser` both contain the same `userID` value.
```swift
let result = try await Amplify.Auth.signUp(...)

result.userID // "abc123"
switch result {
case .confirmUser(_, _, let userID):
  userID // "abc123"
...
}
```
The associated value on `.confirmUser` is no longer necessary, but removing would be a breaking change. Something to consider for vNext.

#### Misc. Changes
Moved from `wait(for:)` to `await fulfillment(of:)` for two test cases that were being problematic. 

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Service/Helpers/SignUpOutputResponse+Helper.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Service/Helpers/SignUpOutputResponse+Helper.swift
@@ -13,14 +13,15 @@ extension SignUpOutputResponse {
 
     var authResponse: AuthSignUpResult {
         if self.userConfirmed {
-            return .init(.done)
+            return .init(nextStep: .done, userID: userSub)
         }
         return AuthSignUpResult(
-            .confirmUser(
+            nextStep: .confirmUser(
                 codeDeliveryDetails?.toAuthCodeDeliveryDetails(),
                 nil,
                 userSub
-            )
+            ),
+            userID: userSub
         )
     }
 }
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Service/Helpers/SignUpOutputResponse+Helper.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Service/Helpers/SignUpOutputResponse+Helper.swift
@@ -13,10 +13,10 @@ extension SignUpOutputResponse {
 
     var authResponse: AuthSignUpResult {
         if self.userConfirmed {
-            return .init(nextStep: .done, userID: userSub)
+            return .init(.done, userID: userSub)
         }
         return AuthSignUpResult(
-            nextStep: .confirmUser(
+            .confirmUser(
                 codeDeliveryDetails?.toAuthCodeDeliveryDetails(),
                 nil,
                 userSub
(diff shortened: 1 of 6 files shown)


dnys1 (approved):

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Task/AWSAuthConfirmSignUpTask.swift
```diff
@@ -41,7 +41,7 @@ class AWSAuthConfirmSignUpTask: AuthConfirmSignUpTask, DefaultLogger {
                                            environment: userPoolEnvironment)
             _ = try await client.confirmSignUp(input: input)
             log.verbose("Received success")
-            return AuthSignUpResult(.done)
+            return AuthSignUpResult(.done, userID: nil)
```
harsh62, line 44: Do you really need this change? `userID` is set to `nil` by default in the `init`.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Service/Helpers/SignUpOutputResponse+Helper.swift
```diff
@@ -13,14 +13,15 @@ extension SignUpOutputResponse {
 
     var authResponse: AuthSignUpResult {
         if self.userConfirmed {
-            return .init(.done)
+            return .init(.done, userID: userSub)
```
harsh62, line 16: Same here. 
> Do you really need this change? userID is set to nil by default in the init.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Task/AWSAuthConfirmSignUpTask.swift
atierian: We don't *need* it. But I believe it provides some value over omitting it.

I would've preferred to have `AuthSignUpResult.init` without a default argument for `userID`, but doing so would be a breaking change*. Explicitly passing the `nil` here provides some added context: "_we know there's no `userID` available when creating this_" vs. "_we may know that, or we may have missed that the argument exists_."

If explicitly passing `nil` has a drawback I'm not seeing (other than verbosity), I'm happy to reconsider.

*_we could add an overload with different labels or use `@_disfavoredOverload`, but they'd essentially be the same thing._

harsh62: Makes sense.. Thanks for explaining.

harsh62 (approved):

Issue #3176, opened 2023-08-25
Not getting correct error message from Google signin.


AmbrishPatel94:
### Describe the bug

I am use following Amplify version 2.0.0-Next Major, I register new user on amplify and try to login with amplify google signin "Amplify.Auth.signInWithWebUI(for: .google, presentationAnchor: presentationAnchor, options: .preferPrivateSession())" with same user then that time i got error response "invaid_request". But i need to get response like this "invalid_request: PreSignUp failed with error {"error":{"code":409,"custom_code":"002","message":"U2FsdGVkX19xJYJt/Dckptj0KKXuBfrD3uhZScr1bBvqYF2Q29iGb7iC16MDCwLwbIVs1tNcaHQreEIopKwg9onymeCveyiCwHNzjpmPciQ3XSiQ7Oi+pVW4KyB5lUXC"}}."

### Steps To Reproduce

```swift
I register new user on amplify and try to login with amplify google signin "Amplify.Auth.signInWithWebUI(for: .google, presentationAnchor: presentationAnchor, options: .preferPrivateSession())" with same user then that time i got error response "invaid_request".
```

### Expected behavior

I need to get response like this "invalid_request: PreSignUp failed with error {"error":{"code":409,"custom_code":"002","message":"U2FsdGVkX19xJYJt/Dckptj0KKXuBfrD3uhZScr1bBvqYF2Q29iGb7iC16MDCwLwbIVs1tNcaHQreEIopKwg9onymeCveyiCwHNzjpmPciQ3XSiQ7Oi+pVW4KyB5lUXC"}}."

### Amplify Framework Version

2.0.0-Next Major

### Amplify Categories

Auth

### Dependency manager

Swift PM

### Swift version

5

### CLI version

10.6.1

### Xcode version

14.2

### Relevant log output

```shell
<details>
<summary>Log Messages</summary>


INSERT LOG MESSAGES HERE
```
</details>
```


### Is this a regression?

Yes

### Regression additional context

_No response_

### Platforms

iOS

### OS Version

13.0

### Device

iphone 13

### Specific to simulators

_No response_

### Additional context

_No response_

**Comment by atierian** [comment]
Hey @AmbrishPatel94, thanks for opening this issue.
If I understand correctly, you're receiving an error with only `invalid_request` but are expecting an error with more information like in the example you provided:
```
[...]

PR #3183, opened 2023-08-28
fix(auth): hostedui extract error_description query pararm

atierian:
## Issue \#

- https://github.com/aws-amplify/amplify-swift/issues/3176

## Description

If an `error` query parameter is included in the HostedUI callback, we currently throw a `.service` error with the value of the `error` key, omitting important information that may be needed by the caller. With this change, we now attempt to extract and use the value from `error_description`; if it doesn't exist, we fall back to the `error` we know is there. The fall back to `error` is necessary because `error_description` is optional according to the [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1).

This change brings Amplify Swift in parity with Android, Flutter, and JS.

- [Amplify Android](https://github.com/aws-amplify/amplify-android/blob/aac074ab49b44ccea0c9eb12d996f7033c20e3b7/aws-auth-cognito/src/main/java/com/amplifyframework/auth/cognito/HostedUIClient.kt#L105-L108)
- [Amplify Flutter](https://github.com/aws-amplify/amplify-flutter/blob/cd8c84a5f9cc4db998b1befa03145ff8f5a73cd3/packages/auth/amplify_auth_cognito_dart/lib/src/flows/hosted_ui/hosted_ui_platform.dart#L205-L212)
- [Amplify JS](https://github.com/aws-amplify/amplify-js/blob/8d1f79a936a2a07c4ea893957b2c2703011c39dc/packages/auth/src/OAuth/OAuth.ts#L237-L239)

In the linked issue, we can see a specific use case the current behavior is blocking - client side logic based on information contained in an error thrown by a `PreSignUp_ExternalProvider` Lambda trigger. [[Service Documentation](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-sign-up.html#aws-lambda-triggers-pre-registration-example-3)]

### Testing
In addition to the added test cases, this change we manually tested with a `PreSignUp` Lambda trigger:
```js
exports.handler = (event, context, callback) => {
    const payload = JSON.stringify(
        { code: "409", custom_code: "002", message: "<placeholder>" }
    );
    const error = Error(payload);
    callback(error, event);
};
```
and the following client side code:
```swift
do {
    _ = try await Amplify.Auth.signInWithWebUI(
        for: .google, // this can be any social provider
        presentationAnchor: window
    )
} catch let error as AuthError {
>> breakpoint
} catch {
    print("Unexpected error: \(error)")
}
```
Breaking in the first `catch` block and `po`'ing  the error -

Targeting v2.16.0 (current latest release), we see:
```
(lldb) po error
▿ AuthError: invalid_request
Recovery suggestion: Received an error message from the service
  ▿ service : 3 elements
    - .0 : "invalid_request"
    - .1 : "Received an error message from the service"
    - .2 : nil
```

Targeting branch `hostedui_errordescription`, we see:
```
(lldb) po error
▿ AuthError: invalid_request PreSignUp+failed+with+error+{"code":"409","custom_code":"002","message":"<placeholder>"}.+
Recovery suggestion: Received an error message from the service
  ▿ service : 3 elements
    - .0 : "invalid_request PreSignUp+failed+with+error+{\"code\":\"409\",\"custom_code\":\"002\",\"message\":\"<placeholder>\"}.+"
    - .1 : "Received an error message from the service"
    - .2 : nil
```

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Support/HostedUI/HostedUIASWebAuthenticationSession.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Support/HostedUI/HostedUIASWebAuthenticationSession.swift
@@ -33,9 +33,9 @@ class HostedUIASWebAuthenticationSession: NSObject, HostedUISessionBehavior {
                     if let error = queryItems.first(where: { $0.name == "error" })?.value {
                         let errorDescription = queryItems.first(
                             where: { $0.name == "error_description" }
-                        )?.value?.trim() ?? error
-
-                        callback(.failure(.serviceMessage(errorDescription)))
+                        )?.value?.trim() ?? ""
+                        let message = "\(error) \(errorDescription)"
+                        callback(.failure(.serviceMessage(message)))
                         return
                     }
                     callback(.success(queryItems))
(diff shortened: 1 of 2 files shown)


harsh62 (approved):
LGTM

PR #3189, opened 2023-08-28
fix(Auth): Adding the ability to keep old preferred methods

harsh62:
## Description
Fixing the logic of the udpateMFAPreference API to make sure old preferred values stay intact. 

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/MFAPreference.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/MFAPreference.swift
@@ -27,10 +27,10 @@ public enum MFAPreference {
 
 extension MFAPreference {
 
-    func smsSetting(isCurrentlyPreferred: Bool? = nil) -> CognitoIdentityProviderClientTypes.SMSMfaSettingsType? {
+    func smsSetting(isCurrentlyPreferred: Bool = false) -> CognitoIdentityProviderClientTypes.SMSMfaSettingsType {
         switch self {
         case .enabled:
-            return .init(enabled: true, preferredMfa: isCurrentlyPreferred ?? false)
+            return .init(enabled: true, preferredMfa: isCurrentlyPreferred)
         case .preferred:
             return .init(enabled: true, preferredMfa: true)
         case .notPreferred:
@@ -40,7 +40,7 @@ extension MFAPreference {
         }
     }
 
-    func softwareTokenSetting(isCurrentlyPreferred: Bool? = nil) -> CognitoIdentityProviderClientTypes.SoftwareTokenMfaSettingsType? {
+    func softwareTokenSetting(isCurrentlyPreferred: Bool? = nil) -> CognitoIdentityProviderClientTypes.SoftwareTokenMfaSettingsType {
         switch self {
         case .enabled:
             return .init(enabled: true, preferredMfa: isCurrentlyPreferred ?? false)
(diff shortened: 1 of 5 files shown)


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/MFAPreference.swift
```diff
@@ -26,11 +26,11 @@ public enum MFAPreference {
 }
 
 extension MFAPreference {
-    
-    var smsSetting: CognitoIdentityProviderClientTypes.SMSMfaSettingsType? {
+
+    func smsSetting(isCurrentlyPreferred: Bool? = nil) -> CognitoIdentityProviderClientTypes.SMSMfaSettingsType? {
```
ruisebas, line 30: any particular reason for defaulting to `nil` instead of just `false`?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Task/UpdateMFAPreferenceTask.swift
```diff
@@ -61,10 +61,12 @@ class UpdateMFAPreferenceTask: AuthUpdateMFAPreferenceTask, DefaultLogger {
 
     func updateMFAPreference(with accessToken: String) async throws {
         let userPoolService = try userPoolFactory()
+        let currentPreference = try await userPoolService.getUser(input: .init(accessToken: accessToken))
+        let preferredMFAType = MFAType(rawValue: currentPreference.preferredMfaSetting ?? "")
```
atierian, line 65: For this transformation, you have to go from `String?` --> `String` --> `MFAType?`.
Using `map` here more accurately represents the transformation because it removes the need for a default value (`preferredMFASetting ?? ""`). 

This may not seem like a big deal now because there's currently no way for `""` to represent a valid `MFAType`. This could be problematic if we add our own `MFAType.init?(rawValue:)` in the future instead of using the synthesized one.

```suggestion
        let preferredMFAType = currentPreference.preferredMfaSetting.map(MFAType.init(rawValue:))
```

atierian (changes requested):

ruisebas (dismissed):

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/MFAPreference.swift
```diff
@@ -39,11 +39,11 @@ extension MFAPreference {
             return .init(enabled: false)
         }
     }
-    
-    var softwareTokenSetting: CognitoIdentityProviderClientTypes.SoftwareTokenMfaSettingsType? {
+
+    func softwareTokenSetting(isCurrentlyPreferred: Bool? = nil) -> CognitoIdentityProviderClientTypes.SoftwareTokenMfaSettingsType {
```
atierian, line 43: Any reason not to make this `isCurrentlyPreferred: Bool = false` like you did with `smsSetting`?

atierian (approved):

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/MFAPreference.swift
harsh62: Ahh.. Sorry.. Missed that one.. Fixed.

Issue #3184, opened 2023-08-28
AuthTokenURLInterceptor overriding Content-Type header to application/json always


sandeepklr:
### Describe the bug

Relevant amplifyconfiguration.json api configuration blob:

```
    "api": {
        "plugins": {
            "awsAPIPlugin": {
                "AbcApi": {
                    "endpointType": "REST",
                    "endpoint": "https://yyyyyyyyyy.execute-api.us-west-2.amazonaws.com/dev",
                    "region": "us-west-2",
                    "authorizationType": "OPENID_CONNECT"
                }
            }
        }
    }
```

I'm trying to use the Amplify API plugin to make `multipart/form-data` requests to API Gateway. From documentation listed here[1], I'd have expected to be able to do the following:

```
let headers = ["content-type": "multipart/form-data; boundary=....."]
let request = RESTRequest(path: ...., headers: headers, body: ...)
```

Ideally this would allow me to pass in the custom content-type while constructing a RESTRequest object[2], and then eventually override the default `application/json` content-type used while constructing a URLRequest object[3].

However, when the interceptors are run the AuthTokenURLInterceptor kicks and always overwrites the content-type header to set it to `application/json` [4, 5]. 

Is there a reason the AuthTokenURLInterceptor needs to set content-type? 

As a workaround, I could just create another URLInterceptor and set the Content-Type there but given that the rest of the Amplify library is set up to accept custom headers it seems redundant to do that. 

[1] https://docs.amplify.aws/lib/restapi/authz/q/platform/ios/#none
[2] https://github.com/aws-amplify/amplify-swift/blob/9d6fa8b8b9b03493932b442b0c054bcb30209385/Amplify/Categories/API/Request/RESTRequest.swift#L35-L39
[3] https://github.com/aws-amplify/amplify-swift/blob/9d6fa8b8b9b03493932b442b0c054bcb30209385/AmplifyPlugins/API/Sources/AWSAPIPlugin/Support/Utils/RESTOperationRequestUtils.swift#L61-L66
[4] https://github.com/aws-amplify/amplify-swift/blob/9d6fa8b8b9b03493932b442b0c054bcb30209385/AmplifyPlugins/API/Sources/AWSAPIPlugin/Interceptor/RequestInterceptor/AuthTokenURLRequestInterceptor.swift#L35-L36
[5] Debug details:

URLRequest in AWSRESTOperation _before_ AuthTokenURLInterceptor runs:

```
Printing description of urlRequest:
[...]

PR #3190, opened 2023-08-29
fix(api): change request interceptors applying logic

5d:
## Issue \#

https://github.com/aws-amplify/amplify-swift/issues/3184
## Description

The existing implementation of the URLRequest interceptor proceeds as follows:

1. Applying customized headers.
2. Applying library predefined interceptors.
3. Applying user-defined interceptors.

However, this approach has certain issues. For instance, one of the Amplify predefined interceptors requires the calculation of the Sigv4 authentication header for the request. This calculation involves certain headers. Making it impossible for customers to modify these header fields within their own custom interceptors.

After some discussions, we have divided the library predefined interceptors into prelude and postlude interceptors, and the interceptor logic should be as follows:

1. Applying Amplify prelude interceptors.
2. Applying customized headers.
3. Applying customer-defined interceptors.
4. Applying Amplify postlude interceptors.

This updated logic ensures that customer-defined interceptors will be applied with the expected customized headers and the postlude interceptors will operate on the correct request after customer-defined interceptors have been applied.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/API/Sources/AWSAPIPlugin/Configuration/AWSAPIEndpointInterceptors.swift
+++ b/AmplifyPlugins/API/Sources/AWSAPIPlugin/Configuration/AWSAPIEndpointInterceptors.swift
@@ -16,11 +16,16 @@ struct AWSAPIEndpointInterceptors {
     let apiAuthProviderFactory: APIAuthProviderFactory
     let authService: AWSAuthServiceBehavior?
 
-    var amplifyInterceptors: [URLRequestInterceptor] = []
-
-    var checksumInterceptors: [URLRequestInterceptor] = []
+    /// The order of interceptor decoration is as follows:
+    /// 1. **prelude interceptors**: append default auth info
+    /// 2. **cutomize headers**: headers provided when constructing the request
+    /// 3. **customer interceptors**: customer defined interceptors
+    /// 4. **postlude interceptors**: calculate checksums, signatures
+    var preludeInterceptors: [URLRequestInterceptor] = []
 
     var interceptors: [URLRequestInterceptor] = []
+
+    var postludeInterceptors: [URLRequestInterceptor] = []
 
     init(endpointName: APIEndpointName,
          apiAuthProviderFactory: APIAuthProviderFactory,
@@ -46,7 +51,7 @@ struct AWSAPIEndpointInterceptors {
         case .apiKey(let apiKeyConfig):
             let provider = BasicAPIKeyProvider(apiKey: apiKeyConfig.apiKey)
             let interceptor = APIKeyURLRequestInterceptor(apiKeyProvider: provider)
-            amplifyInterceptors.append(interceptor)
+            preludeInterceptors.append(interceptor)
         case .awsIAM(let iamConfig):
             guard let authService = authService else {
                 throw PluginError.pluginConfigurationError("AuthService is not set for IAM",
@@ -56,7 +61,7 @@ struct AWSAPIEndpointInterceptors {
             let interceptor = IAMURLRequestInterceptor(iamCredentialsProvider: provider,
                                                        region: iamConfig.region,
                                                        endpointType: endpointType)
-            checksumInterceptors.append(interceptor)
+            postludeInterceptors.append(interceptor)
         case .amazonCognitoUserPools:
             guard let authService = authService else {
                 throw PluginError.pluginConfigurationError("AuthService not set for cognito user pools",
@@ -64,7 +69,7 @@ struct AWSAPIEndpointInterceptors {
             }
             let provider = BasicUserPoolTokenProvider(authService: authService)
             let interceptor = AuthTokenURLRequestInterceptor(authTokenProvider: provider)
-            amplifyInterceptors.append(interceptor)
+            preludeInterceptors.append(interceptor)
         case .openIDConnect:
             guard let oidcAuthProvider = apiAuthProviderFactory.oidcAuthProvider() else {
                 throw PluginError.pluginConfigurationError("AuthService not set for OIDC",
@@ -72,7 +77,7 @@ struct AWSAPIEndpointInterceptors {
             }
             let wrappedAuthProvider = AuthTokenProviderWrapper(tokenAuthProvider: oidcAuthProvider)
             let interceptor = AuthTokenURLRequestInterceptor(authTokenProvider: wrappedAuthProvider)
-            amplifyInterceptors.append(interceptor)
+            preludeInterceptors.append(interceptor)
         case .function:
             guard let functionAuthProvider = apiAuthProviderFactory.functionAuthProvider() else {
                 throw PluginError.pluginConfigurationError("AuthService not set for function auth",
@@ -80,7 +85,7 @@ struct AWSAPIEndpointInterceptors {
             }
             let wrappedAuthProvider = AuthTokenProviderWrapper(tokenAuthProvider: functionAuthProvider)
             let interceptor = AuthTokenURLRequestInterceptor(authTokenProvider: wrappedAuthProvider)
-            amplifyInterceptors.append(interceptor)
+            preludeInterceptors.append(interceptor)
         }
     }
 }
--- a/AmplifyPlugins/API/Sources/AWSAPIPlugin/Configuration/AWSAPIEndpointInterceptors.swift
+++ b/AmplifyPlugins/API/Sources/AWSAPIPlugin/Configuration/AWSAPIEndpointInterceptors.swift
@@ -16,6 +16,10 @@ struct AWSAPIEndpointInterceptors {
     let apiAuthProviderFactory: APIAuthProviderFactory
     let authService: AWSAuthServiceBehavior?
 
+    var amplifyInterceptors: [URLRequestInterceptor] = []
+
+    var checksumInterceptors: [URLRequestInterceptor] = []
+
     var interceptors: [URLRequestInterceptor] = []
 
     init(endpointName: APIEndpointName,
@@ -42,7 +46,7 @@ struct AWSAPIEndpointInterceptors {
         case .apiKey(let apiKeyConfig):
             let provider = BasicAPIKeyProvider(apiKey: apiKeyConfig.apiKey)
             let interceptor = APIKeyURLRequestInterceptor(apiKeyProvider: provider)
-            addInterceptor(interceptor)
+            amplifyInterceptors.append(interceptor)
         case .awsIAM(let iamConfig):
             guard let authService = authService else {
                 throw PluginError.pluginConfigurationError("AuthService is not set for IAM",
@@ -52,7 +56,7 @@ struct AWSAPIEndpointInterceptors {
             let interceptor = IAMURLRequestInterceptor(iamCredentialsProvider: provider,
                                                        region: iamConfig.region,
                                                        endpointType: endpointType)
-            addInterceptor(interceptor)
+            checksumInterceptors.append(interceptor)
         case .amazonCognitoUserPools:
             guard let authService = authService else {
                 throw PluginError.pluginConfigurationError("AuthService not set for cognito user pools",
@@ -60,7 +64,7 @@ struct AWSAPIEndpointInterceptors {
             }
             let provider = BasicUserPoolTokenProvider(authService: authService)
             let interceptor = AuthTokenURLRequestInterceptor(authTokenProvider: provider)
-            addInterceptor(interceptor)
+            amplifyInterceptors.append(interceptor)
         case .openIDConnect:
             guard let oidcAuthProvider = apiAuthProviderFactory.oidcAuthProvider() else {
                 throw PluginError.pluginConfigurationError("AuthService not set for OIDC",
@@ -68,7 +72,7 @@ struct AWSAPIEndpointInterceptors {
             }
             let wrappedAuthProvider = AuthTokenProviderWrapper(tokenAuthProvider: oidcAuthProvider)
             let interceptor = AuthTokenURLRequestInterceptor(authTokenProvider: wrappedAuthProvider)
-            addInterceptor(interceptor)
+            amplifyInterceptors.append(interceptor)
         case .function:
             guard let functionAuthProvider = apiAuthProviderFactory.functionAuthProvider() else {
                 throw PluginError.pluginConfigurationError("AuthService not set for function auth",
@@ -76,7 +80,7 @@ struct AWSAPIEndpointInterceptors {
             }
             let wrappedAuthProvider = AuthTokenProviderWrapper(tokenAuthProvider: functionAuthProvider)
             let interceptor = AuthTokenURLRequestInterceptor(authTokenProvider: wrappedAuthProvider)
-            addInterceptor(interceptor)
+            amplifyInterceptors.append(interceptor)
         }
     }
 }
--- a/AmplifyPlugins/API/Sources/AWSAPIPlugin/Configuration/AWSAPIEndpointInterceptors.swift
+++ b/AmplifyPlugins/API/Sources/AWSAPIPlugin/Configuration/AWSAPIEndpointInterceptors.swift
@@ -9,6 +9,14 @@ import Amplify
 import Foundation
 import AWSPluginsCore
 
+/// The order of interceptor decoration is as follows:
+/// 1. **prelude interceptors**
+/// 2. **cutomize headers**
+/// 3. **customer interceptors**
+/// 4. **postlude interceptors**
+///
+/// **Prelude** and **postlude** interceptors are used by library maintainers to
+/// integrate essential functionality for a variety of authentication types.
 struct AWSAPIEndpointInterceptors {
     // API name
     let apiEndpointName: APIEndpointName
@@ -16,11 +24,6 @@ struct AWSAPIEndpointInterceptors {
     let apiAuthProviderFactory: APIAuthProviderFactory
     let authService: AWSAuthServiceBehavior?
 
-    /// The order of interceptor decoration is as follows:
-    /// 1. **prelude interceptors**: append default auth info
-    /// 2. **cutomize headers**: headers provided when constructing the request
-    /// 3. **customer interceptors**: customer defined interceptors
-    /// 4. **postlude interceptors**: calculate checksums, signatures
     var preludeInterceptors: [URLRequestInterceptor] = []
 
     var interceptors: [URLRequestInterceptor] = []
--- a/AmplifyPlugins/API/Sources/AWSAPIPlugin/Interceptor/RequestInterceptor/IAMURLRequestInterceptor.swift
+++ b/AmplifyPlugins/API/Sources/AWSAPIPlugin/Interceptor/RequestInterceptor/IAMURLRequestInterceptor.swift
@@ -36,7 +36,6 @@ struct IAMURLRequestInterceptor: URLRequestInterceptor {
             throw APIError.unknown("Could not get host from mutable request", "")
         }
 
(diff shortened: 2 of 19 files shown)


AmplifyPlugins/API/Sources/AWSAPIPlugin/Support/Utils/RESTOperationRequestUtils.swift
```diff
@@ -57,19 +57,31 @@ final class RESTOperationRequestUtils {
...
-        if let headers = headers {
-            for (key, value) in headers {
-                requestHeaders[key] = value
-            }
-        }
-        baseRequest.allHTTPHeaderFields = requestHeaders
+        baseRequest = applyCustomizeRequestHeaders(
+            baseHeaders.merging(headers ?? [:], uniquingKeysWith: { _, new in new }),
+            on: baseRequest
+        )
```
lawmicha, line 65: in this flow (construct -> intercept -> apply customized header), do we still need to apply the `request.headers` here?

AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSRESTOperation.swift
```diff
@@ -116,6 +116,9 @@ final public class AWSRESTOperation: AmplifyOperation<
                 }
             }
 
+            // apply customized request headers
```
lawmicha, line 119:
```suggestion
            // The headers from the request object should override any of the same header modifications done by the intercepters above
```

atierian (changes requested):
> The customized request headers provided during the creation of the RESTRequest should take precedence over interceptors.

This is already happening here before the change.
```swift
var requestHeaders = ["content-type": "application/json"]
if let headers = headers { 
    for (key, value) in headers { // if there's a "content-type" key here
        requestHeaders[key] = value  // it will overwrite the default "application/json" value here
    }
}
```

My understanding is that we need to change...
- [IAMURLRequestInterceptor](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/API/Sources/AWSAPIPlugin/Interceptor/RequestInterceptor/IAMURLRequestInterceptor.swift#L39) 
- [AuthTokenURLRequestInterceptor](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/API/Sources/AWSAPIPlugin/Interceptor/RequestInterceptor/AuthTokenURLRequestInterceptor.swift#L35-L36) 

... to give precedence to callsite supplied headers.

Am I missing something?

5d:
> > The customized request headers provided during the creation of the RESTRequest should take precedence over interceptors.
> 
> This is already happening here before the change.
> 
> ```swift
> var requestHeaders = ["content-type": "application/json"]
> if let headers = headers { 
>     for (key, value) in headers { // if there's a "content-type" key here
>         requestHeaders[key] = value  // it will overwrite the default "application/json" value here
>     }
> }
> ```
> 
> My understanding is that we need to change...
> 
> * [IAMURLRequestInterceptor](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/API/Sources/AWSAPIPlugin/Interceptor/RequestInterceptor/IAMURLRequestInterceptor.swift#L39)
> * [AuthTokenURLRequestInterceptor](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/API/Sources/AWSAPIPlugin/Interceptor/RequestInterceptor/AuthTokenURLRequestInterceptor.swift#L35-L36)
> 
> ... to give precedence to callsite supplied headers.
> 
> Am I missing something?

The logic to build the URLRequest for RESTRequest happends [here](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSRESTOperation.swift#L97-L108). In my opinion, as the interceptors are simply adjust the request to align with their specific requirements, we just need to ensure the customize headers are applied after applying all the interceptors.

AmplifyPlugins/API/Sources/AWSAPIPlugin/Support/Utils/RESTOperationRequestUtils.swift
5d: As [this method](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/API/Sources/AWSAPIPlugin/Support/Utils/RESTOperationRequestUtils.swift#L55-L58) signature implicates, I think we still need to apply the headers.
```
static func constructURLRequest(with url: URL,
                                    operationType: RESTOperationType,
                                    headers: [String: String]?,
                                    requestPayload: Data?) -> URLRequest
```

sandeepklr:
> In my opinion, as the interceptors are simply adjust the request to align with their specific requirements, we just need to ensure the customize headers are applied after applying all the interceptors.

For my edification, can you point to some external documentation that says you need to set a `content-type: application/json header` to use IAM auth or Auth token?

There is probably a requirement to set A content type, at which point would it be cleaner for the Interceptors to check if one has not already been set and then set it based on their individual requirements? Interceptors are passed in a URLRequest object, so should be easy to do. 

By introducing another method to set custom headers fields, there's multiple places in this code where header fields could be set or overwritten, which makes it hard to reason or read. 

Thoughts?

5d:
> > In my opinion, as the interceptors are simply adjust the request to align with their specific requirements, we just need to ensure the customize headers are applied after applying all the interceptors.
> 
> For my edification, can you point to some external documentation that says you need to set a `content-type: application/json header` to use IAM auth or Auth token?


I don't think IAM or auth token need to be in json format. I beleive the `Content-Type` header was added as default just because all Amplify generated APIs are using json.

> 
> There is probably a requirement to set A content type [citation needed], at which point would it be cleaner for the Interceptors to check if one has not already been set and then set it based on their individual requirements? Interceptors are passed in a URLRequest object, so should be easy to do. Concretely, something like this:
> 
> ```
> if  mutableRequest.value(forHTTPHeaderField: "content-type") == nil {
>         mutableRequest.setValue(URLRequestConstants.ContentType.applicationJson,
>                                 forHTTPHeaderField: URLRequestConstants.Header.contentType)
> }
> ```
> 
> By introducing another method to set custom headers fields like this PR does, there's multiple places in this code where header fields could be set or overwritten, which makes it hard to reason or read.
> 
> Thoughts?

The new method is internal method. The key point of the PR is we respect the headers created with `RESTRequest` a higher priority than interceptors.

sandeepklr:
> The new method is internal method. The key point of the PR is we respect the headers created with RESTRequest a higher priority than interceptors.

While this might be an internal method, correct me if I'm wrong- doesn't this have ramifications of the behavior of the library.

Previously you set a "default" value while creating the RESTRequest object, while letting a later user-defined interceptor to dynamically change the value of the header. So something like this: 

```
let request = RESTRequest(path: "/path", ["
       "customHeader": "default-value"
"] )

....

struct CustomInterceptor: URLRequestInterceptor {
    func intercept(_ request: URLRequest) throws -> URLRequest {
           ....
           if (conditionIsTrue) {
                request.setValue("customHeader", "new-value")
           }
    }
```

Now the behavior of the library is changing so that it the value of "customHeader" is going to be set to "default-value", as opposed to "new-value" previously. 

At the very least this needs a documentation change to call out this new behavior, and perhaps even a MV bump  (I'm new to this library so I'm not sure what constitutes a MV bump). 

Just my 2c.

5d:
Thanks for your example @sandeepklr. I got your point. You are right. The current change could  impact current users who rely on custom header fields to inform dynamic decisions within their user interceptors.

The desired request decoration process should follow this sequence:
1. Build the URLRequest
2. Apply interceptors provided by library
3. Apply user-defined interceptors

When we reach step 3, the custom headers should already be integrated into the URLRequest.
One potential solution I'm thinking is creating an additional library interceptor `ApplyCustomizeHeadersInterceptor`. This new interceptor would handle the application of custom headers and execute as the final step within the step 2. Therefore, the user-defined interceptors will remain unaffected, and library interceptors won't override the customized headers.

sandeepklr:
Apologies in advance if I'm misunderstanding this....

While that may work for AuthTokenURLInterceptor, one thing to consider is if that breaks IAMURLRequestInterceptor? 

According to SigV4 specification[1], Content-Type header must included while computing signature, if present. With your approach if someone did the following while using IAM Auth (which is actually what I'm doing in my original bug report sans IAM Auth):

```
let request = RESTRequest(path: "/path", ["
       "Content-Type": "multipart/form-data; boundary=......"
"] )
```

Assuming the library signs Content-Type header, IAMURLRequestInterceptor would sign with `Content-Type: application/json`, while the `ApplyCustomizeHeadersInterceptor` would change the request to `Content-Type": "multipart/form-data; boundary=......` causing it to fail? 

I haven't peeled enough layers of the library back to confirm the following:

1. Does the library include Content-Type while SigV4 signing.
2. Is there an integ test to repro changing Content-Type header after Sigv4 signing, and does the request fail.

-------------


[1] https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html#:~:text=If%20the%20Content%2DType%20header%20is%20present%20in%20the%20request%2C%20you%20must%20add%20it%20to%20the%20CanonicalHeaders%20list.

sandeepklr:
More and more, I think the fact that the IAMURLRequestInterceptor and AuthTokenURLRequestInterceptor setting Content-Type is more headache than it is worth and not intuitive. They shouldn't need to care about the Content-Type, they should only care about the headers that they "own" (Authorization, x-Amz-Date, are examples of headers they could own). Content-Type is not a header they should own and blindly set.

If the library cares about setting a default Content-Type, it should set it before the interceptors run, which it already is based on an earlier comment[1].


[1] https://github.com/aws-amplify/amplify-swift/pull/3190#pullrequestreview-1602693144

5d:
> More and more, I think the fact that the IAMURLRequestInterceptor and AuthTokenURLRequestInterceptor setting Content-Type is more headache than it is worth and not intuitive. They shouldn't need to care about the Content-Type, they should only care about setting the headers that they "own" (Authorization, x-Amz-Date, are examples of headers they could own). Content-Type is not a header they should own and blindly set.
> 
> If the library cares about setting a default Content-Type, it should set it before the interceptors run, which it already is based on an earlier comment[1].
> 
> If the user chooses to change a signed Header value in their custom interceptor later, that is on the user and it's at least not the library doing something wrong. Ideally there'd be a way to detect that and emit a warning log to the user, but whatever.
> 
> [1] [#3190 (review)](https://github.com/aws-amplify/amplify-swift/pull/3190#pullrequestreview-1602693144)

Yep, you and @atierian are correct. I didn't know the sigv4 calculate the signature including the `Content-Type` header. We should have customized header override before calculating the sigv4. I will revert the changes and just remove the logic of setting the `Content-Type` in library provided interceptors.

atierian (approved):

AmplifyPlugins/API/Sources/AWSAPIPlugin/Configuration/AWSAPIEndpointInterceptors.swift
```diff
@@ -16,8 +16,17 @@ struct AWSAPIEndpointInterceptors {
     let apiAuthProviderFactory: APIAuthProviderFactory
     let authService: AWSAuthServiceBehavior?
 
+    /// The order of interceptor decoration is as follows:
```
sandeepklr, line 19: The new prelude / postlude breakdown is great! 

Just a heads up, I think this new order is not b/w compatible with previous order: 

Previously it was:

1/ custom request headers
2/ library defined interceptors (now broken up into prelude and postlude) 
3/ user defined interceptors

Now it's 

1/ prelude interceptors
2/ custom request headers
3/ user defined interceptors
4/ postlude interceptors

Previously prelude interceptors would run after setting custom headers on request.  Now they'll run after running prelude interceptors. 

If this is not a concern, please feel free to disregard.

5d: @sandeepklr Thanks for reviewing.

Yes, the behavior would only change if customer is either:

1. Customizing the headers that prelude interceptors are modifying: 
  - In this case, the prelude interceptors primarily decorate the URLRequest by adding specific `x-amz-*` headers to facilitate the proper functioning of the Amplify API plugin with APIGateway. Customers should not predefine these headers when constructing the REST request.

2. Overriding the headers in customer interceptor that postlude interceptors are modifying:
  - In this case, the postlude interceptor is responsible for computing the signature to ensure the integrity of the URLRequest. If the customer calculates the signature within their interceptor, it will be overwritten by the postlude interceptor.

It's worth noting that, in both cases, we maintain the `User-Agent` header for analytics purposes.

AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSGraphQLOperation.swift
```diff
@@ -87,48 +127,64 @@ final public class AWSGraphQLOperation<R: Decodable>: GraphQLOperation<R> {
...
+    private func getEndpointInterceptors(from request: GraphQLOperationRequest<R>) -> Result<AWSAPIEndpointInterceptors?, APIError> {
+        getEndpointConfig(from: request).flatMap { endpointConfig in
+            do {
+                if let pluginOptions = request.options.pluginOptions as? AWSPluginOptions,
+                   let authType = pluginOptions.authType
+                {
+                    return .success(try pluginConfig.interceptorsForEndpoint(
+                        withConfig: endpointConfig,
+                        authType: authType
+                    ))
```
lawmicha, line 162: No action needed, just some context about this code. This is what provides the ability for DataStore to support models with multi-auth, by allowing DataStore to specific toggle between authType at runtime to attempt the synchronization calls.

When developers use API plugin directly, the configuration's authType is currently the only way to control what's in the interceptors. Developers have to duplicate the api sections in the configuration and change authType to control which whether default interceptor is in prelude/postlude, ie. a separate section for API_KEY, COGNITO_USER_POOL, with a key for each section, the key is then used as the apiName in the request object.

This is a huge customer pain point, especially Amplify CLI may also revert the changes to amplifyconfiguration.json. A runtime solution can potentially be to expose it differently as a public API or document the current mechanism. Solving this however will require some additional thoughts on customer defined interceptors. This code re-creates the interceptors based on authType, which creates a new struct, losing the customer's added interceptors since this is a mutating func on the existing config object. The other customer defined interceptors, such as the lambda or OIDC interceptors should continue to work as expected.

AmplifyPlugins/API/Sources/AWSAPIPlugin/Configuration/AWSAPIEndpointInterceptors.swift
```diff
@@ -16,8 +16,17 @@ struct AWSAPIEndpointInterceptors {
...
+    /// 1. **prelude interceptors**: append default auth info
+    /// 2. **cutomize headers**: headers provided when constructing the request
+    /// 3. **customer interceptors**: customer defined interceptors
```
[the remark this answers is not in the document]
5d: The reasoning behind this is that customers should anticipate seeing construction-time customized headers within their interceptors.

AmplifyPlugins/API/Sources/AWSAPIPlugin/Configuration/AWSAPIEndpointInterceptors.swift
```diff
@@ -16,8 +16,17 @@ struct AWSAPIEndpointInterceptors {
...
+    /// 4. **postlude interceptors**: calculate checksums, signatures
+    var preludeInterceptors: [URLRequestInterceptor] = []
+
     var interceptors: [URLRequestInterceptor] = []
 
+    var postludeInterceptors: [URLRequestInterceptor] = []
```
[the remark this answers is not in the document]
5d: Yes, I think the interceptor is a common concept for HTTP libraries. It's defined as array. 

Concepturally, I think singleton array `[Interceptor]` is more suitable for prelude and postlude interceptors than optional `Interceptor?`. Both can represent empty and existence, however, array is more flexible.

AmplifyPlugins/API/Sources/AWSAPIPlugin/Configuration/AWSAPIEndpointInterceptors.swift
```diff
@@ -16,8 +16,17 @@ struct AWSAPIEndpointInterceptors {
     let apiAuthProviderFactory: APIAuthProviderFactory
     let authService: AWSAuthServiceBehavior?
 
+    /// The order of interceptor decoration is as follows:
+    /// 1. **prelude interceptors**: append default auth info
+    /// 2. **cutomize headers**: headers provided when constructing the request
+    /// 3. **customer interceptors**: customer defined interceptors
+    /// 4. **postlude interceptors**: calculate checksums, signatures
+    var preludeInterceptors: [URLRequestInterceptor] = []
```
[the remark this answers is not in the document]
5d: Thanks, I will update the comment. 

Since the interceptor is mutating the original URLRequest, the applying order is important. That's the key point of this PR. The name `prelude` and `postlude` is to emphsize that this is solely related to the sequence in which interceptors are applied. The comments following each interceptor stage are merely examples, and I plan to remove them to simplify understanding. Our main focus should be on highlighting the interceptor order.

AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSRESTOperation.swift
```diff
@@ -126,6 +101,70 @@ final public class AWSRESTOperation: AmplifyOperation<
...
+    private func applyInterceptor(_ interceptor: URLRequestInterceptor, request: URLRequest) async -> Result<URLRequest, APIError> {
+        do {
+            return .success(try await interceptor.intercept(request))
+        } catch let error as APIError {
+            return .failure(error)
+        } catch {
+            return .failure(
+                APIError.operationError(
+                    "Failed to intercept request fully.",
+                    "Something wrong with the interceptor",
```
lawmicha, line 164: can we provide more info from the error description and recovery message? such as `type(of: interceptor)` or `error.errorDescription` in the error description
```suggestion
                    "Failed to intercept request with \(type(of: interceptor)). Error message: \(error.errorDescription).",
                    "See underlying error for more details",
```

AmplifyPlugins/API/Sources/AWSAPIPlugin/Interceptor/RequestInterceptor/IAMURLRequestInterceptor.swift
```diff
@@ -36,7 +36,6 @@ struct IAMURLRequestInterceptor: URLRequestInterceptor {
             throw APIError.unknown("Could not get host from mutable request", "")
         }
 
-        request.setValue(URLRequestConstants.ContentType.applicationJson, forHTTPHeaderField: URLRequestConstants.Header.contentType)
```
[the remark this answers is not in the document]
5d: This wouldn't pose a problem because we've set it up during the construction of [the URLRequest](https://github.com/aws-amplify/amplify-swift/pull/3190/files#diff-723cc1c8e740c666ddf72add879d38a32dc87851fb68147df7c807cceb7afb6dR61).

AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSRESTOperation.swift
5d: sure, I was just copying the previous code error messages. We definitly can add more info here.

lawmicha (approved):

atierian (approved):

PR #3191, opened 2023-08-30
Revert "feat(auth): add userID property to AuthSignUpResult"

atierian:
Reverts aws-amplify/amplify-swift#3179


atierian:
Reverting this merge temporarily so that we can merge and release https://github.com/aws-amplify/amplify-swift/pull/3189 first.

Issue #3100, opened 2023-07-21
DataStore observeQuery is slow unless AWSAPIPlugin is added


chernushevich:
### Describe the bug

DataStore observeQuery takes anywhere between 1 and 3 seconds to return new snapshot. Unless AWSAPIPlugin is added, then it works instantly

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Follow the tutorial at https://docs.amplify.aws/start/q/integration/ios/, do NOT add AWSAPIPlugin
2. Add Amplify.DataStore.observeQuery to monitor Todos
3. Add a new Todo with Amplify.DataStore.save
4. New items take anywhere between 1 and 3 seconds to appear, visible delays
5. Add and configure AWSAPIPlugin
6. The exact same code now adds new items instantly
```

### Expected behavior

Amplify.DataStore.observeQuery returns new items instantly in offline mode

### Amplify Framework Version

2.15.0

### Amplify Categories

API, DataStore

### Dependency manager

Swift PM

### Swift version

5.9

### CLI version

12.2.0

### Xcode version

15.0 Beta 4 (15A5195m)

### Relevant log output

```shell
<details>
<summary>Log Messages</summary>


INSERT LOG MESSAGES HERE
```
</details>
```


### Is this a regression?

No

### Regression additional context

_No response_

### Platforms

_No response_

### OS Version

macOS Ventura 13.4.1

### Device

iPhone Xr

### Specific to simulators

_No response_

### Additional context

_No response_

chernushevich:
Calling `Amplify.DataStore.stop()` right before Amplify.DataStore.save fixes the issue.

atierian:
We've opened a PR to address this.
- https://github.com/aws-amplify/amplify-swift/pull/3214

We'll follow up to let you know once it's released.
Thanks for reporting the issue @chernushevich!

atierian:
The fix for this was released in [2.25.1](https://github.com/aws-amplify/amplify-swift/releases/tag/2.25.1). Thanks again for reporting this and for your patience!

PR #3214, opened 2023-09-08
fix(datastore): observequery local only collect

atierian:
## Issue \#

- https://github.com/aws-amplify/amplify-swift/issues/3100

## Description

`DataStore.observeQuery(...)` has some internal logic to batch snapshots received based on total count or time elapsed while syncing. This logic is applied when `DataStore` thinks it is hasn't yet completed syncing. 

When `DataStore` is in local only mode (i.e. when `AWSAPIPlugin` isn't being used), it is never "completed syncing", leading to the same throttling logic.

This change ensures that the throttling / collect logic is **not** applied when `DataStore` is in local only mode.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Subscribe/DataStoreObserveQueryOperation.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Subscribe/DataStoreObserveQueryOperation.swift
@@ -211,6 +211,13 @@ class ObserveQueryTaskRunner<M: Model>: InternalTaskRunner, InternalTaskAsyncThr
             })
     }
 
+    /// If the `storageEngine` does sync from remote, the initial batch should
+    /// collect snapshots based on time / snapshots received.
+    /// If it doesn't, it should publish each snapshot without waiting.
+    private var shouldBatchDuringSync: Bool {
+        (storageEngine as? StorageEngine)?.syncsFromRemote ?? true
+    }
+
     // MARK: Observe item changes
 
     /// Subscribe to item changes with two subscribers (During Sync and After Sync). During Sync, the items are filtered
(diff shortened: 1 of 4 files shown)


ruisebas (dismissed):

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Subscribe/DataStoreObserveQueryOperation.swift
```diff
@@ -211,6 +211,13 @@ class ObserveQueryTaskRunner<M: Model>: InternalTaskRunner, InternalTaskAsyncThr
             })
     }
 
+    /// If the `storageEngine` does sync from remote, the initial batch should
+    /// collect snapshots based on time / snapshots received.
+    /// If it doesn't, it should publish each snapshot without waiting.
+    private var shouldBatchDuringSync: Bool {
+        (storageEngine as? StorageEngine)?.syncsFromRemote ?? true
```
lawmicha, line 218: would it make sense to add the computed property `syncsFromRemote` to `StorageEngineBehavior` (potentially a different name like `syncEnabled`)? Right now, using a mock would fail the `storageEngine as? StorageEngine` check, right?

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Subscribe/DataStoreObserveQueryOperation.swift
```diff
@@ -222,12 +229,19 @@ class ObserveQueryTaskRunner<M: Model>: InternalTaskRunner, InternalTaskAsyncThr
...
                 .filter { _ in !self.dispatchedModelSyncedEvent.get() }
                 .filter(self.filterByModelName(mutationEvent:))
                 .filter(self.filterByPredicateMatch(mutationEvent:))
                 .handleEvents(receiveOutput: self.onItemChangeDuringSync(mutationEvent:) )
-                .collect(.byTimeOrCount(self.serialQueue, self.itemsChangedPeriodicPublishTimeInSeconds, self.itemsChangedMaxSize))
+                .collect(
+                    .byTimeOrCount(
+                        self.serialQueue,
+                        self.itemsChangedPeriodicPublishTimeInSeconds,
+                        self.shouldBatchDuringSync ? self.itemsChangedMaxSize : 1
```
lawmicha, line 242: perhaps we can open up an issue on the other DataStore repos to let them verify if they have the same behavior, and if so they can do the similar change here

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Subscribe/DataStoreObserveQueryOperation.swift
atierian: Unless the mock was a `StorageEngine` subclass - which isn't possible as `StorageEngine` is marked `final` - it would fail to cast in that case.

Adding it as a requirement to `StorageEngineBehavior` makes sense.

atierian: Addressed in [dd0e0be](https://github.com/aws-amplify/amplify-swift/pull/3214/commits/dd0e0bed6a61032aefa913abf7a05d236b7acf3f)

lawmicha (approved):

PR #3224, opened 2023-09-15
fix(datastore): Continue initial sync if atleast one model syncs successfully and other models fail due to Unauthorized API error

thisisabhash:
## Issue \#

[3136](https://github.com/aws-amplify/amplify-swift/issues/3136)

## Description

Initial Sync should succeed if atleast one model sync succeeds. In this PR, additional checks are added to catch Unauthorized APIError received for some models during Initial sync and return success.

## General Checklist

- [x] All unit tests pass


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -164,7 +164,16 @@ final class AWSInitialSyncOrchestrator: InitialSyncOrchestrator {
             allMessages.joined(separator: "\n"),
             underlyingError
         )
-        return .failure(syncError)
+        
+        // send success if atleast one model succeeded
+        let syncableModelsCount = ModelRegistry.modelSchemas.filter { $0.isSyncable }.count
+        if syncableModelsCount == syncErrors.count {
+            return .failure(syncError)
```
lawmicha, line 171: This seems like a reasonable change but thinking about this more, and looking at the start of this method, appears that this change will excessively return success. Let's think about the cascading impact of this change. Before we would fail fast, and DataStore will "re-sync" by attemping the entire sync lifecycle on the next DataStore operation. By not failing fast, eventual consistency will be much longer. 

Instead of this change, is it because we are getting an unauthorized error but not checking it properly here?
```
if syncErrors.isEmpty || syncErrors.allSatisfy(isUnauthorizedError) {
            return .successfulVoid
        }
```

In the example https://github.com/aws-amplify/amplify-swift/issues/3136 what is the sync error instance, why doesn't it return success at line 152?

thisisabhash: The bug happens when there is no user signed in and a `Guest` instance is saved locally. 
Now, there are two syncable models : `Guest` and `User`.
`User` gets `onReceiveCompletion()` with error result with `Unauthorized error` and append result to `syncErrors` array.
`Guest` gets `onReceiveCompletion()` with success.

When `makeCompletionResult()` is called - it returns `.failure()` and the `Guest` instance doesn't sync at all.

lawmicha: why does `makeCompletionResult()` return `.failure()` if the syncing User contributes to one syncError which should satisfy `syncErrors.allSatisfy(isUnauthorizedError)` ?

thisisabhash: thanks for mentioning that. i have updated the logic for checking unauthorized error.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/Support/APIError+Unauthorized.swift
```diff
@@ -0,0 +1,62 @@
...
+    /// ```
+    ///
+    /// - Parameter rule: Used to determine if the `APIError` is an unauthorized error.
+    /// - Returns: `true` if unauthorized error, `false` otherwise
+    public func isUnauthorized(rule: UnauthorizedDetermining = .default) -> Bool {
+        rule.isUnauthorized(self)
+    }
+
+    public struct UnauthorizedDetermining {
+        let isUnauthorized: (APIError) -> Bool
```
lawmicha, line 35: Look like this came from https://github.com/aws-amplify/amplify-swift/pull/1994/files#diff-17539a8fec2423e2a770e4dff07c90e0f2e9a2678838f2c1ed139396b9ddcfad I'm trying to recall why we added this and it was because adding a new enum case to APIError was going to be a breaking change. 

I don't recall, did we decide at some point with Auth or Analytics that adding a new enum error case is no longer a major version bump? If this is the case, we can process the response within APIPlugin and directly return `APIError.unauthorized` error. This will simply things greatly

lawmicha: So according to https://github.com/aws-amplify/amplify-swift#semantic-versioning-and-enumeration-cases adding APIError.unauthorized is going to be a minor version bump. I think we should decouple this additon to what you're trying to achieve here since adding a new case will require further discussion, perhaps the unauthorized case should be returned as an underlying AuthError in an APIError (one example of an alternative). 

We don't need all the public interfaces extending APIError (what happens when developers include DataStore and APIPlugin? which public `isUnauthorized` should they use?). The minimum change you will need here is just check if APIError is 401/403/"unauthorized" in operationError, which should all be internal

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/InitialSync/InitialSyncOrchestrator.swift
```diff
@@ -245,6 +246,15 @@ extension AWSInitialSyncOrchestrator {
            case .unauthorized = AppSyncErrorType(errorTypeValue) {
             return true
         }
+        
+        // Check is API error is of unauthorized type
+        if case let .api(amplifyError, _) = datastoreError,
+           let apiError = amplifyError as? APIError {
+            if apiError.isUnauthorized() {
+                return true
+            }
```
lawmicha, line 255: Here, instead of calling isUnauthorized, can we just check `.httpStatus` and `.operationError`? for httpStatus -> 401/403 return true, for operationError -> errorDescription contains "unauthorized" return true

Package.resolved
```diff
@@ -14,8 +14,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/aws-amplify/aws-appsync-realtime-client-ios.git",
       "state" : {
-        "revision" : "c7ec93dcbbcd8abc90c74203937f207a7fcaa611",
-        "version" : "3.1.1"
+        "revision" : "b036e83716789c13a3480eeb292b70caa54114f2",
+        "version" : "3.1.0"
       }
```
lawmicha, line 19: looks like an accidental version downgrade

AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Sync/InitialSync/InitialSyncOrchestratorTests.swift
```diff
@@ -201,6 +202,112 @@ class InitialSyncOrchestratorTests: XCTestCase {
...
+    ///    - It invokes completion callback with success
+    func testInvokesCompletionCallbackWhenUnauthorizedAPIError() async throws {
+        ModelRegistry.reset()
+        PostCommentModelRegistration().registerModels(registry: ModelRegistry.self)
+        let responder = QueryRequestListenerResponder<PaginatedList<AnyModel>> { request, listener in
+            if request.document.contains("SyncPosts") {
+                let urlResponse = HTTPURLResponse(url: URL(string: "https://aws.amazon.com")!,
+                                                 statusCode: 401,
+                                                 httpVersion: nil,
+                                                 headerFields: nil)!
```
lawmicha, line 220: i think this is good for end-to-end unit testing of the SyncOperation, but this PR adds 3 new checks for categorizing `isUnauthorized`, so there should be 3 APIError states to test. To make it more feasible in testing, the method `isUnauthorizedError` can be made internal, and then the unit test can iterate over an array of APIError types. Ideally, this should be extracted out into its own component and reused across other APIError handling logic in DataStore.

lawmicha (approved):

PR #3230, opened 2023-09-20
fix(core): add Foundation HTTP client for watchOS / tvOS

atierian:
## General Checklist

- [ ] All unit tests pass


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
```diff
@@ -110,6 +120,11 @@ extension AWSCognitoAuthPlugin {
                 frameworkMetadata: AmplifyAWSServiceConfiguration.frameworkMetaData(),
                 region: identityPoolConfig.region
             )
+
+            #if os(watchOS) || os(tvOS)
+            configuration.httpClientEngine = FoundationClientEngine()
```
ruisebas, line 125: I now wonder if we should be using the `httpClientEngineProxy` here as well, but that's a story for another time.

ruisebas (approved):

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
```diff
@@ -92,9 +93,18 @@ extension AWSCognitoAuthPlugin {
             )
 
             if var httpClientEngineProxy = httpClientEngineProxy {
-                let sdkEngine = configuration.httpClientEngine
-                httpClientEngineProxy.target = sdkEngine
+                let httpClientEngine: HttpClientEngine
+                #if os(watchOS) || os(tvOS)
```
phantumcode, line 97: nit: should we consider reversing the if/else logic here so that additional platforms don't need to be added in the future?  i.e. `#if os(iOS) || os(macOS)`

atierian: I like it. I was looking at it from the angle of _we know watchOS and tvOS should use `Foundation`_, but it's more correct to say _we only know that the Swift SDK / CRT supports iOS and macOS_.

atierian: On second thought, that works great when we're composing it with an existing `HttpClientEngine`, but gets a little messy when we're using it solo (e.g. a few lines down on 105). In those cases, we'd have to:
1. use `#if os(watchOS) || os(tvOS)`
2. use `#if !os(iOS) && !os(macOS)`
3. 
```swift
#if os(iOS) && os(macOS)
#else
// inject FoundationClientEngine
#endif
```

1. Doesn't solve the issue you mentioned. 2. is difficult to read. 3. is a little confusing, but probably the best option.

What do you think?

phantumcode: I like both options 2 or 3 and will defer to you pick your preference.

atierian: Going with 3. because `if !foo && !bar` conditions make my brain hurt.

Addressed in [c9b6f3b](https://github.com/aws-amplify/amplify-swift/pull/3230/commits/c9b6f3bbff5e2b97cc815f86b10105f75e76dad2). Thanks for the suggestion!

phantumcode (approved):

PR #3261, opened 2023-10-02
fix(analytics): update error handling

atierian:
## Swift SDK Update 0.26.0

Makes necessary changes to get AWSPluginsCore to compile.

## Debt Introduced
- Removed FrameworkMetadata from client configuration. Need to tackle this with user-agent / client engine proxy PR. Tracking on base PR.
- [// TODO: Should some of these really be retried?](https://github.com/aws-amplify/amplify-swift/pull/3261/files#diff-a2f40f4e14e09ea4b5345796410df0323b4df411e720670fca5025818b9b2bd8R12)


AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/Support/Utils/AnalyticsErrorHelper.swift
```diff
@@ -5,22 +5,18 @@
...
 
-import Amplify
-import AWSPinpoint
-import ClientRuntime
 import Foundation
+import Amplify
+import AwsCommonRuntimeKit
 
-class AnalyticsErrorHelper {
+enum AnalyticsErrorHelper {
```
phantumcode, line 12: curious why this was changed from class to enum?

phantumcode (approved):

AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Support/Utils/PinpointRequestsRegistry.swift
```diff
@@ -70,7 +70,11 @@ private struct CustomPinpointHttpClientEngine: HttpClientEngine {
         let currentUserAgent = headers.value(for: userAgentHeader) ?? ""
         headers.update(name: userAgentHeader,
                        value: "\(currentUserAgent)\(userAgentSuffix)")
-        request.headers = headers
+        for header in headers.headers {
+            for value in header.value {
+                request.withHeader(name: header.name, value: value)
```
ruisebas, line 75: Looking at how [this is implemented](https://github.com/smithy-lang/smithy-swift/pull/593/files#diff-d50c8a9f81bf3276120cb94e56812f016f1f0b657f9c28ca2e1adca7d4c5b460), it looks like we'll be adding all the original headers to the new `additionalHeaders` list, which will then be combined with the original headers resulting in unnecessary duplication.

I believe we can just do this instead:
```swift
request.withHeader(
    name: userAgentHeader, 
    value: "\(currentUserAgent)\(userAgentSuffix)"
)
```

But now I wonder, would we be sending two values for the `User-Agent` header then? 🤔. Would that still work?

PR #3322, opened 2023-10-23
fix(Analytics): Handling certain auth errors as retryable errors

ruisebas:
## Description
If guest/unauthenticated access is disabled, attempting to submit cached events with a user signed our or with an expired token will fail and all events will be discarded. This happens because this failure is considered non-retryable.

This PR fixes that by considering the following `AuthError`s to be retryable without affecting the events retry count:
- `.sessionExpired`
-  `.signedOut`
- `.service` with an underlying error of type `AWSCognitoAuthError.invalidAccountTypeException`

This is similar as to how connectivity issues are handled.

## General Checklist
- [X] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/Support/Utils/AnalyticsErrorHelper.swift
+++ b/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/Support/Utils/AnalyticsErrorHelper.swift
@@ -14,6 +14,8 @@ enum AnalyticsErrorHelper {
         switch error {
         case let error as AnalyticsErrorConvertible:
             return error.analyticsError
+        case let error as AuthError:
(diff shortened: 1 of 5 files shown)


AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/Support/Utils/AnalyticsErrorHelper.swift
```diff
@@ -14,6 +14,8 @@ enum AnalyticsErrorHelper {
         switch error {
         case let error as AnalyticsErrorConvertible:
             return error.analyticsError
+        case let error as AuthError:
+            return .configuration(error.errorDescription, error.recoverySuggestion, error)
```
ruisebas, line 18: One can argue that this isn't really a `.configuration` error, but it's also not really an `.unknown` one. And unfortunately those are the only two options 😅

Issue #1308, opened 2021-07-02
Datastore conflict AutoMerge behaves differently from amplify-android


HuiSF:
### Describe the bug

Original issue amplify-flutter [#634](https://github.com/aws-amplify/amplify-flutter/issues/634)

amplify-ios AutoMerge for handling conflict caused by multi-device offline/online syncing behaves differently from amplify-android, where amplify-android seems having the correct behavior.

Detailed description please refer to [this comment](https://github.com/aws-amplify/amplify-flutter/issues/634#issuecomment-873151792)

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Run datastore in two iOS simulators, say device A and device B, both have same data
2. Make device A and B both offline
3. Make changes to a piece of data in device A, quit App, then make device A online, reopen the App to publish the update mutation
4. Make changes to the same piece of data in device B, quit App, then make device B online, reopen the App to publish the update mutation
```

### Expected behavior

The update mutation made in device B when it was offline should be merged into DynamoDB after AutoMerge

### Amplify Framework Version

1.11.0

### Amplify Categories

DataStore

### Dependency manager

Cocoapods

### Swift version

5

### CLI version

5.1.0

### Xcode version

12.5.1

### Device

iPhone 12 simulators

### iOS Version

iOS 14.5

chrisbonifacio:
@HuiSF sorry the delay, we've triaged the issue and added it to our backlog for improving behavior parity between platforms.

atierian:
This fix has been released in [2.25.3](https://github.com/aws-amplify/amplify-swift/releases/tag/2.25.3) and [1.30.6](https://github.com/aws-amplify/amplify-swift/releases/tag/1.30.6)

Thanks for reporting this and for your patience!

PR #3377, opened 2023-11-23
fix(datastore): sync pending mutation events with latest synced metadata

5d:
## Issue \#

https://github.com/aws-amplify/amplify-swift/issues/1308

## Description

- This patch aligns with the existing [amplify-android](https://github.com/aws-amplify/amplify-android/blob/main/aws-datastore/src/main/java/com/amplifyframework/datastore/syncengine/MutationProcessor.java#L287-L296) logic:
  - Apply metadata (version) from incoming mutation events when there is an pending mutation.
  - When syncing with AppSync, retrieve the latest synchronized metadata (version) and submit it along with the pending mutation event.

## General Checklist

- [x] All unit tests pass


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
```diff
@@ -128,7 +131,7 @@ class SyncMutationToCloudOperation: AsynchronousOperation {
                 request = GraphQLRequest<MutationSyncResult>.deleteMutation(of: model,
                                                                             modelSchema: modelSchema,
                                                                             where: graphQLFilter,
-                                                                            version: mutationEvent.version)
+                                                                            version: latestSyncMetadata?.version)
```
lawmicha, line 134: should we check the optionality of `latestSyncMetadata` earlier than at the time of constructing the request? what should we do if it's nil?

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/ReconcileAndLocalSaveOperation.swift
```diff
@@ -133,16 +133,21 @@ class ReconcileAndLocalSaveOperation: AsynchronousOperation {
...
+                        self.queryPendingMutations(withModels: dispositions.map(\.remoteModel.model))
+                            .map { pendingMutations in (pendingMutations, dispositions) }
+                    }
+                    .map { (pendingMutations, dispositions) in
+                        self.separateDispositions(pendingMutations: pendingMutations, dispositions: dispositions)
+                    }
+                    .flatMap { (dispositions, dispositionOnlyApplyMetadata) in
+                        self.waitAllPublisherFinishes(publishers: dispositionOnlyApplyMetadata.map(self.saveMetadata(disposition:)))
+                            .flatMap { _ in self.applyRemoteModelsDispositions(dispositions) }
                     }
```
lawmicha, line 151: can you add doc comment for the high level details of this `reconcile(remoteModels)` method? So we have a clear understanding of what is logic (and the proposed in this PR) and why this order of operations is correct

lawmicha (commented):
Some follow up questions, first and formost would be: Should MutationEvent's version property be deprecated?

Exploring if we removed `version` from MutationEvent:
- OutgoingMutationQueue processes existing pending mutations by reconciling the version from the incoming mutation response through `MutationEvent.reconcilePendingMutationEventsVersion`. Is this still needed if we always get the latest sync metadata from storage? 
- ReconcileAndLocalSaveOperation will call `notify(savedModel:mutationType)` and the hub payload is a MutationEvent. if the answer to the first question is that `version` is deprecated, then this will impact customers listening to hub events with MutationEvent payloads. 
- Similar question to above, MutationEvent is also the event for `DataStore.observe` API. Asking the deprecated question in another way: Are other platform's observe API returning data that contains the `version`  as well?
- Some other areas of interest is AWSMutationDatabaseAdapter

5d:
> Some follow up questions, first and formost would be: Should MutationEvent's version property be deprecated?
> 
> Exploring if we removed `version` from MutationEvent:
> 
> * OutgoingMutationQueue processes existing pending mutations by reconciling the version from the incoming mutation response through `MutationEvent.reconcilePendingMutationEventsVersion`. Is this still needed if we always get the latest sync metadata from storage?
> * ReconcileAndLocalSaveOperation will call `notify(savedModel:mutationType)` and the hub payload is a MutationEvent. if the answer to the first question is that `version` is deprecated, then this will impact customers listening to hub events with MutationEvent payloads.
> * Similar question to above, MutationEvent is also the event for `DataStore.observe` API. Asking the deprecated question in another way: Are other platform's observe API returning data that contains the `version`  as well?
> * Some other areas of interest is AWSMutationDatabaseAdapter

I think for now, the safest way to do is mark it as deprecated field. If we remove the `version` field, shouldn't it be a major version change? For those customer who still using the mutation event version, the version will have the same value and meaning as before.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -201,8 +201,10 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
         }
 
         Task {
+            let latestSyncMetadata = try? storageAdapter.queryMutationSyncMetadata(for: mutationEvent.modelId, modelName: mutationEvent.modelName)
             let syncMutationToCloudOperation = await SyncMutationToCloudOperation(
                 mutationEvent: mutationEvent,
+                latestSyncMetadata: latestSyncMetadata,
```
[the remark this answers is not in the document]
lawmicha: I think the options are..
1. When the MutationEvent is queried from the MutationEvent table, we also do a second query to get from MutationSyncMetadata, and store the version in the MutationEvent (probably do a hard copy). MutationEvent gets passed to SyncMutationToCloudOperation. This option means SyncMutationToCloudOperation doesn't have to change. MutationEvent is a "snapshot" of the model, version, etc. We're just making sure as we dequeue, the "snapshots" we work off always have the latest version. This also means MutationEvent.version is not being deprecated, just rehydrated when syncing.
2. Hydrate the version before creating the SyncMutationToCloudOperation, timing wise is the same as option 1. SyncMutationToCloudOperation's logic should use `latestSyncMetadata.version` and not use mutationEvent.version. It is odd to have "version" properties, so deprecating MutationEvent.version makes sense. However there are other implications of deprecating MutationEvent.version (discussion in another thread).
3. Hydrate the version within the execution of SyncMutationToCloudOperation, this way we optimize on getting the latest version at all times. If we hydrate it before, option 1 and 2, we don't know the delta between that version and the time that the SyncMutationToCloudOperation is actually executing its `main()` function. 

I think option 3 will have the smallest time gap between sending the request and getting the latest version for that model

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
lawmicha: Makes sense, i forgot create mutations will be `nil` version. Seems like if query failed with error, we should still continue to send the mutation and let the conflict handler resolve the conflicts.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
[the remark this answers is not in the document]
5d: I see your point. The option#3 is the most optimized one. 

Initially, I attempted to pass the async escape closure to `SyncMutationToCloudOperation` in order to retrieve the `latestSyncMetadata`. This would be executed at the very last moment when generating the API request for the mutation event. However, I observed that this operation involves a lock and spans across various code snippets. Introducing an async closure within a lock context would need many modifications. Consequently, I opted to take a step back and simplify the change by resolving the `latestSyncMetadata` separately and passing it in.

Amplify/Categories/DataStore/Subscribe/MutationEvent.swift
```diff
@@ -17,6 +17,8 @@ public struct MutationEvent: Model {
     public var json: String
     public var mutationType: String
     public var createdAt: Temporal.DateTime
+
+    @available(*, deprecated, message: "version field is deprecated")
```
lawmicha, line 21: > I think for now, the safest way to do is mark it as deprecated field. If we remove the `version` field, shouldn't it be a major version change? For those customer who still using the mutation event version, the version will have the same value and meaning as before.

I'm not sure if marking as deprecated in this PR will provide any benefits to the customer or our codebase (without it cascading to additional changes). My question was more on whether we should do it or not, I am thinking now that deprecating `MutationEvent.version` is going to be a big effort out of scope of this PR, it is still the public facing contract for providing events through DataStore.observe API and other channels like the Hub. 

If it should be deprecated, we should make sure we know why we are doing this, perhaps customer doesn't care about the version. This would also be a big initiative to align the APIs across platforms, or effort to align with JS's API: https://docs.amplify.aws/javascript/build-a-backend/more-features/datastore/real-time/

The steps i'd suggest for this is..
- Keep MutationEvent.version not deprecated
- Treat MutationEvent's version as a snapshot of the version at the time it was hydrated. This means in OutgoingMutationQueue, we should stop modifying the snapshot version for the consecutive save scenario (remove MutationEvent.reconcilePendingMutationEventsVersion and calling code). We're getting the latest version in SyncToCloudMutationOperation anyways.
- Make sure all notify calls to the Hub and event propagation to DataStore.observe subscribers get a MutationEvent that contains the latest version
- There may be other areas that I missed, one way to find these is to remove the `version` from MutationEvent completely and see where the code depends on it. (Then revert removing the `version` field back of course)

lawmicha: > Make sure all notify calls to the Hub and event propagation to DataStore.observe subscribers get a MutationEvent that contains the latest version

we looked over this and 
- OutgoingMutationQueue sends an OutboxMutationProcessedEvent which has a version coming from the remote model's version metadata, so it is the latest
- Reconcile path will send a MutationEvent that is reconstructed from the remote model's version as well. 
Both places use the latest version for creating the MutationEvent object for the public facing APis

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/AWSMutationDatabaseAdapter/AWSMutationDatabaseAdapter+MutationEventIngester.swift
```diff
@@ -33,21 +33,6 @@ extension AWSMutationDatabaseAdapter: MutationEventIngester {
...
-        // We don't want to query MutationSync<AnyModel> because a) we already have the model, and b) delete mutations
-        // are submitted *after* the delete has already been applied to the local data store, meaning there is no model
-        // to query.
-        var mutationEvent = mutationEvent
-        do {
-            // TODO: Refactor this so that it's clear that the storage engine is not responsible for setting the version
-            // perhaps as simple as renaming to `submit(unversionedMutationEvent:)` or similar
-            let syncMetadata = try storageAdapter.queryMutationSyncMetadata(for: mutationEvent.modelId,
-                                                                               modelName: mutationEvent.modelName)
-            mutationEvent.version = syncMetadata?.version
```
lawmicha, line 46: Down below on line 196 creates a new MutationEvent based on a "candidate" mutation event and sets the version from the candidate. This line isn't needed anymore, right? If not needed, we should remove it as well since it will be confusing to the reader in the future to understand why we have logic around the version field in this file at all.

AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Sync/MutationQueue/SyncMutationToCloudOperationTests.swift
```diff
@@ -67,12 +67,15 @@ class SyncMutationToCloudOperationTests: XCTestCase {
...
 
-        let operation = await SyncMutationToCloudOperation(mutationEvent: mutationEvent,
-                                                     api: mockAPIPlugin,
-                                                     authModeStrategy: AWSDefaultAuthModeStrategy(),
-                                                     networkReachabilityPublisher: publisher,
-                                                     currentAttemptNumber: 1,
-                                                     completion: completion)
+        let operation = await SyncMutationToCloudOperation(
+            mutationEvent: mutationEvent,
+            getLatestSyncMetadata: { nil },
```
lawmicha, line 72: We can test passing in a closure that returns non-nil as well and assert the API call is made with the version

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
```diff
@@ -31,13 +32,15 @@ class SyncMutationToCloudOperation: AsynchronousOperation {
     private var authTypesIterator: AWSAuthorizationTypeIterator?
 
     init(mutationEvent: MutationEvent,
+         getLatestSyncMetadata: @escaping () -> MutationSyncMetadata?,
```
lawmicha, line 35: are there some benefits to passing in this closure over passing in the storageAdapter instance?

AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Sync/MutationQueue/SyncMutationToCloudOperationTests.swift
5d: will update the test!

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/AWSMutationDatabaseAdapter/AWSMutationDatabaseAdapter+MutationEventIngester.swift
5d: do you mean the logic inside `getResolvedEvent`? This is about merging 2 MutationEvents into one. Even the version is not used when submitting to AppSync. As we still keep the merging logic, it makes sense to take the candidate version over the old mutation event version. It might help with debugging.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/SyncMutationToCloudOperation.swift
5d: The closure acting as an interface here. It could simplify testing. For current impelmentation, the `storageAdapter` is the hidden denpendency here. 

Btw, I realized the SQL query is not an async operation and I moved the get version logic to the very moment of creating the API request.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/AWSMutationDatabaseAdapter/AWSMutationDatabaseAdapter+MutationEventIngester.swift
lawmicha: Yes inside `getResolvedEvent`, is `resolvedEvent.version = candidate.version` always setting nil? If we remove the current code block (from lines 36-50), would the candidate version always be nil? If so, it is confusing why we will set the version during the merging logic

5d: oh, I see. The candidate was created in this block of code. Yeah, I will remove that version assignment.

lawmicha (approved):

Issue #3004, opened 2023-06-02
AWSCognitoNetworkPreferences cannot be found after upgrading to 2.x


grutts:
### Describe the bug

We are upgrading our library from 1.x to 2.11.5

Our existing code will not build, as AWSCognitoNetworkPreferences can no longer be found.

We've read the upgrade instructions and amplify codebase but can't find references to the network preferences or an alternative configuration.

```
import Amplify
import AWSCognitoAuthPlugin

try Amplify.add(plugin: AWSCognitoAuthPlugin(networkPreferences: AWSCognitoNetworkPreferences(
        // Limit number of retries and timeout to prevent spamming of backend after iPad reconnects to internet
        maxRetryCount: 2,
        timeoutIntervalForRequest: 60,
        timeoutIntervalForResource: 60
)))
try Amplify.configure()
```

### Steps To Reproduce

```swift
Run the above code
```

### Expected behavior

It continues to work

### Amplify Framework Version

2.11.5

### Amplify Categories

Auth

### Dependency manager

Swift PM

### Swift version

5

### CLI version

n/a

### Xcode version

14.2

### Is this a regression?

No

### Device

iPad 10 simulator

### iOS Version

15

grutts:
Hi please could you explain why this is tagged as a feature request and not a bug? This code worked in version 1.X

atierian:
Generally, adding a new API into a major version would be considered new feature. But you're right, it looks like v2 has removed the ability to configure these values, and with that has removed support for this use case.

grutts:
OK thank you for correcting the label. Do you have an estimate of when this will be prioritised? This bug blocks us from upgrading the library - so it would be good if it was fixed with ample buffer before V1 of the library left maintenance (May 31, 2024)

harsh62:
@grutts Our team is reviewing the effort and changes that would needed to support this feature in Amplify V2. We will post an update on the issue in the next couple of weeks.

grutts:
Hi @harsh62 do you have an update please?

grutts:
Hello, do you have an estimate when this will be fixed now that https://github.com/awslabs/aws-sdk-swift/issues/1065 has been merged?

grutts:
Thanks @atierian - that's a very helpful update

grutts:
Hi @harsh62 thanks for the update. We should be able to test this w/c 4 December

5d:
[...]

PR #3379, opened 2023-11-24
feat(Auth): Adding network preferences

harsh62:
## Issue \#

#3004 

## Description

This PR aims to add network preferences (`AWSCognitoNetworkPreferences `) back to the Auth plugin which were accidentally removed while release Amplify V2. 

This preference helps control
* Max number of retries the underlying SDK client will make
* The connection timeout that the underlying SDK client should set. 

Example configuration: 

```
        let plugin = AWSCognitoAuthPlugin(
            networkPreferences: .init(
                maxRetryCount: 2,
                timeoutIntervalForRequest: 60,
                timeoutIntervalForResource: 60))
        try Amplify.add(plugin: plugin)
```

NOTE: Resource time out is only applicable to HostedUI ATM, because the underlying Swift SDK does not support resource timeouts. Hence `timeoutIntervalForResource` has been made an optional.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoNetworkPreferences.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoNetworkPreferences.swift
@@ -9,22 +9,21 @@ import Foundation
 
 public struct AWSCognitoNetworkPreferences {
 
-    /// The maximum number of retries for failed requests. The value needs to be between 0 and 10 inclusive.
-    /// If set to higher than 10, it becomes 10.
+    /// The maximum number of retries for failed requests.
     public let maxRetryCount: UInt32
 
     /// The timeout interval to use when waiting for additional data.
-    public let timeoutIntervalForRequest: Double
+    public let timeoutIntervalForRequest: TimeInterval
 
     /// The maximum amount of time that a resource request should be allowed to take.
     /// 
     /// NOTE: This value is only applicable to HostedUI because the underlying Swift SDK does
     /// not support resource timeouts
-    public let timeoutIntervalForResource: Double?
+    public let timeoutIntervalForResource: TimeInterval?
 
     public init(maxRetryCount: UInt32,
-                timeoutIntervalForRequest: Double,
-                timeoutIntervalForResource: Double? = nil) {
+                timeoutIntervalForRequest: TimeInterval,
+                timeoutIntervalForResource: TimeInterval? = nil) {
         self.maxRetryCount = maxRetryCount
         self.timeoutIntervalForRequest = timeoutIntervalForRequest
         self.timeoutIntervalForResource = timeoutIntervalForResource


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoNetworkPreferences.swift
```diff
@@ -0,0 +1,32 @@
...
+import Foundation
+
+public struct AWSCognitoNetworkPreferences {
+
+    /// The maximum number of retries for failed requests. The value needs to be between 0 and 10 inclusive.
+    /// If set to higher than 10, it becomes 10.
+    public let maxRetryCount: UInt32
+
+    /// The timeout interval to use when waiting for additional data.
+    public let timeoutIntervalForRequest: Double
```
lawmicha, line 17: should this just be `TimeInterval` type since that's always specified in seconds? If it's just a Double, would suggest something that says "seconds" in there `timeoutRequestInSeconds`

lawmicha (commented):
questions regarding the types in AWSCognitoNetworkPreferences. Not sure if there's an opportunity here to change some of the types of those properties considering we are adding back something that was taken out by accident.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
```diff
@@ -102,6 +102,15 @@ extension AWSCognitoAuthPlugin {
                 configuration.httpClientEngine = .userAgentEngine(for: configuration)
             }
 
+            if let requestTimeout = networkPreferences?.timeoutIntervalForRequest {
+                let requestTimeOutMs = requestTimeout * 1_000
+                configuration.connectTimeoutMs = UInt32(requestTimeOutMs)
+            }
+
+            if let maxRetryUnwrapped = networkPreferences?.maxRetryCount {
+                configuration.retryStrategyOptions = RetryStrategyOptions(maxRetriesBase:  Int(maxRetryUnwrapped))
```
lawmicha, line 111: Not related to this PR, question for my knowledge, when/where in the plugin validates the 10 limit for retry count?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoNetworkPreferences.swift
ruisebas: `TimeInterval` **is** a `Double` anyway, so I vote to change the type to be explicit on what we expect (i.e. seconds).

```swift
typealias TimeInterval = Double
```

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoNetworkPreferences.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    /// The maximum amount of time that a resource request should be allowed to take.
+    /// 
+    /// NOTE: Resource time out is only applicable to HostedUI ATM, because the underlying Swift SDK does
```
ruisebas, line 21:
```suggestion
    /// NOTE: This value is only applicable to HostedUI because the underlying Swift SDK does
```

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
ruisebas: Should _we_ do the validation then?

I guess not because this is network-related, but for example if we know that the service will absolutely fail if we retry more than 10 times, then we might wanna add the validation.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Models/AWSCognitoNetworkPreferences.swift
lawmicha: As @ruisebas  mentioned, `Double` to `TimeInterval` is non-breaking change since it's just a typealias.

No naming changes, TimeInterval is expected to be in seconds, other places have names like "backgroundTrackingTimeout", "timeout", "sessionRefreshInterval", etc.

Amplify core provides a public extension for TimeInterval: https://github.com/aws-amplify/amplify-swift/blob/main/Amplify/Core/Support/TimeInterval%2BHelper.swift This seems useful if we changed the type to TimeInterval.

If we agree on above, looks like we can just do it now since it's non-breaking

lawmicha (approved):

ruisebas (approved):

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/AWSCognitoAuthPlugin+Configure.swift
ruisebas: Hmmm I kinda disagree with "the service's behaviour may change in the future" statement, on a conceptual basis.

If we know the service fails or does not support a certain thing, I don't think we should still allow that very thing to be attempted. It's like setting ourselves (or Amplify's customers) for failure.

But again, I don't really know the specifics of why this particular use case was implemented in the first place, so I'm fine with the removal of it if you guys think it doesn't apply anymore.

PR #3403, opened 2023-12-06
fix(Analytics): Fixing session duration being reported for non-stopped sessions

ruisebas:
## Description
According to [Pinpoint docs](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-events.html), the session duration: 
> is required if a value is provided for the StopTimestamp.

However we are currently sending a value even for non-stopped sessions, by reporting how much time has passed since the session started.

This could result in crashes on watchOS when attempting to submit old cached events with non-stopped sessions, since watchOS's `Int` is actually 32 bit. As an extra precaution, I've added an explicit check to validate that the session duration can be cast.

## General Checklist

- [X] All unit tests pass
- [X] [All integration tests pass](https://github.com/aws-amplify/amplify-swift/actions/runs/7119826273)
- [X] Documentation update for the change if required

### DIFF
--- a/AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/PinpointEvent+PinpointClientTypes.swift
+++ b/AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/PinpointEvent+PinpointClientTypes.swift
@@ -10,7 +10,14 @@ import AWSPluginsCore
 import Foundation
 
 extension PinpointEvent {
-    var clientTypeSession: PinpointClientTypes.Session {
+    var clientTypeSession: PinpointClientTypes.Session? {
+    #if os(watchOS)
+        // If the session duration cannot be represented by Int, return a nil session instead.
+        // This is extremely unlikely to happen since a session's stopTime is set when the app is closed
+        if let duration = session.duration, duration > Int.max {
+            return nil
+        }
+    #endif
         return PinpointClientTypes.Session(duration: Int(session.duration),
                                            id: session.sessionId,
                                            startTimestamp: session.startTime.asISO8601String,
@@ -46,5 +53,14 @@ extension Bundle {
 
     var appVersion: String {
         object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String ?? ""
+    }
+}
+
+private extension Int {
+    init?(_ value: Int64?) {
+        guard let value = value else {
+            return nil
+        }
+        self.init(value)
     }
 }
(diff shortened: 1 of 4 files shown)


AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Session/PinpointSession.swift
```diff
@@ -35,8 +35,11 @@ public class PinpointSession: Codable {
         return stopTime != nil
     }
 
-    var duration: Date.Millisecond {
-        let endTime = stopTime ?? Date()
+    var duration: Date.Millisecond? {
+        /// According to Pinpoint's documentation, `duration` is only required if `stopTime` is not nil.
+        guard let endTime = stopTime else {
+            return nil
+        }
```
atierian, line 42: nit: we can use shorthand shadowing now that our min. support is Swift 5.7
```swift
guard let stopTime else { return nil }
```

AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/PinpointEvent+PinpointClientTypes.swift
```diff
@@ -48,3 +55,12 @@ extension Bundle {
...
 }
+
+private extension Int {
+    init?(_ value: Int64?) {
+        guard let value = value else {
+            return nil
+        }
+        self.init(value)
+    }
+}
```
atierian, line 66: FWIW you can use map for this instead.
```swift
let a: Int64? = 42
let b = a.map(Int.init)
```

AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/PinpointEvent+PinpointClientTypes.swift
```diff
@@ -10,7 +10,14 @@ import AWSPluginsCore
...
 extension PinpointEvent {
-    var clientTypeSession: PinpointClientTypes.Session {
+    var clientTypeSession: PinpointClientTypes.Session? {
+    #if os(watchOS)
+        // If the session duration cannot be represented by Int, return a nil session instead.
+        // This is extremely unlikely to happen since a session's stopTime is set when the app is closed
+        if let duration = session.duration, duration > Int.max {
+            return nil
+        }
+    #endif
```
atierian, line 20: What do you think about using `Int(exactly:)` to handle this conversion generally (not just watchOS)?
[[docs](https://developer.apple.com/documentation/swift/int/init(exactly:)-b1dy)]

PR #3453, opened 2024-01-04
fix(DataStore): auth plugin requirement for single auth rule

lawmicha:
## Issue \#
https://github.com/aws-amplify/amplify-flutter/issues/3967
https://github.com/aws-amplify/amplify-flutter/issues/3967#issuecomment-1875988560

## Description

The auth mode strategy and the auth rules are used in determination of whether the auth plugin is required for DataStore. Some common scenarios are
- No auth rules on the models does not require auth plugin
- Single auth rule on a model should make its determination based on the provider. 
  - OIDC/lambda - does not require auth plugin, but requires auth provider on the APIPLugin
  - APIKey - does not require auth plugin, the key is the configuration of the APIPlugin.
  - UserPools/IAM - requires auth plugin
- Multi-Auth rules with multi-auth enabled, require auth plugin if any of the rules requires auth plugin.

The issue in https://github.com/aws-amplify/amplify-flutter/issues/3967 describes the scenario
- multi-auth rules: OIDC and IAM
- OIDC auth provider on the APIPlugin
- multi-auth not enabled- only want to use the default authorization (OIDC) from the API

The problem was DataStore was requiring auth plugin for this scenario. The change in this PR addresses that by only checking the auth rules if and only if there is one auth rule when multi-auth is not enabled. If there is more than one auth rule, fall back to the API's default auth type. There are also some edge cases that remains the same in the logic and were not changed by this PR: if there are no auth rules, do not require auth plugin. if there is an auth rule but no provider can be found, also fall back to the API's default auth type.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/StorageEngine+SyncRequirement.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/StorageEngine+SyncRequirement.swift
@@ -34,7 +34,9 @@ extension StorageEngine {
                 ))
             }
 
-            let authPluginRequired = StorageEngine.requiresAuthPlugin(api)
+            let authPluginRequired = StorageEngine.requiresAuthPlugin(
+                api,
+                authModeStrategy: dataStoreConfiguration.authModeStrategyType)
             guard authPluginRequired else {
                 syncEngine.start(api: apiGraphQL, auth: nil)
                 return .success(.successfullyInitialized)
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/StorageEngine+SyncRequirement.swift
```diff
@@ -34,7 +34,9 @@ extension StorageEngine {
                 ))
             }
 
-            let authPluginRequired = StorageEngine.requiresAuthPlugin(api)
+            let authPluginRequired = StorageEngine.requiresAuthPlugin(
+                api,
+                authModeStrategy: dataStoreConfiguration.authModeStrategyType)
```
ruisebas, line 39:
```suggestion
                authModeStrategy: dataStoreConfiguration.authModeStrategyType
            )
```

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/StorageEngine+SyncRequirement.swift
```diff
@@ -96,20 +98,45 @@ extension StorageEngine {
...
+        case .default:
+            if authRules.isEmpty {
+                return false
+            }
+            // Only use the auth rule as determination for auth plugin requirement when there is
+            // exactly one. If there is more than one auth rule AND multi-auth is not enabled,
+            // then immediately fall back to using the default auth type configured on the APIPlugin because
+            // we do not have enough information to know which provider to use to make the determination.
+            if authRules.count == 1,
+                let singleAuthRule = authRules.first,
```
ruisebas, line 132:
```suggestion
               let singleAuthRule = authRules.first,
```

ruisebas (dismissed):

phantumcode (approved):

PR #3465, opened 2024-01-10
ci: use alias latest and minimum to represent xcode cmd tool version

5d:
## Description

- add alias `latest` and `minimum` to represent xcode cmd tool version
  - `latest` is xcode 14.3.1
  - `minimum` is xcode 14.0.1

## General Checklist

- [X] All unit tests pass

### DIFF
--- a/.github/composite_actions/get_platform_parameters/action.yml
+++ b/.github/composite_actions/get_platform_parameters/action.yml
@@ -5,62 +5,99 @@ inputs:
     required: true
     type: string
   xcode_version:
-    description: 'The version of Xcode. Valid values are 14.3 and 15.0'
-    required: true
+    description: "The version of Xcode. Available aliases are 'latest' and 'minimum'"
+    default: 'latest'
     type: string
+  destination:
+    description: "The destination associated with the given platform and Xcode version"
+    default: ''
+    type: string
+
 outputs:
   destination:
     description: "The destination associated with the given platform and Xcode version"
-    value: ${{ steps.platform.outputs.destination }}
+    value: ${{ steps.get-destination.outputs.destination }}
   sdk:
     description: "The SDK associated with the given platform"
-    value: ${{ steps.platform.outputs.sdk }}
+    value: ${{ steps.get-sdk.outputs.sdk }}
+  xcode-version:
+    description: "The Xcode version to build with"
+    value: ${{ steps.get-xcode-version.outputs.xcode-version }}
+
 runs:
   using: "composite"
   steps:
-    - id: platform
+    - name: Validate platform
       run: |
-        PLATFORM=${{ inputs.platform }}
-        case $PLATFORM in
+        INPUT_PLATFORM=${{ inputs.platform }}
+        case $INPUT_PLATFORM in
           iOS|tvOS|watchOS|macOS) ;;
-          *) echo "Unsupported platform: $PLATFORM"; exit 1 ;;
+          *) echo "Unsupported platform: $INPUT_PLATFORM"; exit 1 ;;
         esac
+      shell: bash
 
-        XCODE_VERSION=${{ inputs.xcode_version }}
-        case $XCODE_VERSION in
-          14.0.1|14.3|15.0) ;;
-          *) echo "Unsupported Xcode version: $XCODE_VERSION"; exit 1 ;;
+    - id: get-xcode-version
+      run: |
+        LATEST_XCODE_VERSION=14.3.1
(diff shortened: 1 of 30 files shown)


.github/workflows/integ_test_logging.yml
```diff
@@ -39,7 +39,6 @@ jobs:
       scheme: ${{ matrix.platform == 'watchOS' && 'AWSCloudWatchLoggingPluginIntegrationTestsWatch' || 'AWSCloudWatchLoggingPluginIntegrationTests' }}
       platform: ${{ matrix.platform }}
       project_path: ./AmplifyPlugins/Logging/Tests/AWSCloudWatchLoggingPluginHostApp
-      xcode_version: ${{ matrix.platform == 'watchOS' && '15.0' || '14.3' }}
```
ruisebas, line 42: These tests used to run on Xcode 15.0 for watchOS, but now they will use 14.3.
Have you checked it still works?

.github/composite_actions/get_platform_parameters/action.yml
```diff
@@ -5,62 +5,99 @@ inputs:
     required: true
     type: string
   xcode_version:
-    description: 'The version of Xcode. Valid values are 14.3 and 15.0'
-    required: true
+    description: "The version of Xcode. Available aliases are 'latest' and 'minimum'"
+    default: 'latest'
+    type: string
+  destination:
```
ruisebas, line 11: Is this new `destination` input meant to override the one that is automatically determined by the Xcode version?

When is it used or needed?

.github/composite_actions/get_platform_parameters/action.yml
```diff
@@ -5,62 +5,99 @@ inputs:
     required: true
     type: string
   xcode_version:
-    description: 'The version of Xcode. Valid values are 14.3 and 15.0'
-    required: true
+    description: "The version of Xcode. Available aliases are 'latest' and 'minimum'"
```
ruisebas, line 8: Why not just add the new aliases but still support passing an explicit version? 

Because now `"latest`" is not really ***the*** latest (i.e. 15.0), so it's kinda misleading to call it like that.
If we have `latest = 15.0` set, but we explicitly pass `14.3` in our tests invocations, it's clear we need to investigate why we cannot upgrade.

So basically what I'm saying is keeping `14.0.1`,  `14.3` and `15.0` as allowed value, but also supporting `latest` which we consider to be `15.0` and `minimum` which we consider to be `14.0.1`.

Then you can just use `latest` in tests that support it, but set `14.3` in the ones that don't.

ruisebas (changes requested):
Left some comments/suggestions.

Also, have you run all the integration tests suits to check they also pass with the new Xcode version?

.github/composite_actions/get_platform_parameters/action.yml
5d: Yes, it is used to define the exact destiation. In case of `latest` or `minimum`, you can use the default empty string value and the script will try to use the predefined destination mappings.
This input option becomes essential when testing a particular Xcode version that is neither `latest` nor `minimum`. For example, when testing a beta Xcode version and specifying the beta SDK version in the `destination` input.

.github/composite_actions/get_platform_parameters/action.yml
5d: It is possible to pass a specific version. The alias available is only `latest` and `minimum`. You can still pass `15.1` as `xcode_version` input. However, you will also need to specify the `destination` for it.

.github/workflows/integ_test_logging.yml
5d: good catch, will update it.

.github/composite_actions/get_platform_parameters/action.yml
ruisebas: Aah I see, thanks for clarifying that :) 

If setting an explicit destination is indeed an "edge" case (i.e. when testing new beta versions, or for a reduced number of tests), then that's fine. But if we have to write the whole `destination` in many files, then this action becomes irrelevant and exactly what it tried to replace :P

tl;dr: this is fine as long as we don't have to set an explicit xcode version in more than 1 workflow, otherwise we should add the mapped destination for that xcode version and save ourselves one long and messy input.

5d: yeah, if that's the case, we might need to consider adding another alias `beta` and the destination mapping in the script.

ruisebas (approved):

Issue #3259, opened 2023-10-02
In case of errors on some items during sync, no items are saved


gbitaudeau:
### Describe the bug

If during initial sync, AppSync responds with some items on error, these items are `null` in `items` response array. The other items which are not in error are not saved in datastore. Here a redacted result for sync Model named Draw with an error on one item:
```
{
  "data": {
    "syncDraws": {
      "items": [
        {
          "id": "70083950-9b18-4e62-90c6-433ae933def3",
          "drawBatch": {
            "id": "e236042a-7c21-4b82-90b3-d8b73d2c6b99",
            "_deleted": true
          },
          "item": {
            "_deleted": true,
            "id": "fa986c78-ff21-46e2-82ff-8d0a218510bc"
          },
          "owner": "cb8c1cd8-f22c-4e30-8a7c-42fa3e2e16b1",
          "updatedAt": "2023-09-20T09:09:06.134Z"
        },
        null
  },
  "errors": [
    {
      "path": [
        "syncDraws",
        "items",
        17,
        "drawBatch"
      ],
      "data": null,
      "errorType": "Unauthorized",
      "errorInfo": null,
      "locations": [
        {
          "line": 10,
          "column": 7,
          "sourceName": null
        }
      ],
      "message": "Not Authorized to access drawBatch on type Draw"
    }
  ]
}
```

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Consider a schema.graphql where 2 objects are linked together with `@hasMany` and `@belongsTo` and each object have the following auth policy: `@auth(rules: [{ allow: owner}])`.
2. Create on object of each type and linked then together.
3. Change the owner of the object with the `@hasMany` relationship.
4. The sync of the object type with `@belongsTo` is no longer working.
```

### Expected behavior

DataStore's initial sync should reconcile partial GraphQL responses by saving the items in the local Database and emitting the errors to the errorHandler.

### Amplify Framework Version

2.19.0

### Amplify Categories

DataStore

### Dependency manager

Swift PM

### Swift version

5.9

### CLI version

12.4.0

### Xcode version

15.0

### Is this a regression?

No

### Platforms

iOS

### OS Version

iOS 17.0.1

### Device

iPhone

### Additional context

See discussion https://github.com/aws-amplify/amplify-swift/issues/3232#issuecomment-1741312878

lawmicha:
Hi @gbitaudeau, do you have an redacted version of your schema for us to test with?

[...]

PR #3477, opened 2024-01-18
fix(DataStore): retry on session expired requests

lawmicha:
## Issue \#

https://github.com/aws-amplify/amplify-swift/issues/3259
More specifically https://github.com/aws-amplify/amplify-swift/issues/3259#issuecomment-1894217957

> Case 1: Expired Cognito Token

## Description

An expired token will be refreshed automatically when Amplify.API retrieves the token from the token provider, via Amplify.Auth. At this point, it should refresh the token if it is expired. When the token cannot be refreshed because the Refresh token is expired, then API receives an AuthError, which is wrapped in an `APIError.operationError`. The user needs to sign in again. 

DataStore's SyncMutationToCloudOperation uses Amplify.API to make the user pool auth request and receives  `API.opeationError(,,AuthError.sessionExpired)` error. When the session is expired, DataStore shouldn't delete mutation event because the request never made it to the service to perform the authorization check. The PR adds a check for session expired errors and goes into the exponential retry state.

Once the user signs in, and the mutation event gets sent on the next retry, then the response will be handled as normally. Mutation events should only be deleted if the response is Unauthorized, by calling `errorHandler` to notify the developer that we're dropping the event, then deleting it.

## General Checklist

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/RequestRetryablePolicy.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/RequestRetryablePolicy.swift
@@ -40,7 +40,8 @@ class RequestRetryablePolicy: RequestRetryable {
              .timedOut,
              .dataNotAllowed,
              .cannotParseResponse,
-             .networkConnectionLost:
+             .networkConnectionLost,
+             .userAuthenticationRequired:
             let waitMillis = retryDelayInMillseconds(for: attemptNumber)
             return RequestRetryAdvice(shouldRetry: true, retryInterval: .milliseconds(waitMillis))
         default:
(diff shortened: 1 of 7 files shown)


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/RequestRetryablePolicy.swift
```diff
@@ -40,7 +40,8 @@ class RequestRetryablePolicy: RequestRetryable {
              .timedOut,
              .dataNotAllowed,
              .cannotParseResponse,
-             .networkConnectionLost:
+             .networkConnectionLost,
+             .userAuthenticationRequired:
```
thisisabhash, line 44: can we have a unit test for this change?

lawmicha: added!

thisisabhash (approved):

PR #3485, opened 2024-01-27
fix(datastore-v1): correct the filter predicate logic applied to optional fields

5d:
## Description

The GraphQL filter predication not working properly on optional field. It always return `false` on optional field with value `nil`.

## General Checklist

- [x] All unit tests pass


AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderConfirmSignupTests.swift
```diff
@@ -26,7 +26,7 @@ class AuthenticationProviderConfirmSignupTests: BaseAuthenticationProviderTest {
     ///
     func testSuccessfulConfirmSignUp() {
 
-        let mockSignupResult = SignUpResult(signUpState: .confirmed, codeDeliveryDetails: nil)
+        let mockSignupResult = SignUpResult(signUpState: .confirmed, codeDeliveryDetails: nil, userSub: nil)
```
lawmicha, line 29: what's this from?

AmplifyPlugins/Core/AWSPluginsCoreTests/Query/QueryPredicateEvaluateGeneratedBoolTests.swift
```diff
@@ -41,7 +41,7 @@ class QueryPredicateEvaluateGeneratedBoolTests: XCTestCase {
 
         let evaluation = try predicate.evaluate(target: instance.eraseToAnyModel().instance)
 
-        XCTAssertFalse(evaluation)
+        XCTAssertTrue(evaluation)
```
lawmicha, line 44: [nit] let's add either a test comment or improve the test name to `testBooltruenotEqualBoolNil` it's not immediately obvious that the `instance` has a nil Bool due to the default parameter value. Here and other tests that have similar changes

AmplifyPlugins/Auth/AWSCognitoAuthPluginTests/AuthenticationProviderTests/AuthenticationProviderConfirmSignupTests.swift
5d: I think this is from the modle update of AWSCognito

Issue #2982, opened 2023-05-22
Delete publisher is not triggered on Amplify


Umar009:
### Describe the bug

We had integrated amplify in our project(android, iOS and web portal). We sign in user using amplify and store user data on amplify. Issue is when User is using iOS App it worked perfectly fine if we create or delete or edit user item entries from web or android app. But when App is in background or closed and user create new entry and then delete that entry and then open App either from background or fresh launch create event is called but delete event is not called(If user edit that item that event is also called).
Here point to be noted that if Item is already added and then we move app to background or lock the screen and then delete that item then If App comes back from background delete event is triggered.

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Login in web(or android) and iOS with valid user credential 
2. in iOS move App either background and lock the screen or close the App
3. create an item in user data base and then delete that item from web portal
4. open iOS App. (That delete item is not delete from iOS App)
```

### Expected behavior

Delete event must be triggered on amplify and item must be deleted from Amplify data store

### Amplify Framework Version

1.22.2

### Amplify Categories

DataStore

### Dependency manager

Cocoapods

### Swift version

5

### CLI version

10.7.3

### Xcode version

14.2

### Relevant log output

```shell
<details>
<summary>Log Messages</summary>


INSERT LOG MESSAGES HERE
```
</details>
```


### Is this a regression?

No

### Regression additional context

I don't know. I just encounter this issue

### Device

iphone Se 2020

### iOS Version

15.4

### Specific to simulators

_No response_

### Additional context

_No response_

**Comment by Umar009** [comment]
waiting for response

**Comment by lawmicha** [comment]
Hi @Umar009,

I was able to reproduce the issue. I called DataStore.start to establish the subscriptions, and then put the iOS app in the background. Then I performed a create mutation followed by a delete mutation, while the app still in the background. ie.
```
mutation MyMutation {
  createTodo(input: {content: "From AppSync Console"}) {
    updatedAt
    owner
    id
    createdAt
    content
    _version
    _lastChangedAt
    _deleted
  }
}
```
```
mutation MyMutation2 {
  deleteTodo(input: {_version: 1, id: "e097e308-490f-4480-aba6-bcaf6dd41eec"}) {
    _deleted
    _lastChangedAt
    _version
    content
    createdAt
    id
    owner
    updatedAt
  }
}
```

Once I bring the app back to the foreground, I can see both events, one for the onCreate subscription and one for the onDelete subscription:
```
[...]

PR #3489, opened 2024-01-31
fix(DataStore): serialize IncomingAsyncSubscriptionEventPublisher events

lawmicha:
## Issue \#
https://github.com/aws-amplify/amplify-swift/issues/2982#issuecomment-1671713659

## Description
Before the PR changes, the publisher would emit events concurrently which caused some of them to be missed by the subscriber even though the subscriber applied back pressure to process one event at a time (demand of `max(1)`). For more details regarding back-pressure, see https://developer.apple.com/documentation/combine/processing-published-elements-with-subscribers

The change, to publish the events though the same TaskQueue ensures that the events are properly buffered and sent only when the subscriber demands for it.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisherTests.swift
+++ b/AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisherTests.swift
@@ -1,0 +1,61 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+@testable import Amplify
+@testable import AmplifyTestCommon
+@testable import AWSPluginsCore
+@testable import AWSDataStorePlugin
+
+final class IncomingAsyncSubscriptionEventPublisherTests: XCTestCase {
+    var apiPlugin: MockAPICategoryPlugin!
+    override func setUp() {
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Sync/SubscriptionSync/IncomingAsyncSubscriptionEventPublisherTests.swift
```diff
@@ -0,0 +1,61 @@
...
+        asyncEvents.subscribe(subscriber: mapper)
+        let sink = mapper
+            .publisher
+            .sink(
+                receiveCompletion: { _ in },
+                receiveValue: { _ in
+                    expectedEvents.fulfill()
+                }
+            )
+        DispatchQueue.concurrentPerform(iterations: numberOfEvents) { index in
```
thisisabhash, line 54: is it an expectation that events should be received in order?
if so, maybe we can pass the `index` in event?

lawmicha: Yes, the events should be received in the order in which they are sent. I added a test below this one to make sure the usage of the TaskQueue with the subscriber model of demanding 1 item at a time is working as expected; the order is the correct and no events are dropped.

I don't think we can use `DispatchQueue.concurrentPerform`'s index as indication of the order because concurrent perform will dispatch threads with index out of order, so the order of index doesn't represent the order of the events being added to the taskQueue, and subsequentally received by the subscriber

thisisabhash (approved):

PR #3494, opened 2024-02-02
chore(logging): update logging remote constraint ETag caching

phantumcode:
## Description
Update the Logging remote constraint provider to fetch the configuration and look up the `ETag` in the response header.  `If-None-Match` should only be in the request header not the response header, keeping this logic in so that it is not a breaking change to existing customers.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/Configuration/DefaultRemoteLoggingConstraintsProvider.swift
+++ b/AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/Configuration/DefaultRemoteLoggingConstraintsProvider.swift
@@ -58,7 +58,9 @@ public class DefaultRemoteLoggingConstraintsProvider: RemoteLoggingConstraintsPr
         let loggingConstraint = try JSONDecoder().decode(LoggingConstraints.self, from: data)
         loggingConstraintsLocalStore.setLocalLoggingConstraints(loggingConstraints: loggingConstraint)
 
-        if let etag = (response as? HTTPURLResponse)?.value(forHTTPHeaderField: "If-None-Match") {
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/Configuration/DefaultRemoteLoggingConstraintsProvider.swift
```diff
@@ -58,7 +58,9 @@ public class DefaultRemoteLoggingConstraintsProvider: RemoteLoggingConstraintsPr
         let loggingConstraint = try JSONDecoder().decode(LoggingConstraints.self, from: data)
         loggingConstraintsLocalStore.setLocalLoggingConstraints(loggingConstraints: loggingConstraint)
 
-        if let etag = (response as? HTTPURLResponse)?.value(forHTTPHeaderField: "If-None-Match") {
+        if let etag = (response as? HTTPURLResponse)?.value(forHTTPHeaderField: "ETag") {
+            loggingConstraintsLocalStore.setLocalLoggingConstraintsEtag(etag: etag)
+        } else if let etag = (response as? HTTPURLResponse)?.value(forHTTPHeaderField: "If-None-Match") {
```
lawmicha, line 63: i think it's worth adding a comment in code to explain why the etag may be in the `If-None-Match` header, is it because we provided some lambda code to the developer and they may have deployed it already so we want to be backwards compatible? How did we decide to use ETag instead?

phantumcode: Good suggestion, I've updated with code comments.  ETag should be use based on the Http header spec and is consistent with the updated lambda sample code.

lawmicha (approved):

Issue #3252, opened 2023-09-29
MutationEvent.decodeModel don't decode LazyReference on delete mutation 


gbitaudeau:
### Describe the bug

Consider the sample schema (given there : https://docs.amplify.aws/lib/datastore/relational/q/platform/ios/#updated-schema )
```
enum PostStatus {
  ACTIVE
  INACTIVE
}

type Post @model @auth(rules: [{allow: public}]) {
  id: ID!
  title: String!
  rating: Int!
  status: PostStatus!
  # new field with @hasMany
  comments: [Comment] @hasMany
}

# new model
type Comment @model {
  id: ID!
  content: String
  post: Post @belongsTo
}
```
I use the LazyReference (using generatemodelsforlazyloadandcustomselectionset option for codegen)

When I use `Amplify.DataStore.observe` on Comment, if a comment is deleted, the post is not well set on it using `MutationEvent.decodeModel`, so there is no way to know which post loose a comment.

```
var postsSubscription:  AmplifyAsyncThrowingSequence<MutationEvent>?

// Then in the body of your code, subscribe to the subscription
func subscribeToPosts() async {
    let commentsSubscription = Amplify.DataStore.observe(Comment.self)
    self. commentsSubscription = commentsSubscription

    do {
        for try await changes in commentsSubscription {
            let comment = try change.decodeModel(as: Comment.self)
            let post = try await comment.post //eror here
        }
    } catch {
        print("Subscription received error: \(error)")
    }
}
```

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Setup an env with the previously given code
2. Delete a comment
```

### Expected behavior

The LazyReference must load successfully using `let post = try await comment.post`

### Amplify Framework Version

2.19.0

### Amplify Categories

DataStore

### Dependency manager

Swift PM

### Swift version

5.9

### CLI version

12.4.0

### Xcode version

15.0

### Is this a regression?

No

### Platforms

iOS

### OS Version

iOS 17.0.1

### Device

iPhone 14 Pro

### Additional context

I found a workaround in the generated `Comment.init(from decoder: Decoder)`, I replaced:
```
_post = try values.decodeIfPresent(LazyReference<Post>.self, forKey: .post) ?? LazyReference(identifiers: nil)
```
with
```
_post = try values.fixDecodeLazyReference(type: LazyReference< Post >.self, forKey: . post)
```
And I added the following extension:
```
extension KeyedDecodingContainer {
[...]

PR #3515, opened 2024-02-12
fix(DataStore): pass DataStore source for API decoding

harsh62:
## Issue \#

#3252 

### Root Cause Analysis: 
The issue happens when a delete operation happens remotely and a model generated by the API Plugin gets translated into a structure that is not understood by the DataStore decoder. This results in creating `nil` lazy references, which further bubbles up to the customer not able to load data from a lazy reference. 

### Fix:
There are few parts to the fix that has been done
* The encoders have been fixed to properly encode data that is recognized by the registered decoders. 
* The model provider deocoder is provided with a context about the source of the origination of the model, so that only relevant provider is initialized. For example: an AppSyncModelProvider should not be initialized for a DataStore subscription event. 
* The `AWSPluginOptions` has been deprecated and marked as internal use only. Furthermore, a new `AWSDataStorePluign` option has been created that will act as a type to know about DataStore usecases. This plugin option has also been marked as internal. 

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginLazyLoadTests/LL2/AWSDataStoreLazyLoadBlogPostComment8V2Tests.swift
+++ b/AmplifyPlugins/DataStore/Tests/DataStoreHostApp/AWSDataStorePluginLazyLoadTests/LL2/AWSDataStoreLazyLoadBlogPostComment8V2Tests.swift
@@ -322,6 +322,62 @@ final class AWSDataStoreLazyLoadBlogPostComment8V2Tests: AWSDataStoreLazyLoadBas
         await fulfillment(of: [snapshotReceived], timeout: 60)
         querySnapshots.cancel()
     }
+
+    /// Test retrieving and loading of a Post from a deleted comment in DataStore.observe
+    ///
+    /// - Given:  A configured DataStore and API plugin, with Post and Comment model, which have been
+    ///     created and synced.
+    /// - When:
+    ///    - Call mutate API to delete an existing comment on a post
+    /// - Then:
+    ///    - The operation completes successfully with no errors
+    ///    - The observe query should be fired with the relevant mutation event
+    ///    - The model should be decoded from the mutation event
+    ///    - Lazy reference of the post should retrievable and loadable.
+    ///
+    func testObserveToQueryPostFromDeletedComment() async throws {
+        await setup(withModels: BlogPostComment8V2Models())
+        try await startAndWaitForReady()
+        let post = Post(name: "name", randomId: "randomId")
+        _ = try await createAndWaitForSync(post)
+        let comment = Comment(content: "content", post: post)
+        let receivedComment = try await createAndWaitForSync(comment)
+        let mutationEventReceived = expectation(description: "Received mutation event")
+        let mutationEvents = Amplify.DataStore.observe(Comment.self)
+        Task {
+            for try await mutationEvent in mutationEvents {
+                if let receivedComment = try? mutationEvent.decodeModel(as: Comment.self),
+                   receivedComment.id == receivedComment.id {
+                    assertLazyReference(receivedComment._post,
+                                        state: .notLoaded(identifiers: [.init(name: "id", value: post.identifier)]))
+                    guard let loadedPost = try await comment.post else {
+                        XCTFail("Failed to load the post from the comment")
+                        return
+                    }
+                    XCTAssertEqual(loadedPost.id, post.id)
+                    assertLazyReference(comment._post,
+                                        state: .loaded(model: post))
+                    mutationEventReceived.fulfill()
+                } else {
+                    XCTFail("The model should be correctly decoded")
+                }
+            }
+        }
+
+        let deleteRequest = GraphQLRequest<MutationSyncResult>.deleteMutation(
+            of: comment,
+            modelSchema: Comment.schema,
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/API/Tests/AWSAPIPluginTests/Support/Decode/GraphQLResponseDecoderLazyPostComment4V2Tests.swift
```diff
@@ -376,7 +376,12 @@ class GraphQLResponseDecoderLazyPostComment4V2Tests: XCTestCase, SharedTestCases
             ]
         ]
 
+        let expectedEncodedCommentModel = """
+        {"content":"content","createdAt":null,"id":"id","post":{"identifiers":[{"name":"id","value":"\(post.id)"}],"source":"AppSync"},"updatedAt":null}
+        """
```
lawmicha, line 381:
```suggestion
let expectedEncodedCommentModel = """
{
    "content": "content",
    "createdAt": null,
    "id": "id",
    "post": {
        "identifiers": [
            {
                "name": "id",
                "value": "\(post.id)"
            }
        ],
        "source": "AppSync"
    },
    "updatedAt": null
}
"""
```

AmplifyPlugins/Core/AWSPluginsCore/AWSPluginOptions.swift
```diff
@@ -9,15 +9,53 @@ import Foundation
 import Amplify
 
 /// Plugin specific options type
-public struct AWSPluginOptions {
+///
+/// - Warning: Although this has `public` access, it is intended for internal use and should not be used directly
+///   by host applications. The behavior of this may change without warning.
+///   // OR SPI
+public struct AWSDataStorePluginOptions {
```
lawmicha, line 16: Thoughts on the naming of this? It still feels a bit off, I would expect this type to be used with DataStore (even though DataStore provides no types that take in this type, nor `Any` type. I think something more specific to AWSAPIPlugin would be nice, like `AWSAPIPluginOptionsForDataStore` / `AWSAPIPluginDataStoreOptions` ?

AmplifyPlugins/Core/AWSPluginsCore/AWSPluginOptions.swift
```diff
@@ -9,15 +9,53 @@ import Foundation
...
+
+    public static let DataStoreMetadataSource = "DataStore"
```
lawmicha, line 18: I wish there was some surrounding context on this, can we add an extension to a type coming from Amplify Core, and add this static string? Something like
```swift
extension LazyReference {
  public static let DataStoreModelDecoderSource = "DataStore"
}
```

AmplifyPlugins/Core/AWSPluginsCore/AWSPluginOptions.swift
```diff
@@ -9,15 +9,53 @@ import Foundation
...
+///        self.authType = authType
+///    }
+///}
+///```
+@available(*, deprecated, message: "Intended for internal use.")
+public struct AWSPluginOptions {
+    public let authType: AWSAuthorizationType?
+    public let modelName: String?
+    public init(authType: AWSAuthorizationType?,
+                modelName: String) {
```
lawmicha, line 58: [nit-from-my-refactor] Can we keep the property comments as before? ie.
```
   /// authorization type
   public let authType: AWSAuthorizationType?
   /// name of the model
   public let modelName: String?
 ```

File: AmplifyPlugins/Core/AWSPluginsCore/AWSPluginOptions.swift (hunk 2 above)
Comment (lawmicha) [reply AmplifyPlugins/Core/AWSPluginsCore/AWSPluginOptions.swift:16]: we should make a new file for this struct as well

File: AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Storage/StorageEngineTestsLazyPostComment4V2Tests.swift [hunk 5]
```diff
@@ -257,6 +257,11 @@ final class StorageEngineTestsLazyPostComment4V2Tests: StorageEngineTestsBase, S
             XCTFail("Failed to query saved comment")
             return
         }
+        let expectedJSONOutput = """
+        {\"content\":\"content\",\"createdAt\":null,\"id\":\"\(comment.id)\",\"post\":{\"identifiers\":[{\"name\":\"id\",\"value\":\"\(post.id)\"}],\"source\":\"DataStore\"},\"updatedAt\":null}
+        """
```
lawmicha, line 262:
```suggestion
let expectedJSONOutput = """
{
    "content": "content",
    "createdAt": null,
    "id": "\(comment.id)",
    "post": {
        "identifiers": [
            {
                "name": "id",
                "value": "\(post.id)"
            }
        ],
        "source": "DataStore"
    },
    "updatedAt": null
}
"""

```

lawmicha (changes requested):
Can we update the description with some details of the issue and fix? We should add an integration test to `AWSDataStoreLazyLoadBlogPostComment8V2Tests`. there's some existing tests like `AWSDataStoreLazyLoadBlogPostComment8V2Tests.testDelete`, we can write a new one that observes MutationEvents, decode to the model, and lazy load.

lawmicha (approved):
Looks great!

PR #3527, opened 2024-02-16
feat(API): implement AppSyncRealTimeClient and WebSocketClient with URLSession

5d:
## Description

- Removing the dependency of [aws-appsync-realtime-client-ios](https://github.com/aws-amplify/aws-appsync-realtime-client-ios). 
- Utlizing [`URLSessionWebSocketTask`](https://developer.apple.com/documentation/foundation/urlsessionwebsockettask) for AppSync real-time WebSocket connection.
- Reimplemented the `AppSyncRealTimeClient`.

## General Checklist

- [x] All unit tests pass


AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
```diff
@@ -0,0 +1,457 @@
...
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+
+import Foundation
+import Amplify
+import Combine
+@_spi(AmplifySwift) import AWSPluginsCore
```
lawmicha, line 12: should the identifier be more specific? I see a few other ones named `InternalAWSPinpoint`, `LazyReference`, `PredictionsConvertRequestKind`, `KeyChainStore`, how about `Websocket` or `WebsocketClient`? We could technically move the code under the Websocket folder from AWSPluginsCore over to Amplify as public types, just something to consider, not necessarily proposing us to do that in this PR

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
```diff
@@ -0,0 +1,457 @@
...
+    /**
+     Subscribing to a query with unique identifier.
+     - Parameters:
+        - id: unique identifier
+        - query: GraphQL query for subscription
+
+     -  Returns:
+        A never fail data stream for AppSyncSubscriptionEvent.
+     */
+    func subscribe(id: String, query: String) async throws -> AnyPublisher<AppSyncSubscriptionEvent, Never> {
```
lawmicha, line 158: is it typical to accept the `id` instead of generating it internally to the call? Would generating it internally and return it as ` (id: String, publisher: AnyPublisher<AppSyncSubscriptionEvent, Never>)` be better?

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
```diff
@@ -0,0 +1,457 @@
...
+        log.debug("[AppSyncRealTimeClient] Received subscription request id: \(id), query: \(query)")
+        let subscription = AppSyncRealTimeSubscription(id: id, query: query, endpoint: endpoint)
+        subscriptions[id] = subscription
+
+
+        // Placing the actual subscription work in a deferred task and
+        // promptly returning the filtered publisher for downstream consumption of all error messages.
+        defer {
```
lawmicha, line 166: I'm trying to understand the point of `defer` here, does it actually do anything? for example is`defer` block executed before or after this line?
```
return filterAppSyncSubscriptionEvent(with: id)
            .merge(with: (await subscription.publisher).toAppSyncSubscriptionEventStream())
            .eraseToAnyPublisher()
```

If executed after the line, then the code execution would look like this:
```
let publisher = filterAppSyncSubscriptionEvent(with: id)
            .merge(with: (await subscription.publisher).toAppSyncSubscriptionEventStream())
            .eraseToAnyPublisher()

// run defer block

return publisher
```

if executed before, then the `defer` usage seems unnecessary

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
```diff
@@ -0,0 +1,457 @@
...
+            } catch {
+                log.debug("[AppSyncRealTimeClient] Failed to send connectInit message, error: \(error)")
+            }
+        }
+    }
+
+    private func filterAppSyncSubscriptionEvent(
+        with id: String
+    ) -> AnyPublisher<AppSyncSubscriptionEvent, Never> {
+        subject.filter { $0.id == id || $0.type == .connectionError }
```
lawmicha, line 295: some comments for this func would be nice to understand why we filter on `connectionError` types as well. Is it because the event has no `id` so the filter should be on the relevant`id`  OR any `connectionError` events?

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
5d: The order is important here for downstream to receive all the related events. `defer` here is to ensure the block of code start running after this function returns. 

The suggested code snippet should be working as well. As the `subscribe` API is marked as async as well, I just suspect there is a small chance that the defer Task could be ran simultaneously with `suscribe`.

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
5d: We could generate it internally. Is there any benefit to do it internally? The subscriber need the id for unsubscribing. Handling the tuple returns is cumbersome than passing an exsisting one.

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
5d: yes, it's because of `connection_error` doesn't have an `id` field and I want to pass the connection_error to downstreams.

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
5d: sure, we can use a more specific name. `WebSocket` sounds good to me.

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
lawmicha: discussed, since it is used internally, passing `id` simplifies the return type greatly

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeRequestAuth.swift
```diff
@@ -0,0 +1,127 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+
+import Foundation
+
+public enum AppSyncRealTimeRequestAuth {
+    case cognito(CognitoAuth)
+    case apiKey(ApiKey)
+    case iam(IAM)
```
lawmicha, line 14: are we missing AWS_Lambda auth or OIDC here? Or is OIDC auth using `.cognito(CogniotAuth)`? technically Cognito should just be treated as an OIDC provider, as that's what it is

AmplifyPlugins/API/Tests/AWSAPIPluginTests/AppSyncRealTimeClient/AppSyncRealTimeRequestAuthTests.swift
```diff
@@ -0,0 +1,215 @@
...
+
+import XCTest
+@testable import AWSAPIPlugin
+
+let host = UUID().uuidString
+let apiKey = UUID().uuidString
+let date = UUID().uuidString
+let id = UUID().uuidString
+let data = UUID().uuidString
+let token = UUID().uuidString
```
lawmicha, line 17: are these used elsewhere? if not, prefer scoping to where they are used, under AppSyncRealTimeRequestAuthTests ?

AmplifyPlugins/Core/AWSPluginsCore/WebSocket/AmplifyNetworkMonitor.swift
```diff
@@ -0,0 +1,55 @@
...
+public final class AmplifyNetworkMonitor {
+
+    public enum State {
+        case none
+        case online
+        case offline
+    }
+
+    #if DEBUG
+    private let monitor = NWPathMonitor(requiredInterfaceType: .cellular)
```
lawmicha, line 22: leaving a comment here so we don't lose track. I don't think we should be setting only `celluar` unless this was a specific way to test this in DEBUG mode. But removing the hardcode means we should provide a way to test it, such as being able to pass `.cellular` to AmplifyNetworkMonitor for testing. Can we pass it into the `init` as an `AmplifyNetworkMonitor. InterfaceType` ?

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeRequestAuth.swift
5d: yeah, as `AWS_Lambda`, `OIDC` and `CognitoUserPools` are all providing the way to get the authToken, they provids the same auth payload. So I [unified them](https://github.com/aws-amplify/amplify-swift/pull/3527/files#diff-8b7d0ea3a610253777061a06962e1d7bc43b6d3a266892b8102d6c053ccc4bc4R94-R118) at auth payload level. Do you think that is reasonable?

AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeRequest.swift
```diff
@@ -0,0 +1,104 @@
...
+}
+
+
+extension AppSyncRealTimeRequest {
+    public enum Error: Swift.Error {
+        case timeout
+        case limitExceeded
+        case maxSubscriptionsReached
+        case unauthorized
+        case unknown
```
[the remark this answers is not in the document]
5d: yeah, will add additional info in the `unknown` case.

AmplifyPlugins/API/Tests/AWSAPIPluginTests/AppSyncRealTimeClient/AppSyncRealTimeRequestAuthTests.swift
5d: will move these inside Test class.

AmplifyPlugins/Core/AWSPluginsCore/WebSocket/AmplifyNetworkMonitor.swift
5d: 👍 will add the network interfaces in the injection way.

Issue #3549, opened 2024-03-05
DataStore cannot connect to model's sync subscriptions (AWS_LAMBDA auth type)


MuniekMg:
### Describe the bug

For some of our models we are using AWS_LAMBDA auth type. ( e.g. 'HomeModel')

Every operation on this model is authorized by our custom lambda including onCreate, onUpdate and onDelete subscriptions requests.
Currently Amplify just sends request to subscribe to HomeModel (without filter/predicate) and our lambda cannot authorize this. 
Instead Amplify should add DataStore syncExpressions as request's filter, e.g. subscribe to HomeModel onCreate events where HomeModel.address is equal to 'myAddress'

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1. Start Application
2. See error: ConnectionProviderError.unauthorized
```

### Expected behavior

DataStore add syncExpressions as subscriptions request filter

### Amplify Framework Version

2.27.0

### Amplify Categories

DataStore

### Dependency manager

Swift PM

### Swift version

5

### CLI version

10.3.1

### Xcode version

Version 15.2 (15C500b)

### Relevant log output

```shell
ConnectionProviderError.unauthorized
ConnectionProviderError.unauthorized
ConnectionProviderError.unauthorized
ConnectionProviderError.unauthorized
ConnectionProviderError.unauthorized
ConnectionProviderError.unauthorized
ConnectionProviderError.unauthorized
```
```


### Is this a regression?

No

### Regression additional context

_No response_

### Platforms

iOS

### OS Version

iOS17

### Device

iPad 10th gen

### Specific to simulators

_No response_

### Additional context

_No response_

5d:
HI @MuniekMg ,

Thank you for opening the issue and submitting the patch PR. We'll look into both the issue and the proposed implementation and will resume the conversation when we have any updates or need further information.

PR #3550, opened 2024-03-05
fix(DataStore): dataStore cannot connect to model's sync subscriptions (AWS_LAMBDA auth type) 

MuniekMg:
Bug: #3549 

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/Amplify/Categories/API/Operation/RetryableGraphQLOperation.swift
+++ b/Amplify/Categories/API/Operation/RetryableGraphQLOperation.swift
@@ -170,7 +170,7 @@ public final class RetryableGraphQLSubscriptionOperation<Payload: Decodable>: Op
     public var errorListener: OperationErrorListener
     public var resultListener: OperationResultListener
     public var operationFactory: OperationFactory
-    private var filterLimitRetried: Bool = false
+    private var retriedRTFErrors: [RTFError: Bool] = [:]
     
     public init(requestFactory: @escaping RequestFactory,
                 maxRetries: Int,
@@ -195,9 +195,7 @@ public final class RetryableGraphQLSubscriptionOperation<Payload: Decodable>: Op
     }
 
     public func shouldRetry(error: APIError?) -> Bool {
-//        return attempts < maxRetries
-        
-        guard case let .operationError(errorDescription, recoverySuggestion, underlyingError) = error else {
+        guard case let .operationError(_, _, underlyingError) = error else {
             return false
         }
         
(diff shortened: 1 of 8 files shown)


lawmicha:
Hi @MuniekMg, thanks for opening this PR. I believe some AppSync backends were provisioned without the `filter` input available on subscriptions, which is why client side filtering with the sync expression was done on all subscription events coming back instead of translating the sync expression to the filter input for the subscription request. 

Most likely we will have to introduce a new configuration property on `DataStoreConfiguration` such as `subscriptionFiltering: Bool` to allow developers to opt into this feature.

lawmicha:
@MuniekMg, I editted my message to mention i'll check with the team, maybe you missed it since I had editted sometime later, sorry for the extra work! I was able to check with the team and found out the approach that Amplify JS library is taking. 

1. We attempt the subscription with the filter and catch ther error
2. Handle the specific "RTF" error as a retry without the filter for backwards compatibility.

References
https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/sync/processors/subscription.ts#L467-L492

https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/sync/processors/subscription.ts#L677-L682C10

I think the approach of adding a new configuration flag will require documentation and developers to know about it. This feature was launched in April 2022 so all new customers would have to set the flag which would be a pretty poor DX. Retrying without the filter for the customers with the old provisioning appears to be the performance trade-off we are willing to make.

MuniekMg:
@lawmicha  I have removed my last commit a and crated new one with a possible solution (work in progress) for retrying subscriptions request (but without filter) but it looks a little messy, and I'm not sure whether I should continue this approach. 
Could you maybe help me choose a better solution?

Amplify/Categories/API/Operation/RetryableGraphQLOperation.swift
```diff
@@ -178,9 +195,35 @@ public final class RetryableGraphQLSubscriptionOperation<Payload: Decodable>: Op
...
+            switch authError {
+            case .signedOut, .notAuthorized:
+                return attempts < maxRetries
+            default:
+                return false
+            }
+        }
+        
+        // TODO: - How to distinguish errors?
+        // TODO: - Handle other errors
```
lawmicha, line 214: looks like there is a whole list of errors to catch based on the strings https://github.com/aws-amplify/amplify-js/blob/main/packages/datastore/src/sync/processors/subscription.ts#L692-L698

Amplify/Categories/API/Operation/RetryableGraphQLOperation.swift
```diff
@@ -178,9 +195,35 @@ public final class RetryableGraphQLSubscriptionOperation<Payload: Decodable>: Op
...
+        if error.debugDescription.contains("Filters combination exceed maximum limit 10 for subscription.") &&
+            filterLimitRetried == false {
+            
+            // Just to be sure that endless retry won't happen
+            filterLimitRetried = true
+            maxRetries += 1
```
lawmicha, line 220: is `maxRetries` updated here to account for the `attempt` being incremented for the RTF retry case? So the overall `attempt` to `maxRetries` counts will continue to represent the number of auth types to attempt? (maxRetries is set to the number of auth types to attempt in multi-auth rules scenarios).

MuniekMg:
> Sorry for the delay, we're working on migrating the `Amplify.API` API calls to the async await version, which will drastically change the outer components like the RetryableGraphQLSubscriptionOperation (cc: @5d) We should consider this use case when designing the new retryable mechanism. The code appears to be on the right track, I'll leave some comments in line so we don't lose track of any necessary changes.

@lawmicha No problem :) I have just finished implementing the same logic like in JS implementation (reconnecting in case of any RTF error) and added more tests. It would be great if You could merge those changes while waiting for your new implementation

Issue #3553, opened 2024-03-06
Created and deleted model on one device appears as created on the other


MuniekMg:
### Describe the bug

This is the same bug as [2982](https://github.com/aws-amplify/amplify-swift/issues/2982)
This bug was recently closed and solved but I still can reproduce this issue

### Steps To Reproduce

```swift
1. We have two phones: PhoneA and PhoneB
2. App is open only on PhoneA
3. on PhoneA add model and then remove this model (DataStore)
4. Open app on PhoneB
5. PhoneB will receive onCreate but not onDelete
```

### Expected behavior

in step 5 phoneB should receive:
- receive onCreate but then onDelete
or
- receive nothing

### Amplify Framework Version

2.27.0

### Amplify Categories

DataStore

### Dependency manager

Swift PM

### Swift version

5

### CLI version

12.7.0

### Xcode version

Version 15.2 (15C500b)

MuniekMg:
@5d This is how model looks in DynamoDB table:

"id","__typename","_deleted","_lastChangedAt","_ttl","_version","address","createdAt","name","owner","updatedAt"

"20DE73FE-C4D3-4E91-90AF-E52DCA7952D2",
"Space",
"true",
"1709749749259",
"1712341749",
"2",
"bbb",
"2024-03-06T18:29:04.110Z",
"aaa",
"ca740603-9c2c-4e44-9e69-1adcbfe0cd8c::ca740603-9c2c-4e44-9e69-1adcbfe0cd8c",
"2024-03-06T18:29:04.110Z"

5d:
Thanks @MuniekMg,

The record did mark as deleted. I've reviewed PR https://github.com/aws-amplify/amplify-swift/pull/3554. LGTM

PR #3554, opened 2024-03-06
fix(DataStore): created and deleted model on one device appears as created on the other

MuniekMg:
Bug: #3553 

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/RemoteSyncReconciler.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/RemoteSyncReconciler.swift
@@ -43,7 +43,43 @@ struct RemoteSyncReconciler {
     static func getDispositions(_ remoteModels: [RemoteModel],
                                 localMetadatas: [LocalMetadata]) -> [Disposition] {
         
-        let remoteModelsGroupedById = remoteModels.reduce(into: [:]) { $0[$1.model.id] = ($0[$1.model.id] ?? []) + [$1] }
+    /// Reconciles the incoming `remoteModels` against the local metadata to get the disposition
+    ///
+    /// GroupBy the remoteModels by model identifier and apply only the latest version of the remoteModel
+    ///
+    /// - Parameters:
+    ///   - remoteModels: models retrieved from the remote store
+    ///   - localMetadatas: metadata retrieved from the local store
+    /// - Returns: disposition of models to apply locally
+    static func getDispositions(
+        _ remoteModels: [RemoteModel],
+        localMetadatas: [LocalMetadata]
+    ) -> [Disposition] {
+        let remoteModelsGroupByIdentifier = remoteModels.reduce([String: [RemoteModel]]()) {
+            $0.merging([
+                $1.model.identifier: ($0[$1.model.identifier] ?? []) + [$1]
+            ], uniquingKeysWith: { $1 })
+        }
+
+        let optimizedRemoteModels = remoteModelsGroupByIdentifier.values.compactMap {
+            $0.sorted(by: { $0.syncMetadata.version > $1.syncMetadata.version }).first
+        }
+
+        guard !optimizedRemoteModels.isEmpty else {
+            return []
+        }
+
+        guard !localMetadatas.isEmpty else {
+            return optimizedRemoteModels.compactMap { getDisposition($0, localMetadata: nil) }
+        }
+
+        let metadataBymodelId = localMetadatas.reduce(into: [:]) { $0[$1.modelId] = $1 }
+        let dispositions = optimizedRemoteModels.compactMap {
+            getDisposition($0, localMetadata: metadataBymodelId[$0.model.identifier])
+        }
+
+        return dispositions
+    }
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/RemoteSyncReconciler.swift
```diff
@@ -42,18 +42,35 @@ struct RemoteSyncReconciler {
     /// - Returns: disposition of models to apply locally
     static func getDispositions(_ remoteModels: [RemoteModel],
                                 localMetadatas: [LocalMetadata]) -> [Disposition] {
-        guard !remoteModels.isEmpty else {
-            return []
-        }
-
-        guard !localMetadatas.isEmpty else {
-            return remoteModels.compactMap { getDisposition($0, localMetadata: nil) }
```
MuniekMg, line 50: This line was causing the bug:

here we were geting an empty localMetadatas and remoteModels with two elements for the same model (the same ID):
eg. [
    ModelA - id: aaa, version: 1, deleted: false,
    ModelA - id: aaa, version: 2, deleted: true
]

and then "deteled true disposition" was dropped so only "create disposition" was returned

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/RemoteSyncReconciler.swift
```diff
@@ -42,18 +42,35 @@ struct RemoteSyncReconciler {
...
+        
+        let remoteModelsGroupedById = remoteModels.reduce(into: [:]) { $0[$1.model.id] = ($0[$1.model.id] ?? []) + [$1] }
```
MuniekMg, line 46: Why we get multiple remoteModels for the same instance? Maybe we should just take remoteModel with highest version for each model instance and ignore the rest?

5d: yeah, take the hightest version sounds good to me.

here is the suggest code:
```suggestion
    /// Reconciles the incoming `remoteModels` against the local metadata to get the disposition
    ///
    /// GroupBy the remoteModels by model identifier and apply only the latest version of the remoteModel
    ///
    /// - Parameters:
    ///   - remoteModels: models retrieved from the remote store
    ///   - localMetadatas: metadata retrieved from the local store
    /// - Returns: disposition of models to apply locally
    static func getDispositions(
        _ remoteModels: [RemoteModel],
        localMetadatas: [LocalMetadata]
    ) -> [Disposition] {
        let remoteModelsGroupByIdentifier = remoteModels.reduce([String: [RemoteModel]]()) {
            $0.merging([
                $1.model.identifier: ($0[$1.model.identifier] ?? []) + [$1]
            ], uniquingKeysWith: { $1 })
        }

        let optimizedRemoteModels = remoteModelsGroupByIdentifier.values.compactMap {
            $0.sorted(by: { $0.syncMetadata.version > $1.syncMetadata.version }).first
        }

        guard !optimizedRemoteModels.isEmpty else {
            return []
        }

        guard !localMetadatas.isEmpty else {
            return optimizedRemoteModels.compactMap { getDisposition($0, localMetadata: nil) }
        }

        let metadataBymodelId = localMetadatas.reduce(into: [:]) { $0[$1.modelId] = $1 }
        let dispositions = optimizedRemoteModels.compactMap {
            getDisposition($0, localMetadata: metadataBymodelId[$0.model.identifier])
        }

        return dispositions
    }
```

5d (approved):

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/SubscriptionSync/ReconcileAndLocalSave/RemoteSyncReconciler.swift
lawmicha: Let me try to summarize this, the reconciliation path for DataStore takes the sync query remote model results and decides to apply the remote model to the local database or drop it. Previously the logic processed one model at a time. We changed this to a batch operation by putting the reconciliation logic under a transaction block, operating on the entire list of remote models. The necessary logic updates to perform the SQL operations more efficiently included constructing a predicate of all the model identifiers as a filter criteria for querying local data that is used for the reconciliation logic. There exists a bug with this optimization when having multiple entries of remote model belong to the same identifier and the changes to the local model ends up as a deleted model. 

Bug example
remoteModels = `[id=111 version=1 deleted=false, id=111 version=2 deleted=true]`

When the disposition of the model is calculated for the create model, it is to create the model with the local database, because there is no local model, nor is there a local model with version less than the remote model.
When the disposition of the model is calculated for the delete model, it is to drop it because no local model yet to exist, so it thinks there is nothing to delete.

The end result is the model is created but not deleted. The fix we are applying in the PR is to optimize the reconciliation by only taking the latest versions of the remote model. In the example, this becomes `[id=111 version=2, deleted=true]`. The latest version represents the final state of the model that we want applied to the local database. In this case, the remote model will be dropped. There is no create disposition so it will not be created either.

Issue #3556, opened 2024-03-07
TaskQueue<Success> async execution is not serial


MuniekMg:
### Describe the bug

While working on other PR I found out that one unit test sometimes fails. This test expected correct order of function executions. 

TaskQueue<Success> has async(...) function which should schedule task and execute this task in serial manner but it doesn't.

### Steps To Reproduce

```swift
1. Add 1000 task to TaskQueue<Void>.async { print(i) }
2. Tasks won't be executed in order
```

### Expected behavior

Tasks are being executed in order

### Amplify Framework Version

2.27.0

### Amplify Categories

DataStore

### Dependency manager

Swift PM

### Swift version

5

### CLI version

10.3.1

### Xcode version

Version 15.2 (15C500b)

### Is this a regression?

No

### Platforms

iOS

### OS Version

iOS 17

### Device

iPad 10th gen

5d:
Hi @MuniekMg ,

Thanks for reporting the issue and submitting the proposed PR. Our team will investigate and review the PR.

5d:
The PR has been released in version 2.29.2. We are grateful for your valuable contribution.

PR #3557, opened 2024-03-07
fix(Core): TaskQueue<Success> async execution is not serial #3556

MuniekMg:
## Issue \#
#3556 

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/Amplify/Core/Support/TaskQueue.swift
+++ b/Amplify/Core/Support/TaskQueue.swift
@@ -8,11 +8,26 @@
 import Foundation
 
 /// A helper for executing asynchronous work serially.
-public actor TaskQueue<Success> {
-    private var previousTask: Task<Success, Error>?
-
-    public init() {}
-
+public class TaskQueue<Success> {
+    typealias Block = @Sendable () async -> Void
+    private var streamContinuation: AsyncStream<Block>.Continuation!
+       
+    public init() {
+        let stream = AsyncStream<Block>.init { continuation in
+            streamContinuation = continuation
+        }
+        
+        Task {
+            for await block in stream {
+                _ = await block()
+            }
+        }
+    }
+    
+    deinit {
+        streamContinuation.finish()
+    }
+    
     /// Serializes asynchronous requests made from an async context
     ///
     /// Given an invocation like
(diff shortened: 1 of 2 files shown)


Amplify/Core/Support/TaskQueue.swift
```diff
@@ -25,17 +40,22 @@ public actor TaskQueue<Success> {
...
+                }
+            }
         }
-        previousTask = currentTask
-        return try await currentTask.value
     }
 
-    public nonisolated func async(block: @Sendable @escaping () async throws -> Success) rethrows {
-        Task {
-            try await sync(block: block)
```
MuniekMg, line 38: This line was causing this bug.

New Task { } here will be first executed on global executor and then after a while will 'hop' to an actor isolation.
This time between 'global executor' and 'actor isolation' was enough for other task to execute out of order

Amplify/Core/Support/TaskQueue.swift
```diff
@@ -25,17 +40,22 @@ public actor TaskQueue<Success> {
...
+    public func async(block: @Sendable @escaping () async -> Success) {
```
5d, line 56: Has the removal of the `throws` keyword from the `block` closure's signature been intentional?

MuniekMg: Yes, because there is no way to throw an error from his function to a caller outside of this class as this function is async. `throws` keyword was indicating that something will catch/rethrow an eventual error but it will won't.

The more I think about it, the more I come to a conclusion that maybe this `throws` was not that bad, but we need to catch eventual errors ourselves inside this function and log them as error?

5d: Certainly. Yes, this asynchronous closure operates on a fire-and-forget basis. Errors occurring within it won't be propagated back to the invoker. 
Since this function is first designed as a public API, I'm particularly concerned about any changes to its signature.

MuniekMg: Exactly. 
Of course your concerns are understandable

5d:
I'll merge this PR into branch `5d/pr-3357` first and then create some fix over the PR before submitting it to the main branch

MuniekMg:
> I'll merge this PR into branch `5d/pr-3357` first and then create some fix over the PR before submitting it to the main branch

@5d Great, thanks :)

PR #3602, opened 2024-04-09
fix(api): add collection type casting in swift 5.7

5d:
## Description

Swift Collection type casting is introduced in `Swift 5.8`. As our library is still supporting `Swift 5.7`, adding backward compatibility support for collection casting.

## General Checklist

- [X] All unit tests pass

### DIFF
--- a/AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSGraphQLSubscriptionTaskRunner.swift
+++ b/AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSGraphQLSubscriptionTaskRunner.swift
@@ -379,14 +379,14 @@ fileprivate func toAPIError<R: Decodable>(_ errors: [Error], type: R.Type) -> AP
     }
 
 #if swift(<5.8)
-    if let errors = castArray(errors, withElementType: AppSyncRealTimeRequest.Error.self) {
+    if let errors = errors.cast(to: AppSyncRealTimeRequest.Error.self) {
         let hasAuthorizationError = errors.contains(where: { $0 == .unauthorized})
         return APIError.operationError(
             errorDescription(hasAuthorizationError),
             "",
             errors.first
         )
-    } else if let errors = castArray(errors, withElementType: GraphQLError.self) {
+    } else if let errors = errors.cast(to: GraphQLError.self) {
         let hasAuthorizationError = errors.map(\.extensions)
             .compactMap { $0.flatMap { $0["errorType"]?.stringValue } }
             .contains(where: { AppSyncErrorType($0) == .unauthorized })
@@ -429,14 +429,3 @@ fileprivate func toAPIError<R: Decodable>(_ errors: [Error], type: R.Type) -> AP
     }
 #endif
 }
-
-#if swift(<5.8)
-fileprivate func castArray<T>(_ arr: Array<Any>, withElementType: T.Type) -> [T]? {
-    arr.reduce([]) { result, ele in
-        if let result, let ele = ele as? T {
-            return result + [ele]
-        }
-        return nil
-    }
-}
-#endif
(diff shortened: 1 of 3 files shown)


phantumcode (dismissed):

AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSGraphQLSubscriptionTaskRunner.swift
```diff
@@ -378,6 +378,31 @@ fileprivate func toAPIError<R: Decodable>(_ errors: [Error], type: R.Type) -> AP
...
+            GraphQLResponseError<R>.error(errors)
+        )
+    } else {
+        return APIError.operationError(
+            errorDescription(),
+            "",
+            errors.first
+        )
+    }
+#else
```
lawmicha, line 405: looks like you can probably reduce this once more
```swift
    let hasAuthorizationError: Bool
    let underlyingError: Error?
#if swift(<5.8)
    if let errors = errors.cast(to: AppSyncRealTimeRequest.Error.self) {
        hasAuthorizationError = errors.contains(where: { $0 == .unauthorized})
        underlyingError = errors.first
    } else if let errors = errors.cast(to: GraphQLError.self) {
        hasAuthorizationError = errors.map(\.extensions)
            .compactMap { $0.flatMap { $0["errorType"]?.stringValue } }
            .contains(where: { AppSyncErrorType($0) == .unauthorized })
        underlyingError = GraphQLResponseError<R>.error(errors)
    } else {
        hasAuthorizationError = false
        underlyingError = errors.first
    }
#else
    if let errors = errors as? [AppSyncRealTimeRequest.Error] {
        hasAuthorizationError = errors.contains(where: { $0 == .unauthorized})
        underlyingError = errors.first
    } else if let errors = errors as? [GraphQLError] {
        hasAuthorizationError = errors.map(\.extensions)
            .compactMap { $0.flatMap { $0["errorType"]?.stringValue } }
            .contains(where: { AppSyncErrorType($0) == .unauthorized })
        underlyingError = errors.first
    } else {
        hasAuthorizationError = false
        underlyingError = errors.first
    }
#endif
    return APIError.operationError(
        errorDescription(hasAuthorizationError),
        "",
        underlyingError
    )
```

Or even further, just use `cast` everywhere. This would avoid having to duplicate business logic while maintaining the Collection type casting in one place. Once we support 5.8 and up, then we can replace usages of `cast` with `as`
```
func cast<T>(to type: T.Type) -> [T]? {
        #if swift(<5.8)
        self.reduce([]) { partialResult, ele in
            if let partialResult, let ele = ele as? T {
                return partialResult + [ele]
            }
            return nil
        }
        #else
        return self as? [T]
        #endif
    }
```

lawmicha (approved):
minor refactoring comment, let me know if you need another approval if adopting it

AmplifyPlugins/API/Sources/AWSAPIPlugin/Operation/AWSGraphQLSubscriptionTaskRunner.swift
ruisebas: I like the idea of having the compiler directive in the `cast` function instead

5d: Including the directive in the `cast` func will make unit testing harder. Once we upgrade the Swift tool version to 5.8, removing the code within the annotation will become straightforward.

ruisebas: Why would it make unit testing harder? We don't need to unit-test that we support both 5.7 and 5.8

lawmicha: Chatted with @5d, the CI set up will cause `cast` to be unit tested against Swift 5.9 so it won't reach the code path. We'd have to refactor the unit test to test against something like:
```swift
func cast<T>(to type: T.Type) -> [T]? {
        #if swift(<5.8)
        return castSwift5_7() // update unit tests to test this function instead
        #else
        return self as? [T]
        #endif
    }
```

We to upgrade to min 5.8 soon so as Di mentioned it will be easier to remove the code later in the current PR

Issue #3677, opened 2024-05-06
Amplify API Unathorized error Error


dan-codes1:
### Describe the bug

I have been experiencing some graphql api bugs since I upgraded my amplify version from 2.26.1 to 2.29.2. 
The error occurs when I try to asynchronously load the lazy attibutes in my structs. E.g., take these schemas:

```graphql
type Message @model @auth(rules: [{allow: private, operations: [read]}, {allow: owner, operations: [create, update, delete]}]) {
    id: ID!
    type: MessageType!
    text: String
    sentAt: AWSDateTime!
    isRead: Boolean!
    sender: User! @hasOne
    space: MessageSpace! @hasOne
    editedAt: AWSDateTime
    modelString: String! @index(name: "messagesByDate", queryField: "messagesByDate", sortKeyFields: ["sentAt"])
    modelStatus: ModelStatus!
}

type MessageSpace @model @auth(rules: [{allow: private, operations: [create, read, update]}]) {
    id: ID!
    userIds: [String!]!
    type: MessageSpaceType!
    lastSentMessageId: String
    hasMessages: Boolean!
    lastModified: AWSDateTime!
    messages: [Message] @hasMany
    modelString: String! @index(name: "messagesSpacesByDate", queryField: "messagesSpacesByDate", sortKeyFields: ["lastModified"])
    modelStatus: ModelStatus!
}
```
When I try to load the `space` attribute from the `message`, it throws this error:
```bash
GraphQLResponseError<Optional<MessageSpace>>: GraphQL service returned a successful response containing errors: [Amplify.GraphQLError(message: "Not Authorized to access getMessageSpace on type Query", locations: Optional([Amplify.GraphQLError.Location(line: 2, column: 3)]), path: Optional([Amplify.JSONValue.string("getMessageSpace")]), extensions: Optional(["errorInfo": Amplify.JSONValue.null, "data": Amplify.JSONValue.null, "errorType": Amplify.JSONValue.string("Unauthorized")]))]
Recovery suggestion: The list of `GraphQLError` contains service-specific messages
```

This is the code and api I use:
```swift
let request = try await Amplify.API.query(request: .get(Message.self, byId: id, authMode: .amazonCognitoUserPools))
let message = try request.get() 
let sender = try await message.sender
let space = try await message.space
```
It throws an error when I try to get the space but not the sender. This only started when I upgraded my amplify project, and started using the new multi auth apis in: #3630 

This is the user schema for the `sender` attribute

[...]

PR #3690, opened 2024-05-13
fix(API): pass authMode used for lazy loading functionality

lawmicha:
## Issue \#

https://github.com/aws-amplify/amplify-swift/issues/3677

## Description

Lazy Loading functionality is built on metadata passed to it during decoding. For example, the hasMany list will contain the parent's identifiers and the apiName to reconstruct the request when lazy loading. The belongsTo lazy model will contain its identifiers and the apiName used to when the child model was retrieved. 

The issue is with the new `authMode` parameter, the `authMode` was not being passed down to the lazy types (https://github.com/aws-amplify/amplify-swift/issues/3677#issuecomment-2108404457)

```swift
// default auth type is APIKey.
let message = try await Amplify.API.query(request: .get(Message.self, byId: id, authMode: .amazonCognitoUserPools)).get() // use userPool auth
let sender = try await message.sender  // with fix, uses user pool auth. without fix, uses default API key auth from config
let space = try await message.space // with fix, uses user pool auth. without fix, uses default API key auth from config
```

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/API/Sources/AWSAPIPlugin/Core/AppSyncListPayload.swift
+++ b/AmplifyPlugins/API/Sources/AWSAPIPlugin/Core/AppSyncListPayload.swift
@@ -46,7 +46,7 @@ public struct AppSyncListPayload: Codable {
         // `GraphQLFilter`.
         guard let filterVariablesData = try? encoder.encode(filterValue),
               let filterVariablesJSON = try? JSONSerialization.jsonObject(with: filterVariablesData)
-                as? [String: Any] else {
+                as? GraphQLFilter else {
 
             assertionFailure("Filter variables is not a valid JSON object: \(filterValue)")
             return nil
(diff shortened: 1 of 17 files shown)


AmplifyPlugins/API/Sources/AWSAPIPlugin/Core/AppSyncListPayload.swift
```diff
@@ -43,7 +46,7 @@ public struct AppSyncListPayload: Codable {
         // `GraphQLFilter`.
         guard let filterVariablesData = try? encoder.encode(filterValue),
               let filterVariablesJSON = try? JSONSerialization.jsonObject(with: filterVariablesData)
-                as? GraphQLFilter else {
+                as? [String: Any] else {
```
thisisabhash, line 49: why this change?

thisisabhash (dismissed):

AmplifyPlugins/API/Sources/AWSAPIPlugin/Core/AppSyncListPayload.swift
lawmicha: i've reverted it, when i was compiling earlier Xcode wasn't picking up GraphQLFilter properly from AWSPluginsCore

thisisabhash (approved):

Issue #3649, opened 2024-04-27
Crash in OutgoingMutationQueue.doStart(api:mutationEventPublisher:reconciliationQueue:)


amruth-movano:
### Describe the bug

Our users are getting repetative crash in - 
OutgoingMutationQueue.doStart(api:mutationEventPublisher:reconciliationQueue:)
Attaching the crash log file

[com.movano.ios.staging_issue_a19b83783e3e409ea2eb7c37819f80e6_crash_session_550b44cf76d64c2097a643748c701d1c_DNE_8_v2_stacktrace.txt](https://github.com/aws-amplify/amplify-swift/files/15137646/com.movano.ios.staging_issue_a19b83783e3e409ea2eb7c37819f80e6_crash_session_550b44cf76d64c2097a643748c701d1c_DNE_8_v2_stacktrace.txt)

### Steps To Reproduce

```swift
--
```

### Expected behavior

App should not crash in amplify library

### Amplify Framework Version

2.25.2

### Amplify Categories

DataStore

### Dependency manager

Swift PM

### Swift version

5.8

### CLI version

NA

### Xcode version

15.0

### Is this a regression?

Yes

### Platforms

iOS

### OS Version

iOS 17.4.1

### Device

iPhone 15 Pro Max

lawmicha:
```
# Crashlytics - Stack trace
# Application: com.movano.ios.staging
# Platform: apple
# Version: 1.0 (240)
# Issue: a19b83783e3e409ea2eb7c37819f80e6
# Session: 550b44cf76d64c2097a643748c701d1c_DNE_8_v2
# Date: Fri Apr 26 2024 09:20:51 GMT+0530 (India Standard Time)

com.apple.main-thread
0  libsystem_kernel.dylib         0x1af8 mach_msg2_trap + 8
1  libsystem_kernel.dylib         0x1890 mach_msg2_internal + 80
2  libsystem_kernel.dylib         0x17a8 mach_msg_overwrite + 436
3  libsystem_kernel.dylib         0x15e8 mach_msg + 24
4  CoreFoundation                 0x3601c __CFRunLoopServiceMachPort + 160
5  CoreFoundation                 0x33f04 __CFRunLoopRun + 1208
6  CoreFoundation                 0x33968 CFRunLoopRunSpecific + 608
7  GraphicsServices               0x34e0 GSEventRunModal + 164
8  UIKitCore                      0x22aedc -[UIApplication _run] + 888
[...]

Issue #3467, opened 2024-01-11
DataStore Crash in OutgoingMutationQueue


fzy-spyro:
### Describe the bug

This is regression/the same crash as in https://github.com/aws-amplify/amplify-swift/issues/3325

Error reported: 

```
OS Version: iOS 16.6.1 (20G81)
Report Version: 104

Exception Type: EXC_BAD_ACCESS (SIGSEGV)
Exception Codes: SEGV_ACCERR at 0x000000000000001a
Crashed Thread: 10

Application Specific Information:
Signal 11, Code 2 >
Attempted to dereference garbage pointer at 0x1a.

Thread 10 Crashed:
0   Foundation                      0x31d969644         __NSOQSchedule
1   Foundation                      0x31d8f89a4         -[NSOperationQueue setSuspended:]
2   SampleApp                       0x202eadc50         OutgoingMutationQueue.doStart (OutgoingMutationQueue.swift:143)
3   SampleApp                       0x202eb0f78         OutgoingMutationQueue.queryMutationEventsFromStorage (OutgoingMutationQueue.swift:349)
4   SampleApp                       0x202e74b90         SQLiteStorageEngineAdapter.query<T> (StorageEngineAdapter+SQLite.swift:339)
5   SampleApp                       0x202e755e4         SQLiteStorageEngineAdapter.query<T> (StorageEngineAdapter+SQLite.swift:308)
6   SampleApp                       0x202e77280         SQLiteStorageEngineAdapter
7   SampleApp                       0x202eadf90         OutgoingMutationQueue.queryMutationEventsFromStorage (OutgoingMutationQueue.swift:336)
8   SampleApp                       0x202ead7b0         OutgoingMutationQueue.doStart (OutgoingMutationQueue.swift:140)
9   SampleApp                       0x202ead0c0         OutgoingMutationQueue.respond (OutgoingMutationQueue.swift:108)
10  SampleApp                       0x202eb3e8c         [inlined] OutgoingMutationQueue.init (OutgoingMutationQueue.swift:77)
11  SampleApp                       0x202eb3e8c         OutgoingMutationQueue.init
12  SampleApp                       0x202e2fdc4         thunk for closure
13  libdispatch.dylib               0x337cda31c         _dispatch_call_block_and_release
14  libdispatch.dylib               0x337cdbea8         _dispatch_client_callout
15  libdispatch.dylib               0x337ce3530         _dispatch_lane_serial_drain
[...]

Issue #3325, opened 2023-10-25
Crash in OutgoingMutationQueue


kubatatami:
### Describe the bug

OutgoingMutationQueue Crashes.

Sample error reported:

```
Exception Type: EXC_BAD_ACCESS (SIGSEGV)
Exception Codes: SEGV_ACCERR at 0x4075e00000000018
Crashed Thread: 15

Application Specific Information:
Signal 11, Code 2 >
Attempted to dereference garbage pointer at 0x4075e00000000018.
```

Sample Stacktraces:

```
Thread 15 Crashed:
0   Foundation                      0x32b5d7a2c         __NSOQSchedule
1   Foundation                      0x32b603f44         -[NSOperationQueue setSuspended:]
2   Evie Beta                       0x202906c78         OutgoingMutationQueue.doStart (OutgoingMutationQueue.swift:142)
3   Evie Beta                       0x202909c70         OutgoingMutationQueue.queryMutationEventsFromStorage (OutgoingMutationQueue.swift:342)
4   Evie Beta                       0x2028cf964         SQLiteStorageEngineAdapter.query<T> (StorageEngineAdapter+SQLite.swift:335)
5   Evie Beta                       0x2028d03b8         SQLiteStorageEngineAdapter.query<T> (StorageEngineAdapter+SQLite.swift:304)
6   Evie Beta                       0x2028d2018         SQLiteStorageEngineAdapter
7   Evie Beta                       0x202906f4c         OutgoingMutationQueue.queryMutationEventsFromStorage (OutgoingMutationQueue.swift:331)
8   Evie Beta                       0x202906804         OutgoingMutationQueue.doStart (OutgoingMutationQueue.swift:141)
9   Evie Beta                       0x202906138         OutgoingMutationQueue.respond (OutgoingMutationQueue.swift:109)
10  Evie Beta                       0x20290c86c         [inlined] OutgoingMutationQueue.init (OutgoingMutationQueue.swift:78)
11  Evie Beta                       0x20290c86c         OutgoingMutationQueue.init
12  Evie Beta                       0x20288c6c0         thunk for closure
13  libdispatch.dylib               0x33d44c6a4         _dispatch_call_block_and_release
14  libdispatch.dylib               0x33d44e2fc         _dispatch_client_callout
15  libdispatch.dylib               0x33d455890         _dispatch_lane_serial_drain
16  libdispatch.dylib               0x33d4563c0         _dispatch_lane_invoke
[...]

PR #3720, opened 2024-05-24
fix(datastore): change OutgoingMutationQueue use TaskQueue for state transitions

5d:
## Issue \#

- https://github.com/aws-amplify/amplify-swift/issues/3325
- https://github.com/aws-amplify/amplify-swift/issues/3467
- https://github.com/aws-amplify/amplify-swift/issues/3649

The crash reports indicate that a `SEGV_ACCERR` occurred due to an attempted dereference of a garbage pointer. In the implementation of the `OutgoingMutationQueue`, the `mutationDispatchQueue` was targeted to the concurrent global queue, with state transition actions being dispatched to this queue. This setup means that `starting` and `stopping` could trigger two state transite operations running concurrently. Specifically, the starting action needs to complete some I/O work, while the stopping action is cleaning up the operation queue, potentially causing a data race.

## Description

In this PR, we modified `mutationDispatchQueue` to use the `TaskQueue` implementation, ensuring that all state transition actions run serially. Also, we set the mutation event `operationQueue` to have the same QOS as the global queue and removed the underlying queue.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
@@ -137,7 +137,7 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
         queryMutationEventsFromStorage { [weak self] in
             guard let self = self else { return }
             guard case .starting = self.stateMachine.state else {
-                self.log.debug("Failed to continue doStart operation, current state(\(self.stateMachine.state)) is not starting")
+                self.log.debug("Unexpected state transition while performing `doStart()` during `.starting` state. Current state: \(self.stateMachine.state).")
                 return
             }
 
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
@@ -27,10 +27,7 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
     private let operationQueue: OperationQueue
 
     /// A DispatchQueue for synchronizing state on the mutation queue
-    private let mutationDispatchQueue = DispatchQueue(
-        label: "com.amazonaws.OutgoingMutationQueue",
-        target: DispatchQueue.global()
-    )
+    private let mutationDispatchQueue = TaskQueue<Void>()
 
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -55,7 +52,7 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
 
         let operationQueue = OperationQueue()
         operationQueue.name = "com.amazonaws.OutgoingMutationOperationQueue"
-        operationQueue.underlyingQueue = mutationDispatchQueue
+        operationQueue.qualityOfService = .default
```
lawmicha, line 55: i think the default is `.default` so no need to set this

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
```diff
@@ -139,6 +136,10 @@ final class OutgoingMutationQueue: OutgoingMutationQueueBehavior {
 
         queryMutationEventsFromStorage { [weak self] in
             guard let self = self else { return }
+            guard case .starting = self.stateMachine.state else {
+                self.log.debug("Failed to continue doStart operation, current state(\(self.stateMachine.state)) is not starting")
```
lawmicha, line 140:
```suggestion
                self.log.debug("Unexpected state transition while performing `doStart()` during `.starting` state. Current state: \(self.stateMachine.state).")
```

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Sync/MutationSync/OutgoingMutationQueue/OutgoingMutationQueue.swift
5d: According to [the doc](https://developer.apple.com/documentation/foundation/operationqueue/1417919-qualityofservice), the default value for the `OperationQueue` is `NSOperationQualityOfServiceBackground`. I set it to the `.default` because `DispatchQueue.global()` also uses a default quality of service (`qos = .default`).

>For queues you create yourself, the default value is NSOperationQualityOfServiceBackground. For the queue returned by the [main](https://developer.apple.com/documentation/foundation/operationqueue/1409193-main) method, the default value is NSOperationQualityOfServiceUserInteractive and cannot be changed.

lawmicha: ah, got it, thanks

lawmicha (approved):

PR #3775, opened 2024-07-11
feat(Storage): Adding subpath strategy to the List operation

ruisebas:
## Description
This PR introduces the `subpathStrategy` parameter to `StorageListRequest.Options` and its corresponding `SubpathStrategy` enum type.

Using this field will affect the results included in the response:
- `subpathStrategy: .include`: Objects that are included in nested subpaths will be included in the response. **This is the default behaviour**
- `subpathStrategy: .exclude`: Objects that are included in nested subpaths are not included in the response, instead their subpaths are included in the `excludedSubpaths` list.
- `subpathStrategy: .exclude(delimitedBy:)`: Same as above, but it allows to set a custom character as delimiter for subpaths. The delimiter parameter defaults to `"/"`, which is the same as `.exclude`.

## General Checklist
- [X] All unit tests pass
- [X] [All integration tests pass](https://github.com/aws-amplify/amplify-swift/actions/runs/9976983528)
- [X] Documentation update for the change if required
  - https://github.com/aws-amplify/docs/pull/7831

### DIFF
--- a/AmplifyPlugins/Storage/Tests/StorageHostApp/AWSS3StoragePluginIntegrationTests/AWSS3StoragePluginListObjectsIntegrationTests.swift
+++ b/AmplifyPlugins/Storage/Tests/StorageHostApp/AWSS3StoragePluginIntegrationTests/AWSS3StoragePluginListObjectsIntegrationTests.swift
@@ -189,4 +189,67 @@ class AWSS3StoragePluginListObjectsIntegrationTests: AWSS3StoragePluginTestBase 
         }
     }
 
+    /// Given: Multiple objects uploaded to a public path
+    /// When: `Amplify.Storage.list` is invoked with `subpathStrategy: .exclude`
+    /// Then: The API should execute successfully and list objects for the given path, without including contens from its subpaths
+    func testList_withSubpathStrategyExclude_shouldExcludeSubpaths() async throws {
+        let path = UUID().uuidString
+        let data = Data(path.utf8)
+        let uniqueStringPath = "public/\(path)"
+
+        // Upload data
+        _ = try await Amplify.Storage.uploadData(path: .fromString(uniqueStringPath + "/test1"), data: data, options: nil).value
+        _ = try await Amplify.Storage.uploadData(path: .fromString(uniqueStringPath + "/test2"), data: data, options: nil).value
+        _ = try await Amplify.Storage.uploadData(path: .fromString(uniqueStringPath + "/subpath1/test"), data: data, options: nil).value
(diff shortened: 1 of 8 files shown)


AmplifyPlugins/Storage/Tests/StorageHostApp/AWSS3StoragePluginIntegrationTests/AWSS3StoragePluginListObjectsIntegrationTests.swift
```diff
@@ -189,4 +189,67 @@ class AWSS3StoragePluginListObjectsIntegrationTests: AWSS3StoragePluginTestBase
...
+    /// When: `Amplify.Storage.list` is invoked with `subpathStrategy: .exclude(delimitedBy:)`
+    /// Then: The API should execute successfully and list objects for the given path, without including contents from any subpath that is determined by the given delimiter
+    func testList_withSubpathStrategyExclude_andCustomDelimiter_shouldExcludeSubpaths() async throws {
+        let path = UUID().uuidString
+        let data = Data(path.utf8)
+        let uniqueStringPath = "public/\(path)"
+
+        // Upload data
+        _ = try await Amplify.Storage.uploadData(path: .fromString(uniqueStringPath + "-test"), data: data, options: nil).value
+        _ = try await Amplify.Storage.uploadData(path: .fromString(uniqueStringPath + "-subpath-test"), data: data, options: nil).value
```
thisisabhash, line 236: what happens when the subpath is `--subpath-test` ?

thisisabhash (approved):
minor comment

AmplifyPlugins/Storage/Tests/StorageHostApp/AWSS3StoragePluginIntegrationTests/AWSS3StoragePluginListObjectsIntegrationTests.swift
ruisebas: In that case, the excluded subpath will be `public/[UUID]--`

---
From the S3 perspective, here's how it works. You have these objects:
```
public/[UUID]-test
public/[UUID]--subpath-test
```

You filter everything that starts with the prefix P*`public/[UUID]-`*P, and both objects still match:

<pre>
<b>public/[UUID]-</b>test
<b>public/[UUID]-</b>-subpath-test 
</pre>

But since you set the delimiter to `-`, any object that contains it **after** the given prefix is gets rolled up until that character. So `public/[UUID]--subpath-test` is rolled up to the `public/[UUID]--` common prefix, because the `-` character is the very next character after the prefix. 
This value is then reported as an `excludedSubpath` by Amplify.

Issue #3632, opened 2024-04-25
can not receive update Subscription event on 2.29.2 SDK version


smediomike:
### Describe the bug

after upgrade to 2.92.2, iOS application can not receive update Subscription if application switch to background before. 2.17 SDK version is okay.

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
1, upgrade SDK to 2.29.2 
2, run application and go to background to take one picture
3, upload such picture to s3 storage, as this time, we can receive Subscription Add event, and datastore also can query such new record.
4, delete this item from web by grpahcGL
5, iOS application can not receive such Subscription event; on 2.17 SDK, we can receive such event. and also datastore can not sync such record till to re-start application.

possible root cause: 
our application go to background and then to foreground to upload and wait for update Subscription event. if we do not go to background and and do same steps, such issue do not happen.
```

### Expected behavior

Application switch to background and then go to foreground, it should can receive update/delete Subscription event.

### Amplify Framework Version

2.29.2

### Amplify Categories

DataStore

### Dependency manager

Swift PM

### Swift version

5.0

### CLI version

12.1.1

### Xcode version

15.3

### Relevant log output

```shell
[...]

PR #3800, opened 2024-08-07
fix(api): propagate connectionLost error from websocket client to sync engine

5d:
## Issue \#

https://github.com/aws-amplify/amplify-swift/issues/3632

## Description

When an app using the DataStore plugin switches from foreground to background, the network connection is aborted, and the `URLSessionTaskDelegate` will receive a connection lost error. This error should be propagated back to the remote sync engine to trigger a scheduling restart, ensuring all missing records are retrieved while the app is in the background without a connection.

## General Checklist

- [x] All unit tests pass


AmplifyPlugins/API/Tests/AWSAPIPluginTests/AppSyncRealTimeClient/AppSyncRealTimeClientTests.swift
```diff
@@ -487,4 +487,68 @@ class AppSyncRealTimeClientTests: XCTestCase {
...
+        )
+        let id = UUID().uuidString
+        let query = UUID().uuidString
+
+        let startTriggered = expectation(description: "webSocket writing start event to connection")
+        let errorReceived = expectation(description: "webSocket connection lost error is received")
+
+        await mockWebSocketClient.setStateToConnected()
+        Task {
+            try await Task.sleep(nanoseconds: 80 * 1_000_000)
```
lawmicha, line 508: why do we need delays? does it ensure a deterministic flow of events?

lawmicha (approved):

PR #3807, opened 2024-08-12
chore: Update label actions

tylerjroach:
## Description
* Renames "pending-response" to "pending-community-response"
* Adds "pending-maintainer-response" when new community issue opened
* Removes "pending-maintainer-response" when owner/maintainer comments on issue
* Adds "pending-maintainer-response" when community comments on issue
* removes "pending-maintainer-response" when issue is closed

## General Checklist

- [ ] All unit tests pass


.github/workflows/issue_comment.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   notify:
     runs-on: ubuntu-latest
     permissions: {}
-    if: ${{ !github.event.issue.pull_request && !contains(fromJSON('["MEMBER", "OWNER"]'), github.event.comment.author_association) }}
+    if: ${{ !github.event.issue.pull_request }}
```
ruisebas, line 11: I rather keep this filter to avoid sending notifications for comments from maintainers

.github/workflows/issue_comment.yml
```diff
@@ -19,17 +19,27 @@ jobs:
...
     steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
-      - name: remove unnecessary labels after closing
+      - name: remove pending-community-response when new comment received
+        if: ${{ !contains(fromJSON('["MEMBER", "OWNER"]'), github.event.comment.author_association) && !github.event.issue.pull_request }}
+        shell: bash
+        run: |
+          gh issue edit $ISSUE_NUMBER --repo $REPOSITORY_NAME --remove-label "pending-community-response"
+      - name: add pending-maintainer-response when new community comment received
+        if: ${{ !contains(fromJSON('["MEMBER", "OWNER"]'), github.event.comment.author_association) }}
```
ruisebas, line 37: IIRC for this filter to work you will need to set your [aws-amplify org membership visibility to **Public**](https://github.com/orgs/aws-amplify/people), otherwise the runner won't be able to see it.

ruisebas (changes requested):

.github/workflows/issue_comment.yml
tylerjroach: The if statement was moved down to each of the steps

.github/workflows/issue_comment.yml
tylerjroach: This is no different than the filter above that was already being used?

.github/workflows/issue_comment.yml
ruisebas: This `notify` job only has one step, which just runs the command. Unless i'm blatantly missing something 🤔

.github/workflows/issue_comment.yml
ruisebas: Yeah, everybody in the Swift team made their membership public in order for this to work. You guys will probably need to do the same.

.github/workflows/issue_comment.yml
tylerjroach: Oh sorry, correct.

If you all would prefer, I can remove it. I brought this up with the Android team and we decided to remove it because we wanted to still be notified when a maintainer opened a new issue. This ensures everyone is aware of the issue coming in, not just the person that created it.

.github/workflows/issue_comment.yml
tylerjroach: Got it, I don't remember doing this but we can check. We've had this type of filter used in other places and working well, so maybe we had already done that as well.

.github/workflows/issue_comment.yml
ruisebas: But this action is for comments being added to issues, not for issues being opened.
I _also_ very much like not being notified for comments from us :P. 
But we can check with the team as well

.github/workflows/issue_comment.yml
ruisebas: Cool! I might be mixing things up with the ops metrics lambda needing this permission instead of the GH action 😅 
But I do remember it was needed _somewhere_ 😬

ruisebas (approved):

PR #3817, opened 2024-08-15
feat(Storage): Implementing support for multiple buckets

ruisebas:
## Description
This PR makes the main changes in order to add support for multiple buckets to our existing **Storage** implementation.

The main changes are the following:

### AmplifyOutputsData
Added the new optional `buckets` array to the Storage configuration, plus a convenience initializer for testing purposes.

### AWSS3StoragePlugin
We have pretty much all of the service-related logic encapsulated inside the `AWSS3StorageServiceBehavior` protocol, which is implemented by `AWSS3StorageService`. This implementation, along its many related objects, heavily relies on the assumption of only one bucket and region.

To avoid a complete refactor, we will now have one storage service per bucket. S3 bucket names are globally unique, so we can store them in a dictionary using the bucket name as key for easy retrieval.

On plugin configuration we will only instantiate the "default" storage service,  keeping the existing behaviour. We will only instantiate new services (and store them in the dictionary) when required, to prevent unnecessary memory usage if a customer has many buckets in their configuration but never attempts any Storage operation on most of them. 

I've also added an optional `additionalBucketsByName` dictionary that is set when Amplify is configured with `AmplifyOutputs`. This is needed because we need to be able to perform the bucket lookup when the customer provides a `StorageBucket.fromOutputs(:)` bucket.

Finally, i've changed the type of `authService` to be `AWSAuthCredentialsProviderBehavior`. This was an oversight during the `AmplifyCredentials` refactor, but it's not a breaking change because `AWSAuthService` conforms to both.

### AWSS3Storage<*>Operation
The APIs for uploading and downloading do not directly throw errors, instead they return an `AmplifyTask` that can be used to monitor progress and completion. 

Since retrieving the storage service can throw (e.g. if a wrong bucket name is passed), and the operations need a service in order to execute, I've created the `AWSS3StorageServiceProvider` callback that can throw. We can invoke this provider when the operation is actually running, since we can report errors by then.

### StorageConfiguration
I've removed the `.default` static property and instead crated a new `.init(forBucket:)` that just appends it to the default `sessionIdentifier` in order for it to be unique. This was needed because otherwise the URLSession delegate methods were not being reported to the proper storage service, which made it impossible to properly propagate the results.

## General Checklist
- [ ] Added new tests to cover change, if needed. **TBA in a follow up PR**
- [X] All unit tests pass
- [ ] Documentation update for the change if required **TBA in a follow up PR**

### DIFF
--- a/AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin+StorageBucket.swift
+++ b/AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin+StorageBucket.swift
@@ -1,0 +1,74 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+@_spi(InternalAmplifyConfiguration) import Amplify
+import Foundation
+
+extension AWSS3StoragePlugin {
+    /// Returns a AWSS3StorageServiceBehavior instance for the given StorageBucket
+    func storageService(for bucket: (any StorageBucket)?) throws -> AWSS3StorageServiceBehavior {
+        guard let bucket else {
+            // When no bucket is provided, use the default one
+            return defaultStorageService
+        }
+
+        let bucketInfo = try bucketInfo(from: bucket)
+        guard let storageService = storageServicesByBucket[bucketInfo.bucketName] else {
+            // If no service was found for the bucket, create one
+            let storageService = try createStorageService(
+                authService: authService,
+                bucketInfo: bucketInfo
+            )
+            storageServicesByBucket[bucketInfo.bucketName] = storageService
+            return storageService
+        }
+
+        return storageService
+    }
+
+    /// Returns a AWSS3StorageServiceProvider callback for the given StorageBucket
+    func storageServiceProvider(for bucket: (any StorageBucket)?) -> AWSS3StorageServiceProvider {
+        let storageServiceResolver: () throws -> AWSS3StorageServiceBehavior = { [weak self] in
+            guard let self = self else {
+                throw StorageError.unknown("AWSS3StoragePlugin was deallocated", nil)
+            }
+            return try self.storageService(for: bucket)
+        }
+        return storageServiceResolver
+    }
+
+    /// Returns a valid `BucketInfo` instance from the given StorageBucket
+    private func bucketInfo(from bucket: any StorageBucket) throws -> BucketInfo {
+        switch bucket {
+        case let outputsBucket as OutputsStorageBucket:
+            guard let additionalBucketsByName else {
+                let errorDescription = "Amplify was not configured using an Amplify Outputs file"
+                let recoverySuggestion = "Make sure that `Amplify.configure(with:)` is invoked"
+                throw StorageError.validation("bucket", errorDescription, recoverySuggestion, nil)
+            }
+
+            guard let awsBucket = additionalBucketsByName[outputsBucket.name] else {
+                let errorDescription = "Unable to lookup bucket from provided name in Amplify Outputs"
+                let recoverySuggestion = "Make sure the bucket name exists in the Amplify Outputs file"
+                throw StorageError.validation("bucket", errorDescription, recoverySuggestion, nil)
+            }
+
+            return .init(
+                bucketName: awsBucket.bucketName,
+                region: awsBucket.awsRegion
+            )
+
+        case let resolvedBucket as ResolvedStorageBucket:
+            return resolvedBucket.bucketInfo
+
+        default:
+            let errorDescription = "The specified StorageBucket is not supported"
+            let recoverySuggestion = "Please specify a StorageBucket from the Outputs file or from BucketInfo"
+            throw StorageError.validation("bucket", errorDescription, recoverySuggestion, nil)
+        }
+    }
+}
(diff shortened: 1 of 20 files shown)


AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin+StorageBucket.swift
```diff
@@ -0,0 +1,74 @@
...
+            return try self.storageService(for: bucket)
+        }
+        return storageServiceResolver
+    }
+
+    /// Returns a valid `BucketInfo` instance from the given StorageBucket
+    private func bucketInfo(from bucket: any StorageBucket) throws -> BucketInfo {
+        switch bucket {
+        case let outputsBucket as OutputsStorageBucket:
+            guard let additionalBucketsByName else {
```
phantumcode, line 48: based on the approved API design, on Android, an `InvalidStorageBucketException` is returned for error cases; was there any consideration to provided a specific exception for Swift?

AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin.swift
```diff
@@ -5,21 +5,36 @@
...
 /// The AWSS3StoragePlugin which conforms to the Amplify plugin protocols and implements the Storage
 /// Plugin APIs for AWS S3.
 ///
 /// - Tag: AWSS3StoragePlugin
 final public class AWSS3StoragePlugin: StorageCategoryPlugin {
 
-    /// An instance of the S3 storage service.
-    var storageService: AWSS3StorageServiceBehavior!
+    /// The default S3 storage service.
+    var defaultStorageService: AWSS3StorageServiceBehavior! {
```
phantumcode, line 20: will these cause an exception if the service is force_un

AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin.swift
```diff
@@ -5,21 +5,36 @@
...
+        guard let defaultBucket else {
+            return nil
+        }
+        return storageServicesByBucket[defaultBucket.bucketInfo.bucketName]
+    }
+
+    /// The default bucket
+    var defaultBucket: ResolvedStorageBucket!
```
phantumcode, line 28: is it possible to avoid unsafe unwrapping?

AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin.swift
```diff
@@ -5,21 +5,36 @@
...
+
+    /// A dictionary of S3 storage service instances grouped by a specific bucket
+    var storageServicesByBucket: AtomicDictionary<String, AWSS3StorageServiceBehavior> = [:]
+
+    /// A dictionary of additional Outputs-based buckets, grouped by their names
+    var additionalBucketsByName: [String: AmplifyOutputsData.Storage.Bucket]?
 
     /// An instance of the authentication service.
-    var authService: AWSAuthServiceBehavior!
+    var authService: AWSAuthCredentialsProviderBehavior!
```
phantumcode, line 37: same comment as above, to avoid unsafe unwrapping?

AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin+StorageBucket.swift
ruisebas: The approved API design has `StorageError.validation` being thrown for Swift, not a new error.
This is consistent to what happens when an invalid `StoragePath` is provided, so I didn't explore introducing a new error.

AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/AWSS3StoragePlugin.swift
ruisebas: Yes, but that's the case with the majority of variables in all of our plugins. 
We're basically giving them the outlets treatment: they will be `nil` at initialization, but will be set before the plugins are actually used (on their `configure` methods).
We could make them all optional but then we'd have to unwrap them every single time we want to use them. If a plugin is not configured (i.e. these variables are `nil`), we throw an exception anyway before reaching any of these variables.

AmplifyPlugins/Storage/Sources/AWSS3StoragePlugin/Service/Storage/AWSS3StorageServiceBehavior.swift
```diff
@@ -9,6 +9,7 @@ import Foundation
 import Amplify
 import AWSS3
 
+typealias AWSS3StorageServiceProvider = () throws -> AWSS3StorageServiceBehavior
```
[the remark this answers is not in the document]
ruisebas: I guess the answer depends on what you consider a delegate vs a provider :P 
I went with provider because it's just a callback that gets invoked to retrieve an actual `AWSS3StorageServiceBehavior` instance.
I didn't want to go with a full protocol because then we'd have to deal with `weak`s  and unwrappings.

We could name it `AWSS3StorageServiceRetriever` if that reads better.

phantumcode (approved):

Issue #3855, opened 2024-09-05
'try!' expression unexpectedly raised an error: disk I/O error (code: 10) when device is locked


mmathewsTableau:
### Describe the bug

Whenever our app is locked and aws pinpoint is trying to read the database, we get this error followed by a crash. The error is expected because we have our app set to be NSFileProtectionComplete, which will encrypt the filesystem while the device is locked. However, **our ask is that the Amplify library gracefully handles this scenario instead of crashing the app.** 

### Steps To Reproduce

```swift
Steps to reproduce the behavior:
(this assumes you have an app that uses AWS pinpoint to record events remotely)

1. Set your app to have NSFileProtectionComplete for com.apple.developer.default-data-protection in your entitlements file.
2. Build your app and attach a debugger
3. Put a breakpoint here (https://github.com/aws-amplify/amplify-swift/blob/c13d7b723b236a68d32d04718011ad1f4c6f2f49/AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/LocalStorage/AnalyticsEventSQLStorage.swift#L145)
4. Use your app so to trigger the breakpoint
5. Once you hit the breakpoint, lock your device 
6. Wait 5 seconds... maybe more...
7. Step through the code
```

### Expected behavior

The app should fails to read but app does not crash.

### Amplify Framework Version

2.27.3

### Amplify Categories

Analytics

### Dependency manager

Swift PM

### Swift version

5.9.2

### CLI version

N/A

### Xcode version

15.2

### Relevant log output

```shell
<details>
<summary>Log Messages</summary>


Role:               Background
Exception Type:     EXC_BREAKPOINT 
Exception Subtype:  KERN_INVALID_ADDRESS


Fatal error: 'try!' expression unexpectedly raised an error: disk I/O error (code: 10)

0 libswiftCore.dylib +0xa7d10 _swift_unexpectedError
1 Tableau +0xdc0848           specialized FailableIterator.next() (<compiler-generated>)
2 Tableau +0xdb79f4           protocol witness for AnalyticsEventStorage.getEventsWith(limit:) in conformance AnalyticsEventSQLStorage (<compiler-generated>)
```
</details>
```


### Is this a regression?

No

### Regression additional context

_No response_

### Platforms

iOS

### OS Version

iOS 17.6.1

### Device

iPhone 14.5

### Specific to simulators

No

### Additional context

_No response_

mmathewsTableau:
[...]

PR #3857, opened 2024-09-09
fix(analytics): iterate sqlite rows with failableNext

5d:
## Issue \#

https://github.com/aws-amplify/amplify-swift/issues/3855

## Description

The [default iteration behavior](https://github.com/stephencelis/SQLite.swift/blob/master/Sources/SQLite/Core/Statement.swift#L218) for SQLite statement results uses `try!`, which throws a runtime error that can't be caught by Swift's standard `do..try` mechanism. Since the analytics category runs as a background monitoring thread, it shouldn't cause the app to enter an error state. Instead, use the alternative throwable API [`failableNext`](https://github.com/stephencelis/SQLite.swift/blob/master/Sources/SQLite/Core/Statement.swift#L233) to avoid runtime errors.

## General Checklist

- [X] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/LocalStorage/AnalyticsEventSQLStorage.swift
+++ b/AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/LocalStorage/AnalyticsEventSQLStorage.swift
@@ -142,9 +142,9 @@ class AnalyticsEventSQLStorage: AnalyticsEventStorage {
         ORDER BY timestamp ASC
         LIMIT ?
         """
-        let rows = try dbAdapter.executeQuery(queryStatement, [limit])
+        let rows = try dbAdapter.executeQuery(queryStatement, [limit]).makeIterator()
         var result = [PinpointEvent]()
-        for element in rows {
+        while let element = try rows.failableNext() {
             if let event = PinpointEvent.convertToEvent(element) {
                 result.append(event)
             }


AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/LocalStorage/AnalyticsEventSQLStorage.swift
```diff
@@ -142,12 +142,16 @@ class AnalyticsEventSQLStorage: AnalyticsEventStorage {
...
         LIMIT ?
         """
-        let rows = try dbAdapter.executeQuery(queryStatement, [limit])
+        let rows = try dbAdapter.executeQuery(queryStatement, [limit]).makeIterator()
         var result = [PinpointEvent]()
-        for element in rows {
-            if let event = PinpointEvent.convertToEvent(element) {
-                result.append(event)
+        do {
+            while let element = try rows.failableNext() {
```
atierian, line 148: Is there a scenario where the next row can fail, but subsequent rows can still be successfully retrieved?

If so, would it better (or even possible) to continue iterating through instead of bailing on the first error thrown?

ruisebas: What's funny is that I was gonna suggest the opposite: If get get an error when attempting to retrieve a subsequent row, we should fail the whole operation instead of just returning what we have so far.

The way I see it is that we're explicitly running a SELECT query that we expect to contain the things we've asked, so just giving us back a subset of stuff just because the DB failed at some point sounds incorrect.

ruisebas: Yeah, I think just propagating the error would be enough, since it will be caught and handled by the callers of this method:
- [`AWSPinpointAnalyticsPlugin.flushEvents()`](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/Analytics/Sources/AWSPinpointAnalyticsPlugin/AWSPinpointAnalyticsPlugin%2BClientBehavior.swift#L103)
- [`AnalyticsClient`'s automatic flush of events](https://github.com/aws-amplify/amplify-swift/blob/main/AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/AnalyticsClient.swift#L297)

ruisebas (approved):

PR #3880, opened 2024-09-20
fix(predictions): region handling for creating correct streaming endpoint from region

ppraja22:
## Issue \#

None

## Description
We need to construct the correct streaming endpoint with correct base domain depending on the regions. Not all region belong to commercial partition so the base domain becomes different. We can keep adding new regions in this check to get the desired base domain in future.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Predictions/AWSPredictionsPlugin/Liveness/SPI/LivenessStreamingURL.swift
+++ b/AmplifyPlugins/Predictions/AWSPredictionsPlugin/Liveness/SPI/LivenessStreamingURL.swift
@@ -7,8 +7,20 @@
 
 import Foundation
 
+private let ISO_PARTITION_BASE_DOMAIN: String = "csp.hci.ic.gov"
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/Predictions/AWSPredictionsPlugin/Liveness/SPI/LivenessStreamingURL.swift
```diff
@@ -7,8 +7,20 @@
 
 import Foundation
 
+private let ISO_PARTITION_BASE_DOMAIN: String = "csp.hci.ic.gov"
```
thisisabhash, line 10: nit: can we change the constants to lowerCamelCase?
```
private let isoPartitionBaseDomain: String = "csp.hci.ic.gov"
private let defaultBaseDomain: String = "amazonaws.com"
```

thisisabhash (approved):

Issue #4003, opened 2025-06-09
Xcode 26 build error: Cannot find '__IPHONE_OS_VERSION_MIN_REQUIRED' in scope


minlite:
### Describe the bug

Building a project that has the amplify dependency on Xcode 26 / iOS 26 results in the following error:

`Cannot find '__IPHONE_OS_VERSION_MIN_REQUIRED' in scope`

The culprit code seems to be in the `AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/ASF/ASFAppInfo.swift` file:

```swift
  var targetSDK: String {
        var targetSDK: String = ""
#if os(iOS) || os(watchOS) || os(tvOS)
        targetSDK = "\(__IPHONE_OS_VERSION_MIN_REQUIRED)"
#elseif os(macOS)
        targetSDK = "\(__MAC_OS_X_VERSION_MIN_REQUIRED)"
#else
        targetSDK = "Unknown"
#endif
        return targetSDK
```

### Steps To Reproduce

```swift
1. Download Xcode 26
2. Add amplify sdk as dependency
3. Build
```

### Expected behavior

Code builds without any issues

### Amplify Framework Version

2.42.1

### Amplify Categories

Auth

### Dependency manager

Swift PM

### Swift version

5.3

### CLI version

NA

### Xcode version

26 Dev Beta 1 (17A5241e)

### Relevant log output

```shell
<details>
<summary>Log Messages</summary>


amplify-swift/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/ASF/ASFAppInfo.swift:19:24: error: cannot find '__IPHONE_OS_VERSION_MIN_REQUIRED' in scope
        targetSDK = "\(__IPHONE_OS_VERSION_MIN_REQUIRED)"

</details>
```

### Is this a regression?

Yes

### Platforms

iOS

### OS Version

iOS 26 Dev Beta 1

### Device

iPhone 15 Pro

deivymejiaruiz:
someone fix it?

thisisabhash:
Thank you for posting this - our team will take a look and post updates here.

egesucu:
Issue is also present on 2.48 with the lates 5.X. It's only on Xcode 26 Beta for now.

harsh62:
## Xcode 26 Beta Support

For users experiencing compilation issues, we've created a work-in-progress branch to help with testing: `xcode26-beta`

### Using the WIP Branch

[...]

PR #4013, opened 2025-06-30
Fix accessing Availability.h macros in Swift (Xcode 26 support)

liamnichols:
## Issue \#

#4003

## Description

Starting in Xcode 26, the Availability.h/AvailabilityInternal.h headers that contained the `__IPHONE_OS_VERSION_MIN_REQUIRED` and `__MAC_OS_X_VERSION_MIN_REQUIRED` preprocessor macros are no longer implicitly exposed to Swift code via the `os_availability`/`os_availability_internal` modules which has prevented the Amplify SDK from compiling.

I saw that the `xcode-26` branch just commented out the problematic code as a workaround, but I wanted to propose a more suitable fix that involved explicitly exposing the macro values to the AWSCognitoAuthPlugin target via a new AmplifyAvailability system library target.

The new target defines a module map that exposes the AmplifyAvailability.h header and within this header, we just redefine the Availability.h preprocessor macros via `getIOSVersionMinRequired()` and `getMACOSXVersionMinRequired()`. These new static symbols are now exposed to the Swift code and can be used instead to provide support both in Xcode 26 and older versions.

## General Checklist

- [x] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/ASF/ASFAppInfo.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/ASF/ASFAppInfo.swift
@@ -5,6 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+import AmplifyAvailability
 import Foundation
 
 struct ASFAppInfo: ASFAppInfoBehavior {
@@ -16,9 +17,9 @@ struct ASFAppInfo: ASFAppInfoBehavior {
     var targetSDK: String {
         var targetSDK: String = ""
 #if os(iOS) || os(watchOS) || os(tvOS)
-        targetSDK = "\(__IPHONE_OS_VERSION_MIN_REQUIRED)"
+        targetSDK = "\(getIOSVersionMinRequired())"
 #elseif os(macOS)
-        targetSDK = "\(__MAC_OS_X_VERSION_MIN_REQUIRED)"
+        targetSDK = "\(getMACOSXVersionMinRequired())"
 #else
         targetSDK = "Unknown"
 #endif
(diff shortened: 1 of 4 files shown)


harish-suthar:
@liamnichols This looks a prominent fix any ideas by when can we see this merged and released?

CC: @harsh62

harsh62:
@harish-suthar Thanks for tagging me. 

@liamnichols Thanks for you contribution. The change on a high level looks good. I've approved and ran the workflows to check the health of the PR. 

I will work with my team to get feedback.

harsh62:
@liamnichols @harish-suthar This needs to be build on the minimum version of Xcode supported by Amplify which is Xcode 15. At the moment its failing with:
```
Error: cannot find 'getIOSVersionMinRequired' in scope
```

liamnichols:
> @liamnichols @harish-suthar This needs to be build on the minimum version of Xcode supported by Amplify which is Xcode 15. At the moment its failing with:
> 
> ```
> Error: cannot find 'getIOSVersionMinRequired' in scope
> ```

Hi @harsh62, thanks for pointing this out.. I don't have access to Xcode 15 locally so I hadn't realised this 🙇 I've made some changes to use a regular `.target` instead of `.systemLibrary` that defines a small .c file rather than using inline functions, I think this _should_ work better with Xcode 15 but I still haven't tested it myself. 

If you could re-run CI that would be appreciated 🙏

liamnichols:
Thanks for merging this @harsh62! Would it be possible if you could share timelines for releasing updates to amplify-swift and amplify-ui-swift-liveness that contain this patch? 

No worries if it won't be for a while, I just need to know if it is worth me creating some forks and using them in the meantime or not 🙂

harsh62:
I am gonna try to release this week, with a couple of other fixes in the pipeline. At most it could slip into next week.

PR #4043, opened 2025-08-22
updated RefreshUserPoolTokens to use GetTokensFromRefreshToken API to…

ekjotmultani:
… enable refresh token rotation, also updated test mock clients and added unit tests

## General Checklist

- [ ] All unit tests pass

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/RefreshAuthorizationSession/UserPool/RefreshUserPoolTokens.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/RefreshAuthorizationSession/UserPool/RefreshUserPoolTokens.swift
@@ -28,7 +28,6 @@ struct RefreshUserPoolTokens: Action {
                 return
             }
 
-            let authEnv = try environment.authEnvironment()
             let config = environment.userPoolConfiguration
             let client = try? environment.cognitoUserPoolFactory()
             let existingTokens = existingSignedIndata.cognitoUserPoolTokens
@@ -36,24 +35,27 @@ struct RefreshUserPoolTokens: Action {
             let deviceMetadata = await DeviceMetadataHelper.getDeviceMetadata(
                 for: existingSignedIndata.username,
                 with: environment)
+            
+            let deviceKey: String? = {
+                if case .metadata(let data) = deviceMetadata {
+                    return data.deviceKey
+                }
+                return nil
+            }()
 
-            let asfDeviceId = try await CognitoUserPoolASF.asfDeviceID(
-                for: existingSignedIndata.username,
-                credentialStoreClient: authEnv.credentialsClient)
-
-            let input = await InitiateAuthInput.refreshAuthInput(
-                username: existingSignedIndata.username,
-                refreshToken: existingTokens.refreshToken,
+            let input = GetTokensFromRefreshTokenInput(
+                clientId: config.clientId,
                 clientMetadata: [:],
-                asfDeviceId: asfDeviceId,
-                deviceMetadata: deviceMetadata,
-                environment: environment)
+                clientSecret: config.clientSecret,
+                deviceKey: deviceKey,
+                refreshToken: existingTokens.refreshToken
+            )
 
-            logVerbose("\(#fileID) Starting initiate auth refresh token", environment: environment)
+            logVerbose("\(#fileID) Starting get tokens from refresh token", environment: environment)
 
-            let response = try await client?.initiateAuth(input: input)
+            let response = try await client?.getTokensFromRefreshToken(input: input)
 
-            logVerbose("\(#fileID) Initiate auth response received", environment: environment)
+            logVerbose("\(#fileID) Get tokens from refresh token response received", environment: environment)
 
             guard let authenticationResult = response?.authenticationResult,
                   let idToken = authenticationResult.idToken,
@@ -69,8 +71,7 @@ struct RefreshUserPoolTokens: Action {
             let userPoolTokens = AWSCognitoUserPoolTokens(
                 idToken: idToken,
                 accessToken: accessToken,
-                refreshToken: existingTokens.refreshToken,
-                expiresIn: authenticationResult.expiresIn
+                refreshToken: authenticationResult.refreshToken ?? existingTokens.refreshToken
             )
             let signedInData = SignedInData(
                 signedInDate: existingSignedIndata.signedInDate,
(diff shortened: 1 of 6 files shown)


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/RefreshAuthorizationSession/UserPool/RefreshUserPoolTokens.swift
```diff
@@ -69,8 +71,7 @@ struct RefreshUserPoolTokens: Action {
             let userPoolTokens = AWSCognitoUserPoolTokens(
                 idToken: idToken,
                 accessToken: accessToken,
-                refreshToken: existingTokens.refreshToken,
-                expiresIn: authenticationResult.expiresIn
+                refreshToken: authenticationResult.refreshToken ?? existingTokens.refreshToken
```
harsh62:
```suggestion
guard let authenticationResult = response?.authenticationResult,
        let idToken = authenticationResult.idToken,
        let accessToken = authenticationResult.accessToken,
        let refreshToken = authenticationResult.refreshToken.  <------- change
else {

        let event = RefreshSessionEvent(eventType: .throwError(.invalidTokens))
        await dispatcher.send(event)
        logVerbose("\(#fileID) Sending event \(event.type)", environment: environment)
        return
}

let userPoolTokens = AWSCognitoUserPoolTokens(
            idToken: idToken,
            accessToken: accessToken,
            refreshToken: refreshToken.  <------- change
```

I don't think we should use the old refresh token at this point.. Should throw an error if the refresh token doesn't exist, as we have entered an unknown state because of missing tokens.

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/ActionTests/FetchAuthSession/FetchUserPoolTokens/RefreshUserPoolTokensTests.swift
```diff
@@ -144,4 +144,74 @@ class RefreshUserPoolTokensTests: XCTestCase {
...
+        let expectation = expectation(description: "refreshTokenNoRotation")
+        let identityProviderFactory: BasicSRPAuthEnvironment.CognitoUserPoolFactory = {
+            MockIdentityProvider(
+                mockGetTokensFromRefreshTokenResponse: { _ in
+                    return GetTokensFromRefreshTokenOutput(
+                        authenticationResult: .init(
+                            accessToken: "accessTokenNew",
+                            expiresIn: 100,
+                            idToken: "idTokenNew",
+                            refreshToken: nil))
```
harsh62, line 204: Since the refresh token is `nil`, can that ever happen.. I personally think it shouldn't and this test case should throw and error instead.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/RefreshAuthorizationSession/UserPool/RefreshUserPoolTokens.swift
```diff
@@ -28,32 +28,34 @@ struct RefreshUserPoolTokens: Action {
...
                 for: existingSignedIndata.username,
                 with: environment)
-
-            let asfDeviceId = try await CognitoUserPoolASF.asfDeviceID(
-                for: existingSignedIndata.username,
-                credentialStoreClient: authEnv.credentialsClient)
-
-            let input = await InitiateAuthInput.refreshAuthInput(
-                username: existingSignedIndata.username,
-                refreshToken: existingTokens.refreshToken,
```
harsh62, line 46: We should remove the implementation of `refreshAuthInput` method as its not longer required.

Package.swift
```diff
@@ -9,7 +9,7 @@ let platforms: [SupportedPlatform] = [
     .watchOS(.v9)
 ]
 let dependencies: [Package.Dependency] = [
-    .package(url: "https://github.com/awslabs/aws-sdk-swift", exact: "1.5.14"),
+    .package(url: "https://github.com/awslabs/aws-sdk-swift", exact: "1.5.18"),
```
harsh62, line 12: This is the version required for the patch SDK change?

ekjotmultani: yes the minimum

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/RefreshAuthorizationSession/UserPool/RefreshUserPoolTokens.swift
ekjotmultani: It is already removed in this PR unless I missed something

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/RefreshAuthorizationSession/UserPool/RefreshUserPoolTokens.swift
ekjotmultani: You are right, there's no use for the old token, this was here because its possible the user doesn't have rotation enabled but in that case the returned token would just be their one token anyways

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/ActionTests/FetchAuthSession/FetchUserPoolTokens/RefreshUserPoolTokensTests.swift
ekjotmultani: You are right, while the API can return a null token we should just throw an error in that case

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/RefreshAuthorizationSession/UserPool/RefreshUserPoolTokens.swift
harsh62: I meant the implementation in `InitiateAuthInput+Amplify.swift`

ekjotmultani: got it

PR #4137, opened 2026-02-08
fix(API): Support '+' in query parameters and improve validation logic

rickhohler:
### Summary
Fixed a bug where URL query parameters containing the `+` character were incorrectly flagged as invalid. This was due to:
1.  Reviewing the permitted character set and finding `+` was missing.
2.  Identifying flawed validation logic that checked if *at least one* permitted character was present, rather than ensuring *only* permitted characters were present.

### Changes
*   Added `+` to `permittedQueryParamCharacters` in `RESTOperationRequestUtils.swift`.
*   Updated validation logic to ensure no disallowed characters are present.

### Verification
*   Verified locally with a reproduction test case which failed before the fix and passed after.
*   Ran existing `RESTRequestUtilsTests` to ensure no regressions.


harsh62 (approved):
LGTM

PR #4139, opened 2026-02-10
feat: add logging definitions

thisisabhash:
## Issue \#

None

## Description

This PR adds the Logging API for foundation library.

## General Checklist

- [x] All unit tests pass


AmplifyFoundation/Sources/Logging/AmplifyLogging.swift
```diff
@@ -0,0 +1,31 @@
...
+    internal static let concurrencyQueue = DispatchQueue(label: "com.amplify.foundation.AmplifyLogging")
+    internal static var registeredLogSinks: [String: any LogSinkBehavior] = [:]
+    
+    public static func addSink(_ logSink: any LogSinkBehavior) {
+        return Self.concurrencyQueue.sync {
+            Self.registeredLogSinks[logSink.id] = logSink
+        }
+    }
+    
+    static func removeSink(_ logSink: any LogSinkBehavior) {
```
jvh-aws, line 24: If I remember correctly, we discussed that not including functionality to remove sinks would allow us to simplify the implementation and remove the concurrency queue.

AmplifyFoundation/Sources/Logging/AmplifyLogSink.swift
```diff
@@ -0,0 +1,71 @@
...
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+import Foundation
+import os.log
+
+/// A ready to use implementation of `LogSinkBehavior` which uses `OSLog`
+/// to log messages to the console
+public final class AmplifyLogSink : LogSinkBehavior {
```
jvh-aws: Would another prefix make sense here? It could be something like `OSLogSink`, to clarify the underlying implementation.

AmplifyFoundation/Sources/Logging/AmplifyLogSink.swift
```diff
@@ -0,0 +1,71 @@
...
+        self.logLevel = logLevel
+        self.subsystem = Bundle.main.bundleIdentifier ?? "com.amazonaws.amplify.logSink"
+    }
+    
+    public func isEnabled(for logLevel: LogLevel) -> Bool {
+        logLevel <= self.logLevel
+    }
+    
+    public func emit(message: LogMessage) {
+        if isEnabled(for: message.level) {
```
jvh-aws: I think we're also checking `isEnabled` in `BroadcastLogger`. Does it make sense to do it in only one location?

jvh-aws (commented):
Only a few minor comments

AmplifyFoundation/Sources/Logging/AmplifyLogging.swift
thisisabhash: yup, this is for internal testing

harsh62: Should we remove it once the internal testing is over or you've kept it for unit tests?

AmplifyFoundation/Sources/Logging/AmplifyLogSink.swift
harsh62: _may be_ `AmplifyOSLogSink` ?

AmplifyFoundation/Sources/Logging/Logger.swift
```diff
@@ -0,0 +1,50 @@
...
+    func info(_ message: @autoclosure () -> String, _ error: @autoclosure () -> Error?)
+
+    /// Logs a message at `debug` level
+    func debug(_ message: @autoclosure () -> String, _ error: @autoclosure () -> Error?)
+
+    /// Logs a message at `verbose` level
+    func verbose(_ message: @autoclosure () -> String, _ error: @autoclosure () -> Error?)
+    
+    /// Logs a message at the given level
+    func log(_ logLevel: LogLevel, _ message: @autoclosure () -> String, _ error: @autoclosure () -> Error?)
```
harsh62, line 25: This API definition is forcing callers to always pass nil when no error is present is cumbersome. I can propose like an extension that can help avoid that.. 

```
extension Logger {
    /// We are explicitly passing nil here and the consumer can use this too, if they want to.
    func error(_ message: @autoclosure () -> String) { error(message(), nil) }
```

harsh62: You can see this in your tests too that you've written where you are either passing mocked errors or nil.

AmplifyFoundation/Sources/Logging/LogMessage.swift
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// A struct representing a single log message
+public struct LogMessage {
+    
+    let level: LogLevel
+    let name: String
+    let content: String
+    let error: Error?
```
harsh62: Shouldn't these be explicitly marked as public? In the current state, external consumers implementing `LogSinkBehavior` won't be able to read any of the message fields inside their `emit(message:)` implementation

AmplifyFoundation/Sources/Logging/AmplifyLogging.swift
```diff
@@ -0,0 +1,31 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+public final class AmplifyLogging {
```
harsh62, line 10: I think we should add `private init() {}` to prevent accidental instantiation, since this is purely as a static namespace.

AmplifyFoundation/Sources/Logging/AmplifyLogging.swift
```diff
@@ -0,0 +1,31 @@
...
+    
+    static func removeSink(_ logSink: any LogSinkBehavior) {
+        return Self.concurrencyQueue.sync {
+            Self.registeredLogSinks.removeValue(forKey: logSink.id)
+        }
+    }
+    
+    public static func logger(for name: String) -> Logger {
+        BroadcastLogger(name: name, sinks: Array(registeredLogSinks.values))
+    }
```
harsh62, line 34: I think this should also go through the `concurrencyQueue` like `add` and `remove` are doing it.

AmplifyFoundation/Sources/Logging/BroadcastLogger.swift
```diff
@@ -0,0 +1,50 @@
...
+    func verbose(_ message: @autoclosure () -> String, _ error: @autoclosure () -> (any Error)?) {
+        log(.verbose, message(), error())
+    }
+    
+    func log(_ logLevel: LogLevel, _ message: @autoclosure () -> String, _ error: @autoclosure () -> (any Error)?) {
+        sinks.forEach { sink in
+            if sink.isEnabled(for: logLevel) {
+                sink.emit(message: .init(level: logLevel, name: name, content: message(), error: error()))
+            }
+        }
```
harsh62, line 48: `@autoclosure` arguments are evaluated multiple times: message() and error() are called inside the forEach loop, meaning they are re-evaluated for each sink. For `@autoclosure` parameters, this could invoke the closure N times (once per sink). My suggestion

```suggestion
        let messageValue = message()
        let errorValue = error()
        sinks.forEach { sink in
            if sink.isEnabled(for: logLevel) {
                sink.emit(message: .init(level: logLevel, name: name, content: messageValue, error: errorValue))
            }
        }
```

AmplifyFoundation/Sources/Logging/BroadcastLogger.swift
```diff
@@ -0,0 +1,50 @@
...
+/// protocol that delegates all work to its sinks.
+///
+final class BroadcastLogger {
+    
+    let sinks: [any LogSinkBehavior]
+    let name: String
+
+    init(name: String, sinks: [any LogSinkBehavior]) {
+        self.name = name
+        self.sinks = sinks
```
harsh62, line 18: Sinks are captured at creation time and any sinks added later wouldn't be part of the broadcast. Is this intentional by design?

AmplifyFoundation/Sources/Logging/AmplifyLogSink.swift
```diff
@@ -0,0 +1,71 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+import Foundation
+import os.log
+
+/// A ready to use implementation of `LogSinkBehavior` which uses `OSLog`
+/// to log messages to the console
+public final class AmplifyLogSink : LogSinkBehavior {
+    public var id: String
```
harsh62: Why is this a `var`? An `id` once created shouldn't be changing, right?

AmplifyFoundation/Sources/Logging/AmplifyLogSink.swift
```diff
@@ -0,0 +1,71 @@
...
+        self.id = UUID().uuidString
+        self.logLevel = logLevel
+        self.subsystem = Bundle.main.bundleIdentifier ?? "com.amazonaws.amplify.logSink"
+    }
+    
+    public func isEnabled(for logLevel: LogLevel) -> Bool {
+        logLevel <= self.logLevel
+    }
+    
+    public func emit(message: LogMessage) {
```
harsh62: The method can be refactored into 

```
public func emit(message: LogMessage) {
    let osLog = OSLog(subsystem: subsystem, category: message.name)
    let osLogType: OSLogType
    switch message.level {
    case .error: osLogType = .error
    case .warn, .info: osLogType = .info
    case .debug, .verbose: osLogType = .debug
    case .none: return
    }
    let output = message.content + (message.error?.localizedDescription ?? "")
    os_log("%@", log: osLog, type: osLogType, output)
}
```

AmplifyFoundation/Sources/Logging/AmplifyLogSink.swift
```diff
@@ -0,0 +1,71 @@
...
+        if isEnabled(for: message.level) {
+            let osLog = OSLog(subsystem: subsystem, category: message.name)
+            switch message.level {
+            case .error:
+                os_log(
+                    "%@",
+                    log: osLog,
+                    type: OSLogType.error,
+                    message.content + (message.error?.localizedDescription ?? "")
```
harsh62: Can we consider adding a separator.. This will separate the content and description and would look more verbose. 
```
message.content + (message.error.map { " - \($0.localizedDescription)" } ?? "")
```

AmplifyFoundation/Tests/AmplifyLoggingTests.swift
```diff
@@ -0,0 +1,216 @@
...
+
+import XCTest
+@testable import AmplifyFoundation
+
+class AmplifyLoggingTests: XCTestCase {
+    var logSink: MockLogSink!
+    
+    override func setUp() {
+        logSink = MockLogSink()
+    }
```
harsh62, line 16: Consider adding a tear down

```
override func tearDown() {
    AmplifyLogging.removeSink(logSink)
}
```

AmplifyFoundation/Tests/AmplifyLoggingTests.swift
```diff
@@ -0,0 +1,216 @@
...
+        
+        XCTAssertEqual(logSink.logMessages.count, 0)
+    }
+    
+    
+}
+
+
+/// Mock LogSink for testing which stores the list of log messages in memory
+class MockLogSink : LogSinkBehavior {
```
harsh62:
```suggestion
final class MockLogSink : LogSinkBehavior {
```

AmplifyFoundation/Tests/AmplifyLoggingTests.swift
```diff
@@ -0,0 +1,216 @@
...
+    func isEnabled(for logLevel: AmplifyFoundation.LogLevel) -> Bool {
+        return logLevel <= self.logLevel
+    }
+    
+    func emit(message: AmplifyFoundation.LogMessage) {
+        logMessages.append(message)
+    }
+}
+
+class MockAmplifyError: AmplifyError {
```
harsh62:
```suggestion
final class MockAmplifyError: AmplifyError {
```

AmplifyFoundation/Sources/Logging/AmplifyLogging.swift
thisisabhash: I have kept it for unit tests

AmplifyFoundation/Sources/Logging/BroadcastLogger.swift
thisisabhash: yes, android has the same behavior

PR #4157, opened 2026-03-02
Add client flush test to verify retry behavior works

jvh-aws:
Adding a test that helps verify if the retry behavior on record-level errors works correctly


AmplifyPlugins/Analytics/Sources/AWSKinesisStreamsPlugin/Support/SQLiteRecordStorage.swift
```diff
@@ -26,22 +26,17 @@ actor SQLiteRecordStorage: RecordStorage {
     private static let retryCount = Expression<Int>("retry_count")
     private static let createdAt = Expression<Double>("created_at")
 
-    init(identifier: String, maxRecords: Int, maxBytes: Int64) throws {
+    init(identifier: String, maxRecords: Int, maxBytes: Int64, connection: Connection? = nil) throws {
```
jvh-aws, line 29: @harsh62 allowing to inject connection for testing (in memory DB)

PR #4163, opened 2026-03-03
Request-level errors cause retries, network errors return failure, other errors are silently ignored

jvh-aws:
Change error handling so that

 * request-level errors cause retries
 * network errors or unknown errors throw exceptions
 * sdk errors are silently ignored (and remove the public facing error types respectively)

And add tests to verify the given behavior


AmplifyClients/AmplifyKinesisClient/Sources/Support/KinesisError.swift
```diff
@@ -9,189 +9,102 @@ import AmplifyFoundation
...
-    case kmsThrottling(KMSThrottlingException)
-    /// The request rate for the stream is too high, or the requested data is too large.
-    case provisionedThroughputExceeded(ProvisionedThroughputExceededException)
-    /// The requested resource could not be found.
-    case resourceNotFound(ResourceNotFoundException)
-    /// An unexpected SDK error.
-    case unknown(Error)
-}
-
-extension KinesisPutRecordsSdkError: Error {
```
jvh-aws, line 40: Since we're not exposing the SDK errors, we do not need this type anymore

PR #4192, opened 2026-04-10
fix(ci): include Xcode version in cache keys to prevent stale artifacts

harsh62:
## Summary
- Cache keys for dependencies and build artifacts did not include the Xcode version
- After upgrading `LATEST_XCODE_VERSION` from 16.4 to 26.3, stale cached artifacts from Xcode 16.4 were restored, causing linker failures (`Undefined symbol: SmithyXML.Reader.from`)
- Add `${{ steps.platform.outputs.xcode-version }}` to all cache keys across 5 workflows so caches are automatically invalidated on Xcode version changes

**New key format:**
- `amplify-packages-<xcode-version>-<Package.resolved hash>`
- `Amplify-<platform>-<xcode-version>-build-cache`

**Note:** First CI run after merge will be a cold build (no cache hit) since the key format changed. Subsequent runs will cache normally.

## Test plan
- [ ] Verify builds pass without cache hits on first run
- [ ] Verify cache is saved and restored correctly on subsequent runs
- [ ] Confirm no more `SmithyXML` linker failures

### DIFF
--- a/.github/workflows/integ_test_auth_webauthn.yml
+++ b/.github/workflows/integ_test_auth_webauthn.yml
@@ -50,9 +50,9 @@ jobs:
         uses: actions/cache/restore@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
         with:
           path: ~/Library/Developer/Xcode/DerivedData/Amplify
-          key: amplify-packages-${{ hashFiles('Package.resolved') }}
+          key: amplify-packages-${{ steps.platform.outputs.xcode-version }}-${{ hashFiles('Package.resolved') }}
           restore-keys: |
-            amplify-packages-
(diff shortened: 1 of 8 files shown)


.github/workflows/integ_test_auth_webauthn.yml
```diff
@@ -25,6 +29,13 @@ jobs:
         uses: ./.github/composite_actions/get_platform_parameters
         with:
           platform: iOS
+          xcode_version: ${{ inputs.xcode_version || 'latest' }}
```
thisisabhash, line 32: can just be `xcode_version: ${{ inputs.xcode_version}}` as default is set on line 8

thisisabhash (approved):

PR #4196, opened 2026-04-17
fix(auth): Consistent device metadata keychain key across auth flows

harsh62:
## Summary

- Device metadata keychain key was derived from SignedInData.username (JWT claim), which differs between USER_PASSWORD_AUTH (returns internal UUID) and USER_SRP_AUTH (returns user alias like email). This caused device key lookup failures on subsequent logins when auth flows change, generating duplicate device IDs.
- Thread the original user-typed username (inputUsername) through SignedInData, RespondToAuthChallenge, and the parseResponse/sendRespondToAuth helper chain so ConfirmDevice stores device metadata under the same key InitializeSignInFlow uses for retrieval.
- Normalize the device metadata keychain key to lowercase in generateDeviceMetadataKey. Cognito normalizes usernames to lowercase in JWT claims and challenge parameters, but the user-typed input preserves original casing. Without normalization, storing with inputUsername (mixed case) and retrieving with the JWT username (lowercase) causes a key mismatch.
- inputUsername is optional on both structs for backwards compatibility with existing keychain data.

## Test plan

- [x] SignedInData Codable backwards compat (with and without inputUsername)
- [x] RespondToAuthChallenge propagates inputUsername through challenge chains
- [x] ConfirmDevice stores metadata under inputUsername, falls back to JWT username
- [x] Device metadata keychain key is case-insensitive (lowercased)
- [x] Integration test: AuthRememberDeviceTests.testSuccessfulRememberDevice passes
- [x] All 31 unit tests pass, 0 failures

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/SignIn/SRPAuth/InitiateAuthSRP.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/SignIn/SRPAuth/InitiateAuthSRP.swift
@@ -145,13 +145,15 @@ struct InitiateAuthSRP: Action {
         logVerbose("\(#fileID) InitiateAuth response success", environment: environment)
         if case .customChallenge = response.challengeName {
             let parameters = response.challengeParameters
+            let inputUsername = username
             let username = parameters?["USERNAME"] ?? username
             let respondToAuthChallenge = RespondToAuthChallenge(
                 challenge: .customChallenge,
                 availableChallenges: [],
                 username: username,
                 session: response.session,
-                parameters: parameters
+                parameters: parameters,
+                inputUsername: inputUsername
             )
             return SignInEvent(eventType: .receivedChallenge(respondToAuthChallenge))
         }
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/RespondToAuthChallenge.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/RespondToAuthChallenge.swift
@@ -21,6 +21,8 @@ struct RespondToAuthChallenge: Equatable {
 
     let parameters: [String: String]?
 
+    var inputUsername: String?
+
 }
 
 extension RespondToAuthChallenge {
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/SignedInData.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/SignedInData.swift
@@ -15,12 +15,14 @@ struct SignedInData {
     let deviceMetadata: DeviceMetadata
     let cognitoUserPoolTokens: AWSCognitoUserPoolTokens
     var isRefreshTokenExpired: Bool?
+    var inputUsername: String?
 
     init(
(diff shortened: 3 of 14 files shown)


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/SignedInData.swift
thisisabhash: nit: we can have the new property `inputUsername` in this dictionary

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/RespondToAuthChallenge.swift
```diff
@@ -21,6 +21,8 @@ struct RespondToAuthChallenge: Equatable {
 
     let parameters: [String: String]?
 
+    var inputUsername: String?
```
thisisabhash, line 24: should this be `let` ?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/SignIn/SRPAuth/InitiateAuthSRP.swift
```diff
@@ -145,13 +145,15 @@ struct InitiateAuthSRP: Action {
         logVerbose("\(#fileID) InitiateAuth response success", environment: environment)
         if case .customChallenge = response.challengeName {
             let parameters = response.challengeParameters
+            let inputUsername = username
```
thisisabhash, line 148: do we need a similar change in `InitiateAuthDeviceSRP`?

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/SignedInData.swift
harsh62: Done — added `inputUsername` to the `debugDictionary` in the latest commit.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/StateMachine/CodeGen/Data/RespondToAuthChallenge.swift
harsh62: Good catch. Changed `SignedInData.inputUsername` to `let` since it has an explicit `init`. `RespondToAuthChallenge.inputUsername` stays as `var` because the struct relies on Swift's synthesized memberwise init — `var` properties with `Optional` type get an implicit `nil` default, but `let` properties don't, which would require updating all existing construction sites that don't pass `inputUsername`.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/SignIn/SRPAuth/InitiateAuthSRP.swift
harsh62: No change needed in `InitiateAuthDeviceSRP`. Its `parseResponse` only emits `.respondDevicePasswordVerifier(stateData, response)` which feeds into `VerifyDevicePasswordSRP` — that action already passes `inputUsername` through to `sendRespondToAuth`. `InitiateAuthDeviceSRP` never directly produces `confirmDevice` or `finalizeSignIn` events, so the user input doesn't need to be threaded here.

thisisabhash (approved):

PR #4202, opened 2026-04-23
fix(api): recycle WebSocket and resubscribe on same-online network path change

### LINKED ISSUE #3976: AppSync Subscription Disconnects After 2-3 Minutes (NSPOSIXErrorDomain Code 57)
### Describe the bug

AppSync GraphQL subscriptions established using Amplify.API.subscribe with a custom GraphQLRequest consistently disconnect after a specific period of inactivity (no messages received or sent over the subscription). We observe this reliably when the app is in the foreground and the subscription is active but idle. Our backend setup, involving a custom mutation handled by a Lambda function triggering a custom subscription, follows the patterns described in the Amplify Gen 2 documentation for this type of real-time flow.

The disconnection manifests as a failure in the Combine sink's receiveCompletion block, reporting APIError: Subscription item event failed with error caused by the underlying error Error Domain=NSPOSIXErrorDomain Code=57 "Socket is not connected". The timing of this disconnection appears highly consistent: it occurs almost **exactly at 120 seconds (2 minutes)** of inactivity, or, less frequently but still precisely, at exactly **180 seconds (3 minutes)** of inactivity.

This issue occurs when using the latest stable Amplify Swift libraries, even with simplified subscription lifecycle code. Because the basic setup failed, we subsequently implemented custom client-side retry logic (attempting to re-establish the subscription upon detecting the Code 57 error), however, this retry mechanism itself failed to function correctly. We frequently observed [AWSGraphQLSubscriptionTaskRunner] Failed to unsubscribe logs during these disconnection events, suggesting the SDK's internal state prevented a clean teardown and re-initialization of the subscription, rendering the retry logic ineffective. The SDK's built-in WebSocket keep-alive mechanism does not seem to prevent the initial, specific, timed disconnection. We believe network connectivity is generally good when this occurs. Restarting the app allows reconnection, but the idle disconnect happens again after the same exact interval.

### Steps To Reproduce

```swift
1.  Configure an AWS AppSync backend with:
 A custom GraphQL mutation (e.g., publishMessage) handled by a Lambda function.
 A custom GraphQL subscription (e.g., subscribeToMessages) linked to the mutation via .for(a.ref('publishMessage')) in the schema definition (resource.ts).
2.  In the iOS client app (using Amplify Swift SDK):
 Configure Amplify with AWSAPIPlugin.
 Implement code to subscribe to the custom subscribeToMessages using Amplify.API.subscribe and a GraphQLRequest.
3.  Run the app and successfully establish the subscription (observe .connection(.connected) state is reached).
4.  Keep the app in the foreground and ensure the subscription remains active.
5.  Do not trigger the publishMessage mutation (i.e., ensure no events are sent over the subscription, allowing it to remain idle).
6.  Wait for approximately 2 to 3 minutes.
7.  Observe that the Combine sink's receiveCompletion handler associated with the subscription receives a .failure event containing the NSPOSIXErrorDomain Code=57 "Socket is not connected" error.
```

### Expected behavior

The WebSocket connection for the AppSync GraphQL subscription should remain active indefinitely (as long as the client maintains the subscription) via the SDK's automatic keep-alive mechanisms, without disconnecting due to short periods (2-3 minutes) of inactivity.

### Amplify Framework Version

2.46.1

### Amplify Categories

API

### Dependency manager

Swift PM

### Swift version

5

### CLI version

1.5.0

### Xcode version

16.2.0

### Relevant log output

```shell

```

### Is this a regression?

No

### Regression additional context

_No response_

### Platforms

iOS

### OS Version

iOS 18.3

### Device

iPhone16Pro

### Specific to simulators

_No response_

### Additional context

_No response_

harsh62:
## Summary

Fixes #3976. Resolves zombie AppSync subscriptions after iOS recycles the TCP route while the network remains satisfied (e.g. during a `scenePhase` `inactive → active` transition on the same Wi-Fi network).

## Problem

From the bug report, a customer's app was connected with active subscriptions when:

1. iOS fired a `scenePhase` transition (inactive → active)
2. `NWPathMonitor` re-evaluated and reported `path.status == .satisfied` (network was genuinely still up — REST calls in the same millisecond succeeded)
3. But iOS had internally recycled the TCP route during the phase transition, so the existing `URLSessionWebSocketTask` was bound to a dead path
4. All 5 active subscriptions started throwing `APIError 4` within milliseconds

The same pattern was reproducible deterministically: once this happened, no amount of new `subscribe()` calls would recover, because `AppSyncRealTimeClientFactory.apiToClientCache` kept handing out the same zombied `AppSyncRealTimeClient` whose underlying `WebSocketClient` held the dead socket.

## Root cause

Two layered bugs that together produced the zombie state.

### Bug 1 — `WebSocketClient.onNetworkStateChange` silently dropped path changes

`AmplifyPlugins/Core/AWSPluginsCore/WebSocket/WebSocketClient.swift:280`

```swift
switch stateChange {
case (.online, .offline):  // tear down
case (.offline, .online):  // reconnect
default: break             // ← (.online, .online) silently dropped
}
```

`AmplifyNetworkMonitor.pathUpdateHandler` maps any `path.status == .satisfied` to `.online`, and `NWPathMonitor.pathUpdateHandler` only fires on *real* path changes. So a second `.online` emission while the client was already online is `NWPathMonitor`'s way of signalling: *"the path just changed, but the network is still up."* The switch dropped that signal, and `URLSessionWebSocketTask.state` stayed `.running` indefinitely while reads/writes silently failed on the stale route.

### Bug 2 — `AppSyncRealTimeSubscription.subscribe()` short-circuited on stale `.subscribed` state

`AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeSubscription.swift:60`

```swift
guard state.value != .subscribed else {
    log.debug("[AppSyncRealTimeSubscription-\(id)] Subscription already in subscribed state")
    return
}
```

Even after Bug 1 is fixed and the WebSocket is recycled, `AppSyncRealTimeClient.resumeExistingSubscriptions()` calls `subscription.subscribe()` on each surviving subscription. But each subscription's local `state` is still `.subscribed` from before the recycle, so `subscribe()` early-returns without actually re-sending the `start` request. The AppSync server has no memory of the subscription (its own connection died), so the subscription is silently broken from the server's perspective while the local state insists everything is fine.

### Change 1 — handle `(.online, .online)` as a forced recycle

`AmplifyPlugins/Core/AWSPluginsCore/WebSocket/WebSocketClient.swift`

Added an explicit `case (.online, .online)` that cancels the stale `URLSessionWebSocketTask` with `.invalid`, emits a synthetic `.disconnected(.invalid, nil)` event (so `AppSyncRealTimeClient` transitions to `.connectionDropped`), and calls `createConnectionAndRead()` to re-establish. Guarded by `connection?.state == .running` to avoid recycling a connection that's already torn down (e.g. during app backgrounding followed by a path change).

### Change 2 — reset subscription state before resuming

`AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeSubscription.swift`

Added `prepareForResubscribe()` which sends `.none` through the state subject so a subsequent `subscribe()` call will actually transit `.subscribing → .subscribed` and re-send the `start` request to the server.

### Change 3 — call the reset during resume

`AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift`

`resumeExistingSubscriptions()` now calls `subscription.prepareForResubscribe()` before awaiting `startSubscription(id)`.

## How the end-to-end flow works after the fix

1. NWPath recycle → `AmplifyNetworkMonitor` emits `.online`
2. Publisher scan produces `(.online, .online)`
3. `WebSocketClient.onNetworkStateChange` cancels stale task + sends `.disconnected` + creates fresh connection
4. `AppSyncRealTimeClient` receives `.disconnected` → state becomes `.connectionDropped`
5. `AppSyncRealTimeClient` receives `.connected` → detects `.connectionDropped` state → calls `connect()` → sends `.connectionInit`
6. Server replies `connection_ack` → `onAppSyncRealTimeResponse` calls `resumeExistingSubscriptions()`
7. For each surviving subscription: `prepareForResubscribe()` → `subscribe()` → `start` request sent → `start_ack` received → subscription re-established

Verified by a trace print during development — here's the actual event sequence observed against a real AppSync backend (prints removed before merge):

```
WS event: connected
WS event: string {\"type\":\"connection_ack\",...}
AppSyncSubscriptionEvent received: subscribed
===== emit second .online =====
onNetworkStateChange tuple=(.online, .online) autoConnect=true
WS event: disconnected code=0 reason=nil
WS event: connected
WS event: string {\"type\":\"connection_ack\",...}
AppSyncSubscriptionEvent received: subscribing
WS event: string {\"id\":\"...\",\"type\":\"start_ack\"}
AppSyncSubscriptionEvent received: subscribed    ← recovery ✅
```

## Verification

Three new regression tests were added. All three fail against `main` and pass with this fix. No pre-existing tests regressed.

### Unit tests — `AWSPluginsCoreTests/WebSocketClientTests`

| Test | Purpose | Pre-fix | Post-fix |
|---|---|---|---|
| `testAmplifyNetworkMonitor_whenOnlineEmittedTwice_publishesOnlineOnlineTuple` | Characterize input: prove the real `AmplifyNetworkMonitor` does emit the `(.online, .online)` tuple when two `.online` states are pushed consecutively. Documents the bug's input signal. | passes | passes |
| `testWebSocketClient_whenNetworkPathChangesWhileOnline_shouldRecycleConnection` | Drive `WebSocketClient` via `MockNetworkMonitor` (scan-seeded at `(.online, .online)`) and assert the connection recycles. | fails (15s timeout) | passes |
| `testWebSocketClient_withRealNetworkMonitor_whenPathChangesWhileOnline_shouldRecycle` | Companion to above using the real `AmplifyNetworkMonitor` (scan-seeded at `(.none, .none)`), primed with one `.online` emission to reach steady state, then second `.online` to produce the bug tuple. Rules out the mock seed as an artifact. | fails (15s timeout) | passes |

All 5 pre-existing `WebSocketClientTests` continue to pass.

### Integration test — `AWSAPIPluginFunctionalTests/AppSyncRealTimeClientTests`

| Test | Purpose | Pre-fix | Post-fix |
|---|---|---|---|
| `testSubscribe_afterOnlineToOnlinePathChange_shouldRecycleAndResubscribe` | End-to-end against a real AppSync endpoint. Creates a live subscription, then drives an `(.online, .online)` transition through an injected `AmplifyNetworkMonitor` and asserts the subscription is re-established (server returns a second `start_ack`). | fails (15s timeout on resubscribe expectation) | passes |

All 3 pre-existing functional tests in the suite continue to pass. All 13 `AWSAPIPluginTests/AppSyncRealTimeClientTests` unit tests continue to pass.

### Test plan

- [x] `xcodebuild test -scheme AWSPluginsCore -only-testing:AWSPluginsCoreTests/WebSocketClientTests` — 8/8 pass
- [x] `xcodebuild test -scheme AWSAPIPlugin -only-testing:AWSAPIPluginTests/AppSyncRealTimeClientTests` — 13/13 pass
- [x] `xcodebuild test -project APIHostApp.xcodeproj -scheme AWSAPIPluginFunctionalTests -only-testing:AWSAPIPluginFunctionalTests/AppSyncRealTimeClientTests` — 4/4 pass (requires `GraphQLModelBasedTests-amplifyconfiguration.json`)

## Risk

- `(.online, .online)` path changes can legitimately fire during benign events like Wi-Fi AP roaming on the same SSID, which will cause a reconnect cycle. `RetryWithJitter` already throttles this and `resumeExistingSubscriptions()` will re-establish subscribers transparently.
- `prepareForResubscribe()` is only called from `resumeExistingSubscriptions()`, which itself only fires on `connection_ack` after a reconnect. Not reachable from normal subscription lifecycles.

### DIFF
--- a/AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/AppSyncRealTimeClientTests.swift
+++ b/AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/AppSyncRealTimeClientTests.swift
@@ -169,6 +169,94 @@ class AppSyncRealTimeClientTests: XCTestCase {
         withExtendedLifetime(cancellables) { }
     }
 
+    // End-to-end regression test for https://github.com/aws-amplify/amplify-swift/issues/3976
+    //
+    // Drives a real AppSync subscription through an AmplifyNetworkMonitor we
+    // own, then simulates the scenePhase-triggered NWPath recycle by sending
+    // a second .online state. In the buggy WebSocketClient, the scan produces
+    // (.online, .online), onNetworkStateChange hits `default: break`, the
+    // URLSessionWebSocketTask is left zombied, and no new subscription events
+    // arrive. With the fix, the client tears down and reconnects, and the
+    // subscription is re-established.
+    //
+    // Requires GraphQLModelBasedTests-amplifyconfiguration.json at
+    // ~/.aws-amplify/amplify-ios/testconfiguration/ (same as all other tests
+    // in this file). If missing, test fails in setUp rather than here.
+    func testSubscribe_afterOnlineToOnlinePathChange_shouldRecycleAndResubscribe() async throws {
+        var cancellables = Set<AnyCancellable>()
+
+        let data = try TestConfigHelper.retrieve(
+            forResource: GraphQLModelBasedTests.amplifyConfiguration
+        )
+        let amplifyConfig = try JSONDecoder().decode(JSONValue.self, from: data)
+        let (endpoint, apiKey) = (amplifyConfig.api?.plugins?.awsAPIPlugin?.asObject?.values
+            .map { ($0.endpoint?.stringValue, $0.apiKey?.stringValue) }
+            .first { $0.0 != nil && $0.1 != nil }
+            .map { ($0.0!, $0.1!) })!
+
+        // Inject a real AmplifyNetworkMonitor we can drive directly.
+        let networkMonitor = AmplifyNetworkMonitor()
+
+        let webSocketClient = WebSocketClient(
+            url: AppSyncRealTimeClientFactory.appSyncRealTimeEndpoint(URL(string: endpoint)!),
+            handshakeHttpHeaders: [
+                URLRequestConstants.Header.webSocketSubprotocols: "graphql-ws",
+                URLRequestConstants.Header.userAgent: AmplifyAWSServiceConfiguration.userAgentLib + " (intg-test-3976)"
+            ],
+            interceptor: APIKeyAuthInterceptor(apiKey: apiKey),
+            networkMonitor: networkMonitor
+        )
+        let client = AppSyncRealTimeClient(
+            endpoint: URL(string: endpoint)!,
+            requestInterceptor: APIKeyAuthInterceptor(apiKey: apiKey),
+            webSocketClient: webSocketClient
+        )
+        defer { Task { await client.reset() } }
+
+        // Wait for WebSocketClient's internal sink to attach to the monitor's
+        // publisher (it's kicked off via a Task in init). Prime with .online
+        // AFTER the subscriber is attached so the PassthroughSubject actually
+        // delivers the event. This gets the scan to (.none, .online) —
+        // WebSocketClient will ignore it because autoConnect is still false.
+        try await Task.sleep(nanoseconds: 200_000_000)
+        await networkMonitor.updateState(.online)
+
+        let firstSubscribed = expectation(description: "Initial subscription established")
+        let resubscribedAfterPathChange = expectation(description: "Subscription re-established after (.online, .online)")
+        resubscribedAfterPathChange.assertForOverFulfill = false
+
+        let id = UUID().uuidString
+        var subscribedCount = 0
+        let subscription = try await client.subscribe(
+            id: id,
+            query: Self.appSyncQuery(with: subscriptionRequest)
+        ).sink { event in
+            if case .subscribed = event {
+                subscribedCount += 1
+                if subscribedCount == 1 {
+                    firstSubscribed.fulfill()
+                } else {
+                    resubscribedAfterPathChange.fulfill()
+                }
+            }
+        }
+        cancellables.insert(subscription)
+
+        try await client.connect()
+        await fulfillment(of: [firstSubscribed], timeout: 10)
+
+        // Simulate the path-recycle: second .online emission produces
+        // (.online, .online) through the scan — the exact bug tuple.
+        // In the buggy code, nothing happens; WebSocketClient keeps the
+        // zombie connection. With the fix, it should tear down and reconnect,
+        // and AppSyncRealTimeClient.resumeExistingSubscriptions() should
+        // re-subscribe.
+        await networkMonitor.updateState(.online)
+
+        await fulfillment(of: [resubscribedAfterPathChange], timeout: 15)
+        withExtendedLifetime(cancellables) { }
+    }
+
     private func makeOneSubscription(
         id: String = UUID().uuidString,
         onSubscriptionEvents: ((AppSyncSubscriptionEvent) -> Void)?
--- a/AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/AppSyncRealTimeClientTests.swift
+++ b/AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/AppSyncRealTimeClientTests.swift
@@ -169,19 +169,28 @@ class AppSyncRealTimeClientTests: XCTestCase {
         withExtendedLifetime(cancellables) { }
     }
 
-    // End-to-end regression test for https://github.com/aws-amplify/amplify-swift/issues/3976
-    //
-    // Drives a real AppSync subscription through an AmplifyNetworkMonitor we
-    // own, then simulates the scenePhase-triggered NWPath recycle by sending
-    // a second .online state. In the buggy WebSocketClient, the scan produces
-    // (.online, .online), onNetworkStateChange hits `default: break`, the
-    // URLSessionWebSocketTask is left zombied, and no new subscription events
-    // arrive. With the fix, the client tears down and reconnects, and the
-    // subscription is re-established.
-    //
-    // Requires GraphQLModelBasedTests-amplifyconfiguration.json at
-    // ~/.aws-amplify/amplify-ios/testconfiguration/ (same as all other tests
-    // in this file). If missing, test fails in setUp rather than here.
+    /// End-to-end regression test for https://github.com/aws-amplify/amplify-swift/issues/3976
+    /// against a real AppSync backend. Simulates the scenePhase-triggered
+    /// NWPath recycle by driving two .online states through an injected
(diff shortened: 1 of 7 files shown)


AmplifyPlugins/Core/AWSPluginsCoreTests/WebSocket/WebSocketClientTests.swift
```diff
@@ -139,6 +139,160 @@ class WebSocketClientTests: XCTestCase {
...
+    // on the stale route, but WebSocketClient.onNetworkStateChange hits
+    // `default: break` and never cancels/recreates the connection. Every cached
+    // subscriber then reuses the zombie client.
+    //
+    // This test drives the mock through an (.online, .online) transition while
+    // the client is connected with autoConnectOnNetworkStatusChange = true, and
+    // expects the client to tear down and re-establish the connection.
+    // It should FAIL against the current implementation.
+    func testWebSocketClient_whenNetworkPathChangesWhileOnline_shouldRecycleConnection() async throws {
+        var cancellables = Set<AnyCancellable>()
```
jvh-aws, line 160: Does it make sense to have Given/When/Then comments in this test according to [our guidance](https://github.com/aws-amplify/amplify-swift/blob/main/AGENTS.md#testing)?

AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/AppSyncRealTimeClientTests.swift
```diff
@@ -169,6 +169,103 @@ class AppSyncRealTimeClientTests: XCTestCase {
         withExtendedLifetime(cancellables) { }
     }
 
+    /// End-to-end regression test for https://github.com/aws-amplify/amplify-swift/issues/3976
+    /// against a real AppSync backend. Simulates the scenePhase-triggered
+    /// NWPath recycle by driving two .online states through an injected
+    /// AmplifyNetworkMonitor and asserts that the subscription is actually
+    /// re-established (server issues a second start_ack).
+    ///
+    /// Requires GraphQLModelBasedTests-amplifyconfiguration.json at
```
jvh-aws: I think this comment is obsolete (nit)

AmplifyPlugins/API/Tests/APIHostApp/AWSAPIPluginFunctionalTests/AppSyncRealTimeClientTests.swift
```diff
@@ -169,6 +169,103 @@ class AppSyncRealTimeClientTests: XCTestCase {
...
+        // WebSocketClient will ignore it because autoConnect is still false.
+        try await Task.sleep(nanoseconds: 200_000_000)
+        await networkMonitor.updateState(.online)
+
+        let firstSubscribed = expectation(description: "Initial subscription established")
+        let resubscribedAfterPathChange = expectation(description: "Subscription re-established after (.online, .online)")
+        resubscribedAfterPathChange.assertForOverFulfill = false
+
+        let id = UUID().uuidString
+        var subscribedCount = 0
```
jvh-aws: Should this be `AtomicInt` as in the unit test?

jvh-aws (approved):

PR #4222, opened 2026-05-11
fix(api): emit disconnected directly from didCompleteWithError to trigger auto-retry

### LINKED ISSUE #4220: GraphQL subscriptions can get stuck in connecting after repeated iOS background/foreground transitions
## Summary

We have a repro repo for an iOS GraphQL subscription reconnection issue after app lifecycle transitions.

With the current `feature/multi-subscription-repro` branch, we are now able to reproduce a stronger failure mode that is closer to our production app:

- one or more subscriptions enter `connecting`
- we have not observed them recover without fully terminating the app
- the app does not recover by foregrounding again or by normal in-app actions

This stuck state is not reproducing on the simpler `main` branch of the repro repo. It is currently reproducing only on the multi-subscription variant.

We also observe a difference between build configurations:

- the issue is harder to reproduce in `Debug`
- the issue reproduces more easily in `Release`

## Repro repository

- Repo: `https://github.com/akiramur/amplify-swift-ios-repros`
- Branch: `feature/multi-subscription-repro`
- Commit: `6daca6d404771ab3cdac435976eacf0577958b40`

## Repro shape

This repro app uses:

- Cognito auth with Hosted UI sign-in
- AppSync GraphQL API
- one GraphQL model
- three concurrent `onCreate` subscription workers
- background/foreground restart logic across all workers
- in-app logging for worker connection state and lifecycle transitions

## Environment

- Amplify Swift: `2.58.1`
- Xcode: `26.3`
- iOS: `26.4.2`
- Device: `iPhone 12`
- Repro backend: `Amplify Gen 2 sandbox`

## Actual behavior

After repeating lock/unlock and app switching several times, one or more subscription workers eventually enter `connecting` and remain stuck there.

Once this happens:

- one or more stuck subscription workers do not return to `connected`
- normal app usage does not recover them
- repeated foreground/background transitions do not recover them
- the app remains in a degraded state until it is fully terminated and relaunched

In other words, after the stuck state is reached, a full app restart is the only reliable recovery we have observed.

## Important observation

This behavior is currently reproducing only on the `feature/multi-subscription-repro` branch, which runs multiple concurrent subscription workers.

We are not reproducing the same stuck `connecting` state on the simpler `main` branch of the repro repo.

This multi-subscription branch appears closer to our production shape, where multiple subscriptions and lifecycle-driven restarts are active concurrently.

This suggests the issue may be related to one or more of:

- multiple concurrent subscriptions
- lifecycle-driven restart timing across multiple workers
- reconnection races that do not appear with a single subscription

### Steps To Reproduce

```swift
Steps to reproduce the behavior:

1. Start the backend sandbox and generate `amplify_outputs.json`.
2. Generate the Xcode project and run the app from the `feature/multi-subscription-repro` branch.
3. Sign in with Hosted UI.
4. Confirm all subscription workers reach `connected`.
5. Lock the device screen and leave the app idle for around 2 to 3 minutes.
6. Unlock the device.
7. Repeatedly switch between this app and other apps so the repro app cycles through background and foreground several times after unlocking.
8. Observe the aggregate connection badge, worker cards, and event log.
```

## Expected behavior

After app lock/unlock and repeated background/foreground transitions, all subscription workers should eventually reconnect and return to `connected` without requiring a full app restart.

### Amplify Framework Version

2.58.1

### Amplify Categories

API

### Dependency manager

Swift PM

### Swift version

5.9

### CLI version

13.0.0

### Xcode version

26.3 (17C529)

### Relevant log output

```shell
<details>
<summary>Log Messages</summary>


INSERT LOG MESSAGES HERE

</details>
```

### Is this a regression?

No

### Regression additional context

_No response_

### Platforms

_No response_

### OS Version

26.4.2

### Device

iPhone 12

### Specific to simulators

_No response_

### Additional context

_No response_

thisisabhash:
## Issue \#

#4220

## Summary

Fixes #4220. Resolves GraphQL subscriptions getting permanently stuck in `connecting` state after repeated iOS background/foreground transitions.

## Problem

When iOS suspends the app process, the kernel defuncts TCP flows bound to the WebSocket connection. On foreground return, `didCompleteWithError` fires with `NSPOSIXErrorDomain Code=53 "Software caused connection abort"`. The old code handled this by calling `networkMonitor.updateState(.offline)`, expecting `NWPathMonitor` to later fire `.online` to trigger reconnection. But `NWPathMonitor` only fires on **path changes** — if the network path never actually changed (the socket just died from process suspension), no `.online` event arrives and the client stays dead permanently.

## Root Cause

Reconnection after connection abort depends on `NWPathMonitor` firing `.online`:

```swift
case (NSPOSIXErrorDomain.self, Int(ECONNABORTED)):
    subject.send(.error(WebSocketClient.Error.connectionLost))
    Task { await self?.networkMonitor.updateState(.offline) }
```

This pushes `.offline` to the network monitor scan, but reconnection only happens when `NWPathMonitor` fires `.online` to produce `(.offline, .online)`. If the network path never changed (process suspension doesn't change the network path), no `.online` arrives and the client is permanently stuck.

## Fix

**File:** `WebSocketClient.swift`

Instead of going through the network monitor, directly cancel the stale connection and emit `.disconnected(.invalid, nil)`. This triggers the existing auto-retry mechanism (`retryOnCloseCode(.invalid)`) which creates a new connection — no dependency on `NWPathMonitor`.

```swift
case (NSPOSIXErrorDomain.self, Int(ECONNABORTED)),
     (NSPOSIXErrorDomain.self, 57):
    subject.send(.error(WebSocketClient.Error.connectionLost))
    Task { [weak self] in
        await connection?.cancel(with: .invalid, reason: nil)
        subject.send(.disconnected(.invalid, nil))
    }
```

### Unit test — `AWSPluginsCoreTests/WebSocketClientTests` (1 new test)

| Test | What it verifies |
|------|-----------------|
| `testWebSocketClient_whenConnectionAborts_shouldEmitDisconnectedAndReconnect` | Connection abort emits `.disconnected` and auto-retry reconnects |

### Test plan

```bash
swift test --filter "WebSocketClientTests"  # 9/9 pass (8 existing + 1 new, no regressions)
```


jvh-aws (dismissed):
nit: can we run swiftformat?

PR #4228, opened 2026-06-04
Revert "fix(api): recycle WebSocket and resubscribe on same-online network path change (#4202)"

harsh62:
## Summary

Reverts #4202 (commit `be0a21462`), which is a confirmed regression source for #4220 — GraphQL subscriptions getting stuck in `connecting` after repeated iOS background/foreground transitions.

@akiramur confirmed that v2.56.0 (before #4202 landed) does not reproduce the issue, and the regression was traced to the `(.online, .online)` WebSocket recycle path introduced in #4202.

## What this reverts

- **`WebSocketClient.swift`** — removes the explicit `(.online, .online)` case in `onNetworkStateChange` that cancelled and re-established the connection on a second `.satisfied` emission. Behavior returns to handling only `online↔offline` transitions.
- **`AppSyncRealTimeSubscription.swift` / `AppSyncRealTimeClient.swift`** — removes `prepareForResubscribe()` and its invocation in `resumeExistingSubscriptions()`.
- Regression tests added for #3976 (`WebSocketClientTests`, `AppSyncRealTimeClientTests`).
- `AGENTS.md` wording change.

## ⚠️ Follow-up required

This revert also removes the fix for #3976 (zombie AppSync subscriptions after a same-network TCP route swap). That issue will reappear and **needs a different, non-regressing fix**. Reverting here is the fast path to unblock the #4220 regression; #3976 should be re-addressed in a separate PR.

Re #4220
Re #3976


harsh62:
Follow-up tracked in #4229 — the #3976 fix needs to be re-attempted without regressing #4220.

jvh-aws (approved):

PR #4243, opened 2026-07-01
feat: add event enrichment client

thisisabhash:
## Issue \#

None

## Summary

Adds `AmplifyEventEnrichmentClient` — a standalone on-device event enrichment client that attaches device/app/session/identity context onto analytics events and produces a Pinpoint-compatible JSON envelope (v3.1).

This is the Swift counterpart to the Flutter `amplify_event_enrichment_dart` package ([amplify-flutter#7029](https://github.com/aws-amplify/amplify-flutter/pull/7029)). It enables events sent via Kinesis/Firehose to retain the analytics context that Pinpoint previously provided automatically.

### Key design points

- **Actor-based** with full strict concurrency (`Sendable` everywhere)
- **Zero AWS SDK dependency** — depends only on `AmplifyFoundation`; transport is delegated to the `EventSink` protocol
- **Automatic clientId resolution** — persisted UUID via `UserDefaults` (key: `com.amazonaws.amplify.event_enrichment.client_id`) with read-or-create pattern, matching the Flutter `SharedPreferencesClientIdProvider` approach
- **Automatic lifecycle tracking** — subscribes to platform `NotificationCenter` notifications (same approach as the Pinpoint plugin's `ActivityTracker`) to pause/resume sessions on background/foreground transitions
- **Device metadata aligned with Pinpoint plugin** — uses `UIDevice`/`WKInterfaceDevice`/IOKit matching existing `DeviceInfo` patterns
- **Session timeout** — configurable duration (default 5s) after which a backgrounded session expires and a new one starts on next foreground

### Public API

| Type | Role |
|------|------|
| `AmplifyEventEnrichmentClient` | Main actor — `record()`, session control, global fields, userId, close |
| `EnrichedEvent` | Immutable event model with `toJson()` → Pinpoint envelope |
| `Session` | Session model with `Date` timestamps + duration |
| `EventSink` (protocol) | Transport abstraction (Kinesis, Firehose, custom) |
| `EventEnrichmentClientOptions` | Config: `autoSessionTracking`, `sessionTimeout` |
| `EventEnrichmentError` | `AmplifyError`-conforming error enum |
| `AppMetadata` / `DeviceMetadata` / `SDKMetadata` | Metadata value types |

### Files

- 14 source files in `AmplifyClients/AmplifyEventEnrichmentClient/Sources/`
- 5 test files with 27 unit tests covering: recording, global fields, session lifecycle, timeout, JSON serialization, event sink, activity tracker integration
- `Package.swift` — new target + product added

## Test plan

- [x] `swift build --target AmplifyEventEnrichmentClient` compiles cleanly
- [x] `swift test --filter AmplifyEventEnrichmentClientTests` — 27 tests pass
- [ ] Verify JSON output matches Pinpoint event ingestion format
- [ ] Integration test with `AmplifyKinesisClient` or `AmplifyFirehoseClient` as an `EventSink`
- [ ] Test on iOS device to verify lifecycle notifications fire correctly
- [ ] Test on macOS to verify `NSApplication` notification path

## General Checklist

- [x] All unit tests pass


AmplifyClients/AmplifyEventEnrichmentClient/Sources/Support/DeviceMetadataProvider.swift
```diff
@@ -0,0 +1,90 @@
...
+/// ``PlatformDeviceMetadataProvider`` uses the same system APIs as the
+/// Pinpoint analytics plugin.
+protocol DeviceMetadataProvider: Sendable {
+    /// Returns device metadata for the current platform.
+    @MainActor func getDeviceMetadata() -> DeviceMetadata
+}
+
+/// Default ``DeviceMetadataProvider`` aligned with the Pinpoint plugin's
+/// device information gathering (uses UIDevice/WKInterfaceDevice/IOKit).
+struct PlatformDeviceMetadataProvider: DeviceMetadataProvider {
```
ekjotmultani, line 24: I may have missed it but is this being used?

AmplifyClients/AmplifyEventEnrichmentClient/Sources/Support/ClientIDProvider.swift
```diff
@@ -0,0 +1,26 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+private let clientIDKey = "com.amazonaws.amplify.event_enrichment.client_id"
```
ekjotmultani: I have this as `com.amplifyframework.device_id` on [flutter](https://github.com/aws-amplify/amplify-flutter/pull/7029/changes#diff-3c95228178299985a78f24c3580288778eacba230eac926b6ef56119830d7252R13), we should align

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
```diff
@@ -0,0 +1,202 @@
...
+    public init(
+        appId: String,
+        sdkMetadata: SDKMetadata,
+        appMetadata: AppMetadata? = nil,
+        deviceMetadata: DeviceMetadata? = nil,
+        options: EventEnrichmentClientOptions = EventEnrichmentClientOptions(),
+        sink: (any EventSink)? = nil
+    ) {
+        let resolvedAppMetadata = appMetadata ?? AppMetadata(appId: appId)
+        let resolvedDeviceMetadata = deviceMetadata ?? DeviceMetadata()
```
ekjotmultani: When `deviceMetadata` is nil (the common case for a native caller), this falls back to an empty `DeviceMetadata()`, so every event enters `"device": {}`. It looks like `PlatformDeviceMetadataProvider` was meant to fill this in, but it's never instantiated anywhere. Since `getDeviceMetadata()` is `@MainActor`, resolving it in the synchronous actor init won't work. Could we add an async factory (e.g. a static create(...) that resolves device metadata on the main actor before constructing the actor)? Like:

```
public static func create(appId: String, sdkMetadata: SDKMetadata, ...) async -> AmplifyEventEnrichmentClient {
    let device = await MainActor.run { PlatformDeviceMetadataProvider().getDeviceMetadata() }
    return AmplifyEventEnrichmentClient(appId: appId, sdkMetadata: sdkMetadata, deviceMetadata: device, ...)
}
```

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
```diff
@@ -0,0 +1,202 @@
...
+        for (key, value) in metrics {
+            mergedMetrics[key] = value
+        }
+
+        let session = await sessionManager.session!
+
+        let event = EnrichedEvent(
+            eventId: UUID().uuidString,
+            eventType: eventType,
+            eventTimestamp: Int(Date().timeIntervalSince1970 * 1_000),
```
cadivus: This overflows and crashes on watchOS. watchOS 9+ is a declared target, and Apple Watch runs arm64_32 where `Int` is 32-bit. Epoch millis today (~1_750_000_000_000) is far past `Int32.max`, so this traps at runtime — a hard crash on every `record()` call. `Session.duration` (also `Int` ms) has the same exposure for long sessions.

Use `Int64` for `eventTimestamp` (declared `EnrichedEvent.swift:21`) and `duration` (`Session.swift:22`).

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
cadivus: The init doc on line 53 says device metadata is resolved via `PlatformDeviceMetadataProvider` when nil, but `DeviceMetadata()` here is all-nil and the provider is never called anywhere. When the caller omits `deviceMetadata`, the event carries an empty `device: {}` instead of the real platform/model/OS, and the whole `DeviceMetadataProvider.swift` file is dead code.

Either call the provider on init (it is `@MainActor`, so hop accordingly) or drop it and document that device metadata is caller-supplied.

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
```diff
@@ -0,0 +1,202 @@
...
+            sessionTimeout: options.sessionTimeout,
+            generateId: { UUID().uuidString }
+        )
+        self.sessionManager = sessionManager
+        if options.autoSessionTracking {
+            self.activityTracker = ActivityTracker(
+                onPause: { Task { await sessionManager.handleAppPaused() } },
+                onResume: { Task { await sessionManager.handleAppResumed() } }
+            )
+            Task { await sessionManager.startSession() }
```
cadivus: Starting the session in a detached `Task` races with the lazy start in `record()` (line 113). If `record()` wins it starts session A, then this task runs `startSession()` and — since the state is already `.active` — stops A and creates session B (`SessionManager.swift:58-60`). The first recorded event ends up referencing a session that was immediately replaced, and the ordering is nondeterministic.

Start the session with awaited setup rather than a detached `Task`.

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
```diff
@@ -0,0 +1,202 @@
...
+    private let globalFields: GlobalFieldsManager
+    private let sessionManager: SessionManager
+    private let activityTracker: ActivityTracker?
+    private var userId: String?
+    private var closed = false
+
+    /// Initializes a new event enrichment client.
+    ///
+    /// The `clientId` is resolved automatically from `UserDefaults` using a
+    /// read-or-create pattern with the key `com.amplifyframework.device_id`.
```
cadivus: This comment does not match the code. The actual key (`ClientIDProvider.swift:10`) is `com.amazonaws.amplify.event_enrichment.client_id` — a different string, namespaced to this client, so it is not shared cross-package. Either fix the comment or, if sharing the device ID across clients was intended, fix the key.

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
```diff
@@ -0,0 +1,202 @@
...
+        metrics: [String: Double] = [:]
+    ) async throws -> EnrichedEvent {
+        guard !closed else {
+            throw EventEnrichmentError.clientClosed(
+                "Client has been closed",
+                "Create a new AmplifyEventEnrichmentClient instance."
+            )
+        }
+
+        if await sessionManager.session == nil {
```
cadivus: This lazy start runs regardless of `autoSessionTracking`, so setting it to `false` still auto-creates a session on the first `record()`. Either gate this on the flag or rename it, since today the behavior contradicts the caller opting out.

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
```diff
@@ -0,0 +1,202 @@
...
+    public func startSession() async {
+        await sessionManager.startSession()
+    }
+
+    /// Stops the current session.
+    public func stopSession() async {
+        await sessionManager.stopSession()
+    }
+
+    func handleAppPaused() async {
```
cadivus: `handleAppPaused()` / `handleAppResumed()` are never called — the `ActivityTracker` closures talk to `sessionManager` directly (lines 83-84). Dead code; remove them.

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
```diff
@@ -0,0 +1,202 @@
...
+        for (key, value) in attributes {
+            mergedAttributes[key] = value
+        }
+
+        var mergedMetrics = globalMetrics
+        for (key, value) in metrics {
+            mergedMetrics[key] = value
+        }
+
+        let session = await sessionManager.session!
```
cadivus: Force-unwrapping the session is only safe today because `_session` is never set back to nil after start — fragile. Prefer `guard let` here.

AmplifyClients/AmplifyEventEnrichmentClient/Sources/Support/EnrichedEvent.swift
```diff
@@ -0,0 +1,107 @@
...
+            "application": applicationMap,
+            "client": clientMap,
+            "device": deviceMap,
+            "session": sessionMap,
+        ]
+        if !attributes.isEmpty { map["attributes"] = attributes }
+        if !metrics.isEmpty { map["metrics"] = metrics }
+
+        guard let data = try? JSONSerialization.data(withJSONObject: map, options: [.sortedKeys]) else {
+            return "{}"
```
cadivus: Serialization failures are swallowed and returned as `"{}"`, so a failure silently becomes an empty envelope. Log the error before returning the fallback.

AmplifyClients/AmplifyEventEnrichmentClient/Sources/Support/ActivityTracker.swift
```diff
@@ -0,0 +1,73 @@
...
+    private let onPause: @Sendable () -> Void
+    private let onResume: @Sendable () -> Void
+
+    private static let backgroundNotification: Notification.Name = {
+        #if canImport(WatchKit)
+        WKExtension.applicationDidEnterBackgroundNotification
+        #elseif canImport(UIKit)
+        UIApplication.didEnterBackgroundNotification
+        #elseif canImport(AppKit)
+        NSApplication.didResignActiveNotification
```
cadivus: On macOS, `didResignActiveNotification` / `willBecomeActiveNotification` fire whenever the app merely loses focus, so window switching churns sessions. `didHide` / `willUnhide` (or occlusion state) map more closely to background/foreground.

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
cadivus: Still open: `getDeviceMetadata()` is never called, so omitting `deviceMetadata` still emits empty `device: {}`. Provider remains dead code.

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
cadivus: Still open: `record()` unconditionally calls `startSession()`, so `autoSessionTracking: false` still auto-starts a session. Not gated/renamed.

AmplifyClients/AmplifyEventEnrichmentClient/Sources/AmplifyEventEnrichmentClient.swift
```diff
@@ -0,0 +1,204 @@
...
+/// ## Usage
+///
+/// ```swift
+/// let client = AmplifyEventEnrichmentClient(
+///     appId: "my-app",
+///     sdkMetadata: SDKMetadata(name: "amplify-swift", version: "2.58.0")
+/// )
+///
+/// let event = try await client.record("button_clicked")
+/// print(event.toJson())
```
cadivus: NIT: Doesn't this need to be:
```suggestion
/// print(try event.toJson())
```
? I think this example wouldn't compile

cadivus (dismissed):
I have one nit, the rest looks good.

cadivus (dismissed):

ekjotmultani:
One cross platform consistency item that still stands after the recent fixes: the client id is persisted under com.amazonaws.amplify.event_enrichment.client_id, but the other platforms settled on a single shared key, com.amplifyframework.device_id, stored in plain UserDefaults. The Connect client will read the same key so that enrichment and Connect report the same stable device id for a given install. As written the two Swift packages would generate different ids. Could we switch the key here so the contract holds on Swift too?

AmplifyClients/AmplifyEventEnrichmentClient/Sources/Support/DeviceMetadataProvider.swift
```diff
@@ -0,0 +1,88 @@
...
+    private func model() -> String {
+        #if canImport(WatchKit)
+        return WKInterfaceDevice.current().model
+        #elseif canImport(UIKit)
+        return UIDevice.current.model
+        #elseif canImport(IOKit)
+        return ioKitValue(forKey: "model") ?? "Mac"
+        #else
+        return "Mac"
+        #endif
```
harsh62, line 104: IOKit is also available on VisionOS iPadOS, so I won't hardcode to Mac, and also the else could be named as kind of unknown..

AmplifyClients/AmplifyEventEnrichmentClient/Sources/Support/DeviceMetadataProvider.swift
```diff
@@ -0,0 +1,88 @@
...
+        return (
+            name: UIDevice.current.systemName,
+            version: UIDevice.current.systemVersion
+        )
+        #else
+        return (
+            name: "macOS",
+            version: ProcessInfo.processInfo.operatingSystemVersionString
+        )
+        #endif
```
harsh62, line 57: Similar to https://github.com/aws-amplify/amplify-swift/pull/4243/changes#r3667828512

harsh62 (dismissed):

ekjotmultani (dismissed):

ekjotmultani (approved):

PR #4251, opened 2026-07-16
fix: prevent command injection in WebAuthn/Push test LocalServers

harsh62:
## Issue

The Node test utilities under `AmplifyPlugins/.../LocalServer/index.mjs` (used only to drive the iOS Simulator during WebAuthn and Push Notification UI tests) built shell command strings by interpolating request-body input and executed them with `child_process.exec`, which spawns `/bin/sh`.

For example:

```js
const cmd = `xcrun simctl bootstatus ${deviceId} -b`
childProcess.exec(cmd, ...)
```

A `deviceId` such as `booted; <command>` posted to the localhost endpoints would therefore run arbitrary shell commands. The push server was additionally affected on `/notifications`, where the notification payload was piped through `echo '<json>' | xcrun simctl push ...` in a shell.

Impact is limited to developer machines that manually start these localhost-only test servers; no shipped SDK code is affected. Fixing it removes the footgun and keeps the test tooling safe.

## Changes

Both test servers (`AuthWebAuthnApp` and `PushNotificationHostApp`):

- Replace `exec()` with `execFile()`, passing arguments as an array so no shell is invoked and inputs are never interpreted as shell syntax.
- `/enroll`: split the `&&`-chained command into two sequential `execFile` calls.
- `/notifications`: write the APNS JSON payload to the child process's **stdin** instead of piping it through `echo | ...`.
- Add `deviceId` validation (UUID or `booted`) as defense-in-depth.

No dependency or `package.json` changes — `execFile` is part of core `node:child_process`.

## Verification

Ran both servers locally against a stubbed `xcrun` that records its argv/stdin:

- Valid `deviceId` (`booted` / UUID): commands run with the expected argument vector.
- `deviceId` containing shell metacharacters: rejected with `400`, no command executed.
- Malicious `title` on `/notifications`: delivered verbatim as JSON via stdin, not executed.

A payload designed to `touch /tmp/pwned` produced no file in any case.

### DIFF
--- a/AmplifyPlugins/Auth/Tests/AuthWebAuthnApp/LocalServer/index.mjs
+++ b/AmplifyPlugins/Auth/Tests/AuthWebAuthnApp/LocalServer/index.mjs
@@ -6,11 +6,21 @@ app.use(express.json())
 
 const bundleId = "com.amazon.aws.amplify.swift.AuthWebAuthnApp"
 
-const run = (cmd) => {
+// Simulator device identifiers are either a UUID (UDID) or the literal "booted".
+// Validating up front rejects any value that could be used to smuggle shell
+// metacharacters into the commands below.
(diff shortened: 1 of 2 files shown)


harsh62:
Closing to reopen from a branch on the main repository instead of a fork.

PR #4254, opened 2026-07-22
fix(auth): key device metadata by inputUsername so alias sign-in pools remember devices

dmorrow:
## Issue #

Related to #4196, which fixed the device-metadata keychain key for the *write* path (`ConfirmDevice`) but left several *read* paths keyed off a different value.

## Description

On a user pool configured with alias sign-in (e.g. email alias), `RespondToAuthChallenge` returns a `USERNAME` challenge parameter equal to the user's `sub`, not the value the caller signed in with. #4196 made `ConfirmDevice` store device metadata under `signedInData.inputUsername` (the caller-supplied username), but the sign-in actions still read that metadata back using the echoed `USERNAME` (the sub). The read key never matched the write key, so:

- `DEVICE_KEY` was omitted from the password-verifier response, and Cognito treated every sign-in as a new device — MFA on every sign-in, and a new device record accumulating per attempt.
- On pools where the username *is* the sub, the two values coincide, which is why this isn't seen everywhere.

This PR aligns every device-metadata read (and the two `deviceNotFound` deletes) with the `inputUsername` used by the write, across the SRP, challenge, and device-SRP flows:

- **`VerifyPasswordSRP`** — read and the `deviceNotFound` delete now use `inputUsername`. (The delete previously used the echoed `username`; when a stored device key was stale, Cognito returned `ResourceNotFoundException`, the recovery path deleted the wrong keychain entry, and the retry re-read the same stale key and looped indefinitely.)
- **`VerifySignInChallenge`** — both reads and the `deviceNotFound` delete now use `challenge.inputUsername ?? username`.
- **`UserPoolSignInHelper`** — carries `inputUsername ?? username` into `.initiateDeviceSRP`, so `InitiateAuthDeviceSRP` looks the metadata up (and re-sends `DEVICE_KEY`) under the right key. Without this the device-SRP flow fails with `Missing required parameter DEVICE_KEY`.
- **`VerifyDevicePasswordSRP`** — reads metadata with `inputUsername`; the outgoing request still sends `username` (the echoed value), which is what Cognito expects.

Nothing sent to Cognito changes semantically — `username` still supplies the request `USERNAME` on every request; only the keychain lookups moved to the value the metadata was written under.

### Verification

Manually verified end-to-end on a real device against an alias-sign-in pool (`us-east-1` email alias, MFA via email OTP):

- Fresh device → MFA challenge → `rememberDevice` → subsequent sign-in takes `DEVICE_SRP_AUTH` and completes with no MFA.
- Stale local device key → a single `deviceNotFound` recovery + retry, then a normal MFA challenge (no infinite loop).
- Device removed / password reset → MFA challenge, then re-remembered on confirm.

## General Checklist

- [x] All unit tests pass


dmorrow:
@thisisabhash This fixes a bug in production.  Would it be best for me to submit a bug report as well?  Our iOS users devices are never saved, leading to MFA every time.

harsh62:
All checks passed on https://github.com/aws-amplify/amplify-swift/pull/4263.

harsh62 (approved):

PR #4268, opened 2026-08-07
chore: raise minimum supported platform, Swift, and Xcode versions

harsh62:
## Issue

Apple has required **Xcode 26** and the 26 SDKs for App Store Connect uploads since **April 28, 2026**. The support policy in `README.md` states that Amplify Swift updates its minimum Swift version within 60 days of Apple's change, so this is overdue.

This also lands ahead of AWS SDK for Swift's own proposed bump (~September 2026), so we move by plan rather than being forced by an upstream dependency.

## Description

Platform floors now match Apple's published Xcode 26 support matrix, and the Swift toolchain floor matches AWS SDK for Swift.

| | Before | After |
|---|---|---|
| iOS | 13 | **15** |
| tvOS | 13 | **15** |
| macOS | 12 | 12 (unchanged) |
| watchOS | 9 | 9 (unchanged) |
| visionOS | undeclared / "Preview" | **1.0, GA** |
| swift-tools-version | 5.9 | **6.0** |
| Minimum Xcode | 16.0 | **26.0** |

Notes on the choices:

- **watchOS stays at 9.** Apple's Xcode 26 floor is watchOS 8, and the AWS SDK proposal moves to 8. We are already above that floor, so lowering it would *widen* support rather than raise it.
- **macOS stays at 12** for the same reason.
- **visionOS was never declared in `Package.swift`.** It was documented as "Preview" and built in CI, but resolved implicitly via iOS compatibility. It is now explicitly declared and marked GA.

### Swift 6 language mode is intentionally NOT in this PR

Raising `swift-tools-version` to 6.0 silently flips the default language mode to Swift 6. To keep this change behavior-neutral, `swiftLanguageModes: [.v5]` pins the existing behavior.

Measured separately: the Swift 6 migration is ~350 genuine strict-concurrency sites (concentrated in Storage, DataStore, and Auth). That lands in a follow-up so it can be reviewed on its own.

### CI

- Runners `macos-15` → `macos-26`
- Latest Xcode `26.3` → `26.5`; minimum `16.0.0` → `26.0.1`
- Removed the Xcode 16 legs and the duplicate hardcoded `16.4.0` pin that drifted independently of the central mapping
- Simulator devices updated to those actually present on `macos-26` (`iPhone 17 Pro Max`, `Apple Watch Series 11`) — the previous `iPhone 16 Pro Max` / `Series 10` devices do not exist on that image
- Minimum-Xcode legs run against the **26.2** runtime: the image ships 26.0 SDKs but installs no 26.0 runtime. The Xcode version is what is being validated; the runtime only needs to be at or above our deployment floors.
- **visionOS added to the minimum-supported-version matrix**, which previously never exercised the platform the docs call supported
- **New advisory nightly** against the `xcode-27` preview image for early warning on the next Xcode. Non-blocking (`continue-on-error`), so it never gates a PR.

### Docs

Corrected the stale **Xcode 13.4** requirement in `CONTRIBUTING.md` (12 major versions out of date).

### Semver

Treated as a **minor** release, not a breaking change, consistent with how AWS SDK for Swift is handling their equivalent bump.

## How did you test these changes?

Locally on Xcode 26.5 (Swift 6.3.2):

- `swift build` — clean, 0 errors
- `swift test` — **0 failures**
- `xcodebuild build -scheme Amplify-Package` verified end-to-end for **iOS**, **macOS**, and **visionOS**
- All four floors verified to compile against the Xcode 26.5 SDKs: `ios15.0`, `tvos15.0`, `watchos9.0`, `xros1.0`
- `swiftformat --lint . --swiftversion 6.0` — clean. Three test files needed reformatting because the 6.0 bump activates the `preferCountWhere` rule; verified they were clean at 5.9 beforehand, so CI would have failed without this.
- `swiftlint` — 0 new violations in changed files (366 pre-existing warnings elsewhere are untouched)
- All 76 workflow/action YAML files parse

tvOS and watchOS full-package builds are left to CI, which has those simulator runtimes; the local machine does not. Compiler-level verification at those floors is included above.

## Documentation

- [x] `README.md` — platform table, Swift version, Xcode requirement
- [x] `CONTRIBUTING.md` — stale Xcode version
- [x] `AGENTS.md` / `CLAUDE.md`
- [x] `README-combine-support.md`

## Checklist

- [x] PR description included
- [x] `yarn lint` passed (swiftformat + swiftlint equivalents run)
- [x] Unit tests added/updated — no new tests; this is a configuration change validated by building and testing every supported platform at the new floors
- [x] Documentation updated


thisisabhash (dismissed):

harsh62:
Superseded by #4271, which contains the same version bump rebuilt cleanly off main without the unrelated CI changes that were tangled in here.

PR #4269, opened 2026-08-11
ci: seed the SwiftPM dependencies cache and raise job timeouts

harsh62:
## Issue

CI jobs are being killed at their timeout with **no build output at all**. Job logs end after a few lines and cleanup terminates orphaned `git` and `git-remote-http` processes — the jobs were still cloning dependencies, not compiling or testing.

Root cause is that **the shared SwiftPM dependencies cache has never populated**. `gh api .../actions/caches` reports `total_count: 0`.

Every unit test, integration test, and build job only ever *restores* `amplify-packages-*`. The single step that writes it lives in `build_scheme.yml` gated on `github.ref_name == 'main'` — but the only workflow that reaches that step with saving enabled is `build_amplify_swift_platforms.yml`, which declares `branches-ignore: [main]` and is otherwise only reachable from the deploy workflows that run on `release`. **The write is unreachable.**

So all ~190 jobs on every run independently clone the full dependency graph (~34 packages: aws-sdk-swift, smithy-swift, aws-crt-swift, swift-nio, …) and starve each other.

### 1. Seed the cache from `main` (new workflow)

`seed_dependencies_cache.yml` resolves dependencies on `main` and saves the cache, so PR jobs restore instead of cloning. It runs only when `Package.resolved`/`Package.swift` change, and its concurrency group prevents two seeding runs racing the same key.

It prunes the bare git mirrors under `repositories/` before saving: only `checkouts/` is needed once working copies exist, and consumers build with `-disableAutomaticPackageResolution` on a cache hit. **That takes the entry from ~4.3 GB to ~830 MB**, which matters against the repository-wide 10 GB cache budget.

### 2. Drop the Xcode version from the cache key

```diff
- key: amplify-packages-${{ steps.platform.outputs.xcode-version }}-${{ hashFiles('Package.resolved') }}
+ key: amplify-packages-${{ hashFiles('Package.resolved') }}
```

These checkouts are plain source and not toolchain-specific — `Package.resolved` already pins them exactly. Keying on Xcode meant **every Xcode bump invalidated the entire cache at once**, including via `restore-keys`. Applied to all five workflows that share the key so they keep sharing it.

### 3. Raise timeouts 30 → 45 minutes

Including the nine callers that hardcoded `30` and so overrode the reusable-workflow default. The cache is only seeded from `main` by design, so PR runs must tolerate a cold start.

### 4. Run the Xcode preview build on PRs

Moves the Xcode 27 preview build from a nightly schedule to `pull_request` + `push: main`, so next-Xcode breakages surface on the PR that introduces them rather than on a run nobody is watching. Every job keeps `continue-on-error` — a beta toolchain on a capacity-constrained preview image is not a suitable merge gate.

## Why this is a separate PR

This was found while working on the minimum-version bump (#4268), but it is an independent pre-existing bug and that PR cannot go green until the cache is seeded from `main`. Splitting it breaks the circular dependency and keeps both diffs reviewable. This PR touches **only** `.github/`.

## How did you test these changes?

- Verified the cache is empty (`total_count: 0`) and that `build_amplify_swift_platforms.yml`'s `branches-ignore: [main]` makes the existing save step unreachable.
- Reproduced the stall locally with an empty clone directory and sampled the stalled process. The stack confirms the mechanism:
  ```
  Workspace.loadPackageGraph → Workspace._resolve
    → Workspace.updateDependenciesCheckouts → Workspace.checkoutRepository
      → GitRepository.checkout(revision:) → AsyncProcess.waitUntilExit()
  ```
  `xcodebuild` is blocked in SwiftPM waiting on `git` subprocesses checking out working copies — not simulator boot, not compilation. SwiftPM emits no progress during this phase, which is why the jobs look hung.
- Confirmed the local run completes and passes when given enough time (`** TEST SUCCEEDED **`), with the tests themselves taking 3.4 seconds.
- Verified `xcodebuild -resolvePackageDependencies` populates the cache directory with 31 checkouts, that resolution still succeeds after pruning `repositories/`, and that a real consumer build (`AWSPredictionsPlugin` for macOS) succeeds against the pruned directory with `-disableAutomaticPackageResolution`.
- All 77 workflow/action YAML files parse.

The seeding workflow cannot run until it is on `main` — that is the point of this PR. The first `main` run after merge will populate the cache; subsequent PRs get a warm start.

## Documentation

No customer-facing changes — CI configuration only.

## Checklist

- [x] PR description included
- [x] Unit tests added/updated — not applicable; CI configuration change, validated by YAML parsing and by locally reproducing and diagnosing the stall
- [x] Documentation updated — inline comments explain why the key omits the Xcode version and why `repositories/` is pruned

### DIFF
--- a/.github/composite_actions/run_xcodebuild/action.yml
+++ b/.github/composite_actions/run_xcodebuild/action.yml
@@ -39,6 +39,46 @@ inputs:
 runs:
   using: "composite"
   steps:
+    # See run_xcodebuild_test/action.yml for the rationale: resolution runs as an
+    # explicitly bounded, retried step so it cannot hang silently inside the build
+    # and consume the job's entire timeout with no output.
+    - name: Resolve package dependencies for ${{ inputs.scheme }}
+      timeout-minutes: 20
+      shell: bash
+      env:
+        PROJECT_PATH: ${{ inputs.project_path }}
+        XCODE_PATH: ${{ inputs.xcode_path }}
+        CLONED_SOURCE_PACKAGES_PATH: ${{ inputs.cloned_source_packages_path }}
+      run: |
+        set -o pipefail
+        if [ -n "$PROJECT_PATH" ]; then cd "$PROJECT_PATH"; fi
+        if [ -n "$XCODE_PATH" ]; then sudo xcode-select -s "$XCODE_PATH"; fi
+
+        clonedPath=""
+        if [ -n "$CLONED_SOURCE_PACKAGES_PATH" ]; then
+          clonedPath="-clonedSourcePackagesDirPath $CLONED_SOURCE_PACKAGES_PATH"
+        fi
+
+        for attempt in 1 2; do
+          echo "::group::Resolving package dependencies (attempt $attempt)"
+          if xcodebuild -resolvePackageDependencies \
+               -scheme '${{ inputs.scheme }}' \
+               $clonedPath; then
+            echo "::endgroup::"
+            echo "Dependencies resolved on attempt $attempt."
+            exit 0
+          fi
+          echo "::endgroup::"
+          echo "Resolution attempt $attempt failed."
+          # Drop partial mirror state so the retry starts clean.
+          if [ -n "$CLONED_SOURCE_PACKAGES_PATH" ] && [ "$attempt" -eq 1 ]; then
+            rm -rf "${CLONED_SOURCE_PACKAGES_PATH%/}/repositories" || true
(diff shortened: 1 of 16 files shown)


harsh62:
Superseded. The cache-seeding approach is being dropped: seeding only works once the workflow is on the default branch, and GitHub scopes branch caches so it could never help sibling PRs. Replacing it with a branch that makes dependency resolution resilient whether or not a cache exists.

PR #4274, opened 2026-08-18
fix(api): percent-encode semicolons in REST query parameter values

### LINKED ISSUE #3955: Using a semicolon in a value in QueryParameters suddenly doesn't work anymore
### Describe the bug

I'm calling a REST API with some query parameters ([String : String]).  The API call worked for months now.  One value in the query is a string like "X;Y".
The server now receives only X and not the entire string. 
I put this query dict in a RESTRequest that is simply passed to Amplify.API.get.

Any ideas what happened ? It might coincide with the update to 2.45.4 I did a few days ago.

PS: This suddenly became a huge issue, I realized I have values like this in 99% of the API calls.  Having a ";" in any value in the query parameters, simply removes everything what's after the 1st semicolon.
I tried encoding it (%3B), not encoding it, nothing works.  Amplify just kills it.

### Steps To Reproduce

```swift
- Access any REST API call with Amplify that requires queryParameters and send a value with a semicolon somewhere in the middle.
```

### Expected behavior

The server should receive the entire string, not only what's before the semicolon.

### Amplify Framework Version

2.45.4

### Amplify Categories

API

### Dependency manager

Swift PM

VarshithaPamisetty:
## Description

`URLComponents` leaves `;` unencoded since it's a legal query sub-delimiter, but servers like API Gateway treat it as a separator, so a REST query value `"X;Y"` was truncated to `"X"`. This encodes `;` to `%3B` after the query is built so the value arrives intact.

## Testing

- Regression tests: `;` (raw and pre-encoded) percent-encoded in the URL and through the plugin request path. Both fail without the change.
- AWSAPIPlugin unit suite 227/227; `swiftformat`/`swiftlint` clean; builds on macOS and iOS.
- Couldn't run the REST integration tests locally since they need a provisioned AWS backend I don't have access to.

Fixes #3955

### DIFF
--- a/AmplifyPlugins/API/Sources/AWSAPIPlugin/Support/Utils/RESTOperationRequestUtils.swift
+++ b/AmplifyPlugins/API/Sources/AWSAPIPlugin/Support/Utils/RESTOperationRequestUtils.swift
@@ -35,8 +35,9 @@ final class RESTOperationRequestUtils {
 
         if let queryParameters {
             components.queryItems = prepareQueryParamsForSigning(params: queryParameters)
-            // `URLComponents` leaves `;` unencoded, so servers may treat it as a separator.
-            // Encode it — https://github.com/aws-amplify/amplify-swift/issues/3955
+            // `URLComponents` leaves `;` unencoded, but API Gateway treats it as a query
+            // parameter delimiter and drops everything after it, so encode it explicitly.
+            // https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html#api-gateway-known-issues-rest-apis
             if let percentEncodedQuery = components.percentEncodedQuery {
                 components.percentEncodedQuery = percentEncodedQuery.replacingOccurrences(of: ";", with: "%3B")
             }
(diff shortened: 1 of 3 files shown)


AmplifyPlugins/API/Sources/AWSAPIPlugin/Support/Utils/RESTOperationRequestUtils.swift
```diff
@@ -35,6 +35,11 @@ final class RESTOperationRequestUtils {
 
         if let queryParameters {
             components.queryItems = prepareQueryParamsForSigning(params: queryParameters)
+            // `URLComponents` leaves `;` unencoded, so servers may treat it as a separator.
```
thisisabhash: can we add a reference to the API Gateway known issue in the comment for this change as `;` is a valid delimiter for query parameters.
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html#api-gateway-known-issues-rest-apis

thisisabhash: A similar fix on Android is needed to encode a raw `;` in the query parameters.

VarshithaPamisetty: Added the API Gateway known-issue reference to the comment and yes I'll track the same `;`-encoding fix for Android as a follow-up.

thisisabhash (approved):

PR #4281, opened 2026-08-31
fix(auth): Refresh rejected identity ID

### LINKED ISSUE #4278: Cached identity pool id is never evicted after "Access to Identity is forbidden" (NotAuthorizedException from GetCredentialsForIdentity) — session wedges permanently
### Describe the bug

When `GetCredentialsForIdentity` rejects the cached identity pool id with `NotAuthorizedException` ("Access to Identity 'us-east-1:…' is forbidden."), the plugin never evicts the cached identity id. The authorization state machine lands in `.error(.sessionError(.notAuthorized, credentials))` **with the poisoned credentials retained inside the error state**, and every subsequent `fetchAuthSession` re-presents the exact same cached identity id:

- `FetchAuthSessionOperationHelper.fetch` handles `case .error(.sessionError(_, let credentials))` by calling `refreshIfRequired(existingCredentials: credentials, …)` — same blob, same identity id.
- `InitializeRefreshSession` always pairs the *stored* `identityID` with the stored/refreshed tokens (`.refreshCognitoUserPoolWithIdentityId`, `.refreshAWSCredentialsWithUserPool`); there is no path that falls back to a fresh `GetId` after the identity was rejected.

The error is terminal server-side (the identity is linked to a different login, disabled, or deleted), so the device retries forever and the session never recovers. For a signed-in user this surfaces as `getUserSub()` / `getCognitoTokens()` returning the forbidden error on every launch while `isSignedIn` stays `true` — a permanently wedged boot. The only recovery is app-side: call `Amplify.Auth.signOut()` (which deletes the whole credential blob) and force re-authentication.

This is the same defect class as aws-amplify/amplify-js#5696, which was fixed there by evicting the cached identity id when this error is received.

We verified this against production telemetry: affected devices (predominantly account switchers whose local state ended up holding an identity id from a previous account) hit the error on every boot, on multiple days, and never self-recover; devices on an app version that reacts to this error with an explicit `signOut()` recover immediately.

### Steps To Reproduce

1. Configure Auth with user pool + identity pool.
2. Sign in; let the plugin store `userPoolAndIdentityPool(signedInData, identityID, credentials)`.
3. Arrange for the stored identity id to be rejected by Cognito — e.g. delete the identity in the identity pool console, or reproduce a state where the stored identity id belongs to a different login than the stored tokens.
4. Call `Amplify.Auth.fetchAuthSession()` (or relaunch) repeatedly.

Observed: every call performs `GetCredentialsForIdentity` with the same rejected identity id and fails with `NotAuthorizedException` ("Access to Identity … is forbidden"). The cached identity id is never discarded; no `GetId` with the current logins map is ever attempted.

### Expected behavior

On `NotAuthorizedException` from `GetCredentialsForIdentity`, the plugin should evict the cached identity id and retry with a fresh `GetId` using the current logins map (mirroring the amplify-js fix in aws-amplify/amplify-js#5696). At minimum for the signed-in case, the terminal error should not permanently retain the rejected identity id in the credential store.

### Amplify Framework Version

2.53.3

### Amplify Categories

Auth

### Dependency manager

Swift PM

### Xcode version

26

### Relevant log output

```shell
AuthError: notAuthorized
"Access to Identity 'us-east-1:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx' is forbidden."
(identical on every fetchAuthSession / launch; never recovers without an explicit Amplify.Auth.signOut())
```

### Is this a regression?

No — reproduced on 2.53.3; the relevant code paths (`FetchAuthSessionOperationHelper`, `InitializeRefreshSession`, `AuthorizationState.Resolver` `.error` handling) have behaved this way for a long time.

cadivus:
## Issue #
Fixes #4278

## Description
Refresh rejected cached identity IDs with a new `GetId` request, then retry credential fetching once. This prevents signed-in sessions from remaining wedged after `NotAuthorizedException` or `ResourceNotFoundException`.

## Testing
- `FetchAuthAWSCredentialsTests`: 8 passed
- iOS Simulator: recovered with a new identity after deleting the cached identity

### DIFF
--- a/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/FetchAuthorizationSession/AWSCredentials/FetchAuthAWSCredentials.swift
+++ b/AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/FetchAuthorizationSession/AWSCredentials/FetchAuthAWSCredentials.swift
@@ -32,6 +32,24 @@ struct FetchAuthAWSCredentials: Action {
             return
         }
 
+        await fetchCredentials(
+            withIdentityID: identityID,
+            canRetryWithFreshIdentityID: true,
+            client: client,
+            authZEnvironment: authZEnvironment,
+            dispatcher: dispatcher,
+            environment: environment
+        )
+    }
+
+    private func fetchCredentials(
+        withIdentityID identityID: String,
+        canRetryWithFreshIdentityID: Bool,
+        client: CognitoIdentityBehavior,
+        authZEnvironment: AuthorizationEnvironment,
+        dispatcher: EventDispatcher,
+        environment: Environment
+    ) async {
         let getCredentialsInput = GetCredentialsForIdentityInput(
             identityId: identityID,
             logins: loginsMap
@@ -68,10 +86,58 @@ struct FetchAuthAWSCredentials: Action {
             await dispatcher.send(event)
 
         } catch {
+            // Rejected identity ID: evict and retry via fresh GetId.
+            if canRetryWithFreshIdentityID, isStaleIdentityError(error) {
+                await retryWithFreshIdentityID(
+                    client: client,
+                    authZEnvironment: authZEnvironment,
+                    dispatcher: dispatcher,
+                    environment: environment
+                )
+                return
+            }
             let event = FetchAuthSessionEvent(eventType: .throwError(.service(error)))
             logVerbose("\(#fileID) Sending event \(event.type)", environment: environment)
             await dispatcher.send(event)
         }
+    }
+
+    private func retryWithFreshIdentityID(
+        client: CognitoIdentityBehavior,
+        authZEnvironment: AuthorizationEnvironment,
+        dispatcher: EventDispatcher,
+        environment: Environment
+    ) async {
+        let getIdInput = GetIdInput(
+            identityPoolId: authZEnvironment.identityPoolConfiguration.poolId,
+            logins: loginsMap
+        )
+        do {
+            let response = try await client.getId(input: getIdInput)
+            guard let freshIdentityID = response.identityId else {
+                let event = FetchAuthSessionEvent(eventType: .throwError(.invalidIdentityID))
+                await dispatcher.send(event)
+                return
+            }
+            await fetchCredentials(
+                withIdentityID: freshIdentityID,
+                canRetryWithFreshIdentityID: false,
+                client: client,
+                authZEnvironment: authZEnvironment,
+                dispatcher: dispatcher,
+                environment: environment
+            )
+        } catch {
+            let event = FetchAuthSessionEvent(eventType: .throwError(.service(error)))
+            logVerbose("\(#fileID) Sending event \(event.type)", environment: environment)
+            await dispatcher.send(event)
+        }
+    }
+
+    // Codes returned when the identity ID is stale.
+    private func isStaleIdentityError(_ error: Error) -> Bool {
+        error is AWSCognitoIdentity.NotAuthorizedException ||
+        error is AWSCognitoIdentity.ResourceNotFoundException
     }
 }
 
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/FetchAuthorizationSession/AWSCredentials/FetchAuthAWSCredentials.swift
```diff
@@ -67,12 +85,60 @@ struct FetchAuthAWSCredentials: Action {
             logVerbose("\(#fileID) Sending event \(event.type)", environment: environment)
             await dispatcher.send(event)
 
+        } catch {
```
thisisabhash, line 88: `NotAuthorizedException` here is ambiguous for authenticated sessions. When `loginsMap` is
non-empty, `GetCredentialsForIdentity` fails with `NotAuthorized` most often because the
login token is expired/invalid, not because the identity ID is stale. Re-running `GetId`
with the same invalid logins can't recover that — it just adds two network calls before
surfacing the same error. Recommend gating the `NotAuthorized` retry to guest flows (empty
`loginsMap`), and letting authenticated `NotAuthorized` surface immediately so the app
re-authenticates. `ResourceNotFoundException` recovery is correct in both cases.

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/ActionTests/FetchAuthSession/FetchAWSCredentials/FetchAuthAWSCredentialsTests.swift
```diff
@@ -193,4 +193,136 @@ class FetchAuthAWSCredentialsTests: XCTestCase {
...
+                }
+            },
+            environment: authEnvironment
+        )
+
+        await fulfillment(of: [errored], timeout: 0.1)
+        XCTAssertEqual(getIdCalled.count, 1)
+        XCTAssertEqual(credentialsCallCount.count, 2)
+    }
+
```
thisisabhash, line 315: Nit: stray blank line

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/FetchAuthorizationSession/AWSCredentials/FetchAuthAWSCredentials.swift
```diff
@@ -67,12 +85,60 @@ struct FetchAuthAWSCredentials: Action {
...
+            // Rejected identity ID: evict and retry via fresh GetId.
```
thisisabhash: The stale ID is only replaced on the success path via the `.fetchedAWSCredentials` payload.
If the retry fails, the stale identity stays cached and the next fetch repeats the failure.
Suggest rewording the comment to match the code.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/FetchAuthorizationSession/AWSCredentials/FetchAuthAWSCredentials.swift
cadivus: Fixed: NotAuthorized now retries only for the identity-forbidden message, preserving signed-in recovery while invalid or expired tokens surface immediately.

AmplifyPlugins/Auth/Tests/AWSCognitoAuthPluginUnitTests/ActionTests/FetchAuthSession/FetchAWSCredentials/FetchAuthAWSCredentialsTests.swift
cadivus: Removed.

AmplifyPlugins/Auth/Sources/AWSCognitoAuthPlugin/Actions/FetchAuthorizationSession/AWSCredentials/FetchAuthAWSCredentials.swift
cadivus: Updated the comment.

harsh62 (approved):

PR #4292, opened 2026-09-03
fix(core): report an unconfigured Auth category instead of aborting

harsh62:
Slice **4 of 38** in the Swift 6 language-mode migration — 3 file(s).

Stacked on `swift6/03-async-sequence-cancellation`. Reference (do not merge): #4283.

### DIFF
--- a/Amplify/Categories/Auth/AuthCategory.swift
+++ b/Amplify/Categories/Auth/AuthCategory.swift
@@ -5,7 +5,11 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-public final class AuthCategory: Category {
+/// - Note: `@unchecked Sendable` to satisfy the `Sendable` requirement that the category behavior
+///   protocol now carries. `plugins` is populated during `Amplify.configure()` and only read
+///   afterwards, and the conformance must be declared here because the behavior conformance lives
+///   in an extension in another file.
+public final class AuthCategory: Category, @unchecked Sendable {
 
     public let categoryType =  CategoryType.auth
 
@@ -41,6 +45,17 @@ public final class AuthCategory: Category {
     }
 
     var isConfigured = false
+
+    /// `true` when `Amplify.configure()` has run for this category and a plugin is registered.
+    ///
+    /// Reading ``plugin`` in either of the opposite states trips a `preconditionFailure`, which aborts
+    /// the process rather than failing the call. Exposed over `@_spi` so the AWS plugin modules can
+    /// report an error instead: they hold long-lived clients that can outlive `Amplify.reset()`, and for
+    /// those a missing Auth category is a recoverable condition, not a programmer error.
+    @_spi(InternalAmplifyConfiguration)
+    public var isConfiguredWithPlugin: Bool {
+        isConfigured && !plugins.isEmpty
+    }
 
     // MARK: - Plugin handling
 
--- a/Amplify/Categories/Auth/AuthCategory.swift
+++ b/Amplify/Categories/Auth/AuthCategory.swift
@@ -6,9 +6,15 @@
 //
 
 /// - Note: `@unchecked Sendable` to satisfy the `Sendable` requirement that the category behavior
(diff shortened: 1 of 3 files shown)


Amplify/Categories/Auth/AuthCategory.swift
```diff
@@ -5,7 +5,11 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-public final class AuthCategory: Category {
+/// - Note: `@unchecked Sendable` to satisfy the `Sendable` requirement that the category behavior
+///   protocol now carries. `plugins` is populated during `Amplify.configure()` and only read
```
thisisabhash: `@unchecked Sendable` on `AuthCategory` is questionable-but-defensible. `plugins`, `isConfigured`, and (in slice 05-adjacent code) the config lifecycle mutate the dictionary in `add(plugin:)`/`removeValue(forKey:)` with no lock. The Sendable-conformance doc comment leans on the "populated during configure(), read afterwards" contract; that contract is not enforced anywhere and `removePlugin(for:)`/`reset()` do mutate `plugins` later. If any category currently races `configure()` and `plugin` access, this annotation now hides that race. Non-blocking because the pre-existing code already assumed this contract, but flag it and either add locking to `plugins`/`isConfigured` in a follow-up or document the contract more forcefully.

Amplify/Categories/Auth/AuthCategory.swift
```diff
@@ -42,6 +46,17 @@ public final class AuthCategory: Category {
...
     var isConfigured = false
 
+    /// `true` when `Amplify.configure()` has run for this category and a plugin is registered.
+    ///
+    /// Reading ``plugin`` in either of the opposite states trips a `preconditionFailure`, which aborts
+    /// the process rather than failing the call. Exposed over `@_spi` so the AWS plugin modules can
+    /// report an error instead: they hold long-lived clients that can outlive `Amplify.reset()`, and for
+    /// those a missing Auth category is a recoverable condition, not a programmer error.
+    @_spi(InternalAmplifyConfiguration)
+    public var isConfiguredWithPlugin: Bool {
```
thisisabhash: `isConfiguredWithPlugin` reads `isConfigured` and `plugins` without synchronization; if a concurrent `configure()` is running, the read is a data race. In practice `configure()` runs once early, and after it returns the state is stable, so this is fine given the surrounding contract — same caveat as the `@unchecked Sendable` note.

AmplifyPlugins/Core/AmplifyCredentials/AmplifyAWSCredentialsProvider.swift
```diff
@@ -14,8 +14,29 @@ import SmithyIdentity
...
 
+    /// Fetches the auth session, reporting an unconfigured Auth category as a thrown error.
+    ///
+    /// Reading `Amplify.Auth` with no plugin registered hits a `preconditionFailure`, which aborts the
+    /// process rather than failing the request. A credentials provider is reachable from long-lived
+    /// clients that can outlive `Amplify.reset()` — the cached `PinpointContext` in `AWSPinpointFactory`
+    /// is one, since nothing ever clears it — so this is a state the provider can genuinely be called in,
+    /// and it should surface as an error the caller can handle.
+    private static func fetchAuthSession() async throws -> AuthSession {
+        guard Amplify.Auth.isConfiguredWithPlugin else {
```
thisisabhash: The gate is TOCTOU with respect to `Amplify.reset()`: if reset runs between `isConfiguredWithPlugin` returning `true` and `fetchAuthSession()` reading `Amplify.Auth.plugin`, you still hit the `preconditionFailure`. That is exactly the scenario the doc-comment names as motivation, so the guard reduces but doesn't eliminate the crash. Recommend either (a) documenting this as best-effort in the credential provider, or (b) having `Auth.plugin` itself throw rather than trap when accessed post-reset. Non-blocking for this slice, but worth acknowledging in the description.

thisisabhash (approved):

PR #4293, opened 2026-09-03
fix(analytics): keep ActivityTracker registration synchronous

harsh62:
Slice **5 of 38** in the Swift 6 language-mode migration — 2 file(s).

Stacked on `swift6/04-credentials-unconfigured-auth`. Reference (do not merge): #4283.

### DIFF
--- a/AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Session/ActivityTracking/ActivityTracker.swift
+++ b/AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Session/ActivityTracking/ActivityTracker.swift
@@ -61,12 +61,15 @@ extension ApplicationState: DefaultLogger {
     }
 }
 
-protocol ActivityTrackerBehaviour: AnyObject {
+/// - Note: `Sendable` because the session client holding this is `Sendable`.
+protocol ActivityTrackerBehaviour: AnyObject, Sendable {
     var backgroundTrackingTimeout: TimeInterval { get set }
     func beginActivityTracking(_ listener: @escaping (ApplicationState) -> Void)
 }
 
-class ActivityTracker: ActivityTrackerBehaviour {
+/// - Note: `final` and `@unchecked Sendable`: the background-tracking state is only touched from the
+///   main actor, and the state machine it forwards to is internally synchronized.
+final class ActivityTracker: ActivityTrackerBehaviour, @unchecked Sendable {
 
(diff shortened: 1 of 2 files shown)


AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Session/ActivityTracking/ActivityTracker.swift
```diff
@@ -61,12 +61,15 @@ extension ApplicationState: DefaultLogger {
...
+/// - Note: `Sendable` because the session client holding this is `Sendable`.
+protocol ActivityTrackerBehaviour: AnyObject, Sendable {
     var backgroundTrackingTimeout: TimeInterval { get set }
     func beginActivityTracking(_ listener: @escaping (ApplicationState) -> Void)
 }
 
-class ActivityTracker: ActivityTrackerBehaviour {
+/// - Note: `final` and `@unchecked Sendable`: the background-tracking state is only touched from the
+///   main actor, and the state machine it forwards to is internally synchronized.
+final class ActivityTracker: ActivityTrackerBehaviour, @unchecked Sendable {
```
thisisabhash, line 84: `final class ActivityTracker … @unchecked Sendable`: the sendability assertion hides real mutable state — `backgroundTrackingTimeout` (settable non-isolated `var`), `backgroundTask`, `backgroundTimer`, and `stateMachineSubscriberToken` — none of which are protected by a lock. The doc comment says the state is "only touched from the main actor", but neither `backgroundTrackingTimeout`'s setter nor `beginActivityTracking(_:)`/`init` are actually main-actor-isolated. `SessionClient.swift:87-88` calls both of these off-main-actor (from `SessionClient`, which is not `@MainActor`), while `beginBackgroundTracking`/`stopBackgroundTracking`/`handleApplicationStateChange` are `@MainActor`. So writes to `backgroundTask`/`backgroundTimer` from the main-actor methods can race with reads/writes of `backgroundTrackingTimeout`/`stateMachineSubscriberToken` from wherever the owner runs. The `@unchecked Sendable` claim in the doc-comment ("only touched from the main actor") does not match the actual code.
  - Recommend either: (a) mark `ActivityTracker` `@MainActor` and have `SessionClient` hop when configuring it, or (b) actually serialize the mutable fields with a lock. As-is, this annotation hides a plausible data race.

thisisabhash (approved):

PR #4294, opened 2026-09-03
fix(api): stop AppSyncRealTimeClient's sink from inheriting actor isolation

harsh62:
Slice **6 of 38** in the Swift 6 language-mode migration — 1 file(s).

Stacked on `swift6/05-activity-tracker-registration`. Reference (do not merge): #4283.

### DIFF
--- a/AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
+++ b/AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
@@ -50,7 +50,13 @@ actor AppSyncRealTimeClient: AppSyncRealTimeClientProtocol {
     /// and connection-drop paths.
     private var isReconnecting = false
     /// Writable data stream convert WebSocketEvent to AppSyncRealTimeResponse
-    nonisolated let subject = PassthroughSubject<Result<AppSyncRealTimeResponse, Error>, Never>()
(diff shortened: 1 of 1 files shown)


AmplifyPlugins/API/Sources/AWSAPIPlugin/AppSyncRealTimeClient/AppSyncRealTimeClient.swift
```diff
@@ -250,20 +257,24 @@ actor AppSyncRealTimeClient: AppSyncRealTimeClientProtocol {
...
-            self?.log.debug("[AppSyncRealTimeClient] WebSocketClient terminated")
-        } receiveValue: { webSocketEvent in
-            Task { [weak self] in
-                let task = Task { [weak self] in
-                    await self?.onWebSocketEvent(webSocketEvent)
-                }
-                await self?.storeInCancellables(task.toAnyCancellable)
-            }
+        let cancellable = await webSocketClient.publisher.sink { @Sendable [weak self] _ in
+            Task { [weak self] in await self?.logWebSocketClientTerminated() }
```
thisisabhash: the completion handler previously called `self?.log.debug(...)` synchronously (`log` is `nonisolated`, line 539, so no hop was needed). Routing it through the actor-isolated `logWebSocketClientTerminated()` inside a `Task` makes the terminated log fire asynchronously with an actor hop. Purely a debug log, harmless, but it is a small behavior change and the hop is avoidable since `log` is already `nonisolated`.

thisisabhash (approved):

PR #4295, opened 2026-09-03
fix(core): make DevMenu concurrency-safe

harsh62:
Slice **7 of 38** in the Swift 6 language-mode migration — 3 file(s).

Stacked on `swift6/06-appsync-sink-isolation`. Reference (do not merge): #4283.

### DIFF
--- a/Amplify/DevMenu/Trigger/LongPressGestureRecognizer.swift
+++ b/Amplify/DevMenu/Trigger/LongPressGestureRecognizer.swift
@@ -10,7 +10,13 @@ import Foundation
 import UIKit
 
 /// A class for recognizing long press gesture which notifies a `TriggerDelegate` of the event
-class LongPressGestureRecognizer: NSObject, TriggerRecognizer, UIGestureRecognizerDelegate {
+///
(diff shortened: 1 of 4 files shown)


Amplify/DevMenu/Trigger/LongPressGestureRecognizer.swift
```diff
@@ -10,7 +10,13 @@ import Foundation
 import UIKit
 
 /// A class for recognizing long press gesture which notifies a `TriggerDelegate` of the event
-class LongPressGestureRecognizer: NSObject, TriggerRecognizer, UIGestureRecognizerDelegate {
+///
+/// - Note: `@preconcurrency` on the `TriggerRecognizer` conformance: this type is main-actor-isolated
```
thisisabhash: the doc comment states the type "is main-actor-isolated," but the class carries no `@MainActor` annotation. The `@preconcurrency` conformance is still correct/harmless; the comment just overstates the actual isolation.

thisisabhash (approved):

PR #4303, opened 2026-09-03
feat(datastore)!: @Sendable DataStore callbacks

harsh62:
Slice **15 of 38** in the Swift 6 language-mode migration — 7 file(s).

Stacked on `swift6/14-lazy-loading-sendable`. Reference (do not merge): #4283.

### DIFF
--- a/Amplify/Categories/DataStore/Query/QueryPredicate.swift
+++ b/Amplify/Categories/DataStore/Query/QueryPredicate.swift
@@ -8,7 +8,8 @@
 import Foundation
 
 /// Protocol that indicates concrete types conforming to it can be used a predicate member.
-public protocol QueryPredicate: Evaluable, Encodable {}
+/// - Note: `Sendable` because predicates are carried into the storage engine's async query paths.
+public protocol QueryPredicate: Evaluable, Encodable, Sendable {}
 
 public enum QueryPredicateGroupType: String, Encodable {
     case and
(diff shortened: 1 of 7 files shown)


Amplify/Categories/DataStore/Query/QueryPredicate.swift
```diff
@@ -33,7 +34,10 @@ public enum QueryPredicateConstant: QueryPredicate, Encodable {
     }
 }
 
-public class QueryPredicateGroup: QueryPredicate, Encodable {
+/// - Note: `@unchecked Sendable` rather than `final`: `type` and `predicates` are mutated while a
```
thisisabhash: `QueryPredicateGroup: @unchecked Sendable` while `type` and `predicates` remain `internal(set) var` and `and(_:)` / `or(_:)` do `predicates.append(...)` on `self` (lines 48-63). The comment justifies this by "mutated while being composed" — but nothing structurally prevents a group from being handed to two tasks (e.g. cached on a request that is retried on multiple queues) and mutated concurrently. The compiler is being asked to trust a convention that isn't documented at the call sites. If we don't want to introduce a lock here, the safer shape is to make `type`/`predicates` `let` and have `and`/`or` return a new group; keeping them mutable and unchecked is exactly the pattern the review brief flags as unsound.

Amplify/Categories/DataStore/Query/QuerySortInput.swift
```diff
@@ -8,12 +8,13 @@
 import Foundation
 
 /// A simple enum that holding a sorting direction for a field that can be applied to queries.
-public enum QuerySortBy {
+public enum QuerySortBy: Sendable {
```
thisisabhash, line 11: QuerySortBy: Sendable over non-Sendable any CodingKey will break the v6 build at slice 38.

PR #4305, opened 2026-09-03
feat(storage)!: @Sendable progress listeners

harsh62:
Slice **17 of 38** in the Swift 6 language-mode migration — 5 file(s).

Stacked on `swift6/16-hub-sendable`. Reference (do not merge): #4283.

### DIFF
--- a/Amplify/Categories/Storage/StoragePath.swift
+++ b/Amplify/Categories/Storage/StoragePath.swift
@@ -7,12 +7,14 @@
 
 import Foundation
 
-public typealias IdentityIDPathResolver = (String) -> String
+/// - Note: `@Sendable` because resolvers are invoked from the storage plugins async work.
+public typealias IdentityIDPathResolver = @Sendable (String) -> String
 
 /// Protocol that provides a closure to resolve the storage path.
 ///
 /// - Tag: StoragePath
-public protocol StoragePath {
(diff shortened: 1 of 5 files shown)


Amplify/Categories/Storage/StoragePath.swift
```diff
@@ -7,12 +7,14 @@
...
-public typealias IdentityIDPathResolver = (String) -> String
+/// - Note: `@Sendable` because resolvers are invoked from the storage plugins async work.
+public typealias IdentityIDPathResolver = @Sendable (String) -> String
 
 /// Protocol that provides a closure to resolve the storage path.
 ///
 /// - Tag: StoragePath
-public protocol StoragePath {
+/// - Note: `Sendable` because paths are stored on storage requests, which cross task boundaries.
+public protocol StoragePath: Sendable {
```
thisisabhash, line 22: StoragePath` is now declared `Sendable`, but the protocol requirement `var resolve: (Input) -> String { get }` is still a plain (non-`@Sendable`) closure. `StringStoragePath.resolve` is tightened to `@Sendable` (line 37), but external conformers can satisfy the requirement with a non-`Sendable` closure and still claim `Sendable`, defeating the guarantee. Tightening the protocol requirement to `@Sendable (Input) -> String` would enforce it at the type system layer.

thisisabhash (approved):

PR #4307, opened 2026-09-03
feat(core): Sendable for the remaining category models

harsh62:
Slice **19 of 38** in the Swift 6 language-mode migration — 21 file(s).

Stacked on `swift6/18-auth-category-sendable`. Reference (do not merge): #4283.

### DIFF
--- a/Amplify/Categories/API/APICategory.swift
+++ b/Amplify/Categories/API/APICategory.swift
@@ -6,7 +6,11 @@
 //
 
 /// The API category provides a solution for making HTTP requests to REST and GraphQL endpoints.
-public final class APICategory: Category {
+/// - Note: `@unchecked Sendable` to satisfy the `Sendable` requirement that the category behavior
+///   protocol now carries. `plugins` is populated during `Amplify.configure()` and only read
+///   afterwards, and the conformance must be declared here because the behavior conformance lives
+///   in an extension in another file.
+public final class APICategory: Category, @unchecked Sendable {
     /// The category type for API
     public var categoryType: CategoryType {
         .api
--- a/Amplify/Categories/API/APICategory.swift
+++ b/Amplify/Categories/API/APICategory.swift
@@ -7,9 +7,13 @@
 
 /// The API category provides a solution for making HTTP requests to REST and GraphQL endpoints.
 /// - Note: `@unchecked Sendable` to satisfy the `Sendable` requirement that the category behavior
-///   protocol now carries. `plugins` is populated during `Amplify.configure()` and only read
-///   afterwards, and the conformance must be declared here because the behavior conformance lives
-///   in an extension in another file.
+///   protocol now carries.
+///
+///   Unchecked in the literal sense: `plugins` and `isConfigured` are plain mutable state with no lock,
+///   and `isConfigured` is `public`, so callers can flip it. `add(plugin:)` cannot race a configured
+///   category — it throws once `isConfigured` is set — but `removePlugin(for:)` has no such guard, so it
+///   can race a concurrent read. The exposure predates this annotation, which only stops the compiler
+///   from asking about it.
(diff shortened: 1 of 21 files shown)


Amplify/Categories/API/APICategory.swift
```diff
@@ -6,7 +6,11 @@
 //
 
 /// The API category provides a solution for making HTTP requests to REST and GraphQL endpoints.
-public final class APICategory: Category {
+/// - Note: `@unchecked Sendable` to satisfy the `Sendable` requirement that the category behavior
+///   protocol now carries. `plugins` is populated during `Amplify.configure()` and only read
+///   afterwards, and the conformance must be declared here because the behavior conformance lives
+///   in an extension in another file.
+public final class APICategory: Category, @unchecked Sendable {
```
thisisabhash, line 17: `APICategory: @unchecked Sendable` while `plugins = [PluginKey: APICategoryPlugin]()` (line 19) and `isConfigured` (line 53) remain plain mutable `var`s. The comment says "populated during `Amplify.configure()` and only read afterwards", but `add(plugin:)` / `removePlugin(for:)` (lines 80, 104) still mutate `plugins` and there's no lock.

Amplify/Categories/Logging/LoggingCategory.swift
```diff
@@ -8,7 +8,10 @@
 import Foundation
 
 /// AWS Amplify writes console logs through Logger. You can use Logger in your apps for the same purpose.
-public final class LoggingCategory: Category {
+/// - Note: `@unchecked Sendable` because `_logLevel` and the configuration state are
+///   mutable but every access is serialized by `lock`. The `Logger` conformance lives in
+///   an extension in another file, so the conformance must be declared here.
+public final class LoggingCategory: Category, @unchecked Sendable {
```
thisisabhash, line 20: `LoggingCategory: @unchecked Sendable`, and only `_logLevel` is guarded by `lock` (lines 47-58). `configurationState` (line 60) and `plugins` (line 68) are still plain `var`s mutated after init (`add(plugin:)` at line 96, `removePlugin` at line 119). Given this class advertises `@unchecked Sendable`, `configurationState`/`plugins` really should be inside the same lock — otherwise the annotation is a lie about the exact fields the review brief targets.

Amplify/Categories/API/Operation/NondeterminsticOperation.swift
```diff
@@ -16,25 +17,43 @@ enum NondeterminsticOperationError: Error {
...
+    /// Immutable now that it is assigned once in `init`, which is what removes it from the
+    /// mutable state the lock below has to cover.
+    private let shouldTryNextOnError: OnError
+
+    /// Guards `_cancellables` and `_task`.
+    private let lock = NSLock()
+    private var _cancellables = Set<AnyCancellable>()
+    private var _task: Task<Void, Never>?
+
+    private var cancellables: Set<AnyCancellable> {
```
thisisabhash, line 40: cancellables`/`task` are exposed via computed properties that each take `lock` on every access. `cancel()` at line 118 reads `task` then writes `cancellables` — two separate lock acquisitions, so a concurrent `execute()` can re-assign `task` in between; the assignment inside `execute()` at line 84 uses `self?.task = Task { ... }` where the RHS is evaluated first (starting the task) and *then* the lock is taken to store it. Net effect: an `execute()` racing with `cancel()` can leave a live task behind whose reference was overwritten. Same shape as an existing behavior on `_task` in `RetryableGraphQLSubscriptionOperation`. Not a regression from before — the pre-change code had exactly this window unprotected — but if the point of adding the lock is to make cancellation reliable, one `withLock` covering read-then-write in `cancel()` (and swap-into-nil semantics) would actually deliver it.

thisisabhash (approved):

PR #4313, opened 2026-09-03
feat(datastore): Sendable across the DataStore plugin

harsh62:
Slice **25 of 38** in the Swift 6 language-mode migration — 34 file(s).

Stacked on `swift6/24-auth-plugin-sendable`. Reference (do not merge): #4283.

### DIFF
--- a/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
+++ b/AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
@@ -13,7 +13,10 @@ import SQLite
 // swiftlint:disable type_body_length file_length
 /// [SQLite](https://sqlite.org) `StorageEngineAdapter` implementation. This class provides
 /// an integration layer between the AppSyncLocal `StorageEngine` and SQLite for local storage.
-final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
+/// - Note: `@unchecked Sendable` to satisfy `StorageEngineAdapter`'s `Sendable` requirement.
+///   `connection` is established during setup and replaced only by `clear`/`setUp`, never
+///   concurrently with queries.
+final class SQLiteStorageEngineAdapter: StorageEngineAdapter, @unchecked Sendable {
     var connection: Connection?
     var dbFilePath: URL?
     static let dbVersionKey = "com.amazonaws.DataStore.dbVersion"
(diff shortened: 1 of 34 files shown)


AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/SQLite/StorageEngineAdapter+SQLite.swift
```diff
@@ -13,7 +13,10 @@ import SQLite
 // swiftlint:disable type_body_length file_length
 /// [SQLite](https://sqlite.org) `StorageEngineAdapter` implementation. This class provides
 /// an integration layer between the AppSyncLocal `StorageEngine` and SQLite for local storage.
-final class SQLiteStorageEngineAdapter: StorageEngineAdapter {
+/// - Note: `@unchecked Sendable` to satisfy `StorageEngineAdapter`'s `Sendable` requirement.
+///   `connection` is established during setup and replaced only by `clear`/`setUp`, never
+///   concurrently with queries.
+final class SQLiteStorageEngineAdapter: StorageEngineAdapter, @unchecked Sendable {
```
thisisabhash, line 27: `@unchecked Sendable` on `SQLiteStorageEngineAdapter` while `var connection: Connection?` and `var dbFilePath: URL?` have no lock and no serial queue. The comment claims `connection` is "replaced only by `clear`/`setUp`, never concurrently with queries," but nothing enforces that: the adapter is called synchronously from many concurrent reconciliation/query operations, and `clear()`/`setUp()` reassign `connection` from a different context. Concurrent reads-during-reassignment are an unsynchronized race in principle.

AmplifyPlugins/DataStore/Sources/AWSDataStorePlugin/Storage/StorageEngine.swift
```diff
@@ -21,7 +21,10 @@ typealias StorageEngineBehaviorFactory =
     ) throws -> StorageEngineBehavior
 
 // swiftlint:disable type_body_length
-final class StorageEngine: StorageEngineBehavior {
+/// - Note: `@unchecked Sendable` to satisfy `StorageEngineBehavior`'s `Sendable` requirement.
+///   `syncEngine` is established during start-up and cleared on `clear`, not written concurrently
+///   with reads.
+final class StorageEngine: StorageEngineBehavior, @unchecked Sendable {
```
thisisabhash, line 34: `@unchecked Sendable` with plain mutable vars (`syncEngine`, `signInListener`, `iSyncEngineSink`, `iStorageEnginePublisher`) and no serial queue or lock. `syncEngine` is read on the hot mutation path (save/delete forward to it) and written during start-up/`clear`; the comment asserts these don't overlap but the class has no synchronization primitive to back that up.

thisisabhash (approved):

PR #4316, opened 2026-09-03
feat(predictions): Sendable across the Predictions plugin

harsh62:
Slice **28 of 38** in the Swift 6 language-mode migration — 26 file(s).

Stacked on `swift6/27-storage-plugin-sendable`. Reference (do not merge): #4283.

### DIFF
--- a/AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/MultiService/InterpretTextMultiService.swift
+++ b/AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/MultiService/InterpretTextMultiService.swift
@@ -7,7 +7,9 @@
 
 import Amplify
 
-class InterpretTextMultiService: MultiServiceBehavior {
+/// - Note: `final` and `@unchecked Sendable` to satisfy `MultiServiceBehavior`; the two service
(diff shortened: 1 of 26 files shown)


AmplifyPlugins/Predictions/AWSPredictionsPlugin/Service/MultiService/InterpretTextMultiService.swift
```diff
@@ -7,7 +7,9 @@
 
 import Amplify
 
-class InterpretTextMultiService: MultiServiceBehavior {
+/// - Note: `final` and `@unchecked Sendable` to satisfy `MultiServiceBehavior`; the two service
+///   references are set up before any call races them.
+final class InterpretTextMultiService: MultiServiceBehavior, @unchecked Sendable {
```
thisisabhash, line 20: `weak var coreMLService` / `weak var predictionsService` are still `var`. The refactor assumes they are wired up before racing in a task group; that is true in current call sites but not enforced. Consider making them `let` at init and dropping `weak` where the lifetime already outlives the operation, or documenting the ordering more precisely.

thisisabhash (approved):

PR #4317, opened 2026-09-03
feat(logging): Sendable across the Logging plugin

harsh62:
Slice **29 of 38** in the Swift 6 language-mode migration — 8 file(s).

Stacked on `swift6/28-predictions-plugin-sendable`. Reference (do not merge): #4283.

### DIFF
--- a/AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/AWSCloudWatchLoggingCategoryClient.swift
+++ b/AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/AWSCloudWatchLoggingCategoryClient.swift
@@ -20,7 +20,9 @@ import SmithyIdentity
 /// the application's authentication state.
 ///
 /// - Tag: CloudWatchLoggingCategoryClient
-final class AWSCloudWatchLoggingCategoryClient {
+/// - Note: `@unchecked Sendable`: the client starts detached tasks that touch its state, which is
+///   guarded by its lock.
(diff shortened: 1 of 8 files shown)


AmplifyPlugins/Logging/Sources/AWSCloudWatchLoggingPlugin/AWSCloudWatchLoggingCategoryClient.swift
```diff
@@ -20,7 +20,9 @@ import SmithyIdentity
 /// the application's authentication state.
 ///
 /// - Tag: CloudWatchLoggingCategoryClient
-final class AWSCloudWatchLoggingCategoryClient {
+/// - Note: `@unchecked Sendable`: the client starts detached tasks that touch its state, which is
+///   guarded by its lock.
+final class AWSCloudWatchLoggingCategoryClient: @unchecked Sendable {
```
thisisabhash, line 26: the `@unchecked Sendable` doc note says state "is guarded by its lock," but writes to `private var enabled` at `init` (line 53), `enable()` (124), `disable()` (133), and the read at `flushLogs()` (204) all happen outside `lock.execute`. Reads inside `logger(forCategory:namespace:logLevel:)` (line 165) do go through the lock, so there is an inconsistent locking discipline on `enabled`. Bool tearing is unlikely on Apple targets, but the note oversells the guarantee — either wrap the remaining sites in `lock.execute` or reword the note. Pre-existing, but worth tightening since the migration is explicitly asserting the invariant now.

thisisabhash (approved):

PR #4318, opened 2026-09-03
feat(analytics): Sendable across Pinpoint, Analytics, Geo and Push

harsh62:
Slice **30 of 38** in the Swift 6 language-mode migration — 16 file(s).

Stacked on `swift6/29-logging-plugin-sendable`. Reference (do not merge): #4283.

### DIFF
--- a/AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/PinpointEvent.swift
+++ b/AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/PinpointEvent.swift
@@ -10,8 +10,11 @@ import Foundation
 
 private typealias Constants = AWSPinpointAnalytics.Constants.Event
 
+/// - Note: `final` and `@unchecked Sendable`: the identity and session fields are `let`; the lazy
+///   attribute/metric dictionaries are populated while the event is being built, before it is handed
+///   to the recorder.
 @_spi(InternalAWSPinpoint)
-public class PinpointEvent: AnalyticsPropertiesModel {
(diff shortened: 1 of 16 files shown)


AmplifyPlugins/Internal/Sources/InternalAWSPinpoint/Analytics/PinpointEvent.swift
```diff
@@ -10,8 +10,11 @@ import Foundation
 
 private typealias Constants = AWSPinpointAnalytics.Constants.Event
 
+/// - Note: `final` and `@unchecked Sendable`: the identity and session fields are `let`; the lazy
+///   attribute/metric dictionaries are populated while the event is being built, before it is handed
+///   to the recorder.
 @_spi(InternalAWSPinpoint)
-public class PinpointEvent: AnalyticsPropertiesModel {
+public final class PinpointEvent: AnalyticsPropertiesModel, @unchecked Sendable {
```
thisisabhash, line 25: `PinpointEvent` is declared `@unchecked Sendable`, but `attributes`/`metrics` are `private(set) lazy var` dictionaries with public `addAttribute(_:forKey:)` / `addMetric(_:forKey:)` mutators. The note claims mutations happen "before it is handed to the recorder," but that is a caller-contract assumption — nothing in the type prevents a caller from adding attributes after enqueue. Since events flow through the `AnalyticsClient` actor, they can readily cross task boundaries. Recommend either freezing the maps at construction (dictionaries provided to `init`) or funneling mutation through a lock. Pre-existing race semantics, but the `Sendable` assertion is what makes this compile-clean going forward.

thisisabhash (approved):

PR #4326, opened 2026-09-03
feat!: enable the Swift 6 language mode

harsh62:
Slice **38 of 38** in the Swift 6 language-mode migration — 2 file(s).

Stacked on `swift6/37-tests-remaining`. Reference (do not merge): #4283.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -733,5 +733,5 @@ let package = Package(
     // Adopting the Swift 6 language mode is tracked separately from this version bump.
     // Raising swift-tools-version to 6.0 would otherwise switch the default language mode
     // to Swift 6, so it is pinned to v5 here to keep this change behavior-neutral.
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -733,5 +733,5 @@ let package = Package(
     // Adopting the Swift 6 language mode is tracked separately from this version bump.
     // Raising swift-tools-version to 6.0 would otherwise switch the default language mode
     // to Swift 6, so it is pinned to v5 here to keep this change behavior-neutral.
```
thisisabhash: the comment block immediately above the changed line is now stale/self-contradictory: it still says the mode "is pinned to v5 here to keep this change behavior-neutral," directly above `swiftLanguageModes: [.v6]`. Should be updated or removed so the manifest doesn't claim v5.

thisisabhash (approved):

Issue #3716, opened 2024-05-22
Crash in save<Models> from DataStoreBaseBehavior


amruth-movano:
### Describe the bug

Hi Team,
Today I got crash while running the ios app.
The crash was in extension class DataStoreBaseBehavior- 

func save<M: Model>(_ models: [M], where predicate: QueryPredicate?) async throws -> [M] {
        let results = try await withThrowingTaskGroup(of: M.self) { group in
            for model in models {
                group.addTask {
                    try await save(model, where: predicate) // Crash line
                }
            }

            return try await group.reduce(into: []) { array, model in
                array.append(model)
            }
        }

        return results
    }

### Steps To Reproduce

```swift
I got it on app launch.

Crash error - 

Thread 94: Fatal error: SWIFT TASK CONTINUATION MISUSE: save(_:modelSchema:where:) tried to resume its continuation more than once, throwing DataStoreError: The operation couldn’t be completed. (SQLite.Result error 0.)
Recovery suggestion: The operation couldn’t be completed. (SQLite.Result error 0.)
Caused by:
cannot rollback - no transaction is active (code: 1)!
```

### Expected behavior

App should not crash in save models

### Amplify Framework Version

2.33.6

### Amplify Categories

DataStore

### Dependency manager

Swift PM

### Swift version

5.9.2

### CLI version

12.11.1

### Xcode version

15.2

### Relevant log output

```shell
Thread 94: Fatal error: SWIFT TASK CONTINUATION MISUSE: save(_:modelSchema:where:) tried to resume its continuation more than once, throwing DataStoreError: The operation couldn’t be completed. (SQLite.Result error 0.)
Recovery suggestion: The operation couldn’t be completed. (SQLite.Result error 0.)
Caused by:
cannot rollback - no transaction is active (code: 1)!
```

### Is this a regression?

No

### Platforms

iOS

### OS Version

iOS 17.5

### Device

iPhone 15 Plus

phantumcode:
@amruth-movano Thanks for submitting your issue.  To help reduce produce the issue, can you provide your model/schema and the request parameters that resulted in the crash?

amruth-movano:
@phantumcode Hi, currently it is not reproducible, so not sure how can I help wth parameters which were resulted in crash

phantumcode:
@amruth-movano Are you able to provide your model schema or a example schema or test app that can help reproduce the issue?

amruth-movano:
@phantumcode 
[...]

PR #4335, opened 2026-09-09
fix(DataStore): prevent crash in save() when the transaction fails to commit

VarshithaPamisetty:
## Description

`StorageEngine.save` invoked `completion` from both inside the DB transaction (the
async sync hand-off) and the outer `catch`, so a transaction that failed to close
resumed the `CheckedContinuation` twice → `SWIFT TASK CONTINUATION MISUSE` crash.
It now captures the sync engine inside the transaction and runs the hand-off and all
completion calls after the transaction closes, so `completion` fires exactly once. A
close failure returns a failure. A retry is safe because saves are idempotent by id:
saving the same model again updates that row instead of adding another.

This only fixes the crash. The SQLite "cannot rollback" race that triggers it is still there, but save fails now instead of crashing.

## Testing

- Regression: double-resume crashes on `main`; passes here. A dedicated test asserts completion runs once, so a symptom-only fix would fail it.
- Real in-memory SQLite: no sync engine → save fails and the row is rolled back (absent); saving the same model twice leaves a single row; normal save persists and is queryable.
- Unit suite passing; `swiftformat`/`swiftlint` clean; macOS/iOS/tvOS/watchOS/visionOS + real AppSync sync verified.
- Couldn't repro the spontaneous crash on demand since it's a SQLite timing race, so the test forces the same double-completion deterministically.

Refs #3716


AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Storage/StorageEngineTestsSaveDoubleResume.swift
```diff
@@ -0,0 +1,237 @@
...
+        ModelRegistry.register(modelType: Post.self)
+    }
+
+    private func makeStorageEngine(
+        adapter: StorageEngineAdapter,
+        syncEngine: RemoteSyncEngineBehavior? = {
+            let engine = MockRemoteSyncEngine()
+            engine.syncing = true
+            return engine
+        }()
```
thisisabhash, line 84: setUp() never calls super.setUp() and there's no tearDown() resetting ModelRegistry — leaks global model-registration state across the suite, risking order-dependent failures.

AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Storage/StorageEngineTestsSaveDoubleResume.swift
```diff
@@ -0,0 +1,237 @@
...
+    /// - Given: a real SQLite store and a sync engine
+    /// - When: a model is saved normally
+    /// - Then: the save succeeds, persists locally, and invokes completion exactly once
+    func testNormalSaveThroughRealSQLitePersistsAndCompletesOnce() async throws {
+        let plugin = makePlugin(makeStorageEngine(adapter: try realInMemoryAdapter()))
+        let post = Post(title: "e2e-3716", content: "normal save", createdAt: .now())
+
+        let saved = try await plugin.save(post)
+        XCTAssertEqual(saved.id, post.id)
+        XCTAssertEqual(saved.title, "e2e-3716")
```
thisisabhash, line 146: Double-resume detection relies on fixed sleeps (Task.sleep 300ms / Thread.sleep 0.3s) rather than a deterministic signal → timing-flaky and slow. On a loaded CI runner a regression could slip past the 300ms window. Also testSaveWithTransactionCloseFailureDoesNotDoubleResume only asserts an error was thrown, never the completion count.

VarshithaPamisetty: Dropped both sleeps, hand-off now caught by an inverted expectation and the double-resume by `assertForOverFulfill`. Also renamed the async test.

AmplifyPlugins/DataStore/Tests/AWSDataStorePluginTests/Storage/StorageEngineTestsSaveDoubleResume.swift
VarshithaPamisetty: Thanks for pointing it out. Added super.setUp() and a tearDown() that calls ModelRegistry.reset(), so the model registration no longer leaks into other tests.

thisisabhash (approved):

